PR #5, opened 2024-06-26
Build as a purelib by replacing the `_extras` module with a Python implementation

gmarkall:
The `_extras` module only serves to fix up the ABI for passing an IPC handle to `cuIpcOpenMemHandle`, which expects a structure passed by value. It seems that this requirement can be accommodated using the ctypes `Structure` class, so we can replace the functionality of the `_extras` module with it.

Some changes needed to be made to accommodate the way in which we now represent an IPC handle, in the tests and in how we serialize them.

This change makes `numba-cuda` a pure Python package, with no C / C++ build / extension modules needed.

The CI and build changes are aimed at building a purelib wheel / noarch Python conda package.


pyproject.toml
```diff
@@ -4,6 +4,7 @@ cmake.verbose = true
 ninja.make-fallback = true
 build-dir = "build/{wheel_tag}"
 wheel.packages = ["numba_cuda"]
+wheel.platlib= ""
```
gmarkall, line 7:
```suggestion
wheel.platlib= ""
```
```suggestion
wheel.platlib = ""
```

gmarkall:
I just merged this even though CI is not running because the lack of it is making it hard for people to try out the package by building it locally.

PR #8, opened 2024-07-03
let `opt` respect `config.OPT`, also change its default value to `None`

dlee992:
fixes https://github.com/numba/numba/issues/9469

Design:
Based on Graham's comment in the original issue, I choose to set `opt` default value to `None`, then give users a chance to change `opt` option for a single jitted func. If users don't specify it, then `opt` is decided by `NUMBA_OPT`. The rule is:

1. if `NUMBA_OPT` is `0`, then `opt` is `False`;
2. otherwise, it's `True`.

TODOs:

- [x] change comments for `cuda.jit` function.
- [x] change documentation if needed: didn't find any documentation needs to rewrite.
- [x] revise `numba.cuda.compile` as well.
- [x] add test cases.

Let me watch how this new CI works!


dlee992:
Looks like CI won't automatically run for my commit.

gmarkall:
> Looks like CI won't automatically run for my commit.

It's presently not working, I'm awaiting a fix on the CI.

numba_cuda/numba/cuda/decorators.py
```diff
@@ -71,6 +74,7 @@ def jit(func_or_sig=None, device=False, inline=False, link=[], debug=None,
         raise DeprecationError(msg)
 
     debug = config.CUDA_DEBUGINFO_DEFAULT if debug is None else debug
+    opt = (True if config.OPT != 0 else False) if opt is None else opt
```
gmarkall, line 77: I think this is the same as

```suggestion
    opt = (config.OPT != 0) if opt is None else opt
```

numba_cuda/numba/cuda/decorators.py
```diff
@@ -44,7 +44,10 @@ def jit(func_or_sig=None, device=False, inline=False, link=[], debug=None,
        Useful for increasing occupancy.
     :param opt: Whether to compile from LLVM IR to PTX with optimization
                 enabled. When ``True``, ``-opt=3`` is passed to NVVM. When
-                ``False``, ``-opt=0`` is passed to NVVM. Defaults to ``True``.
+                ``False``, ``-opt=0`` is passed to NVVM. Defaults to ``None``.
+                The default value will be reset by environment variable
+                ``NUMBA_OPT``. If ``0``, ``opt`` is reset as ``False``;
+                otherwise, it is reset to ``True``.
```
gmarkall, line 50: I think the original documentation needlessly goes into implementation details - users shouldn't even have to know what NVVM is, necessarily. I think it would be simpler to understand as:

> Whether to compile with optimization enabled. If unspecified, the ``OPT`` configuration variable is consulted; all non-zero values will enable optimization.

gmarkall:
Thanks for the PR - I think the idea is good in principle, so if you'd like to add tests I can test manually if CI isn't working before your PR is ready.

numba_cuda/numba/cuda/decorators.py
dlee992: simplified the doc. also added ``NUMBAOPT`` in it.

dlee992:
Looks like this `compile` function should be also revised as `cuda.jit`. Quite similar.
```python
@global_compiler_lock
def compile(pyfunc, sig, debug=False, lineinfo=False, device=True,
            fastmath=False, cc=None, opt=True, abi="c", abi_info=None,
            output='ptx'):
```
Add this to TODOs.

gmarkall:
> Looks like this `compile` function should be also revised as `cuda.jit`

Ah, good catch!

dlee992:
Suddenly, this is kinda out of control (since `_OptLevel` is an internal class in `config`, and have to get its `_raw_value` field, otherwise need to add a method in that class), but I tried to manage it.
need some reviews.

Now, perhaps we need a `NUMBA_CUDA_OPT` separately, which can simplify code a bit. At the very beginning, I think this should be a one-line change PR : )

gmarkall:
@dlee992 From your previous comment it sounds like you think there's a problem here, but I can't see what it is - if I review this and it looked good to me, would you be happy with it? (I have already skimmed the PR and can't see an obvious conceptual problem)

dlee992:
Yes, please review this.

I just met the same issue when I write tests: https://github.com/numba/numba/issues/9646. 
And wonder if numba core can have a better way or we can use a separate numba ENVVAR to control this OPT behaviour, but we can discuss it after this PR.

gmarkall:
I hope you don't mind that I merged in develop and pushed to this branch - this is because I got the CI setup working again, and wanted to give this a test run.

numba_cuda/numba/cuda/compiler.py
```diff
@@ -307,8 +307,11 @@ def compile(pyfunc, sig, debug=False, lineinfo=False, device=True,
     if output not in ("ptx", "ltoir"):
         raise NotImplementedError(f'Unsupported output type: {output}')
 
+    debug = config.CUDA_DEBUGINFO_DEFAULT if debug is None else debug
```
gmarkall, line 310: Whilst I think this change makes sense, I think `debug=False` needs changing to `debug=None` in the kwargs as well if we're making this change.

numba_cuda/numba/cuda/compiler.py
```diff
@@ -307,8 +307,11 @@ def compile(pyfunc, sig, debug=False, lineinfo=False, device=True,
...
+    opt = (config.OPT._raw_value != 0) if opt is None else opt
```
gmarkall, line 311: I think the opt level is an int subclass, so we can avoid using a private property:
```suggestion
    opt = (config.OPT != 0) if opt is None else opt
```

numba_cuda/numba/cuda/compiler.py
gmarkall: (I think a similar change would need making for the other `compile_` functions beneath, too)

numba_cuda/numba/cuda/decorators.py
```diff
@@ -71,11 +71,12 @@ def jit(func_or_sig=None, device=False, inline=False, link=[], debug=None,
         raise DeprecationError(msg)
 
     debug = config.CUDA_DEBUGINFO_DEFAULT if debug is None else debug
+    opt = (config.OPT._raw_value != 0) if opt is None else opt
```
gmarkall, line 74: Also needs to avoid using the raw value.

numba_cuda/numba/cuda/simulator/api.py
```diff
@@ -91,6 +92,9 @@ def jit(func_or_sig=None, device=False, debug=False, argtypes=None,
     if link is not None:
         raise NotImplementedError('Cannot link PTX in the simulator')
 
+    debug = config.CUDA_DEBUGINFO_DEFAULT if debug is None else debug
+    opt = (config.OPT._raw_value != 0) if opt is None else opt
```
gmarkall, line 96: `opt` isn't used so I don't think the logic on this line needs to be added here.

I've run flake8 locally and I'm not sure why it doesn't pick this up - maybe because it's a parameter?

gmarkall (changes requested):
Thanks for the PR! I think the changes are in general good and the tests make sense. I have left a couple of comments on the diff. I didn't like using underscore-prefixed properties to do with the opt level, and I think it can be avoided - unfortunately I can't see a way to avoid using `_OptLevel` in the tests though.

Thanks for clearning up all the "debug with opt" warnings, too. I think there is one more case to clean up in `test_raise_in_device_function`:

```diff
diff --git a/numba_cuda/numba/cuda/tests/cudapy/test_exception.py b/numba_cuda/numba/cuda/tests/cudapy/test_exception.py
index 8891010..42f3107 100644
--- a/numba_cuda/numba/cuda/tests/cudapy/test_exception.py
+++ b/numba_cuda/numba/cuda/tests/cudapy/test_exception.py
@@ -160,7 +160,7 @@ class TestException(CUDATestCase):
         def f():
             raise ValueError(msg)
 
-        @cuda.jit(debug=True)
+        @cuda.jit(debug=True, opt=False)
         def kernel():
             f()
 
```

if you'd like to add that one as well?

Otherwise, I think this is looking good - many thanks!

numba_cuda/numba/cuda/simulator/api.py
dlee992: removed `opt` line.

numba_cuda/numba/cuda/decorators.py
dlee992: removed `_raw_value`.

numba_cuda/numba/cuda/compiler.py
dlee992: as it is.

numba_cuda/numba/cuda/compiler.py
dlee992: change `debug=None` in signature for `compile` and 3 wrappers.

dlee992:
> Thanks for clearning up all the "debug with opt" warnings, too. I think there is one more case to clean up in test_raise_in_device_function:

Also added `opt=False` in this test.

Pushed all requested changes in latest commit. 

My local `python -m numba.runtests numba.cuda.tests -m` passed.

gmarkall:
Thanks for your efforts here! I don't have cudasim working in CI yet, but cudasim tests pass for me locally.

PR #13, opened 2024-07-05
Search for nvdisasm in the conda environment / `CUDA_HOME` / default CUDA install location in addition to the path

### LINKED ISSUE #9: [BUG] SASS disassembly always tries to use `nvdisasm` on the `PATH`
See: https://github.com/NVIDIA/numba-cuda/blob/e3658d4ac6475d8d7f615aefb2e07e39aee3912e/numba_cuda/numba/cuda/codegen.py#L27-L29

It would be better to search for CUDA command line tools in the more normal way, e.g. from `CUDA_HOME` / `CUDA_PATH` first.

brycelelbach:
Fixes #9.

I'm not sure how good the unit test coverage is for this, or the other path search functionality. Ideally we'd need contained tests that cover:
- Conda
- A default CUDA install (e.g. `/usr/local/cuda`):
  - With the full toolkit.
  - With only libraries, no `nvdisasm` etc.
  - With only libraries, but `nvdisasm` in PATH.
- CUDA_HOME:
   - With the full toolkit.
  - With only libraries, no `nvdisasm` etc.
  - With only libraries, but `nvdisasm` in PATH.

### DIFF
--- a/numba_cuda/numba/cuda/codegen.py
+++ b/numba_cuda/numba/cuda/codegen.py
@@ -4,6 +4,7 @@ from numba.core import config, serialize
 from numba.core.codegen import Codegen, CodeLibrary
 from .cudadrv import devices, driver, nvvm, runtime
 from numba.cuda.cudadrv.libs import get_cudalib
+from numba.cuda.cuda_paths import get_cuda_paths
 
 import os
 import subprocess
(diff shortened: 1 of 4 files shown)


gmarkall:
Just merged `develop` and pushed so that CI can run on this PR.

gmarkall:
Looks like there's some issue with `nvdisasm` tests with this PR - I haven't looked into it yet, but let me know if any guidance would be helpful.

gmarkall:
Rebased on main - I'm trying to get everything away from develop now as it no longer seems necessary.

gmarkall:
See also: https://github.com/NVIDIA/cuda-python/issues/762

kkraus14:
Going to close this as this will be replaced by pathfinder in the near future

PR #16, opened 2024-07-16
Replace scikit-build-core / CMake with setuptools

gmarkall:
Using scikit-build-core and CMake was overkill for copying two files into `site-packages`. This change replaces them with setuptools, and also modifies some of the package names to use a dash (i.e. `numba-cuda`) which I think is more "standard" in the places in which I've done it here.

Manual testing suggests that the packages built before and after this change are identical, with a couple of small exceptions:

- The `LICENSE` file is placed in the root by setuptools instead of a folder called `licenses`.
- Metadata differs slightly (`WHEEL`, `RECORD`, `METADATA`) due to them being generated slightly differently between the build tools.
- setuptools also includes a `top_level.txt` file.

Many thanks to @vyasr for pointing me to the solution to copy files to `site-packages` with setuptools, as used in
[`rapids-dask-dependency`](https://github.com/rapidsai/rapids-dask-dependency/).


ci/build_wheel.sh
```diff
@@ -8,4 +8,4 @@ mkdir -p ./dist
 python -m pip wheel . --wheel-dir=./dist -vvv --disable-pip-version-check --no-deps
 
 rapids-logger "Upload Wheel"
-RAPIDS_PY_WHEEL_NAME="numba_cuda" rapids-upload-wheels-to-s3 ./dist
+RAPIDS_PY_WHEEL_NAME="numba-cuda" rapids-upload-wheels-to-s3 ./dist
```
vyasr, line 11: I believe that this needs to stay with an underscore. This gets used for the wheel filename, which is always normalized to have underscores even if the package name has a hyphen IIRC.

ci/test_wheel.sh
```diff
@@ -15,7 +15,7 @@ rapids-logger "Download Wheel"
 RAPIDS_PY_WHEEL_NAME="numba_cuda" rapids-download-wheels-from-s3 ./dist/
 
 rapids-logger "Install wheel"
-python -m pip install $(echo ./dist/numba_cuda*.whl)
+python -m pip install $(echo ./dist/numba-cuda*.whl)
```
vyasr, line 18: Same as above, I think this needs to stay underscored.

setup.py
```diff
@@ -0,0 +1,24 @@
...
+from setuptools.command.build_py import build_py
+
+
+# Adapted from https://stackoverflow.com/a/71137790
+class build_py_with_redirector(build_py):  # noqa: N801
+    """Include the redirector files in the generated wheel."""
+
+    def copy_redirector_file(self, extension):
+        redirector = f"_numba_cuda_redirector.{extension}"
+        source = pathlib.Path("site-packages") / redirector
```
vyasr, line 14: Nitpick: specializing the function on the filename or the path this way isn't my favorite. I would rather see this function accept two arguments, 1) the input file path relative to the package root and 2) the target directory. The latter can default to "." so that the destination defaults to `self.build_lib` and you can call the function with just the input filename below.

vyasr (changes requested):

ci/build_wheel.sh
gmarkall: Thanks, I've undone this now.

ci/test_wheel.sh
gmarkall: Thanks, undone as above.

setup.py
gmarkall: Changed to follow the suggestion, and verified locally that that change didn't result in a change to the built wheel.

vyasr (approved):
My review has no power here, but this LGTM.

gmarkall:
> My review has no power here, but this LGTM.

On the contrary, I would not merge it without your approval :slightly_smiling_face:

PR #17, opened 2024-07-16
Add NRT c++ functions

brandon-b-miller:
This PR adds device side implementations of some of the NRT c-api. Combined with the change to the CUDATargetContext object, this allows launching of kernels like this:

```python
from numba import cuda
import numpy as np


@cuda.jit
def f(x):
    return x[:5]

@cuda.jit('void()', link=['nrt.cu'])
def g():
    x = cuda.shared.array(10, dtype=np.int32)
    f(x)

g[1,1]()
```

Several things are as of the time of writing missing:
- Currently, no memsys is used and it remains to be discussed how we'd like to expose it, knowing that it may have to be used outside of numba to free objects that persist after the kernel finishes executing.
- ~~No tests are currently included~~ [UPDATE] Tests are now added. We have `test_nrt.py` that mainly enables NRT and test that a refcounted variable can successfully pass in and return from a second function. We also have another test that mocks up a device side allocation and test that allocation statistics can be correctly collected.
- ~~We should discuss how and when we'd like to include these functions in the link. Perhaps some of its compilation might be cached and also it might not be necessary to add it to every link, only the ones that require nrt.~~ [UPDATE] NRT functions are now only linked when any of the NRT specific functions are found in the jitted PTX.


numba_cuda/numba/cuda/runtime/nrt.cu
```diff
@@ -0,0 +1,107 @@
...
+  NRT_MemInfo_destroy(mi);
+}
+
+/*
+  c++ version of the NRT_decref function that usually is added to
+  the final kernel link in PTX form by numba. This version may be
+  used by c++ APIs that accept ownership of live objects and must
+  manage them going forward.
+*/
+extern "C" __device__ void NRT_decref(NRT_MemInfo* mi)
```
brandon-b-miller, line 96: these two functions are intended to replace the ones dynamically generated here
https://github.com/numba/numba/blob/main/numba/core/runtime/nrtdynmod.py

numba_cuda/numba/cuda/runtime/nrt.cu
```diff
@@ -0,0 +1,107 @@
...
+{
+  mi->refct--;
+  if (mi->refct == 0) { NRT_MemInfo_call_dtor(mi); }
```
brycelelbach, line 99: Currently, you are implicitly using `cuda::memory_order_seq_cst`. You can instead use acquire/release semantics here, which will have less synchronization overhead.

Replace this code with:

```
if (mi->refct.fetch_sub(1, cuda::memory_order_acq_rel)) { NRT_MemInfo_call_dtor(mi); }
```

`fetch_sub` is a read-modify-write operation that will atomically replace the current value of the atomic with the result of subtracting the argument (in this case `1`) from the current value. 

With acquire-release semantics, you are guaranteed that no reads or writes of the atomic in the current thread can be reordered before the `fetch_sub` loads or after the `fetch_sub` stores. You're also guaranteed that all writes in other threads that decrement the same atomic are visible before `fetch_sub` modifies the atomic.

numba_cuda/numba/cuda/runtime/nrt.cu
```diff
@@ -0,0 +1,107 @@
...
+}
+
+#endif
+
+extern "C" __device__ void NRT_incref(NRT_MemInfo* mi)
+{
+  mi->refct++;
```
brycelelbach, line 106: Right now, this is using sequential consistency, which is very expensive device side.

We don't actually need that, because we don't need to enforce any memory ordering when we increment. That is, we don't need to guarantee that any writes from other threads are visible to us, or that any other writes in our thread are visible to other threads. We just need the reference count to increase.

This can be replaced with `mi->refct.fetch_add(1, cuda::memory_order_relaxed)`. That will be substantially faster in GPU code, because it doesn't need any memory ordering barriers.

gmarkall:
Some additional changes needed for adding the `nrt` test module: https://github.com/gmarkall/numba-cuda/commit/9bb662d7ca8e5319bc1507f1aa01133cad24c532

gmarkall:
I tried to run the test but I had to work around a couple of other issues:

- `np.empty()` has a kwarg, which presently doesn't work with the CUDA target - I just added a non-kwarg equivalent in the test: https://github.com/gmarkall/numba-cuda/commit/8a4f991a1ed7f102099c515dad9df8ac947fbbed
- The NRT functions aren't linked in. A quick hack (that shouldn't need much refinement) to include it when NRT is used: https://github.com/gmarkall/numba-cuda/commit/0352f231893b6c55cc86bfb9afe2f966c2c9435a

At that point, running the test with:

```
$ python ../numba/runtests.py numba.cuda.tests.nrt.test_nrt
```

is giving me:

```
numba.cuda.cudadrv.driver.LinkerError: [300] Call to cuLinkComplete results in CUDA_ERROR_INVALID_SOURCE
error   : Undefined reference to 'NRT_MemInfo_data_fast' in '<cudapy-ptx>'
error   : Undefined reference to 'NRT_MemInfo_alloc_aligned' in '<cudapy-ptx>'
```

numba_cuda/numba/cuda/runtime/nrt.cu
gmarkall: (for reference) The implementation of these in the CPU target also emits a release fence prior to the write: https://github.com/numba/numba/blob/fca98287a8fc9da1825ec4e6b570ef8eeaf4605c/numba/core/runtime/nrtdynmod.py#L84-L94

numba_cuda/numba/cuda/runtime/nrt.cu
```diff
@@ -0,0 +1,107 @@
+#ifndef _NRT_H
+#define _NRT_H
+
+#include <cuda/atomic>
+
+typedef __device__ void (*NRT_dtor_function)(void* ptr, size_t size, void* info);
```
gmarkall, line 6: Whilst running the example from the issue description, I get:

```
nrt.cu(6): warning #1835-D: attribute "__device__" does not apply here
  typedef __device__ void (*NRT_dtor_function)(void* ptr, size_t size, void* info);
```

gmarkall:
WIth the changes from my branch above, and syncing before the original example exits:

```python
from numba import cuda
import numpy as np


@cuda.jit
def f(x):
    return x[:5]


@cuda.jit
def g():
    x = cuda.shared.array(10, dtype=np.int32)
    f(x)


g[1,1]()
cuda.synchronize()
```

we get:

```
numba.cuda.cudadrv.driver.CudaAPIError: [700] Call to cuCtxSynchronize results in UNKNOWN_CUDA_ERROR
```

I presume this also happens for you, but without the synchronize at the end, Python exits before the error is detected?

gmarkall:
I see, it looks like the meminfo is null:

```
========= Invalid __global__ atomic of size 8 bytes
=========     at NRT_incref+0xb0
=========     by thread (0,0,0) in block (0,0,0)
=========     Address 0x0 is out of bounds
=========     and is 8,703,180,800 bytes before the nearest allocation at 0x206c00000 of size 8,388,864 bytes
=========     Device Frame:cudapy::__main__::g[abi:v1,cw51cXTLSUwv1sCUt9Ww0FGw1NRRwGKFLSgtAWVgaQPKRwoMW3A4A4MbLeahKhGJDEkANeHAJJA8YlutDgtedSsFaDSpAyWBPONaTQA_3d]+0x40 in /home/gmarkall/numbadev/issues/numba-cuda-17/original_repro.py:10
```

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -115,6 +118,10 @@ def __init__(self, py_func, argtypes, link=None, debug=False,
                                              'cpp_function_wrappers.cu')
             link.append(functions_cu_path)
 
+        if "NRT_" in asm:
```
brandon-b-miller, line 121: This code is a quick and dirty way to identify when the nrt c++ functions need to be added to the link, but there's room for improvement here. 

High level, I usually try and avoid having to inspect something that I created, because it seems better to capture whatever the needed information is during the creation process. So in this case my instinct leans towards deciding to link at the time the NRT functions are inserted rather than here. 

However numba has the additional optimization pass that can prune out NRT API calls, meaning that we could end up in a situation where we decide to link because NRT calls were inserted, but ultimately don't need them because they were optimized out.

This pushes me back towards inspecting the final object, though I think we should inspect the most optimized possible state of the asm in the end. An improvement beyond what's written might be to check if the full function names are contained in the asm, but I still think it's possible to break that. 

@gmarkall could we inspect some of the intermediate files that come after the PTX stage, but before link time, maybe for certain symbols, and decide to link or not based on that?

numba_cuda/numba/cuda/target.py
```diff
@@ -67,6 +67,7 @@ def resolve_value_type(self, val):
 class CUDATargetContext(BaseContext):
     implement_powi_as_math_call = True
     strict_alignment = True
+    enable_nrt = True
```
brandon-b-miller, line 70: This should be opt-in for now with a numba configuration option

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -388,7 +395,15 @@ def _prepare_args(self, ty, val, stream, retr, kernelargs):
...
-            meminfo = ctypes.c_void_p(0)
+            # hack: allocate too much space to hold the meminfo and set it all to 1
+            dbuf = cuda.to_device(np.ones(1024, dtype=np.uint8))
+            
+            # hack: don't garbage collect up the array containing the meminfo
+            def inner():
+                print(dbuf)
+            retr.append(inner)
+
+            meminfo = ctypes.c_void_p(dbuf.gpu_data._mem.device_ctypes_pointer.value)
```
brandon-b-miller, line 406: I am thinking this whole block of changes should really be a call to `NRT_MemInfo_new` building around the pointer to the array data

numba_cuda/numba/cuda/dispatcher.py
dlee992: why not respect `enable_nrt = True` in `CUDATargetContext` to decide if link `nrt.cu`?

I think if the original Numba IR has `NRT_..` calls, the optimized version won't remove them all in most cases.

gmarkall: A good point fron @dlee992 - If we link the NRT functions when NRT is enabled and they turn out not to be needed, then they'll be dropped by the linker. So maybe it is just fine to just link the NRT code all the time if we have NRT enabled.

numba_cuda/numba/cuda/dispatcher.py
gmarkall: Would you be calling it on the host and copying the initialized struct to the GPU?

numba_cuda/numba/cuda/dispatcher.py
dlee992: I'm thinking how to port https://github.com/numba/numba/pull/8458 to numba-cuda. 

I think more `xxx.cu` will show up, perhaps you can revise the linking process to a generic way, as Testhound did with `link_to_library_functions`. Does it make  sense?
```python
        # Link to the helper library functions if needed
        link_to_library_functions(helper_functions, 'cuda_helperlib.cu')
        # Link to the CUDA FP16 math library functions if needed
        res = link_to_library_functions(cuda_fp16_math_funcs,
                                        'cpp_function_wrappers.cu',
                                        '__numba_wrapper_')
```

cc @gmarkall

gmarkall: I think that would be OK for now... Thinking about the trajectory where more and more .cu files show up, we should maybe have another mechanism adding files to the link as we generate the calls to them, but for now I think using the `link_to_library_functions` method that scans the assembly will be OK and allows us to move forward.

gmarkall:
From CI: 

```
numba.cuda.cudadrv.error.NvrtcError: NVRTC Compilation failure whilst compiling nrt.cu:

nrt.cu(4): catastrophic error: cannot open source file "cuda/atomic"
```

I wonder if there are additional CUDA toolkit packages on which this introduces a depdencdency. Perhaps we need to install more in the CI environment (or perhaps the NVRTC include path is just set incorrectly).

brandon-b-miller:
> From CI:
> 
> ```
> numba.cuda.cudadrv.error.NvrtcError: NVRTC Compilation failure whilst compiling nrt.cu:
> 
> nrt.cu(4): catastrophic error: cannot open source file "cuda/atomic"
> ```
> 
> I wonder if there are additional CUDA toolkit packages on which this introduces a depdencdency. Perhaps we need to install more in the CI environment (or perhaps the NVRTC include path is just set incorrectly).

I think `cuda/atomic` is meant to come in from libcudacxx. I'm not sure numba has this dependency yet. I will follow up here

gmarkall:
> I think `cuda/atomic` is meant to come in from libcudacxx. I'm not sure numba has this dependency yet. I will follow up here

I think we need to install / depend on the `cuda-cccl` package.

gmarkall:
I think Numba is still not finding the headers because it's not necessarily looking in the right place for them - we have an environment variable to set the CUDA include path, but we probably need to find a way for Numba to find these headers when they're installed elsewhere.

brandon-b-miller:
> I think Numba is still not finding the headers because it's not necessarily looking in the right place for them - we have an environment variable to set the CUDA include path, but we probably need to find a way for Numba to find these headers when they're installed elsewhere.

Looks like when the cccl package is installed, there's a number of places the files can end up. I have a hunch that the real way of fixing this is to rework things in terms of cmake so that `cccl` (as well as other pieces of cuda that are relied on) are discoverable in any environment seamlessly. Then we can embed the paths from cmake into a file and numba can pick them up from there. I will seek some feedback on this idea. 

Without that, we might be stuck with a separate environment variable in the short term.

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -237,7 +237,7 @@ def compile(src, name, cc):
 
     cudadrv_path = os.path.dirname(os.path.abspath(__file__))
     numba_cuda_path = os.path.dirname(cudadrv_path)
-    numba_include = f'-I{numba_cuda_path}'
+    numba_include = f'-I{numba_cuda_path} -I{_get_include_dir()}'
```
isVoid, line 240: It looks like it's a bug here to return the result of `_get_include_dir()`. Because it returns a named tuple, but `numba_include` takes the path to the include dir.

Secondly, the include path seems to conflict with the `include` flag that's already passing into nvrtc. There should be a single source of CTK include, whether its site installation or conda.

My experiment seems to show that for conda installed includes, a more robust path is `$CONDA_PREFIX/targets/<this_machine_arch>/include`.

Finally, cmake uses a different way to find CTK at compile time, which greps output from nvcc compiling an empty file. Adopting this to Numba would be a great change and is certainly out of scope here - but would it be a correct move forward?

numba_cuda/numba/cuda/target.py
isVoid: Which I assume a separate CI item should be introduced to separately test NRT enabled and disabled?

brandon-b-miller:
With @isVoid 's changes, the tests now pass 🎉

numba_cuda/numba/cuda/tests/nrt/test_nrt.py
```diff
@@ -0,0 +1,108 @@
...
+        def impl(shape, dtype):
+            return numpy_empty_nd(shape, dtype, retty)
+        return impl
+    else:
+        msg = f"Cannot parse input types to function np.empty({shape}, {dtype})"
+        raise errors.TypingError(msg)
+
+
+@unittest.skip
+class TestNrtRefCt(EnableNRTStatsMixin, TestCase):
```
gmarkall, line 44: Should this be

```suggestion
class TestNrtRefCt(EnableNRTStatsMixin, CUDATestCase):
```

?

gmarkall: I know it's skipped, but oddly it seems not to even get discovered when I'm running the tests locally:

```
$ python -m numba.runtests numba.cuda.tests.nrt.test_nrt -v
test_nrt_launches (numba.cuda.tests.nrt.test_nrt.TestNrtBasic.test_nrt_launches) ... ok
test_nrt_returns_correct (numba.cuda.tests.nrt.test_nrt.TestNrtBasic.test_nrt_returns_correct) ... ok

Ran 2 tests in 2.926s

OK
```

gmarkall: Maybe this is something odd about `unittest`. If I change it to:

```diff
diff --git a/numba_cuda/numba/cuda/tests/nrt/test_nrt.py b/numba_cuda/numba/cuda/tests/nrt/test_nrt.py
index 6c77c9e..39e9162 100644
--- a/numba_cuda/numba/cuda/tests/nrt/test_nrt.py
+++ b/numba_cuda/numba/cuda/tests/nrt/test_nrt.py
@@ -40,14 +40,14 @@ def ol_cuda_empty(shape, dtype):
         raise errors.TypingError(msg)
 
 
-@unittest.skip
-class TestNrtRefCt(EnableNRTStatsMixin, TestCase):
+class TestNrtRefCt(EnableNRTStatsMixin, CUDATestCase):
 
     def setUp(self):
         # Clean up any NRT-backed objects hanging in a dead reference cycle
         gc.collect()
         super(TestNrtRefCt, self).setUp()
 
+    @unittest.skip("Stats not yet implemented")
     def test_no_return(self):
         """
         Test issue #1291
```

then the test does get discovered and skipped:

```
$ python -m numba.runtests numba.cuda.tests.nrt.test_nrt -v
test_nrt_launches (numba.cuda.tests.nrt.test_nrt.TestNrtBasic.test_nrt_launches) ... ok
test_nrt_returns_correct (numba.cuda.tests.nrt.test_nrt.TestNrtBasic.test_nrt_returns_correct) ... ok
test_no_return (numba.cuda.tests.nrt.test_nrt.TestNrtRefCt.test_no_return)
Test issue #1291 ... skipped 'Stats not yet implemented'

Ran 3 tests in 2.921s

OK (skipped=1)
```

numba_cuda/numba/cuda/tests/nrt/test_nrt.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def setUp(self):
+        # Clean up any NRT-backed objects hanging in a dead reference cycle
+        gc.collect()
+        super(TestNrtRefCt, self).setUp()
+
+    def test_no_return(self):
```
gmarkall, line 51: With the changes below, we can run this test and expect it to fail instead - maybe we should do this so that it can highlight when stats are working?
```suggestion
    @unittest.expectedFailure
    def test_no_return(self):
```

numba_cuda/numba/cuda/tests/nrt/test_nrt.py
```diff
@@ -0,0 +1,108 @@
...
+
+        def impl(shape, dtype):
+            return numpy_empty_nd(shape, dtype, retty)
+        return impl
+    else:
+        msg = f"Cannot parse input types to function np.empty({shape}, {dtype})"
+        raise errors.TypingError(msg)
+
+
+@unittest.skip
```
gmarkall, line 43: I think we should remove this skip, fix up the test case as below, and xfail it:
```suggestion
```

numba_cuda/numba/cuda/tests/nrt/test_nrt.py
```diff
@@ -0,0 +1,108 @@
...
+from numba.cuda.testing import CUDATestCase
+
+from numba import cuda
+
+from numba.core import errors, types
+from numba.core.extending import overload
+from numba.np.arrayobj import (_check_const_str_dtype, is_nonelike,
+                               ty_parse_dtype, ty_parse_shape, numpy_empty_nd)
+
+
```
gmarkall, line 16: I think some explanation here would be good. When I came to this PR, I forgot what the point of this was and had to spend some time figuring it out. I think I wrote this workaround, and if I get confused by it, I think it definitely needs some explanation for others! E.g.:
```suggestion
# Typical tests for allocation use array construction (e.g. np.zeros, np.empty,
# etc.) to induce allocations. These don't work in the CUDA target because they
# need keyword arguments, which are presently not supported properly in the
# CUDA target.
#
# To work around this, we define the cuda_empty() function, that works like
# np.empty(), except that it uses only positional arguments.
#
# Once the CUDA target supports keyword arguments, this workaround will no
# longer be necessary and the tests in this module should be switched to use
# the relevant NumPy functions instead.

```

brandon-b-miller: What do you think of the changes in https://github.com/NVIDIA/numba-cuda/pull/17/commits/8e389f1eb7c376232bea25aca41d573e3e5e47e4 ?

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -256,3 +258,66 @@ def get_debian_pkg_libdevice():
...
+        return system_cuda_include
+    return None
+
+
+def _get_include_dir():
+    """Find the root include directory."""
+    options = [
+        ('Conda environment', get_conda_include_dir()),
+        ('System', get_system_include_dir()),
+        # TODO: add others
```
gmarkall, line 320: I think we need to search the `config.CUDA_INCLUDE_PATH` before these other two options, otherwise users who relied on setting it will find it no longer has any effect.

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -233,12 +232,16 @@ def compile(src, name, cc):
     #   being optimized away.
     major, minor = cc
     arch = f'--gpu-architecture=compute_{major}{minor}'
-    include = f'-I{config.CUDA_INCLUDE_PATH}'
```
gmarkall, line 236: As mentioned in the comment on cuda_paths.py, there seems to be no substitute for using `CUDA_INCLUDE_PATH`.

numba_cuda/numba/cuda/cuda_paths.py
gmarkall: Note existing logic in https://github.com/numba/numba/blob/main/numba/core/config.py#L467-L478

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -241,6 +242,7 @@ def get_cuda_paths():
             'libdevice': _get_libdevice_paths(),
             'cudalib_dir': _get_cudalib_dir(),
             'static_cudalib_dir': _get_static_cudalib_dir(),
+            'include_dir': _get_include_dir(),
```
gmarkall, line 245: The result of this should also be printed out in `numba.cuda.cudadrv.libs.test()`: https://github.com/NVIDIA/numba-cuda/blob/main/numba_cuda/numba/cuda/cudadrv/libs.py#L83

I know it's not libraries (and we might consider renaming that function / file) but the intent there is to debug all of Numba's CUDA file-searching functionality, and I think having the include paths also available in that debug output will be invaluable when people (inevitably, for this sort of thing) run into problems.

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -256,3 +258,66 @@ def get_debian_pkg_libdevice():
...
+    system = platform.system()
+    machine = platform.machine()
+
+    if system == "Linux":
+        arch_to_targets = {
+            'x86_64': 'x86_64-linux',
+            'aarch64': 'sbsa-linux'
+        }
+        return arch_to_targets.get(machine)
+
```
gmarkall, line 281: Are we missing the Windows (`"x64"`?) case here?

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -256,3 +258,66 @@ def get_debian_pkg_libdevice():
...
+    return None
+
+
+def get_conda_include_dir():
+    """
+    Return the include directory in the current conda environment, if one
+    is active and it exists.
+    """
+    conda_prefix = os.environ.get('CONDA_PREFIX')
```
gmarkall, line 290: It's not clear to me that Python is always executed with the conda environment active. The other functions in this file, like `get_conda_ctk()`, check for the presence of `conda-meta` under `sys.prefix` to determine if it's running in a conda environment (and to more robustly defend against `CONDA_PREFIX` being something other than where the current Python interpreter is, I suppose). Is there a strong reason for using `CONDA_PREFIX` instead? If not, can we be consistent with the method used for `get_conda_ctk()`, so that we have less chance of getting two different results for the libraries and the includes?

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -932,7 +970,9 @@ def compile(self, sig):
             kernel = _Kernel(self.py_func, argtypes, **self.targetoptions)
             # We call bind to force codegen, so that there is a cubin to cache
             kernel.bind()
-            self._cache.save_overload(sig, kernel)
+            # We only cache kernels that has no external linkage.
+            if len(kernel._codelibrary._linking_files) == 0:
+                self._cache.save_overload(sig, kernel)
```
gmarkall, line 975: This change should be undone now, because we're not linking the NRT for any non-NRT tests, and we're not caching anything that does use NRT in tests (we can defer support for caching functions that use NRT to a later PR).
```suggestion
            self._cache.save_overload(sig, kernel)
```

numba_cuda/numba/cuda/cuda_paths.py
isVoid: Yes, it should be this:
```python
    if system == "Windows":
        arch_to_targets = {
            'AMD64': 'x64'
        }
        return arch_to_targets.get(machine)
```

numba_cuda/numba/cuda/cuda_paths.py
isVoid: > Is there a strong reason for using CONDA_PREFIX instead?

Not really, I didn't know of the `sys.prefix` way to get conda paths.

numba_cuda/numba/cuda/cuda_paths.py
isVoid: I think there's a dilemma here. The desired solution is to have a mechanism to detect if user has specified `NUMBA_CUDA_INCLUDE_PATH` in configuration, if it's not specified, `config.CUDA_INCLUDE_PATH` returns none and uses other defaults. However, `numba.config` doesn't support such behavior and `config.CUDA_INCLUDE_PATH` always has a value.

numba_cuda/numba/cuda/cudadrv/nvrtc.py
isVoid: Due to the issue mentioned [here](https://github.com/NVIDIA/numba-cuda/pull/17#discussion_r1858108925), we probably should not remove the include path here. Because nvrtc can take multiple paths and search them all, the config that user specified through config can still be picked up by rtc.

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -233,12 +232,16 @@ def compile(src, name, cc):
...
+
+    cuda_include = f"-I{get_cuda_paths()['include_dir'].info}"
```
gmarkall, line 236: Following the suggestion above, it sounds like we can simply augment the existing searching of `config.CUDA_INCLUDE_PATH` by making this

```suggestion

    cuda_include = [f"-I{get_cuda_paths()['include_dir'].info}",
                    f"-I{config.CUDA_INCLUDE_PATH}"]
```

numba_cuda/numba/cuda/cudadrv/nvrtc.py
gmarkall: Good point - looks like we can fix this with https://github.com/NVIDIA/numba-cuda/pull/17/files#r1858379047?

numba_cuda/numba/cuda/cuda_paths.py
isVoid: This is added in https://github.com/NVIDIA/numba-cuda/pull/17/commits/b600fa027a3116bcb92f694461e96ac6fd442f31

numba_cuda/numba/cuda/cuda_paths.py
isVoid: Done in https://github.com/NVIDIA/numba-cuda/pull/17/commits/9b6dda9dd16ec678c303c9fcb733de80e5a4c628

numba_cuda/numba/cuda/cuda_paths.py
isVoid: Also fixed in https://github.com/NVIDIA/numba-cuda/pull/17/commits/9b6dda9dd16ec678c303c9fcb733de80e5a4c628

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -256,3 +258,66 @@ def get_debian_pkg_libdevice():
...
+            # though usually it shouldn't.
+            include_dir = os.path.join(conda_prefix, 'include')
+        if os.path.exists(include_dir):
+            return include_dir
+    return None
+
+
+def get_system_include_dir():
+    """Return the system CUDA include directory, if it exists"""
+    system_cuda_include = '/usr/local/cuda/include'
```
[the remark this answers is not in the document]
isVoid: Now returns `None` for non linux platform, in align with `get_system_ctk`. 
[8f3c5f7](https://github.com/NVIDIA/numba-cuda/pull/17/commits/8f3c5f77587a0d1964232edb1ddc7118b9a5ccdd)

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -256,3 +258,75 @@ def get_debian_pkg_libdevice():
...
+    return
+
+
+def get_system_include_dir():
+    """Return the system CUDA include directory, if it exists"""
+    if sys.platform.startswith('linux'):
+        system_cuda_include = '/usr/local/cuda/include'
+        if os.path.exists(system_cuda_include):
+            return system_cuda_include
+    return
```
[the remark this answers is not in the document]
isVoid: added in https://github.com/NVIDIA/numba-cuda/pull/17/commits/8ec02f2770684614107c0751be190a98fdc39870

numba_cuda/numba/cuda/dispatcher.py
isVoid, line 1: I think there are two things that can also be tested here:
1. Can we execute this function with `compute-sanitizer` to make sure there's no leak?
2. Should we examine the PTX / SASS of this function and determine that the code matches certain NRT pattern? (NRT_IncRef is invoked, for example).

gmarkall: 1. I'm not sure about 1, because I don't know whether it checks allocations made on the device or how that works. I'd be inclined to defer investigating compute-sanitizer until a future PR.
2. I think this is a good idea - I expect we should see calls like that in the PTX, for example. I think for this PR, adding an example or two of a test like that is sufficient, so we can get this initial work merged and then use that as a basis for implementing more exhaustive tests.

isVoid: > I think for this PR, adding an example or two of a test like that is sufficient, so we can get this initial work merged and then use that as a basis for implementing more exhaustive tests.

Ok I will add these tests for this PR.

numba_cuda/numba/cuda/cudadrv/libs.py
```diff
@@ -173,4 +191,14 @@ def test():
         print('\tERROR: failed to find %s:\n%s' % (lib, e))
         failed = True
 
+    include = get_cuda_include_dir()
+    # Check cuda include paths
+    try:
+        print('\tChecking include directory', end='...')
```
gmarkall, line 197: This causes the end of the output of `libs.test()` to be:

```
Finding libdevice from Conda environment (NVIDIA package)
	Located at /home/gmarkall/miniforge3/envs/numbadev/nvvm/libdevice/libdevice.10.bc
	Checking library...	ok
	Checking include directory...	ok
```

It's missing a part that should be something like:

```
Finding include directory from <method>
        Located at <location>
```

so that the output would look like at the end:

```
Finding libdevice from Conda environment (NVIDIA package)
	Located at /home/gmarkall/miniforge3/envs/numbadev/nvvm/libdevice/libdevice.10.bc
	Checking library...	ok
Finding include directory from Conda environment
        Located at /home/gmarkall/miniforge3/envs/numbadev/targets/x86_64-linux/include
	Checking include directory...	ok
```

gmarkall: It might also be handy to print the value of `CUDA_INCLUDE_PATH` here, so that all the information is available to the user (even if we don't test for anything in that location)

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -233,12 +233,19 @@ def compile(src, name, cc):
     #   being optimized away.
     major, minor = cc
     arch = f'--gpu-architecture=compute_{major}{minor}'
-    include = f'-I{config.CUDA_INCLUDE_PATH}'
+
+    cuda_include = [
+        f"-I{get_cuda_paths()['include_dir'].info}",
+        f"-I{config.CUDA_INCLUDE_PATH}"
```
gmarkall, line 239: I think the logic here might need a bit more refinement. I'm thinking about the situation where a user has set `config.CUDA_INCLUDE_PATH`, and has a system toolkit in the default location, but does not have a toolkit installed via conda. In this case, I think the toolkit in the system default location will always be preferred, because it will come first on the path.

Maybe it would simplify things if there were only two ways to find the includes:

- The conda environment (added in this PR)
- From `config.CUDA_INCLUDE_PATH` (already existing)

and not add new functionality to try to find the toolkit from the system default path in this PR?

isVoid: Ah, this makes sense. The config default is the same as system default. So when we don't find conda CTK, we fall back to system default, which is fair solution.

isVoid: This is changed in [1dcfd78](https://github.com/NVIDIA/numba-cuda/pull/17/commits/1dcfd78cd93c8214faf28cea0362cbf2e5c7736f)

numba_cuda/numba/cuda/dispatcher.py
isVoid: Added in https://github.com/NVIDIA/numba-cuda/pull/17/commits/f769cee052ea83a41ee59cac479927a46e949bda

numba_cuda/numba/cuda/cudadrv/libs.py
isVoid: Fixed in https://github.com/NVIDIA/numba-cuda/pull/17/commits/6612e0d49c51e22a1b884fd97b6cc7f51e49c2ae

PR #23, opened 2024-07-29
Use `pynvjitlink` for MVC 

brandon-b-miller:
This PR attempts to move some of the logic inside the `pynvjitlink` [patch.py](https://github.com/rapidsai/pynvjitlink/blob/main/pynvjitlink/patch.py) to work behind `config.CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY` such that numba may perform the patch if necessary rather than `pynvjitlink` itself.


numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2594,9 +2593,18 @@ class Linker(metaclass=ABCMeta):
...
 
     @classmethod
-    def new(cls, max_registers=0, lineinfo=False, cc=None):
+    def new(cls, max_registers=0, lineinfo=False, cc=None, **kwargs):
         if config.CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY:
-            return MVCLinker(max_registers, lineinfo, cc)
+            # TODO: circular
+            from . import runtime
+            driver_ver, runtime_ver = driver.get_version(), runtime.get_version()
+            if driver_ver >= (12, 0) and runtime_ver > driver_ver:
```
gmarkall, line 2601: Rather than this logic, I think we should instead allow the user to specifically configure Numba to use pynvjitlink. I think we could add pynvjitlink as a dependency and think about automatically using it after the introduction of this PR.

gmarkall (changes requested):
Just had a quick skim and left some initial comments / thoughts.

This would also need the Numba tests from pynvjitlink porting in.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2952,6 +2960,217 @@ def complete(self):
...
+
+class LTOIR(LinkableCode):
+    """An LTOIR file in memory"""
+
+    kind = "ltoir"
+    default_name = "<unnamed-ltoir>"
+
+
+@functools.lru_cache(maxsize=1)
+def patch_cuda():
```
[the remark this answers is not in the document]
brandon-b-miller: ah, thanks for the pointer here, I had not realized that these attributes are being created rather than replaced. I will update this.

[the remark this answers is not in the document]
brandon-b-miller: I moved a little code around in the most recent commits to do this hopefully cleanly, curious what you think

numba_cuda/numba/cuda/cudadrv/driver.py
brandon-b-miller: Do you mean exposing pynvjitlink as opt in/out through the config? Or do you mean setting things up so that if pynvjitlink is present in the environment, then use it?

gmarkall: I meant exposing it as opt in/out through the config.

brandon-b-miller: I see. Would this flag have some overlap with `CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY` then, or what would be the interplay between those two? If `CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY=False` then would `USE_PYNVJITLINK` (?) have no effect?

gmarkall: To try and minimise complexity / confusion, I'd go with:

- Keep `CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY` for CUDA 11
- Add a new flag for enabling the pynvjitlink linker
- Raise an error if `CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY` is set for CUDA 12
- Raise an error if the pynvjitlink linker is enabled for CUDA 11

The reasoning is that pynvjitlink enablement is not a 1:1 mapping with supporting or not supporting MVC - it's also to support LTO, so having a flag relating to behaviour you want (enabling MVC) controlling a component that has multiple behaviours seems like an impedance mismatch - therefore for now I'd prefer if the user said what was to happen (enable / disable pynvjitlink) rather than stating what behaviour they want.

I started making a spreadsheet containing all the configurations (CUDA 11, 12, a future 13 that's not compatible with pynvjitlink from 12, etc) x (driver </=/> runtime version) x (LTO requested) x (MVC needed) x (hardware supported by pynvjitlink version) and a summary of which are valid but the complexity got a bit out of control for a quick exercise, so I'm really intent on "have the user say what to do" ahead of "try and choose what to do based on the user's intent" for now.

gmarkall (changes requested):
Thanks for the updates - I just had another pass over this and added some comments on the diff.

numba_cuda/numba/cuda/cudadrv/driver.py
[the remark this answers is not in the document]
gmarkall: The places where the code are looks good to me.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -55,6 +58,44 @@
...
+            "The parse failed with exception:\n"
+            f"{traceback.format_exc()}",
+            RuntimeWarning
+        )
+        return default
+
+
+config.ENABLE_PYNVJITLINK = False
+if _readenv("ENABLE_PYNVJITLINK", bool, False):
+    config.ENABLE_PYNVJITLINK = True
```
gmarkall, line 87: Doing this here causes the config variable to be overwritten if the user set it prior to importing `numba.cuda`. e.g.:

```python
In [1]: from numba.core import config

In [2]: config.ENABLE_PYNVJITLINK = True

In [3]: from numba import cuda

In [4]: config.ENABLE_PYNVJITLINK
Out[4]: False
```

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -55,6 +58,44 @@
...
+
+
+_MVC_ERROR_MESSAGE_CU11 = (
+    "Minor version compatibility requires ptxcompiler and cubinlinker packages "
+    "to be available"
+)
+
+_MVC_ERROR_MESSAGE_CU12 = (
+    "Minor version compatibility requires pynvjitlink package to be available"
+)
```
gmarkall, line 97: This might be confusing if the user requested pynvjitlink in order to use LTO - perhaps "Using pynvjitlink requires the pynvjitlink package to be available" would align more closely with the request that can't be satisfied?

gmarkall (changes requested):
Thanks for the updates - I added a couple more comments on the diff.

numba_cuda/numba/cuda/cudadrv/driver.py
brandon-b-miller: Added some logic that should reflect this.

brandon-b-miller:
> Thanks for the updates - I added a couple more comments on the diff.

Thank you for the review :) I've made some changes that hopefully address your comments.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2626,19 +2686,38 @@ def add_cu_file(self, path):
...
-        elif ext == 'cu':
-            self.add_cu_file(path)
-        else:
-            kind = FILE_EXTENSION_MAP.get(ext, None)
-            if kind is None:
-                raise RuntimeError("Don't know how to link file with extension "
-                                   f".{ext}")
-            self.add_file(path, kind)
+        if isinstance(path_or_code, str):
+            ext = pathlib.Path(path_or_code).suffix
```
gmarkall, line 2692: Are we missing a bit of logic for handling `.ltoir` here? Or has it gone somewhere else? The original I'm looking at is https://github.com/rapidsai/pynvjitlink/blob/a2f23b7c3c237f2cdde3093c845e0453572503eb/pynvjitlink/patch.py#L170-L171

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2930,6 +3003,113 @@ def complete(self):
...
+        ptx, log = nvrtc.compile(cu, name, cc)
+
+        if config.DUMP_ASSEMBLY:
+            print(("ASSEMBLY %s" % name).center(80, "-"))
+            print(ptx)
+            print("=" * 80)
+
+        # Link the program's PTX using the normal linker mechanism
+        ptx_name = os.path.splitext(name)[0] + ".ptx"
+        self.add_ptx(ptx.encode(), ptx_name)
```
gmarkall, line 3103: This looks like a duplicate of the `add_cu()` method from the base `Linker` class - is it needed?

numba_cuda/numba/cuda/cudadrv/enums.py
```diff
@@ -309,7 +309,10 @@
 # Applicable options: PTX compiler options, ::CU_JIT_FALLBACK_STRATEGY
 CU_JIT_INPUT_LIBRARY = 4
 
-CU_JIT_NUM_INPUT_TYPES = 6
+# LTO IR
+CU_JIT_INPUT_LTO_IR = 5
+
+CU_JIT_NUM_INPUT_TYPES = 7
```
gmarkall, line 315: This should not be bumped - there are still 6 input types (it's just that 5 (LTO-IR) was never used before so it didn't appear in Numba)

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -55,6 +58,43 @@
...
+            f"value '{value}' could not be parsed.\n"
+            "The parse failed with exception:\n"
+            f"{traceback.format_exc()}",
+            RuntimeWarning
+        )
+        return default
+
+
+if _readenv("ENABLE_PYNVJITLINK", bool, False):
+    config.ENABLE_PYNVJITLINK = True
```
gmarkall, line 86: Rather than having the `try` / `except ImportError` above, I think we could just try importing `pynvjitlink` here and reporting the error immediately to the user if it can't be imported.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2546,38 +2586,58 @@ def launch_kernel(cufunc_handle,
...
+            )
+        if (
+            config.CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY
+            and config.ENABLE_PYNVJITLINK
+        ):
+            raise ValueError(
+                "can't set both config.ENABLE_PYNVJITLINK "
+                "and config.CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY "
+                "at the same time"
+            )
```
gmarkall, line 2622: I think this check could be pushed up to the top of the file around line 86 where the check for pynvjitlink being enabled / imported is - this would allow us to report a setup issue to the user quickly instead of waiting for them to compile / link something.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2930,6 +3003,113 @@ def complete(self):
...
+    def __init__(
+        self,
+        max_registers=None,
+        lineinfo=False,
+        cc=None,
+        lto=False,
+        additional_flags=None,
+    ):
+        if pynvjitlink_import_err is not None:
+            raise ImportError(_MVC_ERROR_MESSAGE_CU12)
```
gmarkall, line 3016: I think this could be removed / moved to the check for whether pynvjitlink should be enabled above, so we report a setup / config error to the user right away.

gmarkall (changes requested):
Thanks for the changes - I've gone through another round and have some more thoughts on the diff. CI is failing because the logic for setting `config.ENABLE_PYNVJITLINK` is still not quite right.

We also need to port in the tests from pynvjitlink - I suspect there's at least one logic error (to do with adding LTOIR from a file) which the added tests might catch.

numba_cuda/numba/cuda/cudadrv/driver.py
gmarkall: Comment links:
- https://github.com/NVIDIA/numba-cuda/pull/23/files#r1726887376
- https://github.com/NVIDIA/numba-cuda/pull/23/files#r1726884827
- https://github.com/NVIDIA/numba-cuda/pull/23/files#r1726885572

numba_cuda/numba/cuda/cudadrv/mappings.py
```diff
@@ -0,0 +1,23 @@
...
+from . import enums
+if config.CUDA_USE_NVIDIA_BINDING:
+    jitty = cuda.CUjitInputType
+    FILE_EXTENSION_MAP = {
+        'o': jitty.CU_JIT_INPUT_OBJECT,
+        'ptx': jitty.CU_JIT_INPUT_PTX,
+        'a': jitty.CU_JIT_INPUT_LIBRARY,
+        'lib': jitty.CU_JIT_INPUT_LIBRARY,
+        'cubin': jitty.CU_JIT_INPUT_CUBIN,
+        'fatbin': jitty.CU_JIT_INPUT_FATBINARY,
```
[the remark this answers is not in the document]
brandon-b-miller: By my reading of [the docs](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1gc78e5cb421c428676861189048888958) I think this could be CU_JIT_INPUT_NVVM. Maybe `nvJitLinkInputType.NVJITLINK_INPUT_LTOIR` is the correct thing though? I am not sure this is exposed but if not I can see what might need to be done to expose it.

numba_cuda/numba/cuda/cudadrv/enums.py
```diff
@@ -309,6 +309,9 @@
 # Applicable options: PTX compiler options, ::CU_JIT_FALLBACK_STRATEGY
 CU_JIT_INPUT_LIBRARY = 4
 
+# LTO IR
+CU_JIT_INPUT_LTO_IR = 5
```
gmarkall, line 313: Technically this should have been `CU_JIT_INPUT_NVVM` as it's the value from the `CUjitInputType_enum` in cuda.h

numba_cuda/numba/cuda/cudadrv/mappings.py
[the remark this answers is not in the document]
gmarkall: You're right, it should be `CU_JIT_INPUT_NVVM` here - this mapping is using the enum from the driver API's `CUjitInputType`. However in `enums.py` the value for this is called `CU_JIT_INPUT_LTO_IR` in this PR at present. I don't mind which name we use for it as long as we're consistent.

 `nvJitLinkInputType.NVJITLINK_INPUT_LTOIR` is from a different enum, so even though it feels like it would be more correct for the nvjitlink case, we need to be consistent across all the linkers and use the driver enum values.

numba_cuda/numba/cuda/cudadrv/mappings.py
```diff
@@ -0,0 +1,24 @@
+from numba import config
+from cuda import cuda
```
gmarkall, line 2: I just noticed from the (failing) docs build that this requires CUDA Python to be installed - it needs to go inside the `if config.CUDA_USE_NVIDIA_BINDING` below.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2626,19 +2684,42 @@ def add_cu_file(self, path):
...
-        elif ext == 'cu':
-            self.add_cu_file(path)
+    def add_file_guess_ext(self, path_or_code):
+        """
+        Add a file or LinkableCode object to the link. If a file is
+        passed, the type will be inferred from the extension. A LinkableCode
+        object represents a file already in memory.
+        """
+        if isinstance(path_or_code, str):
+            ext = pathlib.Path(path_or_code).suffix.lstrip('.')
```
gmarkall, line 2694: What was the `lstrip()` intended to address? (It seems to have introduced some fails?)

gmarkall (changes requested):
As far as I can tell, the code looks good. Just needs tests adding (I think these can be copied from / inspired by those in pynvjitlink)

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2930,6 +3005,96 @@ def complete(self):
...
+
+        try:
+            fn(data, name)
+        except NvJitLinkError as e:
+            raise LinkerError from e
+
+    def complete(self):
+        try:
+            cubin = self._linker.get_linked_cubin()
+            self._linker._complete = True
```
gmarkall, line 3093: I just noticed whilst looking at #48 I think this is superfluous because pynvjitlink sets it itself when you get the linked cubin: https://github.com/rapidsai/pynvjitlink/blob/main/pynvjitlink/api.py#L85

brandon-b-miller:
https://github.com/NVIDIA/numba-cuda/pull/23/commits/7c384a3b202f8e2e93514edb48d5c372b70a5f74 adds the relevant tests from pynvjitlink, which I suppose for a short time may exist in both places while we deprecate the patching api in pynvjitlink itself. 

It occurs to me that the tests still must be built on the CI test jobs if we'd like to test them there, which now involves compiling a few files with nvcc. This is a problem I still need to solve in this PR I think.

ci/test_conda.sh
```diff
@@ -47,8 +49,13 @@ EXITCODE=0
 trap "EXITCODE=1" ERR
 set +e
 
+if [ "$USE_PYNVJITLINK" == true ]; then
+    rapids-logger "Install pynvjitlink"
+    conda install -c rapidsai pynvjitlink
```
gmarkall, line 54: I think you need to temporarily allow unbound locals for conda operations:
```suggestion
    set +u
    conda install -c rapidsai pynvjitlink
    set -u
```

brandon-b-miller:
The tests now pass, modulo what appears to be an intermittent error on the conda pynvjitlink test job. I do not seem to have permissions to rerun this job.

ci/test_conda_pynvjitlink.sh
```diff
@@ -0,0 +1,77 @@
...
+python -m numba --sysinfo
+
+EXITCODE=0
+trap "EXITCODE=1" ERR
+set +e
+
+
+rapids-logger "Install pynvjitlink"
+set +u
+conda install -c rapidsai pynvjitlink
```
gmarkall, line 53: In order to avoid intermittent failures, I think this can be:
```suggestion
rapids-mamba-retry install -c rapidsai pynvjitlink
```

which has logic to handle all sorts of transient connection issues and retry the install until success.

gmarkall: (Actually I'm going to commit this so the tests will re-run)

gmarkall:
Running the testsuite on my local system seems to be deadlocking at:

```
test_lock (numba.cuda.tests.cudapy.test_dispatcher.TestDispatcher.test_lock)
Test that (lazy) compiling from several threads at once doesn't ...
```

when pynvjitlink is enabled:

```
ENABLE_PYNVJITLINK=1 python -m numba.runtests numba.cuda.tests -v -m
```

gmarkall:
Threads:
```
(gdb) info threads
  Id   Target Id                                            Frame 
* 1    Thread 0x70df86a90740 (LWP 128088) "python"          __futex_abstimed_wait_common64 (private=<optimised out>, 
    cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x70de98395330) at ./nptl/futex-internal.c:57
  2    Thread 0x70df83600640 (LWP 128089) "python"          __futex_abstimed_wait_common64 (private=0, cancel=true, 
    abstime=0x0, op=393, expected=0, futex_word=0x70df859ceb60 <thread_status+96>) at ./nptl/futex-internal.c:57
  3    Thread 0x70df82c00640 (LWP 128090) "python"          __futex_abstimed_wait_common64 (private=0, cancel=true, 
    abstime=0x0, op=393, expected=0, futex_word=0x70df859cebe0 <thread_status+224>) at ./nptl/futex-internal.c:57
  4    Thread 0x70df82200640 (LWP 128091) "python"          __futex_abstimed_wait_common64 (private=0, cancel=true, 
    abstime=0x0, op=393, expected=0, futex_word=0x70df859cec60 <thread_status+352>) at ./nptl/futex-internal.c:57
  5    Thread 0x70df81800640 (LWP 128092) "python"          __futex_abstimed_wait_common64 (private=0, cancel=true, 
    abstime=0x0, op=393, expected=0, futex_word=0x70df859cece0 <thread_status+480>) at ./nptl/futex-internal.c:57
  6    Thread 0x70df80e00640 (LWP 128093) "python"          __futex_abstimed_wait_common64 (private=0, cancel=true, 
    abstime=0x0, op=393, expected=0, futex_word=0x70df859ced60 <thread_status+608>) at ./nptl/futex-internal.c:57
  7    Thread 0x70df78400640 (LWP 128094) "python"          __futex_abstimed_wait_common64 (private=0, cancel=true, 
    abstime=0x0, op=393, expected=0, futex_word=0x70df859cede0 <thread_status+736>) at ./nptl/futex-internal.c:57
  8    Thread 0x70df77a00640 (LWP 128095) "python"          __futex_abstimed_wait_common64 (private=0, cancel=true, 
    abstime=0x0, op=393, expected=0, futex_word=0x70df859cee60 <thread_status+864>) at ./nptl/futex-internal.c:57
  9    Thread 0x70df67000640 (LWP 128096) "python"          __futex_abstimed_wait_common64 (private=0, cancel=true, 
    abstime=0x0, op=393, expected=0, futex_word=0x70df859ceee0 <thread_status+992>) at ./nptl/futex-internal.c:57
  10   Thread 0x70df66600640 (LWP 128097) "python"          __futex_abstimed_wait_common64 (private=0, cancel=true, 
    abstime=0x0, op=393, expected=0, futex_word=0x70df859cef60 <thread_status+1120>) at ./nptl/futex-internal.c:57
  11   Thread 0x70df45c00640 (LWP 128098) "python"          __futex_abstimed_wait_common64 (private=0, cancel=true, 
    abstime=0x0, op=393, expected=0, futex_word=0x70df859cefe0 <thread_status+1248>) at ./nptl/futex-internal.c:57
  12   Thread 0x70df45200640 (LWP 128099) "python"          __futex_abstimed_wait_common64 (private=0, cancel=true, 
    abstime=0x0, op=393, expected=0, futex_word=0x70df859cf060 <thread_status+1376>) at ./nptl/futex-internal.c:57
  13   Thread 0x70df34800640 (LWP 128100) "python"          __futex_abstimed_wait_common64 (private=0, cancel=true, 
    abstime=0x0, op=393, expected=0, futex_word=0x70df859cf0e0 <thread_status+1504>) at ./nptl/futex-internal.c:57
  14   Thread 0x70df33e00640 (LWP 128101) "python"          __futex_abstimed_wait_common64 (private=0, cancel=true, 
    abstime=0x0, op=393, expected=0, futex_word=0x70df859cf160 <thread_status+1632>) at ./nptl/futex-internal.c:57
  15   Thread 0x70df23400640 (LWP 128102) "python"          __futex_abstimed_wait_common64 (private=0, cancel=true, 
    abstime=0x0, op=393, expected=0, futex_word=0x70df859cf1e0 <thread_status+1760>) at ./nptl/futex-internal.c:57
  16   Thread 0x70df12a00640 (LWP 128103) "python"          __futex_abstimed_wait_common64 (private=0, cancel=true, 
    abstime=0x0, op=393, expected=0, futex_word=0x70df859cf260 <thread_status+1888>) at ./nptl/futex-internal.c:57
  17   Thread 0x70df0a000640 (LWP 128104) "python"          __futex_abstimed_wait_common64 (private=0, cancel=true, 
    abstime=0x0, op=393, expected=0, futex_word=0x70df859cf2e0 <thread_status+2016>) at ./nptl/futex-internal.c:57
  18   Thread 0x70df01600640 (LWP 128105) "python"          __futex_abstimed_wait_common64 (private=0, cancel=true, 
    abstime=0x0, op=393, expected=0, futex_word=0x70df859cf360 <thread_status+2144>) at ./nptl/futex-internal.c:57
  19   Thread 0x70def8c00640 (LWP 128106) "python"          __futex_abstimed_wait_common64 (private=0, cancel=true, 
    abstime=0x0, op=393, expected=0, futex_word=0x70df859cf3e0 <thread_status+2272>) at ./nptl/futex-internal.c:57
  20   Thread 0x70def0200640 (LWP 128107) "python"          __futex_abstimed_wait_common64 (private=0, cancel=true, 
    abstime=0x0, op=393, expected=0, futex_word=0x70df859cf460 <thread_status+2400>) at ./nptl/futex-internal.c:57
  21   Thread 0x70ded2200640 (LWP 128117) "cuda00001800007" 0x000070df86918bcf in __GI___poll (fds=0x60aeacfb6e10, 
    nfds=3, timeout=-1) at ../sysdeps/unix/sysv/linux/poll.c:29
  22   Thread 0x70dec3800640 (LWP 129038) "python"          __futex_abstimed_wait_common64 (private=0, cancel=true, 
    abstime=0x0, op=393, expected=0, futex_word=0x60aead59e718) at ./nptl/futex-internal.c:57
  23   Thread 0x70debaa00640 (LWP 129039) "cuda-EvtHandlr"  0x000070df86918bcf in __GI___poll (fds=0x70debc001cf0, 
    nfds=10, timeout=100) at ../sysdeps/unix/sysv/linux/poll.c:29
  24   Thread 0x70dec4200640 (LWP 129045) "python"          futex_wait (private=0, expected=2, 
    futex_word=0x60aeaccc5d58) at ../sysdeps/nptl/futex-internal.h:146
  25   Thread 0x70debb400640 (LWP 129046) "python"          __futex_abstimed_wait_common64 (private=-1153439408, 
    cancel=true, abstime=0x70debb3fea70, op=137, expected=0, futex_word=0x60aeaaeac194 <_PyRuntime+436>)
    at ./nptl/futex-internal.c:57
  26   Thread 0x70deaec00640 (LWP 129047) "python"          __futex_abstimed_wait_common64 (private=<optimised out>, 
    cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x60aeacd6aea0) at ./nptl/futex-internal.c:57
  27   Thread 0x70dead600640 (LWP 129048) "python"          __futex_abstimed_wait_common64 (private=<optimised out>, 
    cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x60aeacd6aea0) at ./nptl/futex-internal.c:57
  28   Thread 0x70deacc00640 (LWP 129049) "python"          __futex_abstimed_wait_common64 (private=<optimised out>, 
    cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x60aeacd6aea0) at ./nptl/futex-internal.c:57
  29   Thread 0x70de9fe00640 (LWP 129050) "python"          __futex_abstimed_wait_common64 (private=<optimised out>, 
    cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x60aeacd6aea0) at ./nptl/futex-internal.c:57
  30   Thread 0x70de9f400640 (LWP 129051) "python"          __futex_abstimed_wait_common64 (private=<optimised out>, 
    cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x60aeacd6aea0) at ./nptl/futex-internal.c:57
  31   Thread 0x70de9ea00640 (LWP 129052) "python"          __futex_abstimed_wait_common64 (private=<optimised out>, 
    cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x60aeacd6aea0) at ./nptl/futex-internal.c:57
  32   Thread 0x70de9e000640 (LWP 129053) "python"          __futex_abstimed_wait_common64 (private=<optimised out>, 
    cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x60aeacd6aea0) at ./nptl/futex-internal.c:57
  33   Thread 0x70de9d600640 (LWP 129054) "python"          __futex_abstimed_wait_common64 (private=<optimised out>, 
    cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x60aeacd6aea0) at ./nptl/futex-internal.c:57
  34   Thread 0x70de9cc00640 (LWP 129055) "python"          __futex_abstimed_wait_common64 (private=<optimised out>, 
    cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x60aeacd6aea0) at ./nptl/futex-internal.c:57
  35   Thread 0x70de97e00640 (LWP 129056) "python"          __futex_abstimed_wait_common64 (private=<optimised out>, 
    cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x60aeacd6aea0) at ./nptl/futex-internal.c:57
  36   Thread 0x70de97400640 (LWP 129057) "python"          __futex_abstimed_wait_common64 (private=<optimised out>, 
    cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x60aeacd6aea0) at ./nptl/futex-internal.c:57
  37   Thread 0x70de96a00640 (LWP 129058) "python"          __futex_abstimed_wait_common64 (private=<optimised out>, 
    cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x60aeacd6aea0) at ./nptl/futex-internal.c:57
  38   Thread 0x70de93e00640 (LWP 129059) "python"          __futex_abstimed_wait_common64 (private=<optimised out>, 
    cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x60aeacd6aea0) at ./nptl/futex-internal.c:57
  39   Thread 0x70de93400640 (LWP 129060) "python"          __futex_abstimed_wait_common64 (private=<optimised out>, 
    cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x60aeacd6aea0) at ./nptl/futex-internal.c:57
```

brandon-b-miller:
Connected offline with @gmarkall , we concluded the above is probably an issue in `nvJitLink`/`pynvjitlink` rather than here.

gmarkall:
> Connected offline with @gmarkall , we concluded the above is probably an issue in `nvJitLink`/`pynvjitlink` rather than here.

It was a combination of mismatched versions (cudadevrt from 12.6 with nvJitLink from 12.5) and nvJitLink not handling this situation gracefully.

gmarkall (approved):
I think this is mergeable. There are some small fixups that I'd like, but would rather push them into a new PR rather than fiddling any more with this, because it's been a heroic effort over a long period of time.

It should also me merged into `main` rather than `develop`, so I've rebased it and it's in #56 - over there I'm just waiting for CI, then will merge and follow up with the fixes.

gmarkall:
Closing as #56 has just been merged.

PR #34, opened 2024-07-30
Fix name of array used in pseudocode to explain how atomic.dec works

brycelelbach:
Docs only.


gmarkall:
I just merged `develop` so that the CI run tests that the docs build OK (and I can inspect the docs build from it).

gmarkall:
Many thanks!

PR #48, opened 2024-08-23
Allow JIT Compile to, and Link from, LTOIR for cuda source input

isVoid:
Depends on #23 

This PR adds the functionality supporting kernel and FFI functions being JIT Compile to and link from LTOIR, allowing better optimization when foreign function is used in Numba-cuda.


numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -3066,6 +3066,32 @@ def add_file(self, path, kind):
...
+            dev = driver.get_device(ac.devnum)
+            cc = dev.compute_capability
+
+        program, log = nvrtc.compile(cu, name, cc, ltoir=self.lto)
+
+        if not self.lto and config.DUMP_ASSEMBLY:
+            # TODO: when linker is configured to generate LTOIR, assembly is not
+            # directly visible via the Linker API. We should provide `-ptx` as
+            # an additional flag to nvjitlink to generate PTX. However, this
+            # could break the linking pipeline. We need to investigate this further.
```
gmarkall, line 3082: To get PTX when doing LTO, [`get_cubin()`](https://github.com/NVIDIA/numba-cuda/blob/main/numba_cuda/numba/cuda/codegen.py#L174) will need to run the linker twice, once without `"-ptx"` (as it does now) and once again with `"-ptx"` (additionally) to get the PTX - it can then dump it in a similar way to how it's dumped here.

gmarkall (changes requested):
As a set of changes on top of #23 this looks good - if you can add the logic to dump assembly correctly when using LTO (and some tests) I think that should be enough for this PR. Do you see anything else that might be needed?

isVoid:
> Do you see anything else that might be needed?

Without looking very deep into the code base, where is `Linker(lto=True)` set in the pipeline? Are they enabled by default through `@cuda.jit`?

> As a set of changes on top of #23 this looks good - if you can add the logic to dump assembly correctly when using LTO (and some tests) I think that should be enough for this PR. 

Ok sounds good.

numba_cuda/numba/cuda/codegen.py
```diff
@@ -192,7 +184,28 @@ def get_cubin(self, cc=None):
...
+        if cubin:
+            return cubin
+
+        if config.DUMP_ASSEMBLY:
+            linker = driver.Linker.new(max_registers=self._max_registers, cc=cc, additional_flags=["-ptx"])
+            self._link_all(linker, cc)
+            ptx = linker.get_linked_ptx().decode('utf-8')
+
+            if config.DUMP_ASSEMBLY:
+                print(("ASSEMBLY (AFTER LTO) %s" % self._name).center(80, '-'))
```
isVoid, line 200: If `dump_assembly=1`, the pipeline will dump PTX twice: once before LTO, and once after LTO. I think it might be helpful to keep both results. So I added `(AFTER LTO)` here for clarity.

gmarkall:
> Without looking very deep into the code base, where is `Linker(lto=True)` set in the pipeline? Are they enabled by default through `@cuda.jit`?

Good point, I think it's missing - I think when #23 adds tests that would have become apparent, and it will be required.

isVoid:
> > Without looking very deep into the code base, where is `Linker(lto=True)` set in the pipeline? Are they enabled by default through `@cuda.jit`?
> 
> Good point, I think it's missing - I think when #23 adds tests that would have become apparent, and it will be required.

In [159e77e](https://github.com/NVIDIA/numba-cuda/pull/48/commits/159e77e57cf167667200010761d62b85e29d268e) I added the line to make lto enabled by default.

gmarkall:
Closing in favour of #62, which is based on a recent `main`.

PR #50, opened 2024-10-03
Elide kernel wrappers by altering device function IR

gmarkall:


gmarkall:
Test successfully on Windows.

PR #51, opened 2024-10-03
Add support for cache-hinted load and store operations

gmarkall:
To-do:

- [ ] Add documentation.
- [ ] Test cases for erroneous arguments. It would be good to check for accidental use on shared or local arrays, but this may not be easy to do.
- [ ] Add additional validations as described in comments in `ld_cache_operator` and `st_cache_operator`.
- [ ] Refactor the implementation - the load and store implementations contain a lot of common code.
- [ ] Decide on whether to support complex, and work out why it presently doesn't work.

### DIFF
--- a/numba_cuda/numba/cuda/cache_hints.py
+++ b/numba_cuda/numba/cuda/cache_hints.py
@@ -5,6 +5,10 @@ from numba.core.extending import intrinsic, overload
 from numba.core.errors import NumbaTypeError
 from numba.cuda.api_util import normalize_indices
 
+# Docs references:
+# https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#data-movement-and-conversion-instructions-ld
+# https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#load-functions-using-cache-hints
+
 
 def ldca(array, i):
     """Generate a `ld.global.ca` instruction for element `i` of an array."""
--- a/numba_cuda/numba/cuda/cache_hints.py
+++ b/numba_cuda/numba/cuda/cache_hints.py
@@ -1,0 +1,226 @@
+from llvmlite import ir
+from numba import types
+from numba.core import cgutils
+from numba.core.extending import intrinsic, overload
+from numba.core.errors import NumbaTypeError
+from numba.cuda.api_util import normalize_indices
+
+
+def ldca(array, i):
+    """Generate a `ld.global.ca` instruction for element `i` of an array."""
+
+
+def ldcg(array, i):
+    """Generate a `ld.global.cg` instruction for element `i` of an array."""
+
+
+def ldcs(array, i):
+    """Generate a `ld.global.cs` instruction for element `i` of an array."""
+
+
+def ldlu(array, i):
+    """Generate a `ld.global.lu` instruction for element `i` of an array."""
+
+
+def ldcv(array, i):
+    """Generate a `ld.global.cv` instruction for element `i` of an array."""
+
+
+def stcg(array, i, value):
+    """Generate a `st.global.cg` instruction for element `i` of an array."""
+
+
+def stcs(array, i, value):
+    """Generate a `st.global.cs` instruction for element `i` of an array."""
+
+
+def stwb(array, i, value):
+    """Generate a `st.global.wb` instruction for element `i` of an array."""
+
+
+def stwt(array, i, value):
+    """Generate a `st.global.wt` instruction for element `i` of an array."""
+
+
+def ld_cache_operator(operator):
+    @intrinsic
+    def impl(typingctx, array, index):
+        if not isinstance(array, types.Array):
+            msg = f"ldcs operates on arrays. Got type {array}"
+            raise NumbaTypeError(msg)
+
+        # Need to validate bitwidth
+
+        # Need to validate indices
+
+        signature = array.dtype(array, index)
+
+        def codegen(context, builder, sig, args):
+            array_type, index_type = sig.args
+            loaded_type = context.get_value_type(array_type.dtype)
+            ptr_type = loaded_type.as_pointer()
+            ldcs_type = ir.FunctionType(loaded_type, [ptr_type])
+
+            array, indices = args
+
+            index_type, indices = normalize_indices(context, builder,
+                                                    index_type, indices,
+                                                    array_type,
+                                                    array_type.dtype)
+            array_struct = context.make_array(array_type)(context, builder,
+                                                          value=array)
+            ptr = cgutils.get_item_pointer(context, builder, array_type,
+                                           array_struct, indices,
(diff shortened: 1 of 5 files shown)


gmarkall:
A note: these also need to work for `CPointer()` types as well as arrays.

gmarkall:
@kaeun97 I started this a while back but never managed to address all the to-do items. Essentially the PR was aimed at adding new functions for cache-hinted load and store operations, but I didn't get the implementation and error handling to be cleaner than the current "prototype" like form. It might be interesting as it is another example of adding new code generation to Numba-CUDA.

What do you think of the PR? Would it be of any interest to you to take it on and complete the to-do list items?

kaeun97:
@gmarkall Happy to continue the work :) Thank you!

gmarkall:
@kaeun97 Thanks! Unfortunately I don't think I have a way to give you permissions to push to this branch / PR, but please feel to open a new PR for the continuation; I'll close this one at that point.

numba_cuda/numba/cuda/cache_hints.py
```diff
@@ -0,0 +1,241 @@
...
+            msg = f"Expected {array.ndim} indices, got a scalar"
+            raise NumbaTypeError(msg)
+        valid_index = True
+
+    if isinstance(index, types.UniTuple):
+        if index.count != array.ndim:
+            msg = f"Expected {array.ndim} indices, got {index.count}"
+            raise NumbaTypeError(msg)
+
+        if all([isinstance(t, types.Integer) for t in index.dtype]):
```
cpcloud, line 82:
```suggestion
        if all(isinstance(t, types.Integer) for t in index.dtype):
```

No reason to create a list if you don't need to.

numba_cuda/numba/cuda/cache_hints.py
```diff
@@ -0,0 +1,241 @@
...
+stcs_intrinsic = st_cache_operator("cs")
+stwb_intrinsic = st_cache_operator("wb")
+stwt_intrinsic = st_cache_operator("wt")
+
+
+@overload(ldca, target='cuda')
+def ol_ldca(array, i):
+    def impl(array, i):
+        return ldca_intrinsic(array, i)
+    return impl
```
cpcloud, line 185: Can't this just be:

```suggestion
	return lcda_intrinsic
```

for each of these?

cpcloud (commented):
Looks good. TIL about these instructions.

numba_cuda/numba/cuda/cache_hints.py
gmarkall: An overload should return a Python function that gets jit-compiled to become the implementation of the function it's overloading. `ldca_intrinsic` is an intrinsic, called from within the compiled function.

I suspect it will not work to return an intrinsic as an overload implementation, but even if it did, it would feel jarring to me to contract a level of abstraction in the implementation here.

cpcloud: It does seem really weird to my eye that a function `foo(*args)` whose only line is to call `bar(*args)` cannot simply be replaced with the call to `bar(*args)`.

This feels like a numbaism perhaps that violates basic substitution rules, but this is of course not a blocking comment.

numba_cuda/numba/cuda/cache_hints.py
kaeun97: Will apply this change in the follow up PR :)

numba_cuda/numba/cuda/cache_hints.py
gmarkall: I think the confusion here comes from reading the code as if it's going to be interpreted by the Python interpreter, rather than seeing it as a form of metaprogramming, which is what functions decorated with `@overload` implement.

An attempt to summarise the pertinent points:

* An `@overload` function returns a Python function that Numba-CUDA compiles.
* An `@intrinsic` function (like `ldca_intrinsic`, generated by [`ld_cache_operator()`](https://github.com/NVIDIA/numba-cuda/pull/51/files#diff-b68f1312d551bef433272d305aa3a378d323ad3fbfc116ec4ea09c2ec1e397bbR90) above), is a function that returns a tuple of `(signature, codegen)`, where:
  * `signature` is the typing signature that Numba-CUDA uses during type inference to determine and validate the function's argument and return types, and
  *  `codegen` is a function that Numba-CUDA calls to generate the LLVM IR for the implementation.
* During the compilation process (when `impl()` is being compiled), the typing and lowering for intrinsics are resolved, and the implementation of the intrinsic generated by the `codegen()` function is inserted into the generated code.
* In the compilation process for `impl()`, type inference and code generation implements `ldca_intrinsic()` as a function that returns a scalar value, and accepts and array and index as arguments. The typing is defined by the [signature on line 96](https://github.com/NVIDIA/numba-cuda/pull/51/files#diff-b68f1312d551bef433272d305aa3a378d323ad3fbfc116ec4ea09c2ec1e397bbR96), and the code generation function follows below it.
* Therefore, if you replace `return impl` with `return ldca_intrinsic` in `ol_ldca()`, you have replaced a function that accepts an array and an index then returns a scalar (`impl(array, i) -> array.dtype`), with one that accepts a typing context and a number of LLVM IR values, and returns a signature and code generation function (`impl(typingctx, array, index) -> (Signature, Function)`).

I hope this clarifies things a bit, but for a more complete understanding I can't see a shortcut that avoids working through the low- and high-level extension API documentation for Numba:

* [Low-level API](https://numba.readthedocs.io/en/stable/extending/low-level.html)
* [High-level API](https://numba.readthedocs.io/en/stable/extending/high-level.html)

That said, I do have a couple of worked examples that show the whole flow in one notebook for each of these APIs for the CUDA target, which may also help:

* [Extending Numba-CUDA with the Low-level API](https://github.com/gmarkall/extending-numba-cuda/blob/main/Extending%20Numba's%20CUDA%20Target.ipynb)
* [Extending Numba-CUDA with the High-level API](https://github.com/gmarkall/extending-numba-cuda/blob/main/Extending%20Numba's%20CUDA%20target%20with%20the%20High-level%20Extension%20API.ipynb)

They may be a little out-of-date and need a couple of bits updating, but the general flow of them is still relevant.

We're using the High-level API here (that `@overload` and `@intrinsic` are part of), but it's probably hard to understand the High-level API without first understanding the Low-level API. The High-level API is intended to make it quicker and easier to write Numba extensions, but in my view the main thing it provides is some shorthand for a lot of Low-level API work.

Finally - what happens if you implement the suggested change (modulo the typo in the name of the function above)? You will get:

```
AssertionError: Implementation function returned by `@overload` has an unexpected type.  Got <intrinsic impl>
```

cc @kaeun97 as the explanation might be helpful in understanding the PR as a whole.

gmarkall:
Closing this as #587 supersedes it.

Issue #58, opened 2024-10-21
[FEA] Env variable for pynvjitlink should have `NUMBA_CUDA_` prefix


gmarkall:
Following the merge of #23 / #56, the variable `ENABLE_PYNVJITLINK` should be `NUMBA_CUDA_ENABLE_PYNVJITLINK` for consistency with other variables.

brandon-b-miller:
Happy to fix this + add the env var to the docs, which I believe still needs to be done.

gmarkall:
I had a couple of fixups to make so I added the rename to it, in #59 - I haven't added any docs yet though.

gmarkall:
This was completed in #59.

PR #59, opened 2024-10-22
Fixups following #23 / #56

gmarkall:
Some small fixups I'd have suggested, and renaming the environment /config variable as per #58.

### DIFF
--- a/ci/test_conda_pynvjitlink.sh
+++ b/ci/test_conda_pynvjitlink.sh
@@ -69,7 +69,7 @@ popd
 
 
 rapids-logger "Run Tests"
-ENABLE_PYNVJITLINK=1 NUMBA_CUDA_TEST_BIN_DIR=$NUMBA_CUDA_TEST_BIN_DIR python -m numba.runtests numba.cuda.tests -v
+NUMBA_CUDA_ENABLE_PYNVJITLINK=1 NUMBA_CUDA_TEST_BIN_DIR=$NUMBA_CUDA_TEST_BIN_DIR python -m numba.runtests numba.cuda.tests -v
 
 popd
 
--- a/ci/test_wheel_pynvjitlink.sh
+++ b/ci/test_wheel_pynvjitlink.sh
@@ -42,6 +42,6 @@ rapids-logger "Show Numba system info"
 python -m numba --sysinfo
 
 rapids-logger "Run Tests"
-ENABLE_PYNVJITLINK=1 NUMBA_CUDA_TEST_BIN_DIR=$NUMBA_CUDA_TEST_BIN_DIR python -m numba.runtests numba.cuda.tests -v
+NUMBA_CUDA_ENABLE_PYNVJITLINK=1 NUMBA_CUDA_TEST_BIN_DIR=$NUMBA_CUDA_TEST_BIN_DIR python -m numba.runtests numba.cuda.tests -v
 
 popd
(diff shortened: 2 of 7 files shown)


brandon-b-miller:
This LGTM but I think we should add docs here as well. What do you think of this patch as something to start out with? 

<details>

```
diff --git a/docs/source/reference/envvars.rst b/docs/source/reference/envvars.rst
index aae4b98..a106f74 100644
--- a/docs/source/reference/envvars.rst
+++ b/docs/source/reference/envvars.rst
@@ -120,4 +120,14 @@ target.
    ``/usr/local/cuda/include``. On Windows, the default is
    ``$env:CUDA_PATH\include``.
 
+.. envvar:: NUMBA_CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY
+   
+      Enable minor version compatibility for the CUDA driver. Requires the
+      ``cubinlinker`` and ``ptxcompiler`` packages to be installed. Provides minor
+      version compatibility for driver versions less than 12.0.
 
+.. envvar:: NUMBA_CUDA_ENABLE_PYNVJITLINK
+
+   Use ``pynvjitlink`` for minor version compatibility. Requires the ``pynvjitlink``
+   package to be installed. Provides minor version compatibility for driver versions
+   greater than 12.0. 
diff --git a/docs/source/user/minor_version_compatibility.rst b/docs/source/user/minor_version_compatibility.rst
index a421c07..879616f 100644
--- a/docs/source/user/minor_version_compatibility.rst
+++ b/docs/source/user/minor_version_compatibility.rst
@@ -65,7 +65,8 @@ MVC support is enabled by setting the environment variable:
 
 .. code:: bash
 
-   export NUMBA_CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY=1
+   export NUMBA_CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY=1 # CUDA 11
+   export NUMBA_CUDA_ENABLE_PYNVJITLINK=1 # CUDA 12
 
 
 or by setting a configuration variable prior to using any CUDA functionality in
@@ -74,7 +75,8 @@ Numba:
 .. code:: python
 
    from numba import config
-   config.CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY = True
+   config.CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY = True # CUDA 11
+   config.CUDA_ENABLE_PYNVJITLINK = True # CUDA 12
 
 
 References
```

</details>

In addition I debated documenting `NUMBA_CUDA_TEST_BIN_DIR`. As a final thought it might now be necessary for users who run the tests locally to `make` the tests in order for them to pass. We might consider documenting this as part of contributing process or similar.

gmarkall:
> What do you think of this patch as something to start out with?

I think that looks perfect as-is.

> In addition I debated documenting `NUMBA_CUDA_TEST_BIN_DIR`. As a final thought it might now be necessary for users who run the tests locally to `make` the tests in order for them to pass. We might consider documenting this as part of contributing process or similar.

Let's leave this for now - I was trying to think if there's any way to streamline this a bit more / make it more transparent for users, since it's not going to be a straightforward explanation to have to run `make` somewhere in someone's install tree if they're running from a package.

gmarkall:
@brandon-b-miller I'm happy with the docs you provided - are you happy for this to be merged?

brandon-b-miller (approved):

PR #60, opened 2024-10-22
#48 rebased: "Allow JIT Compile to, and Link from, LTOIR for cuda source input"

gmarkall:
Following the merge of #23 / #56, this is #48 rebased on `main` for review.

The changes here should be the same as those in #48, except:

- Merge base is `main`
- No commits from the original #23 PR, the content of which is now in `main`.
- Removed the style churn to simplify reading the diff.

cc @isVoid

### DIFF
--- a/numba_cuda/numba/cuda/codegen.py
+++ b/numba_cuda/numba/cuda/codegen.py
@@ -179,17 +179,7 @@ class CUDACodeLibrary(serialize.ReduceMixin, CodeLibrary):
 
         return ltoir
 
-    def get_cubin(self, cc=None):
-        cc = self._ensure_cc(cc)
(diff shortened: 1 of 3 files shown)


isVoid:
Discussed with @gmarkall offline. This PR is actually enabling lto for `CUSource` inputs, and a test should be added for that. Besides, because numba-cuda does not hard depend on pynvjitlink, and that pynvjitlink is required to enable LTO. We should keep LTO disabled by default. User can enable LTO via `cuda.jit(lto=True)`. This control is still important because Numba users can be kernel developers that wants fine grain control over the compilation. In the future, we should consider using a global config to set LTO default values.

PR #62, opened 2024-10-23
Allow CUDA source inputs compiled to LTOIR, and enable pynvjitlinker to link inputs that contains LTOIR

isVoid:
This PR supercedes #60 due to write permission issue.


isVoid:
I'm not able to reproduce this segfault on my v100 machine:
```
test_namedunituple (numba.cuda.tests.cudapy.test_array_args.TestCudaArrayArg.test_namedunituple) ... Fatal Python error: Segmentation fault
...

Extension modules: numpy._core._multiarray_umath, numpy._core._multiarray_tests, numpy.linalg._umath_linalg, numba.core.typeconv._typeconv, numpy.random._common, numpy.random.bit_generator, numpy.random._bounded_integers, numpy.random._mt19937, numpy.random.mtrand, numpy.random._philox, numpy.random._pcg64, numpy.random._sfc64, numpy.random._generator, numba._helperlib, numba._dynfunc, numba._dispatcher, numba.core.runtime._nrt_python, numba.np.ufunc._internal, numba.experimental.jitclass._box, numba.mviewbuf, pynvjitlink._nvjitlinklib, numba.types.itertools (total: 22)
ci/test_conda_pynvjitlink.sh: line 72:  2238 Segmentation fault      (core dumped) ENABLE_PYNVJITLINK=1 NUMBA_CUDA_TEST_BIN_DIR=$NUMBA_CUDA_TEST_BIN_DIR python -m numba.runtests numba.cuda.tests -v
/__w/numba-cuda/numba-cuda
```

isVoid:
In [afcce87](https://github.com/NVIDIA/numba-cuda/pull/62/commits/afcce87b3fc0b7bb35d7d750be2b69e44baea650) I added an additional flag `ignore_nonlto` to the linker so that only LTO-able objects are added when the flag is enabled. This allows the driver is in the correct state when `-ptx` flag is set. And this is also the desired behavior since Numba now dumps the optimized PTX only for the portion that are LTO-abled added to the linker, and raise warning for any source that aren't optimizable.

isVoid:
A subtle case here is that `lto=True` is only enabled for `cuda>12.0` and is tested so. Because this feature depends on pynvjitlink, which is only tested in CTK12.5 environment.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2704,6 +2710,13 @@ def add_file_guess_ext(self, path_or_code):
                         "Don't know how to link file with extension "
                         f"{ext}"
                     )
+                if ignore_nonlto and kind != FILE_EXTENSION_MAP["ltoir"]:
```
isVoid, line 2713: There's a nuance here where fatbin object can also only contains LTOIR. Though that requires building additional bindings for `cuobjdump` and outputting the result.

isVoid: In [bac11f6](https://github.com/NVIDIA/numba-cuda/pull/62/commits/bac11f68a9be548f175abfe9e7698d78eb8c6b7e) I made a new created a new fatbin object that contains both LTOIR and SASS and feed into the linker for testing. I added a helper function that invokes `cuobjdump` to extract object types from the fatbin / object. We allow and fatbin / object that contains LTOIR to be passed into the linker.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -3363,3 +3439,20 @@ def get_version():
     Return the driver version as a tuple of (major, minor)
     """
     return driver.get_version()
+
+
+def inspect_obj_content(objpath: str):
```
gmarkall, line 3444: We need to be prepared for cuobjdump not to be available and handle that case. Similar handling for nvdisasm is at https://github.com/NVIDIA/numba-cuda/blob/main/numba_cuda/numba/cuda/codegen.py#L16

isVoid: Fixed in https://github.com/NVIDIA/numba-cuda/pull/62/commits/224eef4630adcfd5178598e3b8bc40f8423be1e5

gmarkall:
After a merge from `main`, this passes all tests for me locally on Windows.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2683,12 +2686,18 @@ def add_cu_file(self, path):
...
         self.add_cu(cu, os.path.basename(path))
 
-    def add_file_guess_ext(self, path_or_code):
+    def add_file_guess_ext(self, path_or_code, ignore_nonlto=False):
         """
         Add a file or LinkableCode object to the link. If a file is
         passed, the type will be inferred from the extension. A LinkableCode
         object represents a file already in memory.
+
+        When `ignore_nonlto` is set to true, do not add code that are will not
```
gmarkall, line 2695:
```suggestion
        When `ignore_nonlto` is set to true, do not add code that will not
```

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -3086,6 +3156,12 @@ def add_data(self, data, kind, name):
         except NvJitLinkError as e:
             raise LinkerError from e
 
+    def get_linked_ptx(self):
+        try:
+            return self._linker.get_linked_ptx()
+        except NvJitLinkError as e:
+            raise LinkerError from e
+
```
gmarkall, line 3164: I think this is dead code?
```suggestion
```

gmarkall: I was wrong, this is not dead, I just didn't test / inspect locally properly.

numba_cuda/numba/cuda/codegen.py
```diff
@@ -200,11 +189,44 @@ def get_cubin(self, cc=None):
             linker.add_ptx(ptx.encode())
 
         for path in self._linking_files:
-            linker.add_file_guess_ext(path)
+            linker.add_file_guess_ext(path, ignore_nonlto)
```
isVoid, line 192: If user still installs pynvjitlink via `patch_numba_linker`, this call would throw because of mismatched number of argument.

isVoid: Synced offline: users should not use `patch_numba_linker` anymore.

PR #78, opened 2024-12-02
Add NRT Memsys

isVoid:
This PR adds memsys to NRT. A follow up to #17


numba_cuda/numba/cuda/runtime/memsys.cuh
```diff
@@ -0,0 +1,17 @@
...
+      bool enabled;
+      cuda::atomic<size_t, cuda::thread_scope_device> alloc;
+      cuda::atomic<size_t, cuda::thread_scope_device> free;
+      cuda::atomic<size_t, cuda::thread_scope_device> mi_alloc;
+      cuda::atomic<size_t, cuda::thread_scope_device> mi_free;
+    } stats;
+  };
+
+/* The Memory System object */
+__device__ NRT_MemSys* TheMSys;
```
gmarkall, line 15: I *think* this is initialized to 0 (and therefore the checks in nrt.cu like `if(TheMSys && ...)` are always safe. Am I thinking about that correctly?

gmarkall: Also, if this is included into both memsys.cu and nrt.cu, is there a potential ODR violation here? Should this be `extern` in the header and defined in one of those two files?

numba_cuda/numba/cuda/runtime/memsys.cu
```diff
@@ -0,0 +1,92 @@
+#include "memsys.cuh"
+
+namespace detail
+{
+  void __device__ check_memsys()
```
gmarkall, line 5: I think this is a good idea - do we have things set up sufficiently correctly in numba-cuda such that this assertion can fire (e.g. if `debug=True` for a `@cuda.jit` dewcorator) when `TheMSys` is null?

numba_cuda/numba/cuda/runtime/nrt.cu
```diff
@@ -18,29 +20,20 @@ struct MemInfo {
...
 
-/* The Memory System object */
-__device__ NRT_MemSys* TheMSys;
-
 extern "C" __device__ void* NRT_Allocate(size_t size)
 {
   void* ptr = NULL;
   ptr       = malloc(size);
-//  if (TheMSys->stats.enabled) { TheMSys->stats.alloc++; }
+  if (TheMSys && TheMSys->stats.enabled) { TheMSys->stats.alloc++; }
```
wence-, line 36: *nit*: this can be a relaxed atomic, I suspect.

numba_cuda/numba/cuda/runtime/memsys.cu
```diff
@@ -0,0 +1,92 @@
...
+extern "C" __global__ void NRT_MemSys_read(uint64_t *managed_memsys)
+{
+  detail::check_memsys();
+  managed_memsys[0] = TheMSys->stats.alloc;
+  managed_memsys[1] = TheMSys->stats.free;
+  managed_memsys[2] = TheMSys->stats.mi_alloc;
+  managed_memsys[3] = TheMSys->stats.mi_free;
+}
+
+extern "C" __global__ void NRT_MemSys_read_alloc(uint64_t *managed_memsys)
```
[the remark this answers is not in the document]
isVoid: Ah - that's just an oversight since I copied the signature from the above. `managed_result` sounds like a good choice.

numba_cuda/numba/cuda/runtime/memsys.cu
isVoid: I don't think the tests currently cover a use case where the memsys is null. This check only fires when the NRT is enabled, but the memsys is not allocated. However currently we always promise memsys is allocated if NRT is enabled.

numba_cuda/numba/cuda/runtime/memsys.cuh
isVoid: > I think this is initialized to 0 (and therefore the checks in nrt.cu like if(TheMSys && ...) are always safe. Am I thinking about that correctly?

Yes.

> Also, if this is included into both memsys.cu and nrt.cu, is there a potential ODR violation here? Should this be extern in the header and defined in one of those two files?

Currently `memsys.cu` and `nrt.cu` are compiled into two separate modules, so far ODR violation isn't an issue. This is a deliberate design decision to separate memsys stats method from NRT methods. I was hoping this can save us some compilation time when we turn on NRT, but do not turn on statistics. This design is debatable and could change.

numba_cuda/numba/cuda/runtime/memsys.cu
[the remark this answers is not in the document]
isVoid: Done in https://github.com/NVIDIA/numba-cuda/pull/78/commits/685eb64c3dfa280d13fac4c5ac5682225cfc8c31

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -362,11 +362,14 @@ def launch(self, args, griddim, blockdim, stream=0, sharedmem=0):
 
         stream_handle = stream and stream.handle or zero_stream
 
-        rtsys.ensure_allocated(stream_handle)
-        rtsys.set_memsys_to_module(cufunc.module, stream_handle)
-        rtsys.ensure_initialized(stream_handle)
-        if config.CUDA_NRT_STATS:
-            rtsys.memsys_enable_stats(stream_handle)
+        if hasattr(self, "target_context") and self.target_context.enable_nrt:
```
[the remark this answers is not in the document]
gmarkall: I think it is. In future I think we should do something to ensure that NRT and caching can work together.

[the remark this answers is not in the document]
gmarkall: Can we push this into the `_Kernel.bind()` method instead of here? My thinking is that we can then get this check out of the critical path for launching a kernel, which is already quite slow. Eventually we want to jit the call site, so adding more complexity into launches is also going to make that more tricky in the future.

numba_cuda/numba/cuda/runtime/nrt.cu
```diff
@@ -18,29 +20,21 @@ struct MemInfo {
...
-/* The Memory System object */
-__device__ NRT_MemSys* TheMSys;
-
 extern "C" __device__ void* NRT_Allocate(size_t size)
 {
   void* ptr = NULL;
   ptr       = malloc(size);
-//  if (TheMSys->stats.enabled) { TheMSys->stats.alloc++; }
+  if (TheMSys && TheMSys->stats.enabled) { 
+    TheMSys->stats.alloc.fetch_add(1, cuda::memory_order_relaxed) }
```
gmarkall, line 37: I'm a bit surprised this is working on CI - locally I seem to have a syntax error:

```suggestion
    TheMSys->stats.alloc.fetch_add(1, cuda::memory_order_relaxed); }
```

gmarkall: Ah, perhaps CI is not running NRT tests? I also got 

```
FAIL: test_stats_env_var_explicit_on (numba.cuda.tests.nrt.test_nrt.TestNrtStatistics.test_stats_env_var_explicit_on)
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/tests/nrt/test_nrt.py", line 129, in test_stats_env_var_explicit_on
    run_in_subprocess(src, env=env)
  File "/home/gmarkall/numbadev/numba/numba/tests/support.py", line 1136, in run_in_subprocess
    raise AssertionError(msg % (popen.returncode, err.decode()))
AssertionError: process failed with code 1: stderr follows
  File "<string>", line 12, in <module>
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/runtime/nrt.py", line 144, in initialize
    self._single_thread_launch(
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/runtime/nrt.py", line 119, in _single_thread_launch
    func = module.get_function(name)
           ^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'get_function'
```

locally, though I still need to check if I've made a mistake locally here.

numba_cuda/numba/cuda/tests/nrt/test_nrt.py
```diff
@@ -100,11 +73,157 @@ def g(out_ary):
...
+        if self.__stats_state:
+            rtsys.memsys_enable_stats(self._stream)
+        else:
+            rtsys.memsys_disable_stats(self._stream)
+
+    def test_stats_env_var_explicit_on(self):
+        # Checks that explicitly turning the stats on via the env var works.
+        src = """if 1:
+        from numba import cuda
+        from unittest.mock import patch
```
gmarkall, line 99: I think this is now unused:

```suggestion
```

numba_cuda/numba/cuda/tests/nrt/test_nrt.py
```diff
@@ -100,11 +73,157 @@ def g(out_ary):
...
+        from numba.cuda.runtime import rtsys
+        from numba.cuda.tests.nrt.mock_numpy import cuda_arange
+
+        @cuda.jit
+        def foo():
+            x = cuda_arange(10)[0]
+
+        # initialize the NRT before use
+        rtsys.initialize()
```
gmarkall, line 108: I haven't got this test working locally yet (or understood it fully) but it seems we need to allocate before initialization:

```suggestion
        rtsys.allocate()
        rtsys.initialize()
```

numba_cuda/numba/cuda/tests/nrt/test_nrt.py
```diff
@@ -100,11 +73,157 @@ def g(out_ary):
...
+        assert rtsys.memsys_stats_enabled()
```
gmarkall, line 109: It can be a bit hard to see which assertion failed if there's no message to go with it:

```suggestion
        assert rtsys.memsys_stats_enabled(), "Stats not enabled"
```

numba_cuda/numba/cuda/tests/nrt/test_nrt.py
```diff
@@ -100,11 +73,157 @@ def g(out_ary):
...
+        total_alloc = new_stats.alloc - orig_stats.alloc
+        total_free = new_stats.free - orig_stats.free
+        total_mi_alloc = new_stats.mi_alloc - orig_stats.mi_alloc
+        total_mi_free = new_stats.mi_free - orig_stats.mi_free
+
+        expected = 1
+        assert total_alloc == expected
+        assert total_free == expected
+        assert total_mi_alloc == expected
+        assert total_mi_free == expected
```
gmarkall, line 122: Similar to above:

```suggestion
        assert total_alloc == expected, "total_alloc == expected"
        assert total_free == expected, "total_free == expected"
        assert total_mi_alloc == expected, "total_mi_alloc == expected"
        assert total_mi_free == expected, "total_mi_free == expected"
```

numba_cuda/numba/cuda/runtime/nrt.cu
isVoid: I can reproduce this locally now. And yes I don't think the tests are being run in the CI...

numba_cuda/numba/cuda/dispatcher.py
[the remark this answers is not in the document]
isVoid: Done in https://github.com/NVIDIA/numba-cuda/pull/78/commits/5ce81c0256a16da17bcd847c8328d6c785a697f3

numba_cuda/numba/cuda/tests/nrt/test_nrt.py
```diff
@@ -100,11 +73,157 @@ def g(out_ary):
...
+        self.__stats_state = rtsys.memsys_stats_enabled(self._stream)
+
+    def tearDown(self):
+        # Set stats state back to whatever it was before the test ran
+        if self.__stats_state:
+            rtsys.memsys_enable_stats(self._stream)
+        else:
+            rtsys.memsys_disable_stats(self._stream)
+
+    def test_stats_env_var_explicit_on(self):
```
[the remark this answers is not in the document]
isVoid: This test is ported from the CPU target test. It seems to me that the CPU design distinguishes initialization from memsys enablement. When memsys is initialized, the enable flag is set to `False` and needs explicit turning on. With this environment variable, `initialize()` also does the enabling. In the CUDA target design, I also mimicked that behavior.

numba_cuda/numba/cuda/runtime/nrt.py
```diff
@@ -0,0 +1,308 @@
...
+        """
+        from numba.cuda import device_array
+
+        # Check if memsys module is defined
+        if self._memsys_module is None:
+            # Compile the memsys module if not defined
+            self._compile_memsys_module()
+
+        # Allocate space for NRT_MemSys
+        # TODO: determine the size of NRT_MemSys at runtime
```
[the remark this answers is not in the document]
isVoid: Thanks! Is it possible to use a host side variable for `sizeof(NRT_MemSys)`?

numba_cuda/numba/cuda/tests/nrt/test_nrt_refct.py
```diff
@@ -0,0 +1,118 @@
...
+
+class TestNrtRefCt(EnableNRTStatsMixin, CUDATestCase):
+
+    def setUp(self):
+        # Clean up any NRT-backed objects hanging in a dead reference cycle
+        gc.collect()
+        super(TestNrtRefCt, self).setUp()
+
+    def tearDown(self):
+        super(TestNrtRefCt, self).tearDown()
```
gmarkall, line 22: I think these tests were ported from the CPU target - it's not clear to me that the logic applies here as we don't have any connection between objects on the CPU and allocations / refcounts on the GPU. Am I missing something?

numba_cuda/numba/cuda/runtime/memsys.cu
```diff
@@ -0,0 +1,94 @@
...
+
+__device__ size_t memsys_size = sizeof(NRT_MemSys);
+
+namespace detail
+{
+  void __device__ check_memsys()
+  {
+    if (TheMSys == nullptr)
+    {
+      assert(false && "TheMSys pointer is null. Please use NRT_MemSys_set to set pointer first.");
```
gmarkall, line 11: I just checked to see what happens when this assert fires - I get a traceback like:

```
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/cudadrv/driver.py", line 860, in _attempt_allocation
    return allocator()
           ^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/cudadrv/driver.py", line 992, in allocator
    driver.cuMemAllocManaged(byref(ptr), size, flags)
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/cudadrv/driver.py", line 336, in safe_cuda_api_call
    self._check_ctypes_error(fname, retcode)
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/cudadrv/driver.py", line 404, in _check_ctypes_error
    raise CudaAPIError(retcode, msg)
numba.cuda.cudadrv.driver.CudaAPIError: [710] Call to cuMemAllocManaged results in CUDA_ERROR_ASSERT
```

This seems to be what we want. I wonder if there's a way we can make it easier to show assert messages to the user when an assert fires (but that is a complete tangent from this PR, I may make a featuer request / issue later).

gmarkall (approved):
I think this now looks good, and earlier queries are resolved. In addition to reviewing the code, I tried a couple of things:

- Running tests under `compute-sanitizer`. No errors were reported
- Experimenting what happens when the assert in `memsys.cu` fires - this reports a `CUDA_ERROR_ASSERT` back to the user, so we're getting the behaviour that one would expect.
- Experimented manually with the stats counters and inspected their values, which seem as expected (sometimes relative values of the initial and current stats are checked, and I wanted to make sure the absolute values looked reasonable).

I think we need an issue to record the fact that NRT + caching may not work, which I'll create shortly.

numba_cuda/numba/cuda/runtime/nrt.py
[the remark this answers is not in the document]
gmarkall: I couldn't figure out a way to do that with NVRTC, unfortunately.

gmarkall:
Created #105 to record the caching + NRT issue.

@isVoid I'm happy to merge this now, but wanted to check whether you had any thoughts about the `gc.collect()` in `test_nrt_refcount` first - I'm happy for it to be left as-is for now and to merge it, but just wanted to wait for your thoughts before marking this as ready to merge.

numba_cuda/numba/cuda/tests/nrt/test_nrt_refct.py
isVoid: I don't think the `gc.collect()` bits are relavant here, but calling the parent ctor (from `EnableNRTStatsMixin` parent) is. This is a new mixin we created for this PR that enables the memsys stats on test startup and disables on shutdown.

isVoid: Removed in https://github.com/NVIDIA/numba-cuda/pull/78/commits/97b65b27725e968535d22c02201960d6a9aa8ac8.

PR #82, opened 2024-12-04
Fixes for bugs in test suite and conda include detection in v0.0.19

gmarkall:
- Fix tests in `test_debug` when running on Python 3.13 - they did not take that version into account.
- Fix the stream API tests when using the NVIDIA CUDA Python bindings, as they need a different conversion from objects to pointers for stream objects
- Fix conda include dir detection, which could determine that includes were present in the conda environment when they were not, particularly on Windows where the include path would be `$env:CONDA_PREFIX\Library\include`, which is pretty much guaranteed to exist.


gmarkall:
Tested locally on Windows to verify that Numba no longer chooses the conda environment for the include path when there are no CUDA headers installed in the conda environment.

PR #86, opened 2024-12-09
Fix find_spec for old finders that don't implement it.

vyasr:
The code in https://github.com/NVIDIA/numba-cuda/blob/v0.0.17/site-packages/_numba_cuda_redirector.py#L70 is incompatible with legacy finders (pre-Python 3.4) that do not implement the `find_spec` method. This makes numba-cuda incompatible with Google Colab (see https://github.com/googlecolab/colabtools/issues/5000). For the purpose of RAPIDS 24.12, which is an early adopter of numba-cuda, we need this fix to go into a patch release of 0.0.17 because RAPIDS 24.12 is incompatible with numba-cuda>=0.0.18.


vyasr:
Marking as draft until we can figure out our merging/release strategy. Since numba-cuda has only released patch versions up until now, we will either need to add another release segment just for this case, or make a post release. RAPIDS 24.12 is pinned to <0.0.18 due to incompatibilities that introduces, see https://github.com/rapidsai/cudf/pull/17359.

jameslamb:
> we will either need to add another release segment just for this case, or make a post release.

Copying what I wrote offline... I think we should release this as `0.0.17.1` (adding another component to the version number just for this purpose). This type of change deserves to be a patch release, not a post release, and I can't think of any negative consequences of using a 4-part version for this... it doesn't mean we're committing to 4-part version numbers forever or even for the next release after this.

gmarkall:
Ah, crumbs. I rebased this to a branch starting on v0.0.17 and Github's now pulling in all the history from the main branch.

gmarkall:
My plan here was:

- Retarget this PR to the `v0.0.17-patches` branch, which starts from the `v0.0.17` tag
- Merge this into that branch
- Bump the version to 0.0.17.1 in a new PR
- Merge the version bump PR, then upload the resulting packages to anaconda.org and pypi

vyasr:
Rebased accordingly.

PR #91, opened 2024-12-18
Allow Pynvjitlinker Enabled via Config at Runtime

isVoid:


numba_cuda/numba/cuda/tests/cudadrv/test_nvjitlink.py
```diff
@@ -251,5 +253,61 @@ def kernel():
                 pass
 
 
+class TestLinkerUsage(CUDATestCase):
```
gmarkall, line 256: These probably need to be skipped if pynvjitlink isn't installed

isVoid: Fixed in https://github.com/NVIDIA/numba-cuda/pull/91/commits/38238bc6bba2288064a6c30fb5e368b085554921

numba_cuda/numba/cuda/tests/cudadrv/test_nvjitlink.py
```diff
@@ -251,5 +260,64 @@ def kernel():
...
+                result = cuda.device_array(1)
+                kernel[1, 1](result)
+                assert result[0] == 3
+        """
+        env = os.environ.copy()
+        env['NUMBA_CUDA_ENABLE_PYNVJITLINK'] = "1"
+        print(env['NUMBA_CUDA_TEST_BIN_DIR'])
+        run_in_subprocess(src, env=env)
+
+    def test_linker_enabled_config(self):
```
gmarkall, line 303: I think this test also needs to run in a subprocess where the `NUMBA_CUDA_ENABLE_PYNVJITLINK` env var is not present. The CI setup at present will already be running this test with that environment variable set to `1` in the cases where this test actually runs - see e.g.

https://github.com/NVIDIA/numba-cuda/blob/f57706fdce83f51f7dd65ad5ed435810299605b2/ci/test_conda_pynvjitlink.sh#L72

gmarkall: It may also be good to have a test of where the config variable is used to disable pynvjitlink, with `CUDA_ENABLE_PYNVJITLINK` set to `False` to ensure it's successfully disabled.

isVoid: Isn't that already exist? 

https://github.com/isVoid/numba-cuda/blob/c9d148b781b9a2d941d4e64e1b8215cbc6d333a2/ci/test_conda.sh#L51-L52

numba_cuda/numba/cuda/tests/cudadrv/test_nvjitlink.py
```diff
@@ -251,5 +260,74 @@ def kernel():
...
+        @cuda.jit(link=[test_device_functions_cubin], lto=True)
+        def kernel(result):
+            result[0] = add_from_numba(1, 2)
+
+        result = cuda.device_array(1)
+        kernel[1, 1](result)
+        assert result[0] == 3
+
+    def test_linker_enabled_config(self):
+        with override_config("CUDA_ENABLE_PYNVJITLINK", True):
```
gmarkall, line 321: With the latest changes, I still think this test needs to run in a subprocess without pynvjitlink enabled by an environment variable - in its present form in CI, this test is running with pynvjitlink enabled by the environment variable, so even if setting the config variable has no effect I think it would still pass.

(If I'm misunderstanding the flow here, please point me in the right direction to understand what's going on :-) )

isVoid: Ok yeah, you are right on this. I think the test should reflect what we are actually trying to achieve - in a separate process, can user use pynvjitlink with numba-cuda by just turning on the config item at runtime? We do need to run it in separate process to make sure Numba is loaded from scratch.

PR #97, opened 2024-12-20
Fixup debug metadata in kernel fixup

### LINKED ISSUE #95: [BUG] Kernel fixup does not adjust debug metadata arguments
The kernel fixup added in #50 removes the return value placeholder from the argument list, but does not remove it from the debug metadata. Thus, a kernel like:

```python
@cuda.jit("void(int32, int32)", debug=True, opt=False)
def k(x, y): 
    x + y
```

will have the following prototype in LLVM IR:

```
define void @"k"(i32 %"arg.x", i32 %"arg.y") !dbg !10
```

(function name demangled for clarity)

but the debug metadata is:

```
!3 = !DIBasicType(encoding: DW_ATE_unsigned, name: "i8", size: 8)
!4 = !DIDerivedType(baseType: !3, size: 64, tag: DW_TAG_pointer_type)
!5 = !DIDerivedType(baseType: !4, size: 64, tag: DW_TAG_pointer_type)
!6 = !DIBasicType(encoding: DW_ATE_signed, name: "int32", size: 32)
!7 = !DIBasicType(encoding: DW_ATE_signed, name: "int32", size: 32)
!8 = !{ !5, !6, !7 }
!9 = !DISubroutineType(types: !8)
!10 = distinct !DISubprogram(file: !1, isDefinition: true, isLocal: false, isOptimized: true,
                             line: 5, linkageName: "k", name: "k", scope: !1, scopeLine: 5, type: !9, unit: !2)
```

The issue being that the arguments in `!8` are `!{ !5, !6, !7 }`. `!5` is the placeholder for a return value (the equivalent of `void**`), which should have been removed to leave only `!{ !6, !7 }`.

gmarkall:
The first kernel parameter also needs removing from the debug metadata, so that the debug metadata argument types align with the arguments to the fixed-up kernel.

Fixes #95.

### DIFF
--- a/numba_cuda/numba/cuda/compiler.py
+++ b/numba_cuda/numba/cuda/compiler.py
@@ -352,6 +352,18 @@ def kernel_fixup(kernel, debug):
     kernel.return_value = ir.ReturnValue(kernel, ir.VoidType())
     kernel.args = kernel.args[1:]
 
+    # If debug metadata is present, remove the return value from it
+
+    if kernel_metadata := getattr(kernel, 'metadata', None):
+        if dbg_metadata := kernel_metadata.get('dbg', None):
+            for name, value in dbg_metadata.operands:
+                if name == "type":
+                    type_metadata = value
+                    for tm_name, tm_value in type_metadata.operands:
+                        if tm_name == 'types':
+                            types = tm_value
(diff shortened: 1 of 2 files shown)


gmarkall:
This could do with some tests adding.

PR #112, opened 2025-01-21
Pin numba version to `<0.61.0a0`

galipremsagar:
Numba recently had a major release: https://numba.readthedocs.io/en/stable/release/0.61.0-notes.html

Which now has breaking changes:

- Within the new type system, Python and NumPy scalars will be treated and returned separately as different entities through JIT compiled functions.

- The minimum supported Python and NumPy versions have been bumped to 3.10 and 1.24 respectively.

Until `numba-cuda` can support these changes, and to unblock down-stream CIs in RAPIDS repositories, this pinning is required.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ authors = [
 ]
 license = { text = "BSD 2-clause" }
 requires-python = ">=3.9"
-dependencies = ["numba>=0.59.1,<0.61.0"]
+dependencies = ["numba>=0.59.1,<0.61.0a0"]
 
 [project.urls]
 Homepage = "https://github.com/rapidsai/numba-cuda"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ authors = [
 ]
 license = { text = "BSD 2-clause" }
 requires-python = ">=3.9"
-dependencies = ["numba>=0.59.1"]
+dependencies = ["numba>=0.59.1,<0.61.0"]
 
 [project.urls]
 Homepage = "https://github.com/rapidsai/numba-cuda"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -18,7 +18,7 @@ authors = [
 ]
 license = { text = "BSD 2-clause" }
 requires-python = ">=3.9"
-dependencies = ["numba>=0.59.1"]
+dependencies = ["numba>=0.59.1,<0.61.0"]
```
bdice, line 21: Please disallow alphas in upper bounds.
```suggestion
dependencies = ["numba>=0.59.1,<0.61.0a0"]
```

bdice:
If RAPIDS is broken by `numba` 0.61 but `numba-cuda`'s test suite is _not_ broken by `numba` 0.61, we should only pin `numba` in RAPIDS and not in this repo.

bdice:
Discussed with @galipremsagar -- it appears that `numba-cuda` needs to update the `__array__` implementation because `numba-0.61.0` adds support for `numpy` 2.1+, which has breaking changes for `__array__` signatures.

This pinning is needed in the short term to unblock RAPIDS. In the medium term, we should determine whether RAPIDS should pin both `numba` and `numba-cuda`, or if `numba-cuda` intends to continue carrying an upper bound for the supported `numba` versions.

bdice (approved):

bdice:
We need to add tests in `numba-cuda` that exercise the broken `__array__` code paths, as tests currently do not catch this. @galipremsagar Can you help file an issue that scopes that, as a follow-up task?

galipremsagar:
> We need to add tests in `numba-cuda` that exercise the broken `__array__` code paths, as tests currently do not catch this. @galipremsagar Can you help file an issue that scopes that, as a follow-up task?

Done: https://github.com/NVIDIA/numba-cuda/issues/114

Issue #114, opened 2025-01-21
[BUG] __array__ implementation needs signature updates with numpy-2.1.+


galipremsagar:
**Describe the bug**
Currently it appears to be no pytests testing `__array__` implementation with `np.array`. We discovered in `cudf` that `numba-cuda` breaks with `numba-0.61.0` because it adds support for latest versions of numpy(2.1.+)

xref: https://github.com/NVIDIA/numba-cuda/pull/112#issuecomment-2605877705

PR #115, opened 2025-01-22
Fix `__array__()` method for NumPy 2 and remove Numba pin

gmarkall:
This fixes the `__array__()` method by implementing the `copy` kwarg. Numba pins are now removed, as this should fix the compatibility issue first addressed in #112 and described in #114.

### DIFF
--- a/numba_cuda/numba/cuda/tests/cudadrv/test_cuda_ndarray.py
+++ b/numba_cuda/numba/cuda/tests/cudadrv/test_cuda_ndarray.py
@@ -4,6 +4,7 @@ from numba.cuda.cudadrv import devicearray
 from numba import cuda
 from numba.cuda.testing import unittest, CUDATestCase
 from numba.cuda.testing import skip_on_cudasim
+from numba.tests.support import IS_NUMPY_2
 
 
 class TestCudaNDArray(CUDATestCase):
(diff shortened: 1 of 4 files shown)


numba_cuda/numba/cuda/tests/cudadrv/test_cuda_ndarray.py
```diff
@@ -456,6 +457,36 @@ def test_issue_8477(self):
...
+    def test_np_array_dtype(self):
+        dtype = np.int32
+        dev_array = cuda.to_device(np.asarray([1.0, 2.0, 3.0]))
+        host_array = np.array(dev_array, dtype=dtype)
+        np.testing.assert_equal(
+            host_array,
+            dev_array.copy_to_host().astype(dtype)
+        )
+
+    @unittest.skipUnless(IS_NUMPY_2, "NumPy 1.x does not pass copy kwarg")
```
bdice, line 477: Is this a sufficiently specific condition? I think this might only be newer releases like 2.1.0. If we can check both 2.0.0 and 2-latest, that would be good.

gmarkall: I'm not sure I understand what you're saying, but I'll try to answer with some information: these unit tests will run if the NumPy version is 2.0 or greater. I skipped them when NumPy is 1.x because they fail, which results from it not passing the `copy` kwarg.

bdice: Did you test both NumPy 2.0.0 and 2.1.0? I think there may be relevant changes in a minor, rather than major, version.

bdice: If these fail with NumPy 2.0.0, we will need something more specific than `IS_NUMPY_2`. Perhaps something like `IS_NUMPY_2_1_OR_GREATER`.

gmarkall: They pass with NumPy 2.0 and 2.1, and fail with 1.x.

bdice: Okay, great. This is fine then. Approving!

bdice (approved):

Issue #67, opened 2024-11-06
[FEA] Support automatic generation of `link` list for external code


gmarkall:
Numba-cuda extensions (e.g. [nvmath-python](https://developer.nvidia.com/nvmath-python)) are frequently leaning on CUDA C++ implementations to support the core of their functionality.

One current UX limitation is that the kernel author is required to add the list of files to link with a kernel as a keyword argument to the `@cuda.jit` decorator, for example:

```python
@cuda.jit(link=FFT.files)
def f(data):
    ...
```

from [cufftdx_simple_fft_block.py](https://github.com/NVIDIA/nvmath-python/blob/main/examples/device/cufftdx_simple_fft_block.py#L27)

The `FFT` object supplies the files, and is created like:

```python
FFT = fft(fft_type='c2c', 
          size=128,
          precision=np.float32,
          direction='forward',
          elements_per_thread=8,
          ffts_per_block=2,
          execution='Block',
          compiler='numba')
```
and is called inside the kernel as:

```python
FFT(thread_data, shared_mem)
```

Rather than the user being required to link `FFT.files`, Numba should provide a mechanism to obtain and link the list of files at the point of compilation and linking from the `FFT` object (or any implementation of a method, property, object, etc. backed by an extension).

leofang:
Perhaps ask all device functions to implement a ~~method~~ attribute, say, `__numba_cuda_link__` that returns a list of files, if they want numba-cuda to handle the linking?

gmarkall:
Continuing with the `__numba_cuda_link__` idea, I think it might need to be a method that can accept a signature, so that it can return the appropriate files for the given signature.

leofang:
How should the kernel author pass function arguments at the call site if it is a method not attribute?

gmarkall:
In the example above, the kernel author wrote:

```
FFT(thread_data, shared_mem)
```

assuming the Numba types of these are `float32[:]` and `float32[::1]` (for the sake of argument, they could be any Numba type really) I'd expect during compilation time that Numba would be doing the equivalent of calling

```
ltoir = FFT.__numba_cuda_link__(float32[:], float32[::1])
```

where `ltoir` is then an `LTOIR` linkable code object, i.e. an instance of

https://github.com/NVIDIA/numba-cuda/blob/e5f2ef936a6155712020f79e1b39b28f5d41e9ce/numba_cuda/numba/cuda/cudadrv/linkable_code.py#L59-L63

[...]

Issue #46, opened 2024-08-23
[FEA] Support adding additional search paths for `nvrtc.compile`


isVoid:
[UPDATE OF FEATURE REQUIREMENT]
The original intent of the feature request is, actually to support additional search paths for the NVRTC binding. However, I framed it as if we should allow arbitrary compilation flags for NVRTC. The nvrtc bindings are specific to compiling external functions into a format that can be linked with Numba kernels lowered with nvvm. It require certain consistency with the rest of the compilation pipeline (lto - lto, RDC requirements, etc.). Exposing arbitrary compilation flag to user can cause unpredictable breakage.

Extending the existing search paths globally, is however, a bit less risky.

For arbitrary compilation commands, we should instead support it on a per-library basis. Which can be specified partially based on [`cuda.core.ProgramOptions`](https://nvidia.github.io/cuda-python/cuda-core/latest/generated/cuda.core.experimental.ProgramOptions.html) interface. 

[ORIGINAL]
**Is your feature request related to a problem? Please describe.**
Currently `nvrtc.compile` only accepts `src, name, cc` as input arguments. When we compile a CUDA source. we may want to configure additional compile flags such as `-include`. To enhance the capability. We should make Numba's nvrtc binding to be more configurable when we need to compile additional feature with Numba.

**Describe the solution you'd like**
I think there are multiple ways we can achieve this:
- The most straight forward way is to augment the argument list to also accept `**options`, and append to the existing option list. But it may take multiple layers of exposing to make it available to the dispatcher, and maybe it's not the most preferable UX too.
- Via environment variable. Expose the additional includes via `NUMBA_CUDA_RTC_INCLUDE_PATHS` environment variable. This makes `Numba-cuda` easier to adapt under different setup configuration. But may require better documentation on how it should be set.

PR #124, opened 2025-02-12
Support linking code for device functions in declaration

gmarkall:
Allows specifying files to link in the `cuda.declare_device()` declaration, so that it's no longer required for the user to know which files to link.

Changes consist of:

- Adding the `link` kwarg to the `declare_device` function, and automatically linking in any linkable items when the declared function is used.
- Updating the documentation to describe this mechanism, and reflect that it's the recommended way to specify what to link.
- Documents the `LinkableCode` classes, which were previously undocumented.
- Removes some obsolete notices about needing the NVIDIA bindings for linking C/C++ code.
- Adds cffi to the test environment, as it's used by one of the new tests (it should have already been present, really).

I decided to not tackle #67 in its entirety, which also requests that a callback function can be used to generate the implementation, for a couple of reasons:

- I think the existing implementation is of immediate value for Numbast, and all other FFI-calling implementations.
- There is some thought needed about how to handle typing when a callback function is used - for example, whether it's necessary to generalize the typing beyond just the single signature that `declare_device()` presently accepts.

### DIFF
--- a/numba_cuda/numba/cuda/testing.py
+++ b/numba_cuda/numba/cuda/testing.py
@@ -115,14 +115,16 @@ def skip_on_arm(reason):
 def skip_if_cuda_includes_missing(fn):
     # Skip when cuda.h is not available - generally this should indicate
     # whether the CUDA includes are available or not
-    cuda_h = os.path.join(config.CUDA_INCLUDE_PATH, 'cuda.h')
+    cuda_include_path = libs.get_cuda_include_dir()
+    cuda_h = os.path.join(cuda_include_path, 'cuda.h')
     cuda_h_file = (os.path.exists(cuda_h) and os.path.isfile(cuda_h))
     reason = 'CUDA include dir not available on this system'
     return unittest.skipUnless(cuda_h_file, reason)(fn)
 
 
 def skip_if_curand_kernel_missing(fn):
-    curand_kernel_h = os.path.join(config.CUDA_INCLUDE_PATH, 'curand_kernel.h')
+    cuda_include_path = libs.get_cuda_include_dir()
+    curand_kernel_h = os.path.join(cuda_include_path, 'curand_kernel.h')
     curand_kernel_h_file = (os.path.exists(curand_kernel_h) and
                             os.path.isfile(curand_kernel_h))
     reason = 'curand_kernel.h not available on this system'


numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -158,6 +159,16 @@ def link_to_library_functions(library_functions, library_path,
...
 
+        for k, v in cres.fndesc.typemap.items():
+            if not isinstance(v, Function):
+                continue
+
+            if not isinstance(v.typing_key, ExternFunction):
+                continue
+
+            for obj in v.typing_key.link:
+                lib.add_linking_file(obj)
```
isVoid, line 170: I'm trying to understand the code flow of this part a bit deeper. Is NVRTC going to be invoked on every `CUSource` addition?

isVoid (commented):
A second question is about usage does each shim function added must be self-contained (must pass compilation on its own), aka, it must include all necessary header that's referenced by the shim function's impl.

I'm asking this because, in Numbast, there are chances where all shim functions refer to the same header file. An obvious optimization is to fuse all includes into a single header so that we can reduce the latency for RTC to parse the included headers. Just want to know if that's an optimization that's hidden under plain sight here.

numba_cuda/numba/cuda/tests/cudapy/test_device_func.py
isVoid, line 1: I also wonder if it's easy to add a test case with "#include<xx>". The difficult bit being that nvrtc is requires very explicit pointers on where to look for include files. And `compile` function is not very extensible at this point. Ext ref: https://github.com/NVIDIA/numba-cuda/issues/46

numba_cuda/numba/cuda/dispatcher.py
gmarkall: I believe so - is there another way?

gmarkall:
> A second question is about usage / optimization: does each shim function added must be self-contained (must pass compilation on its own), aka, it must include all necessary header that's referenced by the shim function's impl.
> 
> I'm asking this because, in Numbast, there are chances where all shim functions refer to the same header file. An optimization is to fuse all includes into a single header so that we can reduce the latency for RTC to parse the included headers. Just want to know if that's an optimization hidden under plain sight or an over-optimization.

Everything needs to be able to be compiled on its own - trying to make this into some sort of engine for assembling code templates seems a little bit too far of a step to go.

In terms of optimization - I think user kernels would likely only include a small subset of generated shims - if the headers are not fused, then perhaps the latency of parsing the unfused headers is low as they'll contain only minimal code?

Perhaps another way to optimize things is to give a tool like Numbast (or another library that implements code generation with callbacks, using future PR's functionality) the opportunity to have all the used functions passed in one batch so that it can generate a custom fused .cu file so that there's only one compilation needed?

gmarkall:
Another thought - perhaps we should also have an in-memory NVRTC cache - so if the same cu file with the same options is presented, we can retrieve PTX or LTOIR from the cache instead of recompiling - perhaps this would allow the same .cu file to serve as the implementation for many device functions efficiently, with a potentially longer upfront cost on the first call.

numba_cuda/numba/cuda/dispatcher.py
isVoid: Nah - It seems like this is the only way. I think at best what we can do is to cache each CUSource and do some source code analysis to determine which source can be fused and compile in the same pass. Though that's too heavy handed and may actually bring more overhead.

isVoid:
> In terms of optimization - I think user kernels would likely only include a small subset of generated shims - if the headers are not fused, then perhaps the latency of parsing the unfused headers is low as they'll contain only minimal code?

Yes, it seems to me most examples contain 3-4 intrinsics per kernel.

>  the opportunity to have all the used functions passed in one batch so that it can generate a custom fused .cu file so that there's only one compilation needed?

I'm thinking what that would look like: you would make `CUSource` have an additional attribute `fuse`, at link time you would iterate through all CUSource that has that attribute set to `True`, then fuse them before compiling. I guess the order of the source to fuse should respect the original order the user passes in the source. Additionally, there should be a separate global field to specialize the shared headers.

Additionally, you could allow user specializing a `fuse_key` option, allowing the compiler to fuse sources with the same key, a finer control knob. As for headers, they become per-key headers. (Numbast doesn't need this for now, just going wild with design ideas).

isVoid (approved):
This is super useful, thanks!

Issue #107, opened 2025-01-08
[BUG] Debug: grid groups appear to be floats, in cuda-gdb


gmarkall:
**Describe the bug**
For example in the [`sessionize.py`](https://nvidia.github.io/numba-cuda/user/examples.html#dividing-click-data-into-sessions) example, the grid group appears to be represented by a float.

**Steps/Code to reproduce bug**

Execute the script under `cuda-gdb` and break in the kernel, and then step through until `grid` is defined, then print it out:

```
$ cuda-gdb --args python sessionize.py 
NVIDIA (R) cuda-gdb 12.x
...

(cuda-gdb) set cuda break_on_launch application 
(cuda-gdb) run
Starting program: /home/gmarkall/miniforge3/envs/numbadev/bin/python sessionize.py
...

[Switching focus to CUDA kernel 0, grid 1, block (0,0,0), thread (0,0,0), device 0, sm 0, warp 0, lane 0]

CUDA thread hit application kernel entry function breakpoint, __main__::sessionize[abi:v1][abi:cw51cXTLSUwv1sDUaKthoaNgqamjgM2RtqBUBpSCpRooHz0BIAlD0hqSAGr6gUkg_2bce2Wh0WyupWCtDYUgdKAnkGQFY6kAZ5plYTAA_3d_3d](Array<long long, 1, C, mutable, aligned>, Array<long long, 1, C, mutable, aligned>, Array<double, 1, C, mutable, aligned>)
   <<<(1,1,1),(576,1,1)>>> (user_id=..., timestamp=..., results=...) at sessionize.py:12

12	    gid = cuda.grid(1)
(cuda-gdb) n
... a few more "next"s ...

(cuda-gdb) n
39	    grid = cuda.cg.this_grid()
(cuda-gdb) n
40	    grid.sync()
(cuda-gdb) p grid
$6 = 1.1125369292536002e-308
```

**Expected behavior**

I'm not sure what the best representation is (perhaps "this_grid", or something mentioning that it's a group and specifying the handle of the group), but printing out the data as a float is just about the least helpful thing that could possibly be done.

**Environment details (please complete the following information):**

Numba with patches:

```diff
diff --git a/numba/core/cgutils.py b/numba/core/cgutils.py
index 04bb3cb5c..9177e893c 100644
--- a/numba/core/cgutils.py
+++ b/numba/core/cgutils.py
[...]

PR #131, opened 2025-02-21
Fix the debug info of GridGroup type

jiel-nv:
The customized CUDA front end type 'GridGroup' maps to the back end type 'ir.IntType(64)' which is defined in numba.cuda.models.

The numba DIBuilder emit variable types for simple numeric types by looking into back end type which falls through the 'DW_ATE_float' because the 'datamodel.fe_type' is not 'types.Integer'.

So emitting this type through the CUDADIBuilder is needed. Since the type represents the handle of a group object, giving it the encoding of 'DW_ATE_unsigned' should be reasonable.

A test is also added in this change.

This change addresses https://github.com/NVIDIA/numba-cuda/issues/107

### DIFF
--- a/numba_cuda/numba/cuda/debuginfo.py
+++ b/numba_cuda/numba/cuda/debuginfo.py
@@ -1,6 +1,7 @@
 from llvmlite import ir
 from numba.core import types
 from numba.core.debuginfo import DIBuilder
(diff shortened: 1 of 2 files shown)


gmarkall (approved):
Many thanks! Just waiting for CI.

PR #132, opened 2025-02-21
Allow `link` argument to accept a bitcode file

isVoid:
This is a PoC on allowing llvm bitcode being linked into Numba.


gmarkall:
What's the thinking behind implementing this in the linker and not the codegen? I think we discussed this before and from what I recall, I had been approaching thinking about this as doing it in the codegen, adding it to the list of IRs that gets compiled here: https://github.com/NVIDIA/numba-cuda/blob/main/numba_cuda/numba/cuda/codegen.py#L149

isVoid:
Agreed, closing this. Working on a replacement branch.

Issue #129, opened 2025-02-20
[MNT] Replace all internal `Linker`-related classes by `cuda.core.(experimental).Linker`


leofang:
As the first step of #128 I would like to laser-focus on the linker replacement, since it affects multiple projects (nvmath.device, cuda.cooperative, Numbast, nvshmem, RAPIDS, …). I think there are two things that need to happen as part of the transition: One is in pynvjitlink (https://github.com/rapidsai/pynvjitlink/issues/130) and another is below.

I think all Linker related classes in `numba/cuda/cudadrv/driver.py` can be immediately replaced by `cuda.core.Linker`. If this does not work we want to know why, and potentially fix it asap in cuda-core.

brandon-b-miller:
If someone isn't already addressing this I am happy to start taking a look :)

leofang:
Thanks, @brandon-b-miller! No one is working on this, so if you have time to help and give it a shot it'd be great!

tpn:
Just out of interest, is this still the leading issue for tracking blockers related to pynvjitlink deprecation (and thus, the things preventing Windows use re https://github.com/NVIDIA/cccl/issues/3783)?  @brandon-b-miller have you made any progress, or do you need any help?

brandon-b-miller:
Hi @tpn , the cross referenced PR should be ready for review over the next day or two.

ZzEeKkAa:
Since #133 is merged, does it mean that this issue could be closed?

brandon-b-miller:
> Since [#133](https://github.com/NVIDIA/numba-cuda/pull/133) is merged, does it mean that this issue could be closed?

For CUDA 12+ the answer is yes, but for cuda 11 we still have the MVCLinker here

https://github.com/NVIDIA/numba-cuda/blob/main/numba_cuda/numba/cuda/cudadrv/driver.py#L3090

I think once that's no longer needed, this issue can be closed.

brandon-b-miller:
With the removal of `MVCLinker` this can be closed.

PR #133, opened 2025-02-21
Use `cuda.bindings` and `cuda.core` for `nvjitlink`

brandon-b-miller:
WIP
xref https://github.com/NVIDIA/numba-cuda/issues/129

### DIFF
--- a/numba_cuda/numba/cuda/cudadrv/driver.py
+++ b/numba_cuda/numba/cuda/cudadrv/driver.py
@@ -47,6 +47,9 @@ try:
 except ImportError:
     NvJitLinker, NvJitLinkError = None, None
 
+from cuda.bindings.nvjitlink import nvJitLinkError
+from cuda.bindings import nvjitlink as _nvjitlink
+
 USE_NV_BINDING = config.CUDA_USE_NVIDIA_BINDING
 
 if USE_NV_BINDING:
@@ -2594,7 +2597,7 @@ class Linker(metaclass=ABCMeta):
                 "Enabling pynvjitlink requires CUDA 12."
             )
         if config.CUDA_ENABLE_PYNVJITLINK:
-            linker = PyNvJitLinker
+            linker = PyNvJitLinkerNew
 
         elif config.CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY:
             linker = MVCLinker
@@ -3022,6 +3025,143 @@ class CudaPythonLinker(Linker):
         # We return a copy of the cubin because it's owned by the linker
         cubin_ptr = ctypes.cast(cubin_buf, ctypes.POINTER(ctypes.c_char))
         return bytes(np.ctypeslib.as_array(cubin_ptr, shape=(size,)))
+
+
+class PyNvJitLinkerNew(Linker):
+    def __init__(
+        self,
+        max_registers=None,
+        lineinfo=False,
+        cc=None,
+        lto=False,
+        additional_flags=None,
+    ):
+
+        if cc is None:
+            raise RuntimeError("NvJitLink requires CC to be specified")
+        if not any(isinstance(cc, t) for t in [list, tuple]):
+            raise TypeError("`cc` must be a list or tuple of length 2")
+
+        sm_ver = f"{cc[0] * 10 + cc[1]}"
+        arch = f"-arch=sm_{sm_ver}"
+        options = [arch]
+        if max_registers:
+            options.append(f"-maxrregcount={max_registers}")
+        if lineinfo:
+            options.append("-lineinfo")
+        if lto:
+            options.append("-lto")
+        if additional_flags is not None:
+            options.extend(additional_flags)
+
+        self.handle = _nvjitlink.create(len(options), options)
+        self.lto = lto
+        self.options = options
+
+        self._info_log = None
+        self._error_log = None
+        self._complete = False
+
+    @property
+    def info_log(self):
+        return self._info_log
+
+    @property
+    def error_log(self):
+        return self._error_log
+
+    def add_data(self, input_type, data, name):
+        if self._complete:
+            raise nvJitLinkError("Cannot add data to already-completeted link")
+
+        try:
+            _nvjitlink.add_data(
+                self.handle,
+                input_type.value,
+                data,
+                len(data),
+                name
+            )
+        except RuntimeError as e:
+            log_size = _nvjitlink.get_error_log_size(self.handle)
+            log = bytearray(log_size)
+            self._info_log = _nvjitlink.get_error_log(self.handle, log)
+
+            log_size = _nvjitlink.get_info_log_size(self.handle)
+            log = bytearray(log_size)
+            self._error_log = _nvjitlink.get_info_log(self.handle, log)
+            raise nvJitLinkError(f"{e}\n{self.error_log}")
+
+    def add_cubin(self, cubin, name=None):
+        name = name or "unnamed-cubin"
+        self.add_data(_nvjitlink.InputType.CUBIN, cubin, name)
+
+    def add_ptx(self, ptx, name=None):
+        name = name or "unnamed-ptx"
+        self.add_data(_nvjitlink.InputType.PTX, ptx, name)
+
+    def add_ltoir(self, ltoir, name=None):
+        name = name or "unnamed-ltoir"
+        self.add_data(_nvjitlink.InputType.LTOIR, ltoir, name)
+
+    def add_object(self, object_, name=None):
+        name = name or "unnamed-object"
+        self.add_data(_nvjitlink.InputType.OBJECT, object_, name)
+
+    def add_fatbin(self, fatbin, name=None):
+        name = name or "unnamed-fatbin"
+        self.add_data(_nvjitlink.InputType.FATBIN, fatbin, name)
+
+    def add_library(self, library, name=None):
+        self.add_data(_nvjitlink.InputType.LIBRARY, library, name)
+
+    def add_file(self, path, kind):
+        try:
+            with open(path, "rb") as f:
+                data = f.read()
+        except FileNotFoundError:
+            raise LinkerError(f"{path} not found")
+
+        name = pathlib.Path(path).name
+        self.add_data(data, kind, name)
+
+    def get_linked_cubin(self):
+        try:
+            _nvjitlink.complete(self.handle)
+            self._complete = True
+            size = _nvjitlink.get_linked_cubin_size(self.handle)
+            cubin = bytearray(size)
+
+            _nvjitlink.get_linked_cubin(self.handle, cubin)
+            return cubin
+
+        except nvJitLinkError as e:
+            size = _nvjitlink.get_error_log_size(self.handle)
+            log = bytearray(size)
+            self._error_log = _nvjitlink.get_error_log(self.handle, log)
+            raise nvJitLinkError(f"{e}\n{self.error_log}")
+        finally:
+            size = _nvjitlink.get_info_log_size(self.handle)
+            log = bytearray(size)
+            self._info_log = _nvjitlink.get_info_log(self.handle, log)
+
+    def get_linked_ptx(self):
+        try:
+            _nvjitlink.complete(self.handle)
+            self._complete = True
+            size = _nvjitlink.get_linked_ptx_size(self.handle)
+            return _nvjitlink.get_linked_ptx(self.handle, size)
+        except RuntimeError as e:
+            self._error_log = _nvjitlink.get_error_log(self.handle)
+            raise NvJitLinkError(f"{e}\n{self.error_log}")
+        finally:
+            self._info_log = _nvjitlink.get_info_log(self.handle)
+
+    def complete(self):
+        try:
+            return self.get_linked_cubin()
+        except NvJitLinkError as e:
+            raise LinkerError from e
 
 
 class PyNvJitLinker(Linker):


leofang:
Thanks, @brandon-b-miller. Remember our goal is to drop every Linker subclasses inside Numba, in favor of `cuda.core.Linker`. The current PR is not what we want. Also note that to help pynvjitlink to phase out, we already have https://github.com/rapidsai/pynvjitlink/pull/111 which is essentially what this PR does today.

numba_cuda/numba/cuda/codegen.py
```diff
@@ -246,10 +246,11 @@ def get_cufunc(self):
             return cufunc
 
         cubin = self.get_cubin(cc=device.compute_capability)
-        module = ctx.create_module_image(cubin)
+        #module = ctx.create_module_image(cubin)
 
         # Load
-        cufunc = module.get_function(self._entry_name)
+        #cufunc = module.get_function(self._entry_name)
+        cufunc = cubin.get_kernel(self._entry_name)
```
isVoid, line 253: Since we switched to context independent loading API, the `CUDADispatcher.bind` method should probably be renamed since it no longer binds context via calling the `get_cufunc` function.

brandon-b-miller:
@gmarkall @leofang numba-cuda contains [nvjitlink tests](https://github.com/NVIDIA/numba-cuda/blob/df6d56a2f029f61ba4de91653ef5dc9dbff78a6c/numba_cuda/numba/cuda/tests/cudadrv/test_nvjitlink.py#L36), should we maintain support for these as part of this PR or drop them in favor of testing in upstream `cuda-python`?

gmarkall:
> @gmarkall @leofang numba-cuda contains [nvjitlink tests](https://github.com/NVIDIA/numba-cuda/blob/df6d56a2f029f61ba4de91653ef5dc9dbff78a6c/numba_cuda/numba/cuda/tests/cudadrv/test_nvjitlink.py#L36), should we maintain support for these as part of this PR or drop them in favor of testing in upstream `cuda-python`?

I think we need to maintain the tests that test Numba-CUDA's interaction with the linker, like the `TestLinkerUsage` class and the `test_*_with_linkable_code` (tests with names like that). I don't think we need to keep the tests that purely test the `PyNvJitLinker` API like the ones that test passing different flags etc. to it.

gmarkall:
Also, I think we can probably delete the `PyNvJitLinker` class in this PR as well - is there any reason to keep it around?

I'm comfortable with:

- Using `cuda.core.Linker` when the user asks for pynvjitlink or the NVIDIA bindings, and
- Using the ctypes linker otherwise

which is what this PR seems to offer. (correct me if I've read it wrong :slightly_smiling_face:)

brandon-b-miller:
> Also, I think we can probably delete the `PyNvJitLinker` class in this PR as well - is there any reason to keep it around?
> 
> I'm comfortable with:
> 
> * Using `cuda.core.Linker` when the user asks for pynvjitlink or the NVIDIA bindings, and
> * Using the ctypes linker otherwise
> 
> which is what this PR seems to offer. (correct me if I've read it wrong 🙂)

Correct, this is the outcome I am aiming for.

brandon-b-miller:
@gmarkall on second thought, we might need to leave the MVCLinker in in some capacity as long as we're supporting cuda 11. I don't _think_ that `cuda-python` supports the functionality that cubinlinker enables.

gmarkall:
@brandon-b-miller Sorry, yes - I had that in mind but didn't write it down.

brandon-b-miller:
> @brandon-b-miller Sorry, yes - I had that in mind but didn't write it down.

Ok, just to have it written down somewhere, after this PR we will:

**For cuda 11, maintain** the current way of configuring which bindings to use:
- Default ctypes bindings, optional `cuda-python` bindings with `NUMBA_CUDA_USE_NVIDIA_BINDING=1`, optional `MVCLinker` with `NUMBA_CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY=1`. 

**for cuda 12**, we will have:
- Default ctypes bindings, optional `cuda-python` bindings with `NUMBA_CUDA_USE_NVIDIA_BINDING=1`
- Use of pynvjitlink _through_ `cuda-python` if `NUMBA_CUDA_ENBALE_PYNVJITLINK=1`

This will leave us with 3 linkers:
- The ctypes linker which is used by default regardless of cuda version
- the mvc linker which is used in a cuda 11 environment when mvc is required, regardless of what binding is being used
- the new linker which is used in a cuda 12 enviornment either the `cuda-python` bindings _or_ `pynvjitlink` is enabled

gmarkall:
Thanks for the summary! To look a little further ahead, we want to end up with only one linker, which is the new linker. This would be achieved by deprecating / removing the other linkers as soon as appropriate:

* `MVCLinker` can be removed as soon as CUDA 11 support is dropped.
* The `ctypes` linker can be deprecated and removed whenever we can have a hard dependency on `cuda.core` and had tested the new linker in use for a bit to shake out any issues.

Are you in alignment with the above plan @brandon-b-miller ?

brandon-b-miller:
> Thanks for the summary! To look a little further ahead, we want to end up with only one linker, which is the new linker. This would be achieved by deprecating / removing the other linkers as soon as appropriate:
> 
> * `MVCLinker` can be removed as soon as CUDA 11 support is dropped.
> * The `ctypes` linker can be deprecated and removed whenever we can have a hard dependency on `cuda.core` and had tested the new linker in use for a bit to shake out any issues.
> 
> Are you in alignment with the above plan @brandon-b-miller ?

Yup this sounds good to me.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2749,6 +2755,180 @@ def complete(self):
...
+        # TODO - hack
+        from cuda.bindings import nvjitlink
+        from cuda.core.experimental import _linker
+        _linker._nvjitlink_input_types['object'] = nvjitlink.InputType.OBJECT
+        ObjectCode._supported_code_type = (*ObjectCode._supported_code_type, 'object')
+        obj = ObjectCode._init(obj, 'object')
+        self._object_codes.append(obj)
+    
+
+    def add_library(self, lib, name='<cudapy-lib>'):
```
brandon-b-miller, line 2864: There's a few cases in this class where I'm finding I have to do something like this. We discussed wanting to avoid having an `ObjectCode` constructor for every type of possible input, but it seems like numba requires the ability to assemble the eventual inputs to the link from a pretty broad variety of sources.

gmarkall: It seems like we need an API to be able to construct `ObjectCode` instances from inputs of arbitrary kinds - archives, cubins, LTOIR, fatbins, etc.. It seems like the existing interface is aimed at providing something relatively safe, but doesn't allow for the lower-level / fine-grained control we need here to set up and complete a link.

Is that understanding correct?

brandon-b-miller: This is what I've arrived at doing, @leofang and I have had a few conversations about how this capability would fit into the broader goals of the `cuda-python` API, I'd be curious his thoughts.

ZzEeKkAa:
What is the status of the PR? Can get use of it at [nvmath-python](https://github.com/NVIDIA/nvmath-python)

brandon-b-miller:
@ZzEeKkAa I merged/resolved from tree yesterday, I'll push the changes today. This is coming back up in the priority queue now so you can expect some more movement soon. 

As for the status of the PR, I believe last I updated things tests were passing on both cuda 11 and 12, however doing this involved hacking `cuda-python` to support more `ObjectCode` constructors, so we'll either need consensus with the `cuda-python` team on how to proceed there (cc @leofang) or a different solution before the code is closer to a mergeable state.

numba_cuda/numba/cuda/cudadrv/driver.py
leofang: My question is if we really have to keep Numba's `Linker` class? Is it a public API that we must keep indefinitely?

We discussed in a past `cuda.core` design review meeting that Numba should just drop all of these linker classes once `cuda.core` is ready. In particular, we agreed that the `Linker` class can just take all inputs at once, instead of having all kinds of `add_<something>()` APIs to add things one by one. Do we really need to bend backwards here?

leofang:
`cuda-core` needs to be added to the dependency list so that it can be tested in the CI.

numba_cuda/numba/cuda/cudadrv/driver.py
leofang: xref: https://github.com/NVIDIA/cuda-python/issues/629

brandon-b-miller:
with https://github.com/NVIDIA/cuda-python/pull/652, I pass tests locally except for a few caching tests that I believe are failing due to https://github.com/NVIDIA/cuda-python/issues/526.

brandon-b-miller:
Adding the changes from https://github.com/NVIDIA/cuda-python/pull/660 on top of my local copy of `cuda-python` passes the caching tests.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2904,6 +2909,176 @@ def complete(self):
...
+
+    @property
+    def error_log(self):
+        if not self.linker:
+            raise ValueError("Not Initialized")
+        if self._complete:
+            return self._error_log
+        return self.linker.get_error_log()
+
+    def add_ptx(self, ptx, name="<cudapy-ptx>"):
```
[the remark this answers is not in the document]
gmarkall: Providing the name to the linker allows it to be presented in error messages that come from the linker. It would be a pain to debug for that information to be lost, given that it's becoming more common to link multiple code objects together.

brandon-b-miller:
Mostly blocked on upstream `cuda-python` PRs here now.

brandon-b-miller:
This PR will need to depend on a new release of `cuda-core` to pass CI but is otherwise ready for review.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2746,22 +2754,21 @@ def new(
         if config.CUDA_ENABLE_PYNVJITLINK and driver_ver < (12, 0):
             raise ValueError("Enabling pynvjitlink requires CUDA 12.")
         if config.CUDA_ENABLE_PYNVJITLINK:
-            linker = PyNvJitLinker
```
gmarkall, line 2749: `CudaPythonLinker` and `PyNvJitLinker` look like dead code with this change. Is there any reason not to remove them?

gmarkall:
I think the docs on MVC need to be revised as this seems to remove the use of pynvjitlink - in particular, https://nvidia.github.io/numba-cuda/user/minor_version_compatibility.html

gmarkall:
The following:

```python
from numba import cuda

nonexistent = cuda.declare_device("nonexistent", "int32()")


@cuda.jit
def bad_kernel():
    nonexistent()


bad_kernel[1, 1]()
```

gives:

```
  File "/home/gmarkall/numbadev/issues/numba-cuda-133/repro.py", line 1, in <module>
    from numba import cuda
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/__init__.py", line 7, in <module>
    from .device_init import *
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/device_init.py", line 3, in <module>
    from numba.cuda import cg
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/cg.py", line 4, in <module>
    from numba.cuda import nvvmutils
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/nvvmutils.py", line 4, in <module>
    from .cudadrv import nvvm
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/cudadrv/nvvm.py", line 16, in <module>
    from .libs import get_libdevice, open_libdevice, open_cudalib
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/cudadrv/libs.py", line 19, in <module>
    from numba.cuda.cudadrv.driver import locate_driver_and_loader, load_driver
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/cudadrv/driver.py", line 3005
    warnings.warn(f"NVRTC log messages: {'\n'.join(logger.log)}")
                                                                ^
SyntaxError: f-string expression part cannot include a backslash
```

gmarkall:
Actually, it seems the error was nothing to do with my code in particular - it just seems to be an error in general with the parsing of driver.py?

brandon-b-miller:
> ```SyntaxError: f-string expression part cannot include a backslash```

that's weird, not sure why that would occur with these changes. Almost like a python version issue.

brandon-b-miller:
@gmarkall if you are testing locally, you'll need `cuda-python` built from source on main.

gmarkall:
> that's weird, not sure why that would occur with these changes. Almost like a python version issue.

Trying CI now to see if the same issue occurs there.

gmarkall:
> @gmarkall if you are testing locally, you'll need `cuda-python` built from source on main.

Thanks. I'll try again with `cuda-python` built from source on `main`. For now I notice errors are being emitted to stderr / stdout instead of being captured, along with a bit of garbage:

```
$ NUMBA_CUDA_ENABLE_PYNVJITLINK=1 python repro.py 
/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/dispatcher.py:680: NumbaPerformanceWarning: Grid size 1 will likely result in GPU under-utilization due to low occupancy.
  warn(NumbaPerformanceWarning(msg))
@E@error   : Undefined reference to 'nonexistent' in 'None_ptx'
  File "/home/gmarkall/numbadev/issues/numba-cuda-133/repro.py", line 11, in <module>
    bad_kernel[1, 1]()
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 683, in __call__
    return self.dispatcher.call(
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 829, in call
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 837, in _compile_for_args
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba/numba/core/compiler_lock.py", line 35, in _acquire_compile_lock
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 1105, in compile
    kernel.bind()
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 321, in bind
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/codegen.py", line 321, in get_cufunc
    cubin = self.get_cubin(cc=device.compute_capability)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/codegen.py", line 300, in get_cubin
    cubin = linker.complete()
            ^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/cudadrv/driver.py", line 3084, in complete
    result = self.linker.link("cubin")
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/miniforge3/envs/numbadev/lib/python3.11/site-packages/cuda/core/experimental/_linker.py", line 442, in link
    _nvjitlink.complete(self._mnff.handle)
  File "cuda/bindings/nvjitlink.pyx", line 156, in cuda.bindings.nvjitlink.complete
  File "cuda/bindings/nvjitlink.pyx", line 166, in cuda.bindings.nvjitlink.complete
  File "cuda/bindings/nvjitlink.pyx", line 66, in cuda.bindings.nvjitlink.check_status
cuda.bindings.nvjitlink.nvJitLinkError: ERROR_INTERNAL (6)
Linker error log: ERROR 9: finish
```

brandon-b-miller:
@gmarkall are you still having an issue running that reproducer code? Locally I get what I believe the be the expected error (`CUDA_ERROR_INVALID_SOURCE` followed by `error  : Undefined reference to 'nonexistent' in '<unnamed-ptx>'`. I can't seem to reproduce the f-string thing.

numba_cuda/numba/cuda/cudadrv/driver.py
brandon-b-miller: Good call, removed in https://github.com/NVIDIA/numba-cuda/pull/133/commits/c32a8508d6702ed7869fa65fcabd9e214f856cb2.

brandon-b-miller:
I repurposed the `pynvjitlink` specific CI scripts for testing the bindings. @gmarkall what should the fate of `NUMBA_CUDA_ENABLE_PYNVJITLINK` be after this PR?

pyproject.toml
```diff
@@ -18,7 +18,7 @@ authors = [
 ]
 license = { text = "BSD 2-clause" }
 requires-python = ">=3.9"
-dependencies = ["numba>=0.59.1"]
+dependencies = ["numba>=0.59.1", "cuda-bindings>=12.9.0", "cuda-core==0.2.0"]
```
leofang: Let's get #278 merged first, and them add `cuda-core[cu12]==0.2.0` to the new `cu12` optional dependencies.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2904,6 +2909,183 @@ def complete(self):
...
+                max_register_count=self.max_registers,
+                relocatable_device_code=True,
+                include_path=self._include_paths,
+                link_time_optimization=self.lto,
+            ),
+        )
+        with warnings.catch_warnings():
+            warnings.filterwarnings(
+                "ignore", message=("The CUDA driver version is older")
+            )
```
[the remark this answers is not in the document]
leofang: I think it is a downstream (i.e. numba-cuda) issue, not upstream issue. Why would we want to allow compiling into PTX (#262) instead of LTOIR if the assumption is nvJitLink is available always (since we're dropping CUDA 11)? 

The upstream warning is clear. If a user wants to generate PTX whose version is newer than the driver, the driver cannot load it, and we should just not generate PTX in the first place.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2916,6 +2921,183 @@ def complete(self):
         """
 
 
+class CUDALinker(Linker):
```
leofang: I feel I keep forgetting about this: Does numba-cuda's `Linker` class a public API that we must expose? If so, now that people can access `cuda.core.Linker` can we consider deprecating it from numba-cuda?

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2916,6 +2921,183 @@ def complete(self):
...
+        result = self.linker.link("ptx")
+        self.close()
+        self._complete = True
+        return result
+
+    def close(self):
+        self._info_log = self.linker.get_info_log()
+        self._error_log = self.linker.get_error_log()
+        self.linker.close()
+
```
leofang: I don't get why we need these two methods when there's already a `.complete()` below? Other linker classes don't have them?

numba_cuda/numba/cuda/cudadrv/driver.py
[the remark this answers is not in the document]
leofang: Will be resolved by https://github.com/NVIDIA/cuda-python/pull/682.

numba_cuda/numba/cuda/cudadrv/driver.py
gmarkall: It's not a public API - I've only treated it as public in as far as pynvjitlink used to patch it, so it needed to remain consistent.

brandon-b-miller: AFAICT the linker interface is not documented as public anywhere. There's technically nothing signaling to an advanced user that they shouldn't use it though, I suppose the class could use a `_` and I could do some general renaming/docs.

numba_cuda/numba/cuda/cudadrv/driver.py
gmarkall: It's a bit convoluted, but this method is used when LTO is enabled, to provide the PTX that is eventually produced after linking. As only the pynvjitlink linker could do LTO, only the PyNvJitLinker linker had this method before.

numba_cuda/numba/cuda/cudadrv/driver.py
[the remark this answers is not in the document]
brandon-b-miller: Trying to make sure I understand the action item here. If we plan to add `nvjitlink>=12.3` as a hard dependency, should this linker _always_ have `lto` enabled for everything?

pyproject.toml
gmarkall: Does pinning to cuda-core 0.2.0 potentially cause issues for people that want to use later versions, or other libraries that require greater than 0.2.0? Or do we expect to move forward quickly when new cuda-core versions are released?

brandon-b-miller: This 0.2.0 pinning was just temporary while a new release of `cuda-python` is cut, to try and tease out any CI script errors, etc. I expect this will change before merge.

numba_cuda/numba/cuda/codegen.py
```diff
@@ -280,14 +283,17 @@ def get_cubin(self, cc=None):
...
 
             print(("ASSEMBLY (AFTER LTO) %s" % self._name).center(80, "-"))
             print(ptx)
             print("=" * 80)
 
-        linker = driver.Linker.new(
+        linker = driver._Linker.new(
             max_registers=self._max_registers, cc=cc, lto=self._lto
         )
         self._link_all(linker, cc, ignore_nonlto=False)
```
gmarkall, line 296: I can't comment on line 294 / 300 just below as it's not part of the diff, but I suspect if it was changed so that `cubin = linker.complete().code` if the NVIDIA binding is in use, then the other changes to use `<variable>.code` (https://github.com/NVIDIA/numba-cuda/pull/133/files#diff-bf7c3a638421914e7463c6a23bd6293b740ed0c3e22267a7469091f70a9ad746R26, https://github.com/NVIDIA/numba-cuda/pull/133/files#diff-626c1086aae9de03cd89533ef1e1b9f7b1279e834b54aab10ab3c271739def53R1690) wouldn't be needed and the interface would be more uniform across the linkers. (Not a necessary change as a lot of this stuff will churn anyway as we move more to cuda-core / cuda-bindings, but noticed the use of `.code` in other places).

numba_cuda/numba/cuda/cudadrv/driver.py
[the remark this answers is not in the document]
gmarkall: I think use of pynvjitlink and LTO are distinct - it's the use of pynvjitlink that facilitates MVC, and LTO can be off or on with it.

gmarkall:
/ok to test [1585116](https://github.com/NVIDIA/numba-cuda/pull/133/commits/1585116511132e808f677567818690fb6457c1c3)

numba_cuda/numba/cuda/cudadrv/driver.py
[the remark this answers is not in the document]
brandon-b-miller: Is the thing to do just to switch a default somewhere then? Currently the linker should work with `lto=True` but it appears nothing ever sets that in the test suite and there's not a way for the user to do it.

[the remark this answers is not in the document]
brandon-b-miller: I think https://github.com/NVIDIA/numba-cuda/pull/133/commits/c284f614e65ff021cdea5db9376a5871755f7f1e _might_ be what we want - LTO by default but still make it technically possible internally not to

numba_cuda/numba/cuda/cudadrv/driver.py
brandon-b-miller: Renamed/ prefixed things in https://github.com/NVIDIA/numba-cuda/pull/133/commits/9f0fd746f23b38b9a37920c5104fb8b25406e6e3, let me know if this seems like its appropriate here

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2916,6 +2921,184 @@ def complete(self):
         """
 
 
+class _Linker(_LinkerBase):
```
gmarkall, line 2934: It seems like this class is the interface used elsewhere - e.g. in `codegen`, so implying it is private seems a little strange. Should it be
```suggestion
class Linker(_LinkerBase):
```

?

brandon-b-miller: Since we're pushing `cuda.core.Linker` as the public facing linker interface, I tried to avoid having something else called that in the codebase. We don't have to make it private though, it only is due to https://github.com/NVIDIA/numba-cuda/pull/133#discussion_r2130488367.

pyproject.toml
leofang: Yeah we'll probably need to change it to `cuda-core>=0.3.0` before we merge this PR or cut a release here

numba_cuda/numba/cuda/cudadrv/driver.py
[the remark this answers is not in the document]
leofang: btw I think we don't need to set any version constraint for nvJitLink -- we did that already in `cuda-bindings`:
https://github.com/NVIDIA/cuda-python/blob/9ef1562c3902d13884a831b1c6970548d56ef81a/cuda_bindings/pyproject.toml#L37
which `cuda-python` inherits:
https://github.com/NVIDIA/cuda-python/blob/9ef1562c3902d13884a831b1c6970548d56ef81a/cuda_python/setup.py#L22

Bottom line: Instead of installing `cuda-python`, do `cuda-python[all]` or `cuda-bindings[all]`.

numba_cuda/numba/cuda/cudadrv/driver.py
leofang: thx for clarifying

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2741,22 +2760,21 @@ def new(
...
 
-        if linker is PyNvJitLinker:
-            return linker(max_registers, lineinfo, cc, lto, additional_flags)
-        elif additional_flags or lto:
-            raise ValueError("LTO and additional flags require PyNvJitLinker")
-        else:
-            return linker(max_registers, lineinfo, cc)
+        params = (max_registers, lineinfo, cc)
+        if linker is _Linker:
+            params = (*params, driver_ver >= (12, 0), additional_flags)
```
brandon-b-miller: This is a small snag I should have forseen. It's not enough to just check the driver version, we need to know if `nvjitlink` is available to decide if we want to LTO everything. 

Previously we could use the presence of `pynvjitlink` as a proxy for the existence of `nvjitlink`, but the same isn't true with this PR. Currently this line causes CI to fail when we have a cuda 12 driver and an 11.8 ctk, because numba has no way of knowing not to generate LTOIR. 

I can't really see a way of getting around calling a [cuda-python private function](https://github.com/NVIDIA/cuda-python/blob/main/cuda_core/cuda/core/experimental/_linker.py#L32) to find out if `nvjitlink` is usable. The only public option would be to instantiate a real `Linker` instance and check its `backend` just to throw it away. What do you think @kkraus14 @leofang ?

brandon-b-miller: WAR here https://github.com/NVIDIA/numba-cuda/pull/133/commits/143732944ceb14324fc919d72944d38a26739462

brandon-b-miller:
The tests here now pass.

conda/recipes/numba-cuda/meta.yaml
```diff
@@ -29,6 +29,8 @@ requirements:
     - python
     - numba >=0.59.1
     - cuda-bindings
+    - cuda-core ==0.3.0
+    - libnvjitlink
```
kkraus14: I think this would only work with CUDA 12+ and `cuda-bindings 12` pulls this in as a dependency already: https://github.com/conda-forge/cuda-python-feedstock/blob/43add1841aaf2f4c5acabf340adc9eb21fe5cd29/recipe/meta.yaml#L138

numba_cuda/numba/cuda/__init__.py
```diff
@@ -43,6 +42,16 @@
             "bindings."
         )
 
+if ENABLE_PYNVJITLINK:
+    if USE_NV_BINDING:
+        warnings.warn(
+            "Explicitly enabling PyNvJitLink no longer necessary. "
+            "To link through the driver, set the ctypes bindings  "
+            "through NUMBA_CUDA_USE_NVIDIA_BINDING=0. "
```
kkraus14: I think we need to change this warning message a bit. Linking through the driver is now an implementation detail handled by `cuda-core` where we shouldn't tell people to use the ctypes bindings for it.

numba_cuda/numba/cuda/tests/cudadrv/test_linker.py
```diff
@@ -107,7 +109,7 @@ class TestLinker(CUDATestCase):
     @require_context
     def test_linker_basic(self):
         """Simply go through the constructor and destructor"""
-        linker = Linker.new(cc=(5, 3))
+        linker = _Linker.new(cc=(5, 3))
```
kkraus14: Unrelated to this PR but we updated things to 7.5 in #275. Even though this is the simulator would be nice to be consistent:
```suggestion
        linker = _Linker.new(cc=(7, 5))
```

numba_cuda/numba/cuda/__init__.py
brandon-b-miller: How about something like
```suggestion
            "To opt out of nvJitLink usage, set  "
            "NUMBA_CUDA_USE_NVIDIA_BINDING=0. "
```
?

kkraus14: I think the only warning we should be presenting to the user is that if they set `NUMBA_CUDA_USE_NVIDIA_BINDING=0` they are opting into using the deprecated ctypes bindings. I don't think it should be related to nvjitlink.

kkraus14: We could warn along the lines of "If `USE_NV_BINDING` is set, then nvjitlink is used via `cuda.core` instead of `pynvjitlink` so `ENABLE_PYNVJITLINK` is ignored."

brandon-b-miller: Does this capture it?
```suggestion
            "NVIDIA Binding enabled. cuda.core will be used "
            "to link, not pynvjitlink."
```

conda/recipes/numba-cuda/meta.yaml
kkraus14:
```suggestion
```

kkraus14 (approved):

conda/recipes/numba-cuda/meta.yaml
```diff
@@ -29,6 +29,7 @@ requirements:
     - python
     - numba >=0.59.1
     - cuda-bindings
+    - cuda-core ==0.3.0
```
leofang: nit:
```suggestion
    - cuda-core ==0.3.*
```

pyproject.toml
```diff
@@ -23,17 +23,20 @@ dependencies = ["numba>=0.59.1"]
 [project.optional-dependencies]
 cu11 = [
     "cuda-bindings==11.8.*",
+    "cuda-core==0.3.0",
```
leofang:
```suggestion
    "cuda-core==0.3.*",
```

pyproject.toml
```diff
@@ -23,17 +23,20 @@ dependencies = ["numba>=0.59.1"]
...
     "cuda-python==11.8.*",  # supports all CTK 11.x
     "nvidia-cuda-nvcc-cu11",  # for libNVVM
     "nvidia-cuda-runtime-cu11",
     "nvidia-cuda-nvrtc-cu11",
 ]
 cu12 = [
     "cuda-bindings==12.9.*",
+    "cuda-core==0.3.0",
```
leofang:
```suggestion
    "cuda-core==0.3.*",
```

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -387,54 +394,83 @@ def compile(src, name, cc, ltoir=False):
...
+            *includes,
+            "-rdc",
+            "true",
+        ]
+
+        if ltoir:
+            options.append("-dlto")
+
+        if nvrtc_version < (12, 0):
+            options += ["-std=c++17"]
```
leofang: Q: Is this because C++17 was not the default in 11.x?

numba_cuda/numba/cuda/tests/cudadrv/test_linker.py
```diff
@@ -1,13 +1,15 @@
 import numpy as np
 import warnings
+from numba import config
 from numba.cuda.testing import unittest
 from numba.cuda.testing import skip_on_cudasim, skip_if_cuda_includes_missing
 from numba.cuda.testing import CUDATestCase, test_data_dir
-from numba.cuda.cudadrv.driver import CudaAPIError, Linker, LinkerError
-from numba.cuda.cudadrv.error import NvrtcError
+from numba.cuda.cudadrv.driver import CudaAPIError, _Linker, LinkerError
+from cuda.core.experimental._utils.cuda_utils import NVRTCError
```
leofang: It seems we shouldn't hide the exceptions in a private module 😛 Will fix...

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -27,6 +27,9 @@
 # Result code
 nvrtc_result = c_int
 
+if config.CUDA_USE_NVIDIA_BINDING:
+    from cuda.core.experimental import Program, ProgramOptions
```
leofang, line 31: Thanks for also migrating the NVRTC support 🙏

ci/test_conda_ctypes_binding.sh
leofang, line 1: nice 👍

ci/test_wheel.sh
```diff
@@ -10,7 +10,8 @@ package=$(realpath wheel/numba_cuda*.whl)
 echo "Package path: ${package}"
 python -m pip install \
     "${package}[test]" \
-    "cuda-python==${CUDA_VER_MAJOR_MINOR%.*}.*"
+    "cuda-python==${CUDA_VER_MAJOR_MINOR%.*}.*" \
+    "cuda-core==0.3.0"
```
leofang:
```suggestion
    "cuda-core==0.3.*"
```

ci/test_wheel_ctypes_binding.sh
```diff
@@ -11,7 +11,7 @@ echo "Package path: $package"
 python -m pip install \
     "${package}[test]" \
     cuda-python \
-    "pynvjitlink-cu${CUDA_VER_MAJOR}"
+    "cuda-core==0.3.0"
```
leofang: If this CI is not testing NV bindings (`NUMBA_CUDA_USE_NVIDIA_BINDING=0`), shouldn't we just not install cuda-core, same as what's done in `test_conda_ctypes_bindings.sh`?

numba_cuda/numba/cuda/__init__.py
```diff
@@ -44,6 +46,16 @@
             "bindings."
         )
 
+if config.CUDA_ENABLE_PYNVJITLINK:
+    if USE_NV_BINDING:
+        warnings.warn(
+            "Explicitly enabling PyNvJitLink no longer necessary. "
```
bdice: We've never written this name in CamelCase, just all lowercase.
```suggestion
            "Explicitly enabling pynvjitlink is no longer necessary. "
```

numba_cuda/numba/cuda/__init__.py
```diff
@@ -44,6 +46,16 @@
...
+            "NVIDIA Binding enabled. cuda.core will be used "
+            "to link, not pynvjitlink."
```
bdice:
```suggestion
            "NVIDIA bindings are enabled. cuda.core will be used "
            "in place of pynvjitlink."
```

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -58,6 +59,22 @@
...
+    from cuda.core.experimental import (
+        Linker,
+        LinkerOptions,
+        ObjectCode,
+    )
+
+    # There is no definition of the default stream in the Nvidia bindings (nor
+    # is there at the C/C++ level), so we define it here so we don't need to
+    # use a magic number 0 in places where we want the default stream.
+    CU_STREAM_DEFAULT = 0
```
bdice, line 75: This isn't true. There is a default stream defined in the CUDA Python bindings, mirroring the same default value in the C/C++ runtime.

```python
>>> from cuda.bindings import runtime
>>> runtime.cudaStreamDefault
0
```

Also if this is kept, please write NVIDIA in all caps.
```suggestion
    # There is no definition of the default stream in the NVIDIA bindings (nor
    # is there at the C/C++ level), so we define it here so we don't need to
    # use a magic number 0 in places where we want the default stream.
    CU_STREAM_DEFAULT = 0
```

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2735,30 +2764,24 @@ def new(
...
-        ):
-            raise ValueError("Use CUDA_ENABLE_PYNVJITLINK for CUDA >= 12.0 MVC")
-        if config.CUDA_ENABLE_PYNVJITLINK and driver_ver < (12, 0):
-            raise ValueError("Enabling pynvjitlink requires CUDA 12.")
-        if config.CUDA_ENABLE_PYNVJITLINK:
-            linker = PyNvJitLinker
-
-        elif config.CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY:
-            linker = MVCLinker
+        if driver_ver <= (12, 0):
```
bdice: Should this be `<=` or `<`? Did something change in the drivers for 12.1? Can we explicitly use `< (12, 1)` if that is the case?

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2901,7 +2922,143 @@ def complete(self):
...
+class _Linker(_LinkerBase):
+    def __init__(
+        self,
+        max_registers=None,
+        lineinfo=False,
+        cc=None,
+        lto=None,
+        additional_flags=None,
+    ):
+        arch = f"sm_{cc[0] * 10 + cc[1]}"
```
bdice: Here we do `major * 10 + minor` but elsewhere in `nvrtc.py` we are just using string concatenation like `arch = f"sm_{major}{minor}"`. I think I prefer string concatenation, but being consistent is more important.

```suggestion
        arch = f"sm_{cc[0]}{cc[1]}"
```

numba_cuda/numba/cuda/cudadrv/nvrtc.py
bdice: Let's be consistent in how we append. Above we use `options.append` and here we are doing `+= [...]` addition. I'd recommend sticking to `.append`.
```suggestion
        if nvrtc_version < (12, 0):
            options.append("-std=c++17")
```

numba_cuda/numba/cuda/cudadrv/driver.py
gmarkall: I'm not sure how important the distinction is, but that's from the runtime binding rather than the driver binding. Is there one defined in the driver binding?

(for background - that statement was true at one point, but may have become false at some point if a definition were added to the NVIDIA bindings)

numba_cuda/numba/cuda/cudadrv/driver.py
brandon-b-miller: Bad local merge, thanks 🚀 
```suggestion
        if driver_ver < (12, 0):
```

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2735,30 +2764,24 @@ def new(
...
             if USE_NV_BINDING:
-                linker = CudaPythonLinker
+                linker = _Linker
             else:
                 linker = CtypesLinker
 
-        if linker is PyNvJitLinker:
-            return linker(max_registers, lineinfo, cc, lto, additional_flags)
-        elif additional_flags or lto:
-            raise ValueError("LTO and additional flags require PyNvJitLinker")
```
gmarkall, line 2759: Are we now silently ignoring LTO and additional flags if nvjitlink is not in use instead of reporting it as an error?

numba_cuda/numba/cuda/codegen.py
```diff
@@ -271,7 +274,7 @@ def get_cubin(self, cc=None):
             return cubin
 
         if self._lto and config.DUMP_ASSEMBLY:
-            linker = driver.Linker.new(
+            linker = driver._Linker.new(
```
gmarkall, line 277: Note / nit: this feels a little funny because the `new` method is a class method of `_LinkerBase` rather than `_Linker`.

gmarkall:
When I run the test suite locally, I see a warning even though I don't have the environment variable set:

```
$ python -m numba.runtests numba.cuda.tests -vf -m
/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/__init__.py:51: UserWarning: Explicitly enabling PyNvJitLink no longer necessary. NVIDIA Binding enabled. cuda.core will be used to link, not pynvjitlink.
```

then I'm also hitting

```
ERROR: test_cudajit_in_attached_primary_context (numba.cuda.tests.cudadrv.test_context_stack.Test3rdPartyContext.test_cudajit_in_attached_primary_context)
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/tests/cudadrv/test_context_stack.py", line 141, in test_cudajit_in_attached_primary_context
    self.test_attached_primary(do)
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/tests/cudadrv/test_context_stack.py", line 101, in test_attached_primary
    extra_work()
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/tests/cudadrv/test_context_stack.py", line 138, in do
    foo[1, 1](a)
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 688, in __call__
    return self.dispatcher.call(
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 834, in call
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 842, in _compile_for_args
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba/numba/core/compiler_lock.py", line 35, in _acquire_compile_lock
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 1110, in compile
    kernel.bind()
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 324, in bind
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/codegen.py", line 318, in get_cufunc
    cubin = self.get_cubin(cc=device.compute_capability)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/codegen.py", line 296, in get_cubin
    self._link_all(linker, cc, ignore_nonlto=False)
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/codegen.py", line 257, in _link_all
    linker.add_ltoir(ltoir)
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/cudadrv/driver.py", line 2986, in add_ltoir
    obj = ObjectCode.from_ltoir(ltoir, name=name)
          ^^^^^^^^^^^^^^^^^^^^^
AttributeError: type object 'ObjectCode' has no attribute 'from_ltoir'
```

(digging into these a bit locally to see if I can see why)

brandon-b-miller:
For this error `AttributeError: type object 'ObjectCode' has no attribute 'from_ltoir'` do you have `cuda-core==0.3.*` locally?

gmarkall:
Ah, maybe I don't have the right version of `cuda-core` installed.

numba_cuda/numba/cuda/cudadrv/driver.py
brandon-b-miller: Yes, that's a good point. I'll push an update.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2735,30 +2764,24 @@ def new(
...
-        else:
-            return linker(max_registers, lineinfo, cc)
+        params = (max_registers, lineinfo, cc)
+        if linker is _Linker:
+            params = (*params, _have_nvjitlink(), additional_flags)
```
brandon-b-miller: I think that this is slightly off. `_LinkerBase.new` shouldn't drop the `lto` that's passed in. _But_ numba probably wants to turn `lto` on whenever it's able to, meaning it needs to to decide it that's possible at some level. @gmarkall where do you think we should set `lto=_have_nvjitlink if lto is None else lto`? Should it be in `codegen.py`? At the decorator level? Somewhere else?

gmarkall: I think in the decorator, as in https://github.com/NVIDIA/numba-cuda/pull/310/files#diff-9cb327a3d3c0daff71c95e90239ec47e84f4324b6e52b02c3ee1072bd5715054R144

brandon-b-miller: Okay that makes sense. Shall we merge 310 then and then switch it in this PR to use `_have_nvjitlink()` rather than the config variable?

gmarkall: I think merging 310 and switching it to use `_have_nvjitlink()` in this PR sounds good. Following that, we still should ensure we're not dropping the `lto` in `_LinkerBase.new` - in conjunction with the observation at https://github.com/NVIDIA/numba-cuda/pull/133/files#r2167132125 (that codegen.py is calling `_Linker.new()` when the method is in `_LinkerBase`), it suggests the abstraction / design here is a little bit mixed up between `_Linker` and `_LinkerBase`.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2901,7 +2922,143 @@ def complete(self):
...
 
-class MVCLinker(Linker):
+class _Linker(_LinkerBase):
+    def __init__(
+        self,
+        max_registers=None,
+        lineinfo=False,
+        cc=None,
+        lto=None,
+        additional_flags=None,
```
gmarkall, line 2941: All the arguments are only used as positional arguments - the defaults here imply they could be omitted, which isn't the case (for example the `arch = ` line below will fail if you omit `cc`) - perhaps these should all be positional?

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2901,7 +2922,143 @@ def complete(self):
...
+    ):
+        arch = f"sm_{cc[0]}{cc[1]}"
+        if lto is False:
+            lto = None
```
gmarkall: Is this needed? The [`ProgramOptions` docs](https://nvidia.github.io/cuda-python/cuda-core/latest/generated/cuda.core.experimental.ProgramOptions.html#cuda.core.experimental.ProgramOptions.link_time_optimization) seem to indicate that `False` would be OK, and our `nvrtc.compile()` function also accepts `false` (which is the other use of it)

```suggestion
```

numba_cuda/numba/cuda/cudadrv/driver.py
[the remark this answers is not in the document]
gmarkall: It looks like cuda-core 0.3.0 allows the name to be specified - can we pass the names through?

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2901,7 +2922,143 @@ def complete(self):
...
+        if self._complete:
+            return self._error_log
+        return self.linker.get_error_log()
+
+    def add_ptx(self, ptx, name="<cudapy-ptx>"):
+        obj = ObjectCode.from_ptx(ptx)
+        self._object_codes.append(obj)
+
+    def add_cu(self, cu, name="<cudapy-cu>"):
+        with driver.get_active_context() as ac:
```
gmarkall, line 2985: We appear to be missing the logic to dump the assembly code for the compiled `.cu` source here (it was present in `PyNvJitLinker`).

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2901,7 +2922,143 @@ def complete(self):
...
+        self._object_codes.append(obj)
+
+    def add_library(self, lib, name="<cudapy-lib>"):
+        obj = ObjectCode.from_library(lib, name=name)
+        self._object_codes.append(obj)
+
+    def add_file(self, path, kind):
+        try:
+            with open(path, "rb") as f:
+                data = f.read()
```
gmarkall: `cached_file_read()` was introduced so that this didn't cause tests to take a very long time if there were lots of reads of the same file (e.g. when testing with NRT enabled). Can we use it here?

numba_cuda/numba/cuda/cudadrv/driver.py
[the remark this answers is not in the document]
gmarkall: To clarify my comment - I now see the other `add_` methods pass the names through, just `add_ptx()` does not.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2901,7 +2922,143 @@ def complete(self):
...
+            return self._info_log
+        return self.linker.get_info_log()
+
+    @property
+    def error_log(self):
+        if not self.linker:
+            raise ValueError("Not Initialized")
+        if self._complete:
+            return self._error_log
+        return self.linker.get_error_log()
```
gmarkall: I think it's unclear what happens if you do `linker.get_[error|info]_log()` before the link is complete. Perhaps we should raise if you try to get the logs before the link is complete? (I think there is no scenario in the code base where we try to get the log before the link is complete).

numba_cuda/numba/cuda/cudadrv/driver.py
gmarkall: I've just merged 310 so that its changes can be incorporated here.

gmarkall (changes requested):
This is looking pretty close - I've left some comments on the diff - some suggestions for robustness and a couple of items that got dropping in the transition from `PyNvJitLinker` to `_Linker`.

numba_cuda/numba/cuda/cudadrv/driver.py
brandon-b-miller: This does seem to be some kind of bug, passing `False` causes things to explode, as in the debug info tests:

```
numba.cuda.tests.cudapy.test_debuginfo.TestCudaDebugInfo.test_union_poly_types
```
```
Linker error log: ERROR 8 in nvvmCompileProgram callback
```

numba_cuda/numba/cuda/cudadrv/driver.py
brandon-b-miller: https://github.com/NVIDIA/numba-cuda/pull/133/commits/3e12fc3bbb57b49b0f312778db7258dc023345a7

numba_cuda/numba/cuda/cudadrv/driver.py
gmarkall: How are you reproducing this? I have

```diff
diff --git a/numba_cuda/numba/cuda/cudadrv/driver.py b/numba_cuda/numba/cuda/cudadrv/driver.py
index c00c194..d811f58 100644
--- a/numba_cuda/numba/cuda/cudadrv/driver.py
+++ b/numba_cuda/numba/cuda/cudadrv/driver.py
@@ -2932,8 +2932,6 @@ class _Linker(_LinkerBase):
         additional_flags=None,
     ):
         arch = f"sm_{cc[0]}{cc[1]}"
-        if lto is False:
-            lto = None
         self.max_registers = max_registers if max_registers else None
         self.lineinfo = lineinfo
         self.cc = cc
diff --git a/numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py b/numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
index 9bccf98..93c6a59 100644
--- a/numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
+++ b/numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
@@ -14,15 +14,6 @@ class TestCudaDebugInfo(CUDATestCase):
     These tests only checks the compiled PTX for debuginfo section
     """
 
-    def setUp(self):
-        super().setUp()
-        # If we're using LTO then we can't check the PTX in these tests,
-        # because we produce LTO-IR, which is opaque to the user.
-        # Additionally, LTO optimizes away the exception status due to an
-        # oversight in the way we generate it (it is not added to the used
-        # list).
-        self.skip_if_lto("Exceptions not supported with LTO")
-
     def _getasm(self, fn, sig):
         fn.compile(sig)
         return fn.inspect_asm(sig)
```

applied locally, and running the test seems to work:

```
$ python -m numba.runtests numba.cuda.tests.cudapy.test_debuginfo.TestCudaDebugInfo.test_union_poly_types -v
/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/__init__.py:51: UserWarning: Explicitly enabling pynvjitlink is no longer necessary. NVIDIA bindings are enabled. cuda.core will be used in place of pynvjitlink.
  warnings.warn(
test_union_poly_types (numba.cuda.tests.cudapy.test_debuginfo.TestCudaDebugInfo.test_union_poly_types) ... ok

Ran 1 test in 2.014s

OK
```

gmarkall: Also - I think with recent versions of nvjitlink it writes the error log to stderr and it doesn't end up propagating to Python - was there any output on stderr?

numba_cuda/numba/cuda/cudadrv/driver.py
brandon-b-miller: I think `cc` is the only parameter that there's no sensible default value for. Shall I make it required?

numba_cuda/numba/cuda/cudadrv/driver.py
brandon-b-miller: I'll try pushing to CI and see if that teases it out, else it might be something I have wrong locally.

brandon-b-miller: Example failure https://github.com/NVIDIA/numba-cuda/actions/runs/15905592830/job/44859736553?pr=133

numba_cuda/numba/cuda/cudadrv/driver.py
gmarkall: Actually, I think we should just leave this as-is.

numba_cuda/numba/cuda/cudadrv/driver.py
gmarkall: Understood, let's keep the forwarding to `None` in with a comment that it's a workaround.

gmarkall:
I thought the code changes looked good but I'm hitting an error disabling the NVIDIA binding locally now with this PR. For example:

```
$ NUMBA_CUDA_USE_NVIDIA_BINDING=0 python -m numba.runtests numba.cuda.tests -v
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/gmarkall/numbadev/numba/numba/runtests.py", line 9, in <module>
    sys.exit(0 if _main(sys.argv) else 1)
                  ^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba/numba/testing/_runtests.py", line 25, in _main
    return run_tests(argv, defaultTest='numba.tests',
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba/numba/testing/__init__.py", line 54, in run_tests
    prog = NumbaTestProgram(argv=argv,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba/numba/testing/main.py", line 204, in __init__
    super(NumbaTestProgram, self).__init__(*args, **kwargs)
  File "/home/gmarkall/miniforge3/envs/numbadev/lib/python3.11/unittest/main.py", line 101, in __init__
    self.parseArgs(argv)
  File "/home/gmarkall/numbadev/numba/numba/testing/main.py", line 293, in parseArgs
    super(NumbaTestProgram, self).parseArgs(argv)
  File "/home/gmarkall/miniforge3/envs/numbadev/lib/python3.11/unittest/main.py", line 150, in parseArgs
    self.createTests()
  File "/home/gmarkall/miniforge3/envs/numbadev/lib/python3.11/unittest/main.py", line 161, in createTests
    self.test = self.testLoader.loadTestsFromNames(self.testNames,
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/miniforge3/envs/numbadev/lib/python3.11/unittest/loader.py", line 232, in loadTestsFromNames
    suites = [self.loadTestsFromName(name, module) for name in names]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/miniforge3/envs/numbadev/lib/python3.11/unittest/loader.py", line 232, in <listcomp>
    suites = [self.loadTestsFromName(name, module) for name in names]
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/miniforge3/envs/numbadev/lib/python3.11/unittest/loader.py", line 162, in loadTestsFromName
    module = __import__(module_name)
             ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/__init__.py", line 57, in <module>
    raise RuntimeError("nvJitLink requires the NVIDIA CUDA bindings. ")
RuntimeError: nvJitLink requires the NVIDIA CUDA bindings. 
```

Just looking into why this is and why it doesn't seem to occur on the ctypes binding test in CI.

numba_cuda/numba/cuda/__init__.py
```diff
@@ -44,6 +46,16 @@
...
 
+if config.CUDA_ENABLE_PYNVJITLINK:
+    if USE_NV_BINDING:
+        warnings.warn(
+            "Explicitly enabling pynvjitlink is no longer necessary. "
+            "NVIDIA bindings are enabled. cuda.core will be used "
+            "in place of pynvjitlink."
+        )
+    else:
+        raise RuntimeError("nvJitLink requires the NVIDIA CUDA bindings. ")
```
gmarkall, line 66: Because `config.CUDA_ENABLE_PYNVJITLINK` is enabled automatically if it's found in the environment, disabling the NVIDIA bindings if pynvjitlink is installed now leads to this exception being hit.

gmarkall (changes requested):
I think we can solve the issue of it not being possible to disable the NVIDIA bindings if pynvjitlink is installed by keeping track of whether it was enabled automatically, and just ignoring it if it was, like:

```diff
diff --git a/numba_cuda/numba/cuda/__init__.py b/numba_cuda/numba/cuda/__init__.py
index 430b3b7..e944fe0 100644
--- a/numba_cuda/numba/cuda/__init__.py
+++ b/numba_cuda/numba/cuda/__init__.py
@@ -9,6 +9,9 @@ import warnings
 # 1. Config setting "CUDA_ENABLE_PYNVJITLINK" (highest priority)
 # 2. Environment variable "NUMBA_CUDA_ENABLE_PYNVJITLINK"
 # 3. Auto-detection of pynvjitlink module (lowest priority)
+
+pynvjitlink_auto_enabled = False
+
 if getattr(config, "CUDA_ENABLE_PYNVJITLINK", None) is None:
     if (
         _pynvjitlink_enabled_in_env := _readenv(
@@ -17,9 +20,10 @@ if getattr(config, "CUDA_ENABLE_PYNVJITLINK", None) is None:
     ) is not None:
         config.CUDA_ENABLE_PYNVJITLINK = _pynvjitlink_enabled_in_env
     else:
-        config.CUDA_ENABLE_PYNVJITLINK = (
+        pynvjitlink_auto_enabled = (
             importlib.util.find_spec("pynvjitlink") is not None
         )
+        config.CUDA_ENABLE_PYNVJITLINK = pynvjitlink_auto_enabled
 
 # Upstream numba sets CUDA_USE_NVIDIA_BINDING to 0 by default, so it always
 # exists. Override, but not if explicitly set to 0 in the envioronment.
@@ -53,6 +57,11 @@ if config.CUDA_ENABLE_PYNVJITLINK:
             "NVIDIA bindings are enabled. cuda.core will be used "
             "in place of pynvjitlink."
         )
+    elif pynvjitlink_auto_enabled:
+        # Ignore the fact that pynvjitlink is enabled, because that was an
+        # automatic decision based on discovering pynvjitlink was present; the
+        # user didn't ask for it
+        pass
     else:
         raise RuntimeError("nvJitLink requires the NVIDIA CUDA bindings. ")
 
```

Does this seem like a workable solution? (It allows me to disable the NVIDIA binding for testing locally)

leofang:
I need to catch with the progress here, but are we still keeping both cuda.core.Linker and pynvjiink?

brandon-b-miller:
> I need to catch with the progress here, but are we still keeping both cuda.core.Linker and pynvjiink?

No, `pynvjitlink` won't be required. We're just puzzling over what to do for existing users who try and enable it, or have it in their environment after this PR :)

Issue #135, opened 2025-02-25
[BUG] [Metadata] Incorrect kernel argument types when NUMBA_DUMP_LLVM is turned on


jiel-nv:
**Describe the bug**
When NUMBA_DUMP_LLVM is turned on, the debug metadata for kernel arguments still contains return value which should not be in the DISubroutineType.

**Steps/Code to reproduce bug**
```
$ cat bug.py
from numba import cuda

@cuda.jit("void(int32, int32)", debug=True, opt=False)
def f(x, y):
    z = x + y
$ NUMBA_DUMP_LLVM=1 python bug.py | grep "\!8"
!8 = !{ !5, !6, !7 }
!9 = !DISubroutineType(types: !8)
!8 = !{ !5, !6, !7 }
!9 = !DISubroutineType(types: !8)
```
Notice that, even after the "post kernel fixup" dump, the return value still shows up in the DISubroutineType, which was fixed by a previous PR (without the NUMBA_DUMP_LLVM turned on)

**Expected behavior**
If not using the NUMBA_DUMP_LLVM, and simply use the inspect_llvm() method, we do get the expected metadata as following,
```
$ cat bug.py
from numba import cuda
from numba.core import types

sig = (types.int32, types.int32)
@cuda.jit("void(int32, int32)", debug=True, opt=False)
def f(x, y):
    z = x + y

llvm_ir = f.inspect_llvm(sig)
llvmir = f.inspect_llvm(sig).splitlines()
for line in llvmir:
    print(line)
$ python bug.py | grep "\!8"
!8 = !{ !6, !7 }
!9 = !DISubroutineType(types: !8)
```
So something wrong happens when turning on NUMBA_DUMP_LLVM.

jiel-nv:
Another observation, taking the inspect_llvm() version of bug.py, if turning on NUMBA_DUMP_LLVM, the result of inspect_llvm() will also be wrong, i.e. NUMBA_DUMP_LLVM dominates in this case.

```
$ NUMBA_DUMP_LLVM=0 python bug.py | grep "\!8"
!8 = !{ !6, !7 }
!9 = !DISubroutineType(types: !8)
$ NUMBA_DUMP_LLVM=1 python bug.py | grep "\!8"
!8 = !{ !5, !6, !7 }
!9 = !DISubroutineType(types: !8)
!8 = !{ !5, !6, !7 }
!9 = !DISubroutineType(types: !8)
!8 = !{ !5, !6, !7 }
[...]

PR #136, opened 2025-02-25
Emit correct kernel arguments debug metadata types when NUMBA_DUMP_LLVM is on

jiel-nv:
After shifting out the return value type from the operands of 'types' field of the 'DISubroutineType' metadata node, compiler needs to clear the cached string representation of the corresponding MDValue object if NUMBA_DUMP_LLVM is on.

Because at that moment, the cached string has already been filled with the tuple before kernel_fixup() which has the return value type and continue being used when printing out the LLVM IR.

This change fixes https://github.com/NVIDIA/numba-cuda/issues/135


numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
```diff
@@ -298,6 +298,14 @@ def f(x, y):
         match = re.compile(pat).search(llvm_ir)
         self.assertIsNotNone(match, msg=llvm_ir)
 
+    def test_kernel_args_types(self):
+        self._test_kernel_args_types()
+
+    def test_kernel_args_types_dump(self):
+        # see issue#135
+        with override_config('DUMP_LLVM', 1):
+            self._test_kernel_args_types()
```
gmarkall, line 307: This needs the output capturing to prevent the test suite dumping everything to stdout (which pollutes the CI logs and the terminal output when running locally:

```suggestion
            with captured_stdout():
                self._test_kernel_args_types()
```

`captured_stdout` can be imported from `numba.tests.support`.

gmarkall (changes requested):
Many thanks! The fix and test look generally good, with one change needed to keep the test suite output clean. Once that's done I think this should be good to merge.

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
jiel-nv: Thank you Graham for pointing this out. Tweaking another commit.

jiel-nv:
I don't think the build-conda error is related to my change.
`conda: error: argument COMMAND: invalid choice: 'mambabuild'`

gmarkall:
@jiel-nv You're correct, the issue is unrelated to your change. The CI setup has changed a bit recently, so your branch just needed a merge from `main`. I've done this and hopefully things go green again.

gmarkall (approved):

PR #137, opened 2025-02-26
Fix linking of external code from callees

gmarkall:
The original linking implementation for linkable code in device declarations did not consider calls inside callees; this change recurses through the typing to find all calls requiring linkable code.

### DIFF
--- a/numba_cuda/numba/cuda/dispatcher.py
+++ b/numba_cuda/numba/cuda/dispatcher.py
@@ -43,15 +43,33 @@ reshape_funcs = ['nocopy_empty_reshape', 'numba_attempt_nocopy_reshape']
 
 
 def get_cres_link_objects(cres):
+    """Given a compile result, return a set of all linkable code objects that
+    are required for it to be fully linked."""
+
     link_objects = set()
+
+    # The typemap of the function includes calls, so we can traverse it to find
+    # the references we need.
     for name, v in cres.fndesc.typemap.items():
+
+        # CUDADispatchers represent a call to a device function, so we need to
+        # look up the linkable code for those recursively.
         if isinstance(v, cuda_types.CUDADispatcher):
+            # We need to locate the signature of the call so we can find the
+            # correct overload.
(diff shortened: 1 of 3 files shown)


numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -42,6 +42,47 @@
...
+            # correct overload.
+            for call, sig in cres.fndesc.calltypes.items():
+                if isinstance(call, ir.Expr) and call.op == 'call':
+                    # There will likely be multiple calls in the typemap; we
+                    # can uniquely identify the relevant one using its SSA
+                    # name.
+                    if call.func.name == name:
+                        called_cres = v.dispatcher.overloads[sig.args]
+                        called_link_objects = get_cres_link_objects(called_cres)
+                        link_objects.update(called_link_objects)
```
isVoid, line 68: This is cool. I learnt a few things by reading through this section. Do you think the below simplifies the code and reduces the code complexity for a little bit?

```python

device_func_calls = [(name, v) for name, v in cres.fndesc.typemap.items() if isinstance(v, cuda_types.CUDADispatcher)]

call_signatures = [(call.func.name, sig) for call, sig in creas.fndesc.calltypes.items() if isinstance(call, ir.Expr) and call.op == 'call']

call_signature_d = defaultdict(list)
for name, sig in call_signatures:
    call_signature_d[name].append(sig)

for name, v in device_func_calls:
    for sig in call_sinagure_d.get(name, []):
        called_cres = v.dispatcher.overloads[sig.args]
        called_link_objects = get_cres_link_objects(called_cres)
        link_objects.update(called_link_objects)
```

I think this reduces the size of the list for both of the nested for-loop. This is proportional to O(`num_calls`^2), not O(`num_typings`^2)

gmarkall: Thanks - I've incorporated your changes. I don't think there's much of a performance impact (the number of calls and typings won't be that large), but I think your modifications improved the readability of the code.

PR #143, opened 2025-03-03
Implement alignment support for local and shared arrays.

tpn:
This PR adds support for specifying an `alignment=N` keyword argument to the `cuda.local.array()` and `cuda.shared.array()` helpers that can be used within JIT'd CUDA kernels (i.e. functions annotated with `@numba.cuda.jit`.

### DIFF
--- a/numba_cuda/numba/cuda/cudadecl.py
+++ b/numba_cuda/numba/cuda/cudadecl.py
@@ -1,5 +1,5 @@
 import operator
-from numba.core import types
+from numba.core import errors, types
 from numba.core.typing.npydecl import (
     parse_dtype,
     parse_shape,
@@ -33,7 +33,7 @@ register_number_classes(register_global)
 
 class Cuda_array_decl(CallableTemplate):
     def generic(self):
-        def typer(shape, dtype):
+        def typer(shape, dtype, alignment=None):
             # Only integer literals and tuples of integer literals are valid
             # shapes
             if isinstance(shape, types.Integer):
@@ -46,6 +46,16 @@ class Cuda_array_decl(CallableTemplate):
                     return None
             else:
                 return None
+
+            if alignment is not None:
+                permitted = (types.IntegerLiteral, types.NoneType)
+                if not isinstance(alignment, permitted):
+                    msg = "alignment must be a constant integer"
+                    raise errors.RequireLiteralValue(msg)
+
+            # N.B. We don't use alignment for typing; it's not part of
+            #      types.Array.  The value supplied to the array declaration
+            #      is handled in the lowering.
 
             ndim = parse_shape(shape)
             nb_dtype = parse_dtype(dtype)
--- a/numba_cuda/numba/cuda/cudaimpl.py
+++ b/numba_cuda/numba/cuda/cudaimpl.py
@@ -1,6 +1,7 @@
 from functools import reduce
 import operator
 import math
+import struct
 
 from llvmlite import ir
 import llvmlite.binding as ll
@@ -92,10 +93,61 @@ def _get_unique_smem_id(name):
     return "{0}_{1}".format(name, _unique_smem_id)
 
 
+def _validate_alignment(alignment: int):
+    """
+    Ensures that *alignment*, if not None, is a) greater than zero, b) a power
+    of two, and c) a multiple of the size of a pointer.  If any of these
+    conditions are not met, a ValueError is raised.  Otherwise, this
+    function returns None, indicating that the alignment is valid.
+    """
+    if alignment is None:
+        return
+    if not isinstance(alignment, int):
+        raise ValueError("Alignment must be an integer")
+    if alignment <= 0:
+        raise ValueError("Alignment must be positive")
+    if (alignment & (alignment - 1)) != 0:
+        raise ValueError("Alignment must be a power of 2")
+    pointer_size = struct.calcsize("P")
+    if (alignment % pointer_size) != 0:
+        msg = f"Alignment must be a multiple of {pointer_size}"
+        raise ValueError(msg)
+
+
+def _try_extract_and_validate_alignment(sig: types.Tuple):
+    """
+    Extracts and validates the alignment from the supplied signature.
+
+    Returns the alignment if it is present and is an integer literal;
+    otherwise, returns None.
+
+    N.B. Currently, this routine assumes the signature has exactly
+         three arguments, with the alignment (if present) as the third
+         argument, as is the case with the shared and local array
+         helper routines below.
+
+         If this routine is called from new places, you may need to
+         review this implicit assumption.
+    """
+    if len(sig.args) != 3:
+        return None
+
+    alignment_arg = sig.args[2]
+    if not isinstance(alignment_arg, types.IntegerLiteral):
+        return None
+
+    alignment_arg = alignment_arg.literal_value
+    _validate_alignment(alignment_arg)
+    return alignment_arg
+
+
 @lower(cuda.shared.array, types.IntegerLiteral, types.Any)
+@lower(cuda.shared.array, types.IntegerLiteral, types.Any, types.IntegerLiteral)
+@lower(cuda.shared.array, types.IntegerLiteral, types.Any, types.NoneType)
 def cuda_shared_array_integer(context, builder, sig, args):
     length = sig.args[0].literal_value
     dtype = parse_dtype(sig.args[1])
+    alignment = _try_extract_and_validate_alignment(sig)
     return _generic_array(
         context,
         builder,
@@ -104,14 +156,17 @@ def cuda_shared_array_integer(context, builder, sig, args):
         symbol_name=_get_unique_smem_id("_cudapy_smem"),
         addrspace=nvvm.ADDRSPACE_SHARED,
         can_dynsized=True,
+        alignment=alignment,
     )
 
 
-@lower(cuda.shared.array, types.Tuple, types.Any)
-@lower(cuda.shared.array, types.UniTuple, types.Any)
+@lower(cuda.shared.array, types.BaseTuple, types.Any)
+@lower(cuda.shared.array, types.BaseTuple, types.Any, types.IntegerLiteral)
+@lower(cuda.shared.array, types.BaseTuple, types.Any, types.NoneType)
 def cuda_shared_array_tuple(context, builder, sig, args):
     shape = [s.literal_value for s in sig.args[0]]
     dtype = parse_dtype(sig.args[1])
+    alignment = _try_extract_and_validate_alignment(sig)
     return _generic_array(
         context,
         builder,
@@ -120,13 +175,17 @@ def cuda_shared_array_tuple(context, builder, sig, args):
         symbol_name=_get_unique_smem_id("_cudapy_smem"),
         addrspace=nvvm.ADDRSPACE_SHARED,
         can_dynsized=True,
+        alignment=alignment,
     )
 
 
 @lower(cuda.local.array, types.IntegerLiteral, types.Any)
+@lower(cuda.local.array, types.IntegerLiteral, types.Any, types.IntegerLiteral)
+@lower(cuda.local.array, types.IntegerLiteral, types.Any, types.NoneType)
 def cuda_local_array_integer(context, builder, sig, args):
     length = sig.args[0].literal_value
     dtype = parse_dtype(sig.args[1])
+    alignment = _try_extract_and_validate_alignment(sig)
     return _generic_array(
         context,
         builder,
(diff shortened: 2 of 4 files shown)


tpn:
I've removed the dependency on altering the underlying types.Array in numba, it wasn't necessary, as @gmarkall pointed out.

numba_cuda/numba/cuda/cudaimpl.py
```diff
@@ -1041,7 +1112,8 @@ def _generic_array(context, builder, shape, dtype, symbol_name, addrspace,
 
     # Create array object
     ndim = len(shape)
-    aryty = types.Array(dtype=dtype, ndim=ndim, layout='C')
+    aryty = types.Array(dtype=dtype, ndim=ndim, layout='C',
+                        alignment=alignment)
```
gmarkall, line 1116: This is probably the cause of the CI fails, and I think is not needed.

tpn: Ah!  Dang, I missed this one.

tpn: Fixed; pushed.

gmarkall: Whatever was changed did not resolve this - the `alignment` kwarg should not be passed to the `Array` constructor.

tpn: That was wild, whenever I did a fixup commit and squashed it via interactive rebase... the alignment kwarg returned.  I kept it as a separate commit and verified that it is in fact now gone.

numba_cuda/numba/cuda/cudaimpl.py
```diff
@@ -91,35 +92,85 @@ def _get_unique_smem_id(name):
     return "{0}_{1}".format(name, _unique_smem_id)
 
 
+def _validate_alignment(alignment: int):
+    """
+    Ensures that *alignment*, if not None, is a) greater than zero, b) a power
+    of two, and c) a multiple of the size of a pointer.  If any of these
+    conditions are not met, a NumbaValueError is raised.  Otherwise, this
```
tpn, line 99: Nit: s/NumbaValueError/ValueError/

gmarkall (changes requested):
Thanks for the PR! I think this is a good start with some functionality working as expected - however there are some other cases to cover and a few observations on the diff. We might need another iteration afterwards once things have shaped up a bit (and the docs might need an update if they didn't get generated from the source, I will have to check).

numba_cuda/numba/cuda/cudadecl.py
```diff
@@ -39,6 +39,14 @@ def typer(shape, dtype):
             else:
                 return None
 
+            # N.B. We don't do anything with alignment in this routine; it's
```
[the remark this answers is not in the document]
tpn: Fixed!

numba_cuda/numba/cuda/cudaimpl.py
```diff
@@ -981,17 +1032,37 @@ def _generic_array(context, builder, shape, dtype, symbol_name, addrspace,
...
+        # underlying dtype alignment from the ABI and then round it up to
+        # a power of two.  Otherwise, just use the caller's alignment.
+        #
+        # N.B. The caller *could* provide a valid-but-smaller-than-natural
+        #      alignment here; we'll assume the caller knows what they're
+        #      doing and let that through without error.
+
+        if alignment is None:
+            abi_alignment = context.get_abi_alignment(lldtype)
+            # Ensure a power of two alignment.
```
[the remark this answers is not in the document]
tpn: Restored that comment.

numba_cuda/numba/cuda/tests/cudapy/test_array_alignment.py
```diff
@@ -0,0 +1,115 @@
...
+                            alignment=alignment,
+                        )
+                        if i == 0:
+                            shrd[0] = shared_array.ctypes.data
+
+                loc = np.zeros(1, dtype=np.uint64)
+                shrd = np.zeros(1, dtype=np.uint64)
+                f[1, 1](loc, shrd, which)
+
+                if alignment is not None:
```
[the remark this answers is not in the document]
tpn: Made sure we check align in LLVM IR.

numba_cuda/numba/cuda/cudaimpl.py
```diff
@@ -91,35 +92,85 @@ def _get_unique_smem_id(name):
...
-                          can_dynsized=True)
+                          can_dynsized=True, alignment=alignment)
 
 
 @lower(cuda.shared.array, types.Tuple, types.Any)
 @lower(cuda.shared.array, types.UniTuple, types.Any)
+@lower(cuda.shared.array, types.Tuple, types.Any, types.IntegerLiteral)
+@lower(cuda.shared.array, types.UniTuple, types.Any, types.IntegerLiteral)
+@lower(cuda.shared.array, types.Tuple, types.Any, types.NoneType)
+@lower(cuda.shared.array, types.UniTuple, types.Any, types.NoneType)
```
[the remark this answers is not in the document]
tpn: Fixed.

numba_cuda/numba/cuda/cudaimpl.py
```diff
@@ -91,35 +92,85 @@ def _get_unique_smem_id(name):
...
 def cuda_local_array_integer(context, builder, sig, args):
     length = sig.args[0].literal_value
     dtype = parse_dtype(sig.args[1])
+    alignment = None
+    if len(sig.args) == 3:
+        try:
+            alignment = sig.args[2].literal_value
+            _validate_alignment(alignment)
+        except AttributeError:
+            pass
```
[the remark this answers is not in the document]
tpn: Added a new `_try_extract_and_validate_alignment()` helper that cleans this up quite a bit.

numba_cuda/numba/cuda/cudaimpl.py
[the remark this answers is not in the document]
gmarkall: It's not strictly necessary. I think it's just a personal preference at this point - you can maximally reduce the stacking with `types.Any`, but leaving the actual types does make it a little clearer to the reader looking at the function in isolation.

numba_cuda/numba/cuda/cudaimpl.py
```diff
@@ -91,35 +92,95 @@ def _get_unique_smem_id(name):
...
     dtype = parse_dtype(sig.args[1])
+    alignment = _try_extract_and_validate_alignment(sig)
     return _generic_array(context, builder, shape=(length,), dtype=dtype,
                           symbol_name='_cudapy_lmem',
                           addrspace=nvvm.ADDRSPACE_LOCAL,
-                          can_dynsized=False)
+                          can_dynsized=False, alignment=alignment)
+
+
+@lower(cuda.local.array, types.BaseTuple, types.Any)
```
gmarkall, line 173: This function duplicates the `ptx_lmem_alloc_array` function below. In my initial comment below (https://github.com/NVIDIA/numba-cuda/pull/143/files#r2000829665) I had imagined you would add the third argument type to the `@lower` decorators and inserted the handling of the alignment, but your new function does have a clear name. So instead I would suggest deleting the `ptx_lmem_alloc_array` function below instead, so that we don't have two different implementations of the same thing.

numba_cuda/numba/cuda/cudaimpl.py
```diff
@@ -91,35 +92,95 @@ def _get_unique_smem_id(name):
...
+        raise ValueError("Alignment must be positive")
+    if (alignment & (alignment - 1)) != 0:
+        raise ValueError("Alignment must be a power of 2")
+    pointer_size = struct.calcsize("P")
+    if (alignment % pointer_size) != 0:
+        msg = f"Alignment must be a multiple of {pointer_size}"
+        raise ValueError(msg)
+
+
+def _try_extract_and_validate_alignment(sig: types.Tuple):
```
gmarkall, line 116: I think generally for writing utility functions in Numba I'd not encode the assumption that the parameter they're looking for is in a particular argument if a whole signature is passed. I think it's OK in this case as it's specifically used for local and shared memory, and they don't fit a common pattern.

numba_cuda/numba/cuda/cudaimpl.py
tpn: Oh my gosh look at that, completely identical to `cuda_local_array_tuple`!  But yes, I think the fact that it differs in name so significantly from the other three is a good enough reason to keep the new `cuda_local_array_tuple` one instead.  I'll fix.

numba_cuda/numba/cuda/cudaimpl.py
tpn: Yeah no problem with removing the type hint here.

numba_cuda/numba/cuda/tests/cudapy/test_array_alignment.py
```diff
@@ -0,0 +1,115 @@
...
+                shrd = np.zeros(1, dtype=np.uint64)
+
+                with self.assertRaises(ValueError) as raises:
+                    f[1, 1](loc, shrd, which)
+                exc = str(raises.exception)
+                self.assertIn("Alignment must be", exc)
+
+                if NOISY:
+                    print('.', end='', flush=True)
+
```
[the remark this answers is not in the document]
tpn: Sorry, I missed this comment--completely agree, I'm adding tests for that now.

numba_cuda/numba/cuda/tests/cudapy/test_array_alignment.py
```diff
@@ -0,0 +1,115 @@
...
+    def test_invalid_aligments(self):
+        shapes = (1, 50)  # Just test small and large
+        dtypes = (np.uint8, np.uint64)  # Just test smallest and largest
+        # Keep a good selection of invalid alignments to test error cases
+        alignments = (-1, 0, 3, 17, 33)  # Negative, zero, and non-powers of 2
+        array_types = [(0, 'local'), (1, 'shared')]
+
+        # This reduces from 576 to 40 test cases (2×2×2×5)
```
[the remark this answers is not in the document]
tpn: I removed that comment; it was LLM leftovers.

numba_cuda/numba/cuda/tests/cudapy/test_array_alignment.py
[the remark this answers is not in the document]
tpn: Done!

tpn:
@gmarkall added some invalid type alignment tests, as well as tweaking the tests to use a common set of DTYPES that also include a bunch of record types (with and without alignment).

numba_cuda/numba/cuda/cudaimpl.py
```diff
@@ -113,7 +113,7 @@ def _validate_alignment(alignment: int):
         raise ValueError(msg)
 
 
-def _try_extract_and_validate_alignment(sig: types.Tuple):
+def _try_extract_and_validate_alignment(sig):
```
gmarkall, line 116: I apologise for having conveyed that the type hint should be removed - that was not what I intended, and I'm generally comfortable / happy with people adding type hints. I'm not experienced enough with them / in the habit of using them, so I really need to learn from others in this area, and I appreciate efforts to add them.

What I meant was that generally I wouldn't write a utility function to extract an argument from a particular index, because then it can't easily be reused with other functions or be robust against changes in the function signature - e.g. if another function had alignment in the fourth argument, or the keyword arg support for these functions were fixed up so it were possible to write `cuda.local.array(shape, alignment=alignment)`, where the dtype is omitted and assumed to be some default.

tpn:
Hi @gmarkall, I believe this one is ready to go with all requested changes made.

ZzEeKkAa:
These changes conflicts with https://github.com/NVIDIA/numba-cuda/pull/176 , but these changes are more important for nvmath-python. I'll fix my branch after this one is merged.

numba_cuda/numba/cuda/tests/cudapy/test_array_alignment.py
```diff
@@ -0,0 +1,237 @@
...
+                shape=shape,
+                dtype=dtype,
+                alignment=alignment,
+            ):
+
+                @cuda.jit
+                def f(local_array, shared_array, which):
+                    i = cuda.grid(1)
+                    if which == 0:
+                        local_array = cuda.local.array(
```
gmarkall, line 195: I think the test here isn't separately testing local and shared arrays. The reason is that the `ValueError` is going to be produced at compile time, not at run time - so using the same alignment for both local and shared arrays is likely to always produce the `ValueError` from the same kind of array (probably the local array just because it's earlier in the function). This pattern was OK for the test above in the success case because the runtime switch between checking the local vs. shared array behaves as expected.

In a way it doesn't matter because the local and shared arrays both use the same typing from `Cuda_array_decl` and lowering through `_generic_array`, but you may wish to update the test case to test errors compiling local and shared arrays in separate kernels so that the behaviour of the test more closely matches your intent.

gmarkall (commented):
Thanks for the updates - there are a couple of minor questions on the added test - I don't think they necessarily need to be addressed for this to be merged, so just let me know what you want to do (i.e. merge as-is or modify based on the comments).

numba_cuda/numba/cuda/tests/cudapy/test_array_alignment.py
tpn: Good catch, I separated things out into two different kernels and removed the in-kernel `which` predicate logic.

numba_cuda/numba/cuda/tests/cudapy/test_array_alignment.py
```diff
@@ -0,0 +1,237 @@
...
+                        if i == 0:
+                            shared_array[0] = shared_array.ctypes.data
+
+                loc = np.zeros(1, dtype=np.uint64)
+                shrd = np.zeros(1, dtype=np.uint64)
+
+                # The type of error we expect differs between an invalid value
+                # that is still an int, and an invalid type.
+                if isinstance(alignment, int):
+                    with self.assertRaises(ValueError) as raises:
```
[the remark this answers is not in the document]
tpn: No good reason, switched to `assertRaisesRegex`.

tpn:
> Thanks for the updates - there are a couple of minor questions on the added test - I don't think they necessarily need to be addressed for this to be merged, so just let me know what you want to do (i.e. merge as-is or modify based on the comments).

Fixes were easy, all good on my side for the merge!

gmarkall (approved):

PR #144, opened 2025-03-03
[Doc] add paddle to interoperability chapter

HydrogenSulfate:
Same pr in <https://github.com/numba/numba/pull/9834>(can be closed if this pr is merged)

we support `paddle.Tensor.__cuda_array_interface__`(https://github.com/PaddlePaddle/Paddle/pull/68192) and `paddle.to_tensor(tenosr_with_CAI_impl)`(https://github.com/PaddlePaddle/Paddle/pull/69913), so we can convert paddle Tensor from/to other DL frameworks via CAI(based on CAI version 2).

``` py
import paddle
import torch
import numpy as np


for i in range(10):
    pt_ten = torch.randn([3, 3], device="cuda:0")
    pd_ten = paddle.to_tensor(pt_ten)
    pd_ten[0, :] = paddle.randn([3])
    np.testing.assert_allclose(pd_ten.numpy(), pt_ten.detach().cpu().numpy())
```


gmarkall:
Many thanks!

PR #145, opened 2025-03-03
Add Module Setup and Teardown Callback to Linkable Code Interface

### LINKED ISSUE #138: [FEA]Support Custom Module Initialization Callbacks
**Is your feature request related to a problem? Please describe.**
Currently, after the kernel has been compiled, Numba directly creates the module with the kernel entry point and launch it. Third party libraries sometimes needs to initialize the modules with a host function to use their device APIs. Numba should provide a way to insert module initialization callbacks into its kernel launching precedure.

**Describe the solution you'd like**
Just an early proposal, please feel free to suggest alternatives.

[EDIT: 022825]
Offline discussion agreed to that the init callback function can be tied with the linkable code object. Third party library must provide their implementation via the linkable code interface, it can be also tied with necessary initialization steps to the module in these functions.

```python
class LinkableCode:
    data: str
    name: str
    init_callbacks: callable | None
```

[Original Post]
We can try adding a new argument to jit call:
```python
@cuda.jit(mod_init_callback=[callbacks,])
def kernel():
...
```

The callback function should follow this signature:
```python
def callback(mod: ctypes.c_void_p, stream: int):
...
```

Alternatively, we can use the [cuda-python object model](https://nvidia.github.io/cuda-python/cuda-bindings/latest/module/driver.html#cuda.bindings.driver.CUmodule) as the signatures.

```python
def callback(mod: cuda.bindings.driver.CUModule, stream: class cuda.bindings.runtime.cudaStream_t):
...
```

Which later gets invoked on the compiled kernel prior to launch in the specified order.

**Describe alternatives you've considered**
Alternatives are basically different ways to pass in the callback function. Since the compiled module is an implementation detail inside Numba, we can't / shouldn't work on it directly.

Alt1 (pass in via the launch configuration): 
```python
kernel[1, 1, stream, smem, modinit]()
```

Alt2 (pass in via the kernel argument):
```python
class ModInitCallback:
...

kernel[...](ModInitCallback(callback))
```

isVoid:
This PR adds `setup_callback` and `teardown_callback` field to the linkable code interface. When library developers pass in an external module, they can also pass in an external function to initialize the compiled and loaded module. When the cumodule is garbage collected in python, the `teardown_callback` is also invoked to make sure no resources is leaked.

Additionally this PR fixes a bug in module creation under multi-threaded scenario. Currently when multiple threads launch a kernel, it's non-deterministic how many times the compiler is run and modules are reloaded from cache. We implement compiler lock in this case to make sure that only a single instance per kernel is compiled and all subsequent threads should reload from the same instance. This saves compiler resources and makes the pipeline more predictable.

closes #138

### DIFF
--- a/.github/actions/compute-matrix/action.yaml
+++ b/.github/actions/compute-matrix/action.yaml
@@ -20,20 +20,16 @@ runs:
         "
 
         export TEST_MATRIX="
-          # TODO: Use CUDA 11.4.3 below.
           - { CUDA_VER: '11.8.0', ARCH: 'amd64', PY_VER: '3.9',  LINUX_VER: 'rockylinux8', GPU: 'l4',   DRIVER: 'latest' }
           - { CUDA_VER: '11.8.0', ARCH: 'amd64', PY_VER: '3.10', LINUX_VER: 'ubuntu20.04', GPU: 'l4',   DRIVER: 'latest' }
           - { CUDA_VER: '12.0.1', ARCH: 'amd64', PY_VER: '3.11', LINUX_VER: 'ubuntu22.04', GPU: 'l4',   DRIVER: 'latest' }
           - { CUDA_VER: '12.2.2', ARCH: 'amd64', PY_VER: '3.12', LINUX_VER: 'ubuntu22.04', GPU: 'l4',   DRIVER: 'latest' }
-          # TODO: Use CUDA 12.8.0 below.
-          - { CUDA_VER: '12.5.1', ARCH: 'amd64', PY_VER: '3.13', LINUX_VER: 'ubuntu24.04', GPU: 'l4',   DRIVER: 'latest' }
-          # TODO: Use CUDA 11.4.3 below.
+          - { CUDA_VER: '12.8.0', ARCH: 'amd64', PY_VER: '3.13', LINUX_VER: 'ubuntu24.04', GPU: 'l4',   DRIVER: 'latest' }
           - { CUDA_VER: '11.8.0', ARCH: 'arm64', PY_VER: '3.9',  LINUX_VER: 'rockylinux8', GPU: 'a100', DRIVER: 'latest' }
           - { CUDA_VER: '11.8.0', ARCH: 'arm64', PY_VER: '3.10', LINUX_VER: 'ubuntu20.04', GPU: 'a100', DRIVER: 'latest' }
           - { CUDA_VER: '12.0.1', ARCH: 'arm64', PY_VER: '3.11', LINUX_VER: 'ubuntu22.04', GPU: 'a100', DRIVER: 'latest' }
           - { CUDA_VER: '12.2.2', ARCH: 'arm64', PY_VER: '3.12', LINUX_VER: 'ubuntu22.04', GPU: 'a100', DRIVER: 'latest' }
-          # TODO: Use CUDA 12.8.0 below.
-          - { CUDA_VER: '12.5.1', ARCH: 'arm64', PY_VER: '3.13', LINUX_VER: 'ubuntu24.04', GPU: 'a100', DRIVER: 'latest' }
+          - { CUDA_VER: '12.8.0', ARCH: 'arm64', PY_VER: '3.13', LINUX_VER: 'ubuntu24.04', GPU: 'a100', DRIVER: 'latest' }
         "
 
         echo "BUILD_MATRIX=$(
--- a/.github/workflows/conda-python-tests.yaml
+++ b/.github/workflows/conda-python-tests.yaml
@@ -72,27 +72,22 @@ jobs:
             pull-request:
               # amd64
               - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', GPU: 'l4',   DRIVER: 'earliest', DEPENDENCIES: 'oldest' }
-              # TODO: Use CUDA 12.8.0 below.
-              - { ARCH: 'amd64', PY_VER: '3.13', CUDA_VER: '12.5.1', LINUX_VER: 'ubuntu24.04', GPU: 'h100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
+              - { ARCH: 'amd64', PY_VER: '3.13', CUDA_VER: '12.8.0', LINUX_VER: 'ubuntu24.04', GPU: 'h100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
               # arm64
               - { ARCH: 'arm64', PY_VER: '3.11', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', GPU: 'a100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
             nightly:
               # amd64
-              # TODO: Use CUDA 11.4.3 below.
               - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', GPU: 'l4',   DRIVER: 'earliest', DEPENDENCIES: 'oldest' }
               - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu20.04', GPU: 'h100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
               - { ARCH: 'amd64', PY_VER: '3.11', CUDA_VER: '12.0.1', LINUX_VER: 'rockylinux8', GPU: 'l4',   DRIVER: 'latest',   DEPENDENCIES: 'latest' }
               - { ARCH: 'amd64', PY_VER: '3.12', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', GPU: 'l4',   DRIVER: 'latest',   DEPENDENCIES: 'latest' }
-              # TODO: Use CUDA 12.8.0 below.
-              - { ARCH: 'amd64', PY_VER: '3.13', CUDA_VER: '12.5.1', LINUX_VER: 'ubuntu24.04', GPU: 'h100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
+              - { ARCH: 'amd64', PY_VER: '3.13', CUDA_VER: '12.8.0', LINUX_VER: 'ubuntu24.04', GPU: 'h100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
               # arm64
-              # TODO: Use CUDA 11.4.3 below.
               - { ARCH: 'arm64', PY_VER: '3.9',  CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu20.04', GPU: 'a100', DRIVER: 'latest',   DEPENDENCIES: 'oldest' }
               - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', GPU: 'a100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
               - { ARCH: 'arm64', PY_VER: '3.11', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', GPU: 'a100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
               - { ARCH: 'arm64', PY_VER: '3.12', CUDA_VER: '12.2.2', LINUX_VER: 'rockylinux8', GPU: 'a100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
-              # TODO: Use CUDA 12.8.0 below.
-              - { ARCH: 'arm64', PY_VER: '3.13', CUDA_VER: '12.5.1', LINUX_VER: 'ubuntu24.04', GPU: 'a100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
+              - { ARCH: 'arm64', PY_VER: '3.13', CUDA_VER: '12.8.0', LINUX_VER: 'ubuntu24.04', GPU: 'a100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
           "
 
           # Use the nightly matrix for branch tests
--- a/.github/workflows/wheels-test.yaml
+++ b/.github/workflows/wheels-test.yaml
@@ -83,14 +83,12 @@ jobs:
               - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', GPU: 'l4',   DRIVER: 'latest', DEPENDENCIES: 'latest' }
               - { ARCH: 'amd64', PY_VER: '3.11', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', GPU: 'l4',   DRIVER: 'latest', DEPENDENCIES: 'latest' }
               - { ARCH: 'amd64', PY_VER: '3.12', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', GPU: 'l4',   DRIVER: 'latest', DEPENDENCIES: 'latest' }
-              # TODO: Use CUDA 12.8.0 below.
-              - { ARCH: 'amd64', PY_VER: '3.13', CUDA_VER: '12.5.1', LINUX_VER: 'ubuntu24.04', GPU: 'l4',   DRIVER: 'latest', DEPENDENCIES: 'latest' }
+              - { ARCH: 'amd64', PY_VER: '3.13', CUDA_VER: '12.8.0', LINUX_VER: 'ubuntu24.04', GPU: 'l4',   DRIVER: 'latest', DEPENDENCIES: 'latest' }
               - { ARCH: 'arm64', PY_VER: '3.9',  CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', GPU: 'a100', DRIVER: 'latest', DEPENDENCIES: 'oldest' }
               - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', GPU: 'a100', DRIVER: 'latest', DEPENDENCIES: 'latest' }
               - { ARCH: 'arm64', PY_VER: '3.11', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', GPU: 'a100', DRIVER: 'latest', DEPENDENCIES: 'latest' }
               - { ARCH: 'arm64', PY_VER: '3.12', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', GPU: 'a100', DRIVER: 'latest', DEPENDENCIES: 'latest' }
-              # TODO: Use CUDA 12.8.0 below.
-              - { ARCH: 'arm64', PY_VER: '3.13', CUDA_VER: '12.5.1', LINUX_VER: 'ubuntu24.04', GPU: 'a100', DRIVER: 'latest', DEPENDENCIES: 'latest' }
+              - { ARCH: 'arm64', PY_VER: '3.13', CUDA_VER: '12.8.0', LINUX_VER: 'ubuntu24.04', GPU: 'a100', DRIVER: 'latest', DEPENDENCIES: 'latest' }
           "
 
           TEST_MATRIX=$(yq -n 'env(MATRICES) | .[strenv(BUILD_TYPE)]')
--- a/numba_cuda/numba/cuda/cudadrv/nvrtc.py
+++ b/numba_cuda/numba/cuda/cudadrv/nvrtc.py
@@ -6,6 +6,7 @@ from numba.cuda.cudadrv.error import (
     NvrtcSupportError,
 )
 from numba.cuda.cuda_paths import get_cuda_paths
+
 import functools
 import os
 import threading
@@ -65,6 +66,13 @@ class NVRTC:
     NVVM interface. Initialization is protected by a lock and uses the standard
     (for Numba) open_cudalib function to load the NVRTC library.
     """
+
+    _CU11_2ONLY_PROTOTYPES = {
+        # nvrtcResult nvrtcGetNumSupportedArchs(int *numArchs);
+        "nvrtcGetNumSupportedArchs": (nvrtc_result, POINTER(c_int)),
+        # nvrtcResult nvrtcGetSupportedArchs(int *supportedArchs);
+        "nvrtcGetSupportedArchs": (nvrtc_result, POINTER(c_int)),
+    }
 
     _CU12ONLY_PROTOTYPES = {
         # nvrtcResult nvrtcGetLTOIRSize(nvrtcProgram prog, size_t *ltoSizeRet);
@@ -139,6 +147,8 @@ class NVRTC:
 
                 from numba.cuda.cudadrv.runtime import get_version
 
+                if get_version() >= (11, 2):
+                    inst._PROTOTYPES |= inst._CU11_2ONLY_PROTOTYPES
                 if get_version() >= (12, 0):
                     inst._PROTOTYPES |= inst._CU12ONLY_PROTOTYPES
 
@@ -167,6 +177,53 @@ class NVRTC:
                     setattr(inst, name, checked_call)
 
         return cls.__INSTANCE
+
+    def get_supported_archs(self):
+        """
+        Get Supported Architectures by NVRTC as list of arch tuples.
+        """
+        ver = self.get_version()
+        if ver < (11, 0):
+            raise RuntimeError(
+                "Unsupported CUDA version. CUDA 11.0 or higher is required."
+            )
+        elif ver == (11, 0):
+            return [
+                (3, 0),
+                (3, 2),
+                (3, 5),
+                (3, 7),
+                (5, 0),
+                (5, 2),
+                (5, 3),
+                (6, 0),
+                (6, 1),
+                (6, 2),
+                (7, 0),
+                (7, 2),
+                (7, 5),
+            ]
+        elif ver == (11, 1):
+            return [
+                (3, 5),
+                (3, 7),
+                (5, 0),
+                (5, 2),
+                (5, 3),
+                (6, 0),
+                (6, 1),
+                (6, 2),
+                (7, 0),
+                (7, 2),
+                (7, 5),
+                (8, 0),
+            ]
+        else:
+            num = c_int()
+            self.nvrtcGetNumSupportedArchs(byref(num))
+            archs = (c_int * num.value)()
+            self.nvrtcGetSupportedArchs(archs)
+            return [(archs[i] // 10, archs[i] % 10) for i in range(num.value)]
 
     def get_version(self):
         """
@@ -273,12 +330,36 @@ def compile(src, name, cc, ltoir=False):
     nvrtc = NVRTC()
     program = nvrtc.create_program(src, name)
 
+    version = nvrtc.get_version()
+    ver_str = lambda v: ".".join(v)
+    if version < (11, 0):
+        raise RuntimeError(
+            "Unsupported CUDA version. CUDA 11.0 or higher is required."
+        )
+    else:
+        supported_arch = nvrtc.get_supported_archs()
+        try:
+            found = max(filter(lambda v: v <= cc, [v for v in supported_arch]))
+        except ValueError:
+            raise RuntimeError(
+                f"Device compute capability {ver_str(cc)} is less than the "
+                f"minimum supported by NVRTC {ver_str(version)}. Supported "
+                "compute capabilities are "
+                f"{', '.join([ver_str(v) for v in supported_arch])}."
+            )
+
+        if found != cc:
+            warnings.warn(
+                f"Device compute capability {ver_str(cc)} is not supported by "
+                f"NVRTC {ver_str(version)}. Using {ver_str(found)} instead."
+            )
+
     # Compilation options:
     # - Compile for the current device's compute capability.
     # - The CUDA include path is added.
     # - Relocatable Device Code (rdc) is needed to prevent device functions
     #   being optimized away.
-    major, minor = cc
+    major, minor = found
     arch = f"--gpu-architecture=compute_{major}{minor}"
 
     cuda_include = [
--- a/numba_cuda/numba/cuda/cudadrv/nvvm.py
+++ b/numba_cuda/numba/cuda/cudadrv/nvvm.py
@@ -372,8 +372,10 @@ COMPUTE_CAPABILITIES = (
     (12, 0),
 )
 
+
 # Maps CTK version -> (min supported cc, max supported cc) inclusive
-CTK_SUPPORTED = {
+_CUDA_CC_MIN_MAX_SUPPORT = {
+    (11, 1): ((3, 5), (8, 0)),
     (11, 2): ((3, 5), (8, 6)),
     (11, 3): ((3, 5), (8, 6)),
     (11, 4): ((3, 5), (8, 7)),
@@ -395,7 +397,7 @@ CTK_SUPPORTED = {
 def ccs_supported_by_ctk(ctk_version):
     try:
         # For supported versions, we look up the range of supported CCs
-        min_cc, max_cc = CTK_SUPPORTED[ctk_version]
+        min_cc, max_cc = _CUDA_CC_MIN_MAX_SUPPORT[ctk_version]
         return tuple(
             [cc for cc in COMPUTE_CAPABILITIES if min_cc <= cc <= max_cc]
         )
@@ -423,7 +425,7 @@ def get_supported_ccs():
         return _supported_cc
 
     # Ensure the minimum CTK version requirement is met
-    min_cudart = min(CTK_SUPPORTED)
+    min_cudart = min(_CUDA_CC_MIN_MAX_SUPPORT)
     if cudart_version < min_cudart:
         _supported_cc = ()
         ctk_ver = f"{cudart_version[0]}.{cudart_version[1]}"
--- a/numba_cuda/numba/cuda/tests/cudadrv/test_cuda_array_slicing.py
+++ b/numba_cuda/numba/cuda/tests/cudadrv/test_cuda_array_slicing.py
@@ -193,7 +193,7 @@ class CudaArraySlicing(CUDATestCase):
         for i in range(darr.shape[0]):
             np.testing.assert_array_equal(darr[i:i].copy_to_host(), arr[i:i])
         # empty slice of empty slice
-        self.assertFalse(darr[:0][:0].copy_to_host())
+        self.assertFalse(darr[:0][:0].copy_to_host().size > 0)
         # out-of-bound slice just produces empty slices
         np.testing.assert_array_equal(darr[:0][:1].copy_to_host(), arr[:0][:1])
         np.testing.assert_array_equal(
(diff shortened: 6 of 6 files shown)


numba_cuda/numba/cuda/tests/cudadrv/test_module_init.py
```diff
@@ -0,0 +1,57 @@
...
+    retval = num;
+    return 0;
+}
+"""
+
+        def set_fourty_two(mod):
+            # Initialize 42 to global variable `num`
+            res, dptr, size = cuModuleGetGlobal(
+                mod.handle.value, "num".encode()
+            )
```
isVoid, line 28: Right now we assumes `mod` to be a `ctypesmodule`. But this is about to be changed to `ObjectCode` in #133 .

isVoid: Synced offline - we are still going to proceed assuming we use ctypes module for now.

numba_cuda/numba/cuda/cudadrv/managed_module.py
```diff
@@ -0,0 +1,93 @@
...
+        else:
+            key = self._module.handle.value
+        module_obj = ctx.modules.get(key, None)
+
+        if module_obj is not None:
+            weakref.finalize(
+                module_obj,
+                lazy_callback,
+                self._teardown_callbacks,
+                self._module,
```
isVoid, line 80: Here's a subtle point: per finalizer doc.

> It is important to ensure that func, args and kwargs do not own any references to obj, either directly or indirectly, since otherwise obj will never be garbage collected.

`self._module` is a weakref to the `module_obj`, therefore the finalizer does not create a new reference to the module object itself.

docs/source/user/cuda_ffi.rst
```diff
@@ -160,6 +160,19 @@ CUDA C/C++ source code will be compiled with the `NVIDIA Runtime Compiler
 kernel as either PTX or LTOIR, depending on whether LTO is enabled. Other files
 will be passed directly to the CUDA Linker.
 
+For cumodules linked to a linkable code object, additional setup
+and teardown callbacks are available. Setup functions are invoked once for
+every new module loaded. Teardown functions are invoked before the module is
+garbage collected by python. These callbacks are defined as follows:
```
gmarkall, line 166: I think the module is getting unloaded during the GC process or when the refcount has already dropped to zero outside of a reference cycle, so maybe the following is a closer representation of what happens:

```suggestion
every new module loaded. Teardown functions are invoked just prior to
module unloading. These callbacks are defined as follows:
```

docs/source/user/cuda_ffi.rst
```diff
@@ -160,6 +160,19 @@ CUDA C/C++ source code will be compiled with the `NVIDIA Runtime Compiler
...
+
+.. code::
+
+  def setup_callback(
```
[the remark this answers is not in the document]
isVoid: Yes - I was also debating whether we should list two, though their signature is the same. Perhaps we should just do so.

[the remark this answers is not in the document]
gmarkall: I think that would be fine, yeah. Also, I realise I'm confused about how the "correct" stream is supposed to be selected or passed for the teardown callback, but maybe that will become clear with a bit more thinking / review.

gmarkall:
I just merged main into this PR so it can (potentially) succeed on CI.

numba_cuda/numba/cuda/cudadrv/managed_module.py
```diff
@@ -0,0 +1,93 @@
...
+            if not callable(callback):
+                raise TypeError("all callbacks must be callable")
+
+        self._initialized = False
+        self._module = module
+        self._setup_callbacks = setup_callbacks
+        self._teardown_callbacks = teardown_callbacks
+
+    def _init(self, stream):
+        """Eagerly call the setup functions for cumodule on `stream`
```
gmarkall, line 56: I think the setup function isn't really being called on the stream - it's being called, and given a stream to use for CUDA operations. Is that correct?

numba_cuda/numba/cuda/cudadrv/managed_module.py
```diff
@@ -0,0 +1,93 @@
...
+                teardown(module, stream)
+
+        ctx = devices.get_context()
+        if USE_NV_BINDING:
+            key = self._module.handle
+        else:
+            key = self._module.handle.value
+        module_obj = ctx.modules.get(key, None)
+
+        if module_obj is not None:
```
gmarkall, line 75: Should there be an `else` branch that raises an exception if we can't find the module? Are there any circumstances in which we could attempt to set a finalizer for a non-existent module?

numba_cuda/numba/cuda/cudadrv/managed_module.py
```diff
@@ -0,0 +1,93 @@
...
+            weakref.finalize(
```
gmarkall, line 76: I spent a while trying to refresh my memory of what we do with finalizers in Numba - https://github.com/numba/numba/pull/10001 may be worth a quick look in conjunction with thinking about the behaviour here.

numba_cuda/numba/cuda/cudadrv/managed_module.py
```diff
@@ -0,0 +1,93 @@
...
+        self._module._init(stream)
+
+    def lazy_finalize_module(self, stream):
+        self._module._lazy_finalize(stream)
+
+    def __getattr__(self, name):
+        return getattr(self._cufunc, name)
+
+
+class ManagedModule:
```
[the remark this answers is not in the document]
isVoid: Yes - I think the existing class should suffice. Let me rework towards that, it's always good to fewer code to maintain.

numba_cuda/numba/cuda/cudadrv/managed_module.py
isVoid: Correct - at this stage we don't know what operation the user is going to perform inside the callbacks.

numba_cuda/numba/cuda/cudadrv/managed_module.py
isVoid: I'm really not sure. I could perform a print to see if in any scenario within the test suite the module would be non-existent.

docs/source/user/cuda_ffi.rst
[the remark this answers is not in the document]
gmarkall: Yes. Maybe teardown shouldn't take a stream? unloading a module must surely synchronize the device, so there seems not to be much point in trying to do the teardown on a stream. (I know we're not unloading the module now, but we *should* be, and ought to in future)

numba_cuda/numba/cuda/cudadrv/managed_module.py
gmarkall: Thanks - in that case, would this more closely reflect what `_init()` does

```suggestion
        """Call the setup functions for cumodule with the given `stream`
```

numba_cuda/numba/cuda/cudadrv/managed_module.py
gmarkall: I think I'd err on the side of caution, and raise an exception if the module can't be found - if we didn't find one, I think we're going to omit a call to a teardown function that I think would otherwise have been expected to happen.

numba_cuda/numba/cuda/cudadrv/managed_module.py
isVoid: It even took me 30min to understand this refresh PR. I think this is super subtle. Especially the point "the first finalizer registered also calls `atexit.register` to place a function that invokes all finalizers." Is this documented anywhere or just a CPython detail?

It's up to the user to decide whether to inspect the `shutting_down` variable. If they decide to check that, it's probably a good idea to put a reminder about this subtlety. For the most part, if the finalizer is to operate on just the modules, we are probably fine since they are still alive when the finalizer is invoked.

numba_cuda/numba/cuda/cudadrv/managed_module.py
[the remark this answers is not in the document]
isVoid: Addressed in d8f2f23. It appears that this addition is transparent to whether the module type is `CtypesModule` or `CudaPythonModule`. If `CudaPythonModule` is used, the callbacks will receive that instead.

numba_cuda/numba/cuda/cudadrv/managed_module.py
isVoid: After refactoring the code into `Module`, this issue naturally goes away because `self` is an owning object of the module. We no longer need to retrieve it from the context.

numba_cuda/numba/cuda/cudadrv/managed_module.py
isVoid: This is applied to the similar method in `Module` named `setup`.

docs/source/user/cuda_ffi.rst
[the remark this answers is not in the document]
gmarkall: > IIRC we are unloading the module. It is `_module_finalizer` in driver.py no?

Ah, I see the disconnect here... I was thinking we don't unload it as soon as we can (which is when the kernel / jitted function that created it is no longer reachable) because the current context holds on to a reference to all modules. If the context were destroyed then we would probably unload the module, but I think that never happens in practice in any user code.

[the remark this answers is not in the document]
gmarkall: With your proposal nullified, does that mean you settled on the idea of not having teardown take a stream? Or just to think more about it for another way?

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2406,6 +2416,49 @@ def get_function(self, name):
...
+
+        setattr(self, attr, callbacks)
+
+    def set_setup_functions(self, callbacks):
+        """Set the setup callback for this module"""
+        self._set_callables(callbacks, "setup_functions")
+
+    def set_teardown_functions(self, callbacks):
+        """Set the finalize callback for this module"""
+        self._set_callables(callbacks, "teardown_functions")
```
gmarkall, line 2435: Do we need all these extra functions instead of just taking setup and teardown functions as parameters to the `__init__()` function?

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2406,6 +2416,49 @@ def get_function(self, name):
...
+
+    def setup(self, stream):
+        """Call the setup functions for cumodule with the given `stream`"""
+        if self.setup_functions is None:
+            return
+
+        for f in self.setup_functions:
+            f(weakref.proxy(self), stream)
```
gmarkall, line 2443: Is the `weakref.proxy` used because we expect the setup function might retain a reference to this module? Do we expect users to do that? (e.g. is that a requirement in the use cases you've been discussing with potential users of this?)

gmarkall: I guess I'm thinking - is this accidentally just mirroring the `set_finalizers()` function, where we use `weakref.proxy()` to avoid creating a cycle that would prevent us from every unloading the module?

numba_cuda/numba/cuda/tests/cudadrv/test_module_callbacks.py
```diff
@@ -0,0 +1,204 @@
...
+        def kernel():
+            pass
+
+        self.assertEqual(counter, 0)
+        kernel[1, 1]()
+        self.assertEqual(counter, 1)
+        kernel[1, 1]() # cached
+        self.assertEqual(counter, 1)
+
+        wipe_all_modules_in_context()
```
gmarkall, line 46: I feel that for correctness we should also `del kernel`, otherwise it holds a reference to the now-removed module (and context). It shouldn't make a difference for this test case because we don't do anything with it afterwards, but I think it's probably a good idea to ensure that the test case follows a good pattern.

(Also - if we unloaded modules when we should be, doing `del kernel` ought to be enough to force unloading - so perhaps `wipe_all_modules_in_context()` ought to be documented as a hack until we properly unload modules with kernel deletion.

numba_cuda/numba/cuda/tests/cudadrv/test_module_callbacks.py
```diff
@@ -0,0 +1,204 @@
...
+        module = """
+__device__ int num = 0;
+extern "C"
+__device__ int get_num(int &retval) {
+    retval = num;
+    return 0;
+}
+"""
+
+        self.counter = 0
```
gmarkall, line 166: Should we be checking the value of `self.counter` in the test cases in this class as well?

gmarkall (changes requested):
Thanks for the changes - I think we're getting most of the way there now (assuming you're happy with the general direction too). I have made a few comments on the diff.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -1442,6 +1442,13 @@ def create_module_image(self, image):
         self.modules[key] = module
         return weakref.proxy(module)
 
+    def create_module_image_with_callbacks(
```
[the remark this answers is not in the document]
isVoid: Not really - it was based on the idea that I don't want to impact existing API in work. But since this isn't a breaking change, we can go forward with the proposal.

numba_cuda/numba/cuda/tests/cudadrv/test_module_callbacks.py
isVoid: > otherwise it holds a reference to the now-removed module (and context).

Summarizing our discussion offline: essentially because `Function` is not passed around via weakref like modules do, when the context is cleared, the kernel handle still holds onto the function pointer, which is no longer valid because its correponding cumodule has been unloaded. In the short term, we should add a line of `del kernel` in the test code to indicate that the kernel handle is no longer valid to avoid promoting any anti-pattern of use-after-clear for external users.

In the long term, I think we cufunc should also be passed around in weakref and managed by context as well. On module unload, there should be a map that link the cufunc instances to the module that is being unloaded, and release those references from the context as well. When user trys to invoke the kernel after the module is unloaded, the reference to the actual cufunc will become `None` when dereferencing, subsequent user error can be re-surfaced from there. This should provide a better UX to the current `numba.cuda.cudadrv.driver.CudaAPIError: [400] Call to cuLaunchKernel results in CUDA_ERROR_INVALID_HANDLE` if we try to use after free.

isVoid: > if we unloaded modules when we should be, doing del kernel ought to be enough to force unloading

From what you suggest, you seem to be suggesting that there should be a 1:1 mapping from module to functions (which is the case for modules that contains Numba kernels). If we remove references to the kernel (function), we should also remove references to the module. 

The premise of this is whether we regard the `kernel` object as the single frontend how user can interact with the function and module. After we remove this handle, are there (should we) provide any other ways for user to access that module/function? (reloading from cache, perhaps?) If not, then there should be a finalizer for `CUDADispatcher` that looks through the references it holds for Function and Modules and deref them all.

numba_cuda/numba/cuda/tests/cudadrv/test_module_callbacks.py
isVoid: I think the counter tests in the base class is probably sufficient - I'd like to isolate this test only for some actual driver calls.

numba_cuda/numba/cuda/tests/cudadrv/test_module_callbacks.py
```diff
@@ -0,0 +1,204 @@
...
+def wipe_all_modules_in_context():
+    ctx = cuda.current_context()
+    ctx.reset()
+
+
+@unittest.skipIf(
+    config.CUDA_USE_NVIDIA_BINDING,
+    "NV binding support superceded by cuda.bindings."
+)
+class TestModuleCallbacksBasic(CUDATestCase):
```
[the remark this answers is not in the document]
isVoid: Done.

numba_cuda/numba/cuda/tests/cudadrv/test_module_callbacks.py
```diff
@@ -0,0 +1,204 @@
...
+        self.assertEqual(counter, 0)
+        kernel[1, 1]()
+        self.assertEqual(counter, 1)
+        kernel[1, 1]() # cached
+        self.assertEqual(counter, 1)
+
+        wipe_all_modules_in_context()
+        self.assertEqual(counter, 0)
+
+    def test_different_argtypes(self):
```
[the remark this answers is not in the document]
isVoid: I added the modules to a set holding all the modules that's seen in the setup/teardown functions and assert that uniquely there are in total 2 modules.

numba_cuda/numba/cuda/tests/cudadrv/test_module_callbacks.py
gmarkall: > If not, then there should be a finalizer for `CUDADispatcher` that looks through the references it holds for Function and Modules and deref them all.

Yes, this is what I think should happen - although I think I'd suggest putting the finalizer on the `_Kernel` object because they have a 1:1 mapping to modules. When a kernel is finalized, it should remove its module from the context's `modules` dict. But I am not suggesting addressing that in this PR - I think it would be better to get this over the line because it's really a separate issue.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2406,6 +2422,46 @@ def get_function(self, name):
...
 
+    def _set_callables(self, callbacks, attr):
+        if not isinstance(callbacks, list):
+            raise TypeError("callbacks must be a list")
+
+        for f in callbacks:
+            if not callable(f):
+                raise TypeError("callback must be callable")
+
+        setattr(self, attr, callbacks)
```
gmarkall, line 2433: I think this is dead code now?

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2406,6 +2422,46 @@ def get_function(self, name):
...
+
+    def setup(self):
+        """Call the setup functions for cumodule with the given `stream`"""
+        if self.setup_functions is None or self.initialized:
```
gmarkall, line 2437: Should it be an error to call `setup()` twice? The reason I ask is that there's an asymmetry between `setup()` and `_set_finalizers()` - Calling `setup()` twice will result in the setup functions being called once. But calling `_set_finalizers()` twice looks like it will result in the teardown callbacks being run twice at module unload time.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2406,6 +2422,46 @@ def get_function(self, name):
...
+            raise TypeError("callbacks must be a list")
+
+        for f in callbacks:
+            if not callable(f):
+                raise TypeError("callback must be callable")
+
+        setattr(self, attr, callbacks)
+
+    def setup(self):
+        """Call the setup functions for cumodule with the given `stream`"""
```
gmarkall, line 2436:
```suggestion
        """Call the setup functions for cumodule"""
```

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2406,6 +2422,46 @@ def get_function(self, name):
...
+        for f in self.setup_functions:
+            f(self.handle)
+
+        self.initialized = True
+
+    def _set_finalizers(self):
+        """Create finalizers that tears down the cumodule.
+
+        Unlike the setup functions which takes in streams, Numba does not
+        provide a stream object to the teardown function.
```
gmarkall, line 2449:
```suggestion
        """Create finalizers that tear down the module.
```

numba_cuda/numba/cuda/cudadrv/linkable_code.py
```diff
@@ -7,11 +7,17 @@ class LinkableCode:
     :param data: A buffer containing the data to link.
     :param name: The name of the file to be referenced in any compilation or
                  linking errors that may be produced.
+    :param setup_callback Callback function on kernel before launching.
+    :param teardown_callback Callback function on kernel before tearing down.
```
gmarkall, line 11: I also found it a bit tricky to parse the meaning of these - what do you think about:

```suggestion
    :param setup_callback: A function called prior to the launch of a kernel
                           contained within a module that has this code object
                           linked into it.
    :param teardown_callback: A function called just prior to the unloading of
                              a module that has this code object linked into
                              it.
```
?

numba_cuda/numba/cuda/cudadrv/driver.py
gmarkall: I gues `_set_finalizers()` isn't invoked from outside the class, so maybe for that reason the asymmetry doesn't matter.

numba_cuda/numba/cuda/tests/cudadrv/test_module_callbacks.py
```diff
@@ -0,0 +1,201 @@
...
+    def test_jitarg(self):
+        get_num = cuda.declare_device("get_num", "int32()")
+
+        @cuda.jit(link=[self.lib])
+        def kernel(arr):
+            arr[0] = get_num()
+
+        arr = np.zeros(1, np.int32)
+        kernel[1, 1](arr)
+        self.assertEqual(arr[0], 42)
```
gmarkall, line 201: We should also be able to run the test file standalone:

```suggestion
        self.assertEqual(arr[0], 42)
        
if __name__ == '__main__':
    unittest.main()
```

numba_cuda/numba/cuda/tests/cudadrv/test_module_callbacks.py
gmarkall, line 1: With the following local changes I was able to hack this up so that the tests work with both the legacy NVIDIA bindings and the ctypes ones:

```diff
diff --git a/numba_cuda/numba/cuda/tests/cudadrv/test_module_callbacks.py b/numba_cuda/numba/cuda/tests/cudadrv/test_m
odule_callbacks.py
index 53afe20..83eadd5 100644
--- a/numba_cuda/numba/cuda/tests/cudadrv/test_module_callbacks.py
+++ b/numba_cuda/numba/cuda/tests/cudadrv/test_module_callbacks.py
@@ -1,11 +1,14 @@
-import unittest
-
 import numpy as np
 
 from numba import cuda, config
 from numba.cuda.cudadrv.linkable_code import CUSource
-from numba.cuda.testing import CUDATestCase, ContextResettingTestCase
-from numba.cuda.cudadrv.drvapi import cu_module
+from numba.cuda.testing import (CUDATestCase, ContextResettingTestCase,
+                                unittest)
+
+if config.CUDA_USE_NVIDIA_BINDING:
+    from cuda.cuda import CUmodule as cu_module_type
+else:
+    from numba.cuda.cudadrv.drvapi import cu_module as cu_module_type
 
 from cuda.bindings.driver import cuModuleGetGlobal, cuMemcpyHtoD
 
@@ -15,22 +18,18 @@ def wipe_all_modules_in_context():
     ctx.reset()
 
 
-@unittest.skipIf(
-    config.CUDA_USE_NVIDIA_BINDING,
-    "NV binding support superceded by cuda.bindings."
-)
 class TestModuleCallbacksBasic(ContextResettingTestCase):
 
     def test_basic(self):
         counter = 0
 
         def setup(handle):
-            self.assertTrue(isinstance(handle, cu_module))
+            self.assertTrue(isinstance(handle, cu_module_type))
             nonlocal counter
             counter += 1
 
         def teardown(handle):
-            self.assertTrue(isinstance(handle, cu_module))
+            self.assertTrue(isinstance(handle, cu_module_type))
             nonlocal counter
             counter -= 1
 
@@ -58,12 +57,16 @@ class TestModuleCallbacksBasic(ContextResettingTestCase):
         def setup(handle):
             nonlocal counter, setup_seen
             counter += 1
-            setup_seen.add(handle.value)
+            if not config.CUDA_USE_NVIDIA_BINDING:
+                handle = handle.value
+            setup_seen.add(handle)
 
         def teardown(handle):
             nonlocal counter
             counter -= 1
-            teardown_seen.add(handle.value)
+            if not config.CUDA_USE_NVIDIA_BINDING:
+                handle = handle.value
+            teardown_seen.add(handle)
 
         lib = CUSource("", setup_callback=setup, teardown_callback=teardown)
 
@@ -94,12 +97,16 @@ class TestModuleCallbacksBasic(ContextResettingTestCase):
         def setup(handle):
             nonlocal counter, setup_seen
             counter += 1
-            setup_seen.add(handle.value)
+            if not config.CUDA_USE_NVIDIA_BINDING:
+                handle = handle.value
+            setup_seen.add(handle)
 
         def teardown(handle):
             nonlocal counter, teardown_seen
             counter -= 1
-            teardown_seen.add(handle.value)
+            if not config.CUDA_USE_NVIDIA_BINDING:
+                handle = handle.value
+            teardown_seen.add(handle)
 
         lib = CUSource("", setup_callback=setup, teardown_callback=teardown)
 
@@ -167,9 +174,12 @@ __device__ int get_num(int &retval) {
 """
 
         def set_forty_two(handle):
+            if not config.CUDA_USE_NVIDIA_BINDING:
+                handle = handle.value
+
             # Initialize 42 to global variable `num`
             res, dptr, size = cuModuleGetGlobal(
-                handle.value, "num".encode()
+                handle, "num".encode()
             )
 
             arr = np.array([42], np.int32)
@@ -199,3 +209,7 @@ __device__ int get_num(int &retval) {
         arr = np.zeros(1, np.int32)
         kernel[1, 1](arr)
         self.assertEqual(arr[0], 42)
+
+
+if __name__ == '__main__':
+    unittest.main()
```

Would it be possible to update the test so it works with both sets of bindings, as we have implemented support for both the ctypes and the legacy NVIDIA bindings (or make it so it's not possible to use callbacks with the legacy NVIDIA bindings)? The above could do with a bit of a clean up, but it shows where I got to.

gmarkall (changes requested):
Many thanks for the updates - I have added another set of comments on the diff - all the items seem like minor things, the overall picture seems to be looking very good.

numba_cuda/numba/cuda/cudadrv/driver.py
isVoid: I prefer raising an error. The current state quietly skips setup function if it's been initialized once, which may go against what user would expect.

docs/source/user/cuda_ffi.rst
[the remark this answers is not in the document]
isVoid: The most recent version of the design dropped stream support for both setup and teardown function since this wasn't originally the ask of the feature request. And I don't want to over-engineer some unexpected stream behavior for callbacks.

numba_cuda/numba/cuda/tests/cudadrv/test_module_callbacks.py
isVoid: Addressed in  97367ce

numba_cuda/numba/cuda/tests/cudadrv/test_module_callbacks.py
isVoid: https://github.com/NVIDIA/numba-cuda/issues/171

numba_cuda/numba/cuda/cudadrv/linkable_code.py
gmarkall: The CI failure is because the colons after `setup_callback` and `teardown_callback` are missing in the committed version.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2406,6 +2422,36 @@ def get_function(self, name):
...
+        if self.setup_functions is None:
+            return
+
+        for f in self.setup_functions:
+            f(self.handle)
+
+        self.initialized = True
+
+    def _set_finalizers(self):
+        """Create finalizers that tears down the module.
```
gmarkall, line 2439: A tiny nit (which you might want to resolve if you're fixing things up for the `LinkableCode` docstring below):
```suggestion
        """Create finalizers that tear down the module.
```

gmarkall (approved):
I think this is looking good - there's a couple of syntax errors in the `LinkableCode` docstring but I've pushed a fix rather than making use do another iteration. I think this is good to merge once CI has passed.

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -310,6 +310,10 @@ def bind(self):
         """
         cufunc = self._codelibrary.get_cufunc()
 
+        mod = cufunc.module
+        if not mod.initialized:
+            mod.setup()
```
kkraus14, line 315: Is this thread safe? Can this accidentally get run multiple times and cause unintended behavior?

isVoid: Are we discussing about the use of multitheading in free-threaded python scenario? If multiple threads launches the kernel, each thread can indeed see the `initialized` as `False` and therefore run `setup` multiple times.

In general, do we want to start considering for free-threaded python for all new features?

gmarkall: No and yes, I think.

gmarkall: (didn't see the other replies between starting and finishing thinking about this - I was replying to Keith)

isVoid: Discussed offline that we can go with using a lock to make sure thread safety. If there comes another ask to lift such restriction from another library, we'll revisit.

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -304,15 +305,18 @@ def _reduce_states(self):
                     debug=self.debug, lineinfo=self.lineinfo,
                     call_helper=self.call_helper, extensions=self.extensions)
 
+    @module_init_lock
```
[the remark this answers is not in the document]
isVoid: I'm crafting a concurrent test so that we can get a more in depth view of the behavior of setup and teardown with threads. I'll report back.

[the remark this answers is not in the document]
isVoid: When I delved into the multi-threading kernel launches, this is what I discovered:

- The compilation is run once for the first thread that acquired the compilation lock, and the subsequent thread can load the same compiled result from cache.
- Each thread creates its own cumodule pointer and independently invokes the cumoduleLoadDataEx on their own copy of the pointer. So N modules will be created if there are N threads.


### What this means for the setup callback function:
1. Setup function is called N times, if there are N threads. Each invocation with a unique module handle.
2. There is a lock for the setup section, the N invocations will take place in serial. No race condition.

### What this means for the teardown callback function (Today):
- Today, user cannot teardown the kernel. So we assume that all teardown happens after all threads joins and the main thread takes care of the interpreter shutdown and the finalizers. Finalizers are placed in a FILO stack, so they are invoked in serial. No race condition.

### What this means for the teardown callback function (If we implement #171)
- User can delete the kernel handle and unload the module in their thread. The order where the finalizer are invoked are indeterministic. Each finalizer that are invoked with a unique module handle.
  - There maybe race condition if user attempts to access the same address in the teardown function. I'd suggest a safer way to write the teardown is to designate a main thread based on threadId / module handle if global access is required. Implementing a lock for the teardown function should prevent this race condition.
  - There could also be other footgun how user could misuse the teardown function, such as double freeing a pointer in global memory etc. In general, I'd like to defer these discussions to #171 when we plan to implement this.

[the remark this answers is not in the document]
kkraus14: > * Each thread creates its own cumodule pointer and independently invokes the `cumoduleLoadDataEx` on their own copy of the pointer. So `N` modules will be created if there are `N` threads.

Does calling `cumoduleLoadDataEx` consume valuable resources like global device memory? I would assume so.

Do we have one CUDA Context being used by multiple threads?

If the answer is yes to both of these questions, maybe we should implement a lock where the first thread acquires the lock and creates the module and then all future threads can retrieve the module from a cache? Then we presumably only need to call the setup callback function once instead of N times?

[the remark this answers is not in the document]
isVoid: > > * Each thread creates its own cumodule pointer and independently invokes the `cumoduleLoadDataEx` on their own copy of the pointer. So `N` modules will be created if there are `N` threads.
> 
> Does calling `cumoduleLoadDataEx` consume valuable resources like global device memory? I would assume so.
> 
> Do we have one CUDA Context being used by multiple threads?
> 
> If the answer is yes to both of these questions, maybe we should implement a lock where the first thread acquires the lock and creates the module and then all future threads can retrieve the module from a cache? Then we presumably only need to call the setup callback function once instead of N times?

Yes and yes, I think Graham and I touched these questions before and we agreed there were some subtle bug with module creation in Numba. Your suggestion makes sense to me.

gmarkall:
Just circling back to this PR. This fail seems to suggest we hae some issue somewhere:

```
FAIL: test_concurrent_initialization_different_args (numba.cuda.tests.cudadrv.test_module_callbacks.TestMultithreadedCallbacks.test_concurrent_initialization_different_args)
  File "/pyenv/versions/3.12.9/lib/python3.12/site-packages/numba_cuda/numba/cuda/tests/cudadrv/test_module_callbacks.py", line 290, in test_concurrent_initialization_different_args
    self.assertEqual(max_seen_mods, 8)  # 2 kernels per thread
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: 7 != 8
```

gmarkall:
If I put a global compiler lock in the`Dispatcher.compile()` method:

```diff
diff --git a/numba_cuda/numba/cuda/dispatcher.py b/numba_cuda/numba/cuda/dispatcher.py
index e68c3b7..72344aa 100644
--- a/numba_cuda/numba/cuda/dispatcher.py
+++ b/numba_cuda/numba/cuda/dispatcher.py
@@ -1111,6 +1111,7 @@ class CUDADispatcher(Dispatcher, serialize.ReduceMixin):
         self._insert(c_sig, kernel, cuda=True)
         self.overloads[argtypes] = kernel
 
+    @global_compiler_lock
     def compile(self, sig):
         """
         Compile and bind to the current context a version of this kernel
```

then the test fails as:

```
FAIL: test_concurrent_initialization_different_args (numba.cuda.tests.cudadrv.test_module_callbacks.TestMultithreadedCallbacks.test_concurrent_initialization_different_args)
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/tests/cudadrv/test_module_callbacks.py", line 290, in test_concurrent_initialization_different_args
    self.assertEqual(max_seen_mods, 8)  # 2 kernels per thread
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: 2 != 8
```

which is more like what I'd expect given what the test does - there are two kernel compilations (and therefore should be two modules) because of the int parameter and the float parameter. I'm not sure my diff above is the correct fix, but we clearly have a race condition exposed in `Dispatcher.compile()`, which is almost certainly related to the CI fail.

numba_cuda/numba/cuda/dispatcher.py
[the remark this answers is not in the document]
isVoid: I'm happy to address the issue in this PR.

isVoid:
> AssertionError: 7 != 8

I think this means that one thread is only scheduled after 1 out of the other 7 threads have finished compiling, and this thread reloads the module from cache, which means our current test is non-deterministic.

I will implement https://github.com/NVIDIA/numba-cuda/pull/145#issuecomment-2791862779 and get this behavior stable.

gmarkall:
> I think this means that one thread is only scheduled after 1 out of the other 7 threads have finished compiling, and this thread reloads the module from cache, which means our current test is non-deterministic.

I think that's OK - the outcome of the test was non-deterministic because we had a race condition in `Dispatcher.compile()`. Are we thinking about things the same way here?

PR #147, opened 2025-03-05
Move publish step out of `wheels-build.yaml`

KyleFromNVIDIA:
Also change an instance of `mambabuild` to `build`.

Follow-up to https://github.com/NVIDIA/numba-cuda/pull/142#issuecomment-2701214971

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -77,6 +77,25 @@ jobs:
       build_type: pull-request
       script: "ci/build_wheel.sh"
       matrix_filter: ${{ needs.compute-matrix.outputs.BUILD_MATRIX }}
+  # TODO: Remove this job before merging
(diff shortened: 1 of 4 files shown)


KyleFromNVIDIA:
[This job](https://github.com/NVIDIA/numba-cuda/actions/runs/13683543371/job/38261670185?pr=147) shows that the wheels are successfully downloaded from the reusable workflow when required. I am removing the `fake-publish-wheels` job.

gmarkall (approved):

webknjaz:
@gmarkall @KyleFromNVIDIA this is partially what I meant. The move is correct. The next step would be to set up environment and OIDC privileges, get rid of the token and set up trust on PyPI.
Without that, you're not going to get metadata marked as verified and the attestations won't be produced either — they rely on OIDC too.

Issue #65, opened 2024-10-29
[FEA] Support for loading NVVM from a wheel


gmarkall:
NVVM distributed in a wheel is not currently located by the library detection mechanisms. Additional logic to locate NVVM in a wheel is required.

gmarkall:
Patch possibly in https://github.com/NVIDIA/nvmath-python/blob/main/nvmath/device/patch.py

leofang:
Follow-up: #277

PR #155, opened 2025-03-11
Locate nvvm, libdevice, nvrtc, and cudart from nvidia-*-cu12 wheels

brandon-b-miller:
Closes https://github.com/NVIDIA/numba-cuda/issues/66
Closes https://github.com/NVIDIA/numba-cuda/issues/65

WIP, current code finds nvvm/libdevice which is enough to launch kernels, nvrtc support is next. Logic vendored from `nvmath-python`


numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -48,19 +53,58 @@ def _get_nvvm_path_decision():
...
+            dso_dir = os.path.join(
+                sp,
+                "nvidia",
+                "cuda_nvcc",
+                "nvvm",
+                dso_dir
+            )
+            dso_path = os.path.join(dso_dir, dso_path)
+            if os.path.exists(dso_path):
+                return str(Path(dso_path).parent)
```
kkraus14, line 94: I commented on this in https://github.com/NVIDIA/cuda-python/issues/441#issuecomment-2714804149, but we may want to just consider trying to import the `nvidia` package and then subsequently trying to import the `cuda_nvcc` package from the `nvidia` package instead of manually traversing the paths?  We can then use `nvidia.cuda_nvcc.__path__` which always resolve to `sp/nvidia/cuda_nvcc` and will follow the general python rules for which package takes priority properly.

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -48,19 +53,58 @@ def _get_nvvm_path_decision():
...
+        options.append(('System', nvvm_ctk_dir))
+
     by, path = _find_valid_path(options)
     return by, path
 
 
+def _get_nvvm_wheel():
+    site_paths = [
+        site.getusersitepackages()
+    ] + site.getsitepackages() + ["conda", None]
```
kkraus14, line 70: Why do we need to add a `conda` path here? If someone installed the wheel in a conda environment it would presumably be in the `site.getsitepackages()`?

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -48,19 +53,58 @@ def _get_nvvm_path_decision():
...
+        if sys.platform.startswith("linux"):
+            dso_dir = "lib64"
+            # Hack: libnvvm from Linux wheel
+            # does not have any soname (CUDAINST-3183)
+            dso_path = "libnvvm.so"
+        elif sys.platform.startswith("win32"):
+            dso_dir = "bin"
+            dso_path = "nvvm64_40_0.dll"
+        else:
+            raise AssertionError()
```
kkraus14, line 82: Can pull this out of the `site_paths` loop I think?

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -29,9 +30,13 @@ def _get_libdevice_path_decision():
         ('Conda environment', get_conda_ctk()),
         ('Conda environment (NVIDIA package)', get_nvidia_libdevice_ctk()),
         ('CUDA_HOME', get_cuda_home('nvvm', 'libdevice')),
-        ('System', get_system_ctk('nvvm', 'libdevice')),
         ('Debian package', get_debian_pkg_libdevice()),
+        ('NVIDIA NVCC Wheel', get_libdevice_wheel()),
     ]
+    libdevice_ctk_dir = get_system_ctk('nvvm', 'libdevice')
```
gmarkall, line 36: Why did we move the system toolkit after the Debian-packaged version? I think we want to preserve the order if we can.

numba_cuda/numba/cuda/cuda_paths.py
gmarkall: It might also be good to raise the exception with some explanation of what is wrong?

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -48,19 +53,58 @@ def _get_nvvm_path_decision():
...
-    # Search for pattern
-    pat = r'libdevice(\.\d+)*\.bc$'
-    candidates = find_file(re.compile(pat), libdir)
-    # Keep only the max (most recent version) of the bitcode files.
-    out = max(candidates, default=None)
+    if by == "NVIDIA NVCC Wheel":
+        # The NVVM path is a directory, not a file
+        out = os.path.join(libdir, "libdevice.10.bc")
+    else:
+        # Search for pattern
```
gmarkall, line 103: I think it's called `libdevice.10.bc` in all supported toolkit versions, so this logic is probably no longer needed - will check and update here.

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -48,19 +53,58 @@ def _get_nvvm_path_decision():
...
+
 def _get_libdevice_paths():
     by, libdir = _get_libdevice_path_decision()
-    # Search for pattern
-    pat = r'libdevice(\.\d+)*\.bc$'
-    candidates = find_file(re.compile(pat), libdir)
-    # Keep only the max (most recent version) of the bitcode files.
-    out = max(candidates, default=None)
+    if by == "NVIDIA NVCC Wheel":
+        # The NVVM path is a directory, not a file
```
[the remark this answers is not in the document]
gmarkall: I just realised this is a copy/paste error from below.

gmarkall (changes requested):
A few questions on the diff - in addition, do we plan to add a CI config that installs these from wheels so that we know it will continue to work?

numba_cuda/numba/cuda/cuda_paths.py
gmarkall: I just checked - even back in 11.2 it is called `libdevice.10.bc` - so we don't need to search and choose from a set of candidates anymore.

brandon-b-miller:
> A few questions on the diff - in addition, do we plan to add a CI config that installs these from wheels so that we know it will continue to work?

Yes, I'll see about adding a separate CI job for this

ci/test_wheel_deps_wheels.sh
```diff
@@ -0,0 +1,52 @@
...
+RAPIDS_TESTS_DIR=${RAPIDS_TESTS_DIR:-"${PWD}/test-results"}/
+mkdir -p "${RAPIDS_TESTS_DIR}"
+pushd "${RAPIDS_TESTS_DIR}"
+
+rapids-logger "Show Numba system info"
+python -m numba --sysinfo
+
+# remove cuda-nvvm-12-5 leaving libnvvm.so from nvidia-cuda-nvcc-cu12 only 
+apt-get update
+apt remove --purge cuda-nvvm-12-5 -y
```
brandon-b-miller, line 47: This combined with the addition of `nvidia-cuda-nvcc-cu12` was the easiest way I could think of to get to the relevant test environment, but I'm by no means married to it, this would have to be dynamic wrt the minor version as well.

gmarkall: You can get the installed package name with something like

```bash
CUDA_NVVM_PACKAGE=`dpkg --get-selections | grep cuda-nvvm | awk '{print $1}'`
```

ZzEeKkAa:
I've merged this branch with main (fbbc0402ca14b931fd00f2484c0af34730058bfe) and tested on `nvmath-python`. I was able successfully get rid of this patch:

```
    # our device apis only support cuda 12+
    _utils.force_loading_nvrtc("12")
    nvrtc.NVRTC.__new__ = __nvrtc_new__
```

But can't get rid of 
```
    # Patch Numba to support wheels
    _utils.patch_numba_nvvm(nvvm)
```

I'm getting the error:
```
> python ./examples/device/cublasdx_simple_gemm_fp32.py
/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/numba-cuda/numba_cuda/numba/cuda/dispatcher.py:663: NumbaPerformanceWarning: Grid size 1 will likely result in GPU under-utilization due to low occupancy.
  warn(NumbaPerformanceWarning(msg))
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/./examples/device/cublasdx_simple_gemm_fp32.py", line 78, in <module>
    main()
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/./examples/device/cublasdx_simple_gemm_fp32.py", line 68, in main
    f[1, block_dim](a_d, b_d, c_d, alpha, beta, o_d)
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 666, in __call__
    return self.dispatcher.call(args, self.griddim, self.blockdim,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 808, in call
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 816, in _compile_for_args
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 1065, in compile
    kernel = _Kernel(self.py_func, argtypes, **self.targetoptions)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/compiler_lock.py", line 35, in _acquire_compile_lock
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 156, in __init__
    cres = compile_cuda(self.py_func, types.void, self.argtypes,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/compiler_lock.py", line 35, in _acquire_compile_lock
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/numba-cuda/numba_cuda/numba/cuda/compiler.py", line 290, in compile_cuda
    cres = compiler.compile_extra(typingctx=typingctx,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/compiler.py", line 739, in compile_extra
    return pipeline.compile_extra(func)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/compiler.py", line 439, in compile_extra
    return self._compile_bytecode()
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/compiler.py", line 505, in _compile_bytecode
    return self._compile_core()
           ^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/compiler.py", line 481, in _compile_core
    raise e
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/compiler.py", line 473, in _compile_core
    pm.run(self.state)
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/compiler_machinery.py", line 363, in run
    raise e
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/compiler_machinery.py", line 356, in run
    self._runPass(idx, pass_inst, state)
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/compiler_lock.py", line 35, in _acquire_compile_lock
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/compiler_machinery.py", line 311, in _runPass
    mutated |= check(pss.run_pass, internal_state)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/compiler_machinery.py", line 272, in check
    mangled = func(compiler_state)
              ^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/typed_passes.py", line 112, in run_pass
    typemap, return_type, calltypes, errs = type_inference_stage(
                                            ^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/typed_passes.py", line 93, in type_inference_stage
    errs = infer.propagate(raise_errors=raise_errors)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/typeinfer.py", line 1066, in propagate
    errors = self.constraints.propagate(self)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/typeinfer.py", line 160, in propagate
    constraint(typeinfer)
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/typeinfer.py", line 566, in __call__
    self.resolve(typeinfer, typevars, fnty)
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/typeinfer.py", line 589, in resolve
    sig = typeinfer.resolve_call(fnty, pos_args, kw_args)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/typeinfer.py", line 1560, in resolve_call
    return self.context.resolve_function_type(fnty, pos_args, kw_args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/typing/context.py", line 195, in resolve_function_type
    res = self._resolve_user_function_type(func, args, kws)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/typing/context.py", line 247, in _resolve_user_function_type
    return func.get_call_type(self, args, kws)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/types/functions.py", line 538, in get_call_type
    self.dispatcher.get_call_template(args, kws)
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 979, in get_call_template
    self.compile_device(tuple(args))
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 1016, in compile_device
    cres = compile_cuda(self.py_func, return_type, args,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/compiler_lock.py", line 35, in _acquire_compile_lock
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/numba-cuda/numba_cuda/numba/cuda/compiler.py", line 290, in compile_cuda
    cres = compiler.compile_extra(typingctx=typingctx,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/compiler.py", line 739, in compile_extra
    return pipeline.compile_extra(func)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/compiler.py", line 439, in compile_extra
    return self._compile_bytecode()
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/compiler.py", line 505, in _compile_bytecode
    return self._compile_core()
           ^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/compiler.py", line 481, in _compile_core
    raise e
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/compiler.py", line 473, in _compile_core
    pm.run(self.state)
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/compiler_machinery.py", line 363, in run
    raise e
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/compiler_machinery.py", line 356, in run
    self._runPass(idx, pass_inst, state)
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/compiler_lock.py", line 35, in _acquire_compile_lock
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/compiler_machinery.py", line 311, in _runPass
    mutated |= check(pss.run_pass, internal_state)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/compiler_machinery.py", line 272, in check
    mangled = func(compiler_state)
              ^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/typed_passes.py", line 466, in run_pass
    lower = self.lowering_class(targetctx, library, fndesc, interp,
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/lowering.py", line 40, in __init__
    self.module = self.library.create_ir_module(self.fndesc.unique_name)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/nvmath-python/venv/lib/python3.12/site-packages/numba/core/codegen.py", line 574, in create_ir_module
    ir_module = self._codegen._create_empty_module(name)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/numba-cuda/numba_cuda/numba/cuda/codegen.py", line 399, in _create_empty_module
    ir_module.data_layout = nvvm.NVVM().data_layout
                            ^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/numba-cuda/numba_cuda/numba/cuda/cudadrv/nvvm.py", line 139, in __new__
    inst.driver = open_cudalib('nvvm')
                  ^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/numba-cuda/numba_cuda/numba/cuda/cudadrv/libs.py", line 83, in open_cudalib
    path = get_cudalib(lib)
           ^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/numba-cuda/numba_cuda/numba/cuda/cudadrv/libs.py", line 54, in get_cudalib
    return get_cuda_paths()['nvvm'].info or _dllnamepattern % 'nvvm'
           ^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/numba-cuda/numba_cuda/numba/cuda/cuda_paths.py", line 290, in get_cuda_paths
    'nvvm': _get_nvvm_path(),
            ^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/numba-cuda/numba_cuda/numba/cuda/cuda_paths.py", line 263, in _get_nvvm_path
    by, path = _get_nvvm_path_decision()
               ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch.yhavrylko_ent/Projects/nvidia/clean_test/numba-cuda/numba_cuda/numba/cuda/cuda_paths.py", line 60, in _get_nvvm_path_decision
    if os.path.exists(nvvm_ctk_dir):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen genericpath>", line 19, in exists
TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType

```

Context:
pynvjitlink is on and lto set to True

brandon-b-miller:
Hi @ZzEeKkAa , there's a couple pieces of this that are still WIP, I think you'll probably run into bugs right now. I'm working this PR over the next few days so hopefully some more updates soon.

brandon-b-miller:
@ZzEeKkAa would you be able to try again with the latest updates?

numba_cuda/numba/cuda/cuda_paths.py
brandon-b-miller: Done!

ZzEeKkAa:
> @ZzEeKkAa would you be able to try again with the latest updates?

Will do it today, thank you!

.github/workflows/pr.yaml
```diff
@@ -92,6 +93,14 @@ jobs:
...
       script: "ci/test_wheel_pynvjitlink.sh"
       matrix_filter: map(select(.ARCH == "amd64" and .CUDA_VER == "12.5.1" and .PY_VER == "3.12"))
+  test-wheels-deps-wheels:
+    needs:
+      - build-wheels
+    uses: ./.github/workflows/wheels-test.yaml
+    with:
+      build_type: pull-request
+      script: "ci/test_wheel_deps_wheels.sh"
+      matrix_filter: map(select(.ARCH == "amd64" and .CUDA_VER == "12.5.1" and .PY_VER == "3.12"))
```
bdice, line 103:
```suggestion
      # This selects "ARCH=amd64 and CUDA >=12, with the latest supported Python for each CUDA major version".
      matrix_filter: map(select(.ARCH == "amd64" and (.CUDA_VER | split(".") | .[0] | tonumber >= 12))) | group_by(.CUDA_VER|split(".")|map(tonumber)|.[0]) | map(max_by([(.PY_VER|split(".")|map(tonumber)), (.CUDA_VER|split(".")|map(tonumber))]))
```

ci/test_wheel_deps_wheels.sh
```diff
@@ -0,0 +1,55 @@
...
+mkdir -p "${RAPIDS_TESTS_DIR}"
+pushd "${RAPIDS_TESTS_DIR}"
+
+rapids-logger "Show Numba system info"
+python -m numba --sysinfo
+
+# remove cuda-nvvm-12-5 leaving libnvvm.so from nvidia-cuda-nvcc-cu12 only
+apt-get update
+apt remove --purge `dpkg --get-selections | grep cuda-nvvm | awk '{print $1}'` -y
+apt remove --purge `dpkg --get-selections | grep cuda-nvrtc | awk '{print $1}'` -y
```
bdice, line 50: Do we plan to make CUDA wheels a hard dependency for numba-cuda? (I hope so.)

Also, to help with testing this, we can create some images in `rapidsai/ci-imgs` that are like `citestwheel` but using `nvidia/cuda` `base` flavor, which will make it possible to test cases where CUDA wheels are the only source of CUDA libraries. Please feel free to ask for help with this if it would help you.

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -30,6 +33,7 @@ def _get_libdevice_path_decision():
         ("Conda environment (NVIDIA package)", get_nvidia_libdevice_ctk()),
         ("CUDA_HOME", get_cuda_home("nvvm", "libdevice")),
         ("System", get_system_ctk("nvvm", "libdevice")),
+        ("NVIDIA NVCC Wheel", get_libdevice_wheel()),
```
gmarkall, line 36: Shouldn't we have wheels ahead of system packages? Like with the Debian package, if the wheel comes after the system package, you can never get the toolkit you want if it's different from the system one by installing a wheel.

brandon-b-miller: Really good point, will update this.

ci/test_wheel_deps_wheels.sh
brandon-b-miller: If it isn't too much overhead then this would be cleaner I agree, that said, I think it's not the worst outcome to have something that just works here for now. When `cuda-python` becomes a hard dependency, it will handle all the pathfinding, so part of me thinks the long term investment in CI infra for testing these supported library configurations should also just be there.

bdice: Okay, let's defer on this until later.

numba_cuda/numba/cuda/cuda_paths.py
bdice: I think this should be ahead of `CUDA_HOME`, too, at the same level as a conda package.

brandon-b-miller: @bdice @gmarkall should the search priority be global across all the libraries? If so, https://github.com/NVIDIA/numba-cuda/pull/155/commits/4f686e941e33a2c371dd4a7310cb75a7feaff966 might be a good way forward.

ci/test_wheel_deps_wheels.sh
kkraus14: `cuda-python` itself doesn't take a hard dependency on the CUDA wheels: https://github.com/NVIDIA/cuda-python/blob/main/cuda_bindings/pyproject.toml#L38

Regardless, this should punt the problem to `cuda-python` and we should align on how we want to handle it there.

numba_cuda/numba/cuda/cuda_paths.py
kkraus14: Yes it should and we should probably align the search priority with the work @rwgk is doing in cuda-python to build a universal lib finder / path handler.

rwgk: That's the NVIDIA/cuda-python#451 epic.

FYI I took the code from this PR a few weeks ago (NVIDIA/cuda-python#447). Random status update: I added some code on top, that passes all testing for the libraries with cuda-bindings (nvJitLink, nvrtc, nvvm). I'm in the middle of locally testing loading "all" other Nvidia libraries.

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -47,20 +69,143 @@ def _get_nvvm_path_decision():
...
+    for plat, (dso_dir, dso_path) in platform_map.items():
+        if sys.platform.startswith(plat):
+            break
+    else:
+        raise NotImplementedError("Unsupported platform")
+
+    site_paths = [site.getusersitepackages()] + site.getsitepackages()
+
+    for sp in filter(None, site_paths):
+        nvvm_path = Path(sp, "nvidia", "cuda_nvcc", "nvvm", dso_dir, dso_path)
```
kkraus14, line 115: Pinging you offline regarding this

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -47,20 +69,143 @@ def _get_nvvm_path_decision():
...
+    site_paths = [site.getusersitepackages()] + site.getsitepackages()
+    for sp in site_paths:
+        lib_dir = os.path.join(
+            sp,
+            "nvidia",
+            "cuda_nvrtc",
+            ("lib" if sys.platform.startswith("linux") else "bin")
+            if sp
+            else None,
+        )
```
kkraus14, line 156: +1 pinging you offline

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -3,14 +3,34 @@
...
 _env_path_tuple = namedtuple("_env_path_tuple", ["by", "info"])
 
+SEARCH_PRIORITY = [
+    "Conda environment",
+    "Conda environment (NVIDIA package)",
+    "NVIDIA NVCC Wheel",
+    "CUDA_HOME",
+    "System",
+    "Debian package",
+]
```
kkraus14, line 24: What happens if someone has `LD_LIBRARY_PATH` set that points to something like `/usr/local/cuda/lib64`? That would end up bypassing whatever priority we have set here and make the underlying ctypes calls resolve there before resolving to any of these locations. We should probably account for that?

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -25,15 +45,17 @@ def _find_valid_path(options):
...
-    return by, libdir
+    options = _build_options(
+        [
+            ("Conda environment", get_conda_ctk()),
+            ("Conda environment (NVIDIA package)", get_nvidia_libdevice_ctk()),
+            ("CUDA_HOME", get_cuda_home("nvvm", "libdevice")),
+            ("NVIDIA NVCC Wheel", get_libdevice_wheel()),
+            ("System", get_system_ctk("nvvm", "libdevice")),
+            ("Debian package", get_debian_pkg_libdevice()),
+        ]
```
kkraus14, line 56: I know it was already done this way, but it would be nice to not have to resolve every possible path if we can short circuit.

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -47,20 +69,143 @@ def _get_nvvm_path_decision():
...
+    for sp in filter(None, site_paths):
+        nvvm_path = Path(sp, "nvidia", "cuda_nvcc", "nvvm", dso_dir, dso_path)
+        if nvvm_path.exists():
+            return str(nvvm_path.parent)
+
+    return None
+
+
+def detect_nvrtc_major_cuda_version(lib_dir):
+    # TODO - is this a bad idea?
```
kkraus14, line 123: I don't think detecting the cuda major version from a file suffix that presumably maps to the SOVERSION of that file suffix is a particularly good idea. CMake does a relatively good job of handling this and its process is documented here (though it's very nvcc centric): https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -156,7 +139,7 @@ def get_nvrtc_dso_path():
         )
         if lib_dir and os.path.exists(lib_dir):
             try:
-                major = detect_nvrtc_major_cuda_version(lib_dir)
+                major = get_major_cuda_version(lib_dir)
```
kkraus14: You can't do this because you could be running a newer CUDA driver with older library versions. I.E. I could run a CUDA 12 or CUDA 13 or CUDA 14 driver version with CUDA 11 libraries.

If we can load the cuda runtime (assuming it's a dependency already?) we could use the major version from that.

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -45,22 +82,127 @@ def _nvvm_lib_dir():
...
+                dso_dir = os.path.dirname(nvrtc_path)
+                builtins_path = os.path.join(
+                    dso_dir,
+                    [
+                        f
+                        for f in os.listdir(dso_dir)
+                        if re.match("^nvrtc-builtins.*.dll$", f)
+                    ][0],
+                )
+                assert os.path.exists(builtins_path)
```
rwgk, line 199: This will be an opaque error. Easy fix:

```
                assert os.path.exists(builtins_path), builtins_path
```
or slightly more fancy
```
                if not os.path.exists(builtins_path):
                    raise RuntimeError(f'Path does not exist: "{builtins_path}"')
```

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -45,22 +82,127 @@ def _nvvm_lib_dir():
...
+                # This absolute path will
+                # always be correct regardless of the package source
+                nvrtc_path = win32api.GetModuleFileNameW(result._handle)
+                dso_dir = os.path.dirname(nvrtc_path)
+                builtins_path = os.path.join(
+                    dso_dir,
+                    [
+                        f
+                        for f in os.listdir(dso_dir)
+                        if re.match("^nvrtc-builtins.*.dll$", f)
```
rwgk, line 196: Did you consider using `glob.glob(os.path.join(dso_dir, "nvrtc-builtins*.dll"))` here?

I'd `assert` that there is exactly one match to be sure. (On Linux there is `libnvrtc-builtins.alt.so`, maybe one day it'll pop up on Windows as well?)

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -45,22 +82,127 @@ def _nvvm_lib_dir():
...
+def get_major_cuda_version():
+    # TODO: remove once cuda-python is
+    # a hard dependency
+    from numba.cuda.cudadrv.driver import get_version
+
+    return get_version()[0]
+
+
+def get_nvrtc_dso_path():
+    site_paths = [site.getusersitepackages()] + site.getsitepackages()
```
rwgk, line 144: FWIW, in my development branch I gravitated to using `sys.path`. All site packages wind up there, and whatever people want to put there:

https://github.com/NVIDIA/cuda-python/blob/7a0c06870b6260af92f90691f28279cbd40e43eb/cuda_bindings/cuda/bindings/_path_finder/sys_path_find_sub_dirs.py

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -45,22 +82,127 @@ def _nvvm_lib_dir():
...
+    for sp in site_paths:
+        lib_dir = os.path.join(
+            sp,
+            "nvidia",
+            "cuda_nvrtc",
+            ("lib" if sys.platform.startswith("linux") else "bin")
```
rwgk, line 150: I asked ChatGPT "What's best for detecting Windows?" and it gave me this: `sys.platform == "win32"`

(with a long explanation)

(I'd assign that to a bool near the top of the file.)

kkraus14 (approved):

numba_cuda/numba/cuda/cuda_paths.py
brandon-b-miller: This PR just asserts that it exists rather than persisting it to a python global that AFAICT isn't used anywhere. That said, `numba-cuda` isn't running windows tests in CI, and there's no job that's testing this situation on windows. 

I had left it in anyways with the goal of `nvmath-python` being able to _delete_ their patch instead of having to do more additional patching for their supported platforms, but without testing it in our CI, its tough to know if it's working or not...

numba_cuda/numba/cuda/cuda_paths.py
kkraus14: Connected offline and when a resolved path contains a `/` character, it bypasses `LD_LIBRARY_PATH` in the underlying ctypes / `dlopen` calls so this isn't a problem. Resolving.

leofang:
@ZzEeKkAa would it be possible for you to do another quick test with nvmath-python before we merge this? 🙂

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -45,22 +83,126 @@ def _nvvm_lib_dir():
...
+    candidates = find_lib("nvrtc", sys_path)
+    if candidates:
+        return max(candidates)
+
+
+def _get_nvrtc_path_decision():
+    options = _build_options(
+        [
+            ("CUDA_HOME", lambda: get_cuda_home("nvrtc")),
+            ("Conda environment", get_conda_ctk),
```
gmarkall, line 106: I suspect this was missing:
```suggestion
            ("Conda environment", get_conda_ctk),
            ("Conda environment (NVIDIA package)", get_nvidia_cudalib_ctk),
```

gmarkall:
> This seems to cause Numba-CUDA to fail to find NVRTC via the conda package.

I think I found the issue: https://github.com/NVIDIA/numba-cuda/pull/155/files#r2048751189

brandon-b-miller:
> > This seems to cause Numba-CUDA to fail to find NVRTC via the conda package.
> 
> I think I found the issue: https://github.com/NVIDIA/numba-cuda/pull/155/files#r2048751189

Should we add a CI job for this scenario?

gmarkall:
The logs from the wheel-deps test (https://github.com/NVIDIA/numba-cuda/actions/runs/14470658131/job/40584109630?pr=155) show `cudart` and `cudadevrt` being found from the system instead of the wheels - I presume this is not expected?
 
```
Finding cudart from System
	Located at /usr/local/cuda/lib64/libcudart.so.12.8.90
	Trying to open library...	ok
Finding cudadevrt from System
	Located at /usr/local/cuda/lib64/libcudadevrt.a
	Checking library...	ok
```

gmarkall:
> > > This seems to cause Numba-CUDA to fail to find NVRTC via the conda package.
> > 
> > 
> > I think I found the issue: https://github.com/NVIDIA/numba-cuda/pull/155/files#r2048751189
> 
> Should we add a CI job for this scenario?

I don't think so - instead, I think we should add your change that removes the system packages (https://github.com/NVIDIA/numba-cuda/pull/155/files#diff-2b0266d1f4f802428e79f7bb1a00731e3142e2a08ecefd1125c35becdd16bd03R48-R50) to the conda test scripts instead - then the existing CI jobs will be checking this.

gmarkall:
> The logs from the wheel-deps test  show `cudart` and `cudadevrt` being found from the system instead of the wheels - I presume this is not expected?

To follow up, I also see this behaviour occurring with my local pip-installed-packages environment.

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -45,22 +83,127 @@ def _nvvm_lib_dir():
...
+def _get_nvrtc_wheel():
+    dso_path = get_nvrtc_dso_path()
+    if dso_path:
+        try:
+            result = ctypes.CDLL(dso_path, mode=ctypes.RTLD_GLOBAL)
+        except OSError:
+            pass
+        else:
+            if IS_WIN32:
+                import win32api
```
gmarkall, line 183: I think `win32api` is not part of the Python standard library, and this is the first use of it in numba-cuda. Are we guaranteed that it's available on systems in which we end up here? Do we need to add a new dependency to numba-cuda for it?

gmarkall: I guess this comes from the `pywin32` package?

brandon-b-miller: Since we're not testing windows in CI I think it's debateable if we can actually inherit this piece of the patch as well as the other windows logic. What do you think?

gmarkall:
> > The logs from the wheel-deps test  show `cudart` and `cudadevrt` being found from the system instead of the wheels - I presume this is not expected?
> 
> To follow up, I also see this behaviour occurring with my local pip-installed-packages environment.

This persists even after doing `pip install nvidia-cuda-runtime-cu12`

numba_cuda/numba/cuda/cuda_paths.py
gmarkall: I just noticed that `pywin32` is a dependency of `cuda-bindings` (which is a dependency of `cuda-python`), so I think we can reasonably assume it to be around in this configuration for the lifetime of the `cuda_paths.py` file.

gmarkall:
This does not seem to work on Windows:

```
(test-cuda-wheels) PS C:\Users\gmarkall\numbadev\numba-cuda> python -c "from numba import cuda; cuda.cudadrv.libs.test()"
        nvcuda.dll
        \windows\system32\nvcuda.dll
Using loader <class 'ctypes.WinDLL'>
        Trying to load driver...        ok
                Loaded from nvcuda.dll
Finding nvvm from <unknown>
        Located at nvvm.dll
        Trying to open library...       ERROR: failed to open nvvm:
Could not find module 'nvvm.dll' (or one of its dependencies). Try using the full path with constructor syntax.
Finding nvrtc from <unknown>
        Located at nvrtc.dll
        Trying to open library...       ERROR: failed to open nvrtc:
Could not find module 'nvrtc.dll' (or one of its dependencies). Try using the full path with constructor syntax.
Finding cudart from <unknown>
        Located at cudart.dll
        Trying to open library...       ERROR: failed to open cudart:
Could not find module 'cudart.dll' (or one of its dependencies). Try using the full path with constructor syntax.
Finding cudadevrt from <unknown>
        Located at cudadevrt.lib
        Checking library...     ERROR: failed to find cudadevrt:
cudadevrt.lib not found
Finding libdevice from <unknown>
        Located at None
  File "C:\Users\gmarkall\numbadev\numba-cuda\numba_cuda\numba\cuda\cudadrv\libs.py", line 170, in test
    check_static_lib(path)
  File "C:\Users\gmarkall\numbadev\numba-cuda\numba_cuda\numba\cuda\cudadrv\libs.py", line 69, in check_static_lib
    if not os.path.isfile(path):
           ^^^^^^^^^^^^^^^^^^^^
TypeError: _path_isfile: path should be string, bytes, os.PathLike or integer, not NoneType
        Checking library...

(test-cuda-wheels) PS C:\Users\gmarkall\numbadev\numba-cuda> echo $LASTEXITCODE
1

(test-cuda-wheels) PS C:\Users\gmarkall\numbadev\numba-cuda> pip list
Package                  Version   Editable project location
------------------------ --------- -------------------------------------
asttokens                3.0.0
colorama                 0.4.6
cuda-bindings            12.8.0
cuda-python              12.8.0
decorator                5.2.1
exceptiongroup           1.2.2
executing                2.1.0
ipython                  9.1.0
ipython_pygments_lexers  1.1.1
jedi                     0.19.2
llvmlite                 0.44.0
matplotlib-inline        0.1.7
numba                    0.61.2
numba-cuda               0.0.18    C:\Users\gmarkall\numbadev\numba-cuda
numpy                    2.0.2
nvidia-cuda-nvcc-cu12    12.8.93
nvidia-cuda-nvrtc-cu12   12.8.93
nvidia-cuda-runtime-cu12 12.8.90
nvidia-curand-cu12       10.3.9.90
parso                    0.8.4
pickleshare              0.7.5
pip                      25.0.1
prompt_toolkit           3.0.51
pure_eval                0.2.3
Pygments                 2.19.1
pywin32                  310
setuptools               78.1.0
stack_data               0.6.3
traitlets                5.14.3
typing_extensions        4.13.2
wcwidth                  0.2.13
wheel                    0.45.1
```

Are we aiming for Linux support here (for  nvmath-python) or just Linux support (for RAPIDS)?

gmarkall:
(if you saw my previous post, ignore it - I deleted it - I was testing with the wrong branch!)

gmarkall:
Now checking with the correct branch, with the pip wheels on Windows, I get:

```
(test-cuda-wheels) PS C:\Users\gmarkall\numbadev\numba-cuda> python -c "from numba import cuda; cuda.cudadrv.libs.test()"
        nvcuda.dll
        \windows\system32\nvcuda.dll
Using loader <class 'ctypes.WinDLL'>
        Trying to load driver...        ok
                Loaded from nvcuda.dll
  File "C:\Users\gmarkall\numbadev\numba-cuda\numba_cuda\numba\cuda\cudadrv\libs.py", line 159, in test
    path = get_cudalib(lib)
           ^^^^^^^^^^^^^^^^
  File "C:\Users\gmarkall\numbadev\numba-cuda\numba_cuda\numba\cuda\cudadrv\libs.py", line 54, in get_cudalib
    return get_cuda_paths()[lib].info or _dllnamepattern % lib
           ^^^^^^^^^^^^^^^^

... huge traceback ...

  File "C:\Users\gmarkall\numbadev\numba-cuda\numba_cuda\numba\cuda\cudadrv\runtime.py", line 115, in get_version
    self.cudaRuntimeGetVersion(ctypes.byref(rtver))
    ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\gmarkall\numbadev\numba-cuda\numba_cuda\numba\cuda\cudadrv\runtime.py", line 68, in __getattr__
    self._initialize()
  File "C:\Users\gmarkall\numbadev\numba-cuda\numba_cuda\numba\cuda\cudadrv\runtime.py", line 54, in _initialize
    self.lib = open_cudalib("cudart")
               ^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\gmarkall\numbadev\numba-cuda\numba_cuda\numba\cuda\cudadrv\libs.py", line 84, in open_cudalib
    return ctypes.CDLL(path)
           ^^^^^^^^^^^^^^^^^
  File "D:\miniforge\envs\test-cuda-wheels\Lib\ctypes\__init__.py", line 379, in __init__
    self._handle = _dlopen(self._name, mode)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: Could not find module 'cudart.dll' (or one of its dependencies). Try using the full path with constructor syntax.
```

Installed packages:

```
(test-cuda-wheels) PS C:\Users\gmarkall\numbadev\numba-cuda> pip list
Package                  Version   Editable project location
------------------------ --------- -------------------------------------
asttokens                3.0.0
colorama                 0.4.6
cuda-bindings            12.8.0
cuda-python              12.8.0
decorator                5.2.1
exceptiongroup           1.2.2
executing                2.1.0
ipython                  9.1.0
ipython_pygments_lexers  1.1.1
jedi                     0.19.2
llvmlite                 0.44.0
matplotlib-inline        0.1.7
numba                    0.61.2
numba-cuda               0.8.1     C:\Users\gmarkall\numbadev\numba-cuda
numpy                    2.0.2
nvidia-cuda-nvcc-cu12    12.8.93
nvidia-cuda-nvrtc-cu12   12.8.93
nvidia-cuda-runtime-cu12 12.8.90
nvidia-curand-cu12       10.3.9.90
parso                    0.8.4
pickleshare              0.7.5
pip                      25.0.1
prompt_toolkit           3.0.51
pure_eval                0.2.3
Pygments                 2.19.1
pywin32                  310
setuptools               78.1.0
stack_data               0.6.3
traitlets                5.14.3
typing_extensions        4.13.2
wcwidth                  0.2.13
wheel                    0.45.1
```

I think there's some circularity in importing the CUDA runtime library to check the major version whilst trying to determine where the libraries are.

gmarkall:
Do we need to support Windows here (for nvmath-python) or just Linux (e.g. for RAPIDS etc.)?

brandon-b-miller:
> Do we need to support Windows here (for nvmath-python) or just Linux (e.g. for RAPIDS etc.)?

This is the crux of a couple of conversations on this PR. Ultimately we don't have comprehensive testing of windows in CI. The existing code in `numba-cuda` that's conditional on the platform being windows I assume is inherited from upstream `numba` where my understanding is that windows is more thoroughly tested.

I vendored a lot of the windows logic for completeness, but I'm somewhat -1 on keeping it part of the patch, since without testing its a bit of a stab in the dark as to if it works or not (as you are finding via manual testing). 

If we choose _not_ to vendor the windows logic, I'd expect `nvmath-python` will still need to patch numba to maintain existing support, but, the rapids usecase would still be fixed.

brandon-b-miller:
> The logs from the wheel-deps test (https://github.com/NVIDIA/numba-cuda/actions/runs/14470658131/job/40584109630?pr=155) show `cudart` and `cudadevrt` being found from the system instead of the wheels - I presume this is not expected?
> 
> ```
> Finding cudart from System
> 	Located at /usr/local/cuda/lib64/libcudart.so.12.8.90
> 	Trying to open library...	ok
> Finding cudadevrt from System
> 	Located at /usr/local/cuda/lib64/libcudadevrt.a
> 	Checking library...	ok
> ```

https://github.com/NVIDIA/numba-cuda/pull/155/commits/c14644a4f0eff373fa3bb7c122a1a491efdf02d1

rwgk:
@brandon-b-miller Suggested change to PR title:
```diff
-Locate nvvm, libdevice and nvrtc from nvidia-cuda-nvcc-cu12 wheels
+Locate nvvm, libdevice and nvrtc from nvidia-*-cu12 wheels
```

Because:

* nvvm, libdevice → nvidia-cuda-nvcc-cu12
* nvrtc → nvidia-cuda-nvrtc-cu12

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -3,14 +3,50 @@
...
-
+import site
+from pathlib import Path
 from numba.core.config import IS_WIN32
-from numba.misc.findlib import find_lib, find_file
+from numba.misc.findlib import find_lib
 from numba import config
+import ctypes
 
+IS_LINUX = sys.platform == "linux"
```
rwgk, line 13: For all practical purposes, this is equivalent to `not IS_WIN32`

To make this code easier to read, I think it would be better to consistently use `IS_WIN32` (and not introduce this extra variable).

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -3,14 +3,50 @@
...
+    "System",
+    "Debian package",
+]
+
+
+def _priority_index(label):
+    try:
+        return SEARCH_PRIORITY.index(label)
+    except ValueError:
+        return float("inf")
```
rwgk, line 31: This can easily mask bugs that are troublesome to track down (e.g. surprising behavior if there is a typo in the `label`, or a new `label` is introduced elsewhere without updating `SEARCH_PRIORITY`). — I realize this PR is meant to be a stop-gap. Just pointing out. Generally, I'd try to not use such a brittle approach. If there is an easy way to avoid this, that'd be better.

E.g., what happens if you simply remove the `try`-`except`? Do all tests pass?

gmarkall:
The CI runs are still finding the runtime from the system installation: https://github.com/NVIDIA/numba-cuda/actions/runs/14518427298/job/40733627061?pr=155

```
Finding cudart from System
	Located at /usr/local/cuda/lib64/libcudart.so.12.8.90
	Trying to open library...	ok
Finding cudadevrt from System
	Located at /usr/local/cuda/lib64/libcudadevrt.a
	Checking library...	ok
```

The nvidia-cuda-runtime-cu12 wheel also needs installing in the test script, I think.

numba_cuda/numba/cuda/cuda_paths.py
brandon-b-miller: This is a great observation, what do you think of the changes in https://github.com/NVIDIA/numba-cuda/pull/155/commits/8cc37d72d3debef140dc2525ec031aa1e0a7c7d2 ?

numba_cuda/numba/cuda/cuda_paths.py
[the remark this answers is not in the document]
brandon-b-miller: patch added in https://github.com/NVIDIA/numba-cuda/pull/155/commits/d5b68a9b955437c99e0395e88e0b559238574729

brandon-b-miller:
> The CI runs are still finding the runtime from the system installation: https://github.com/NVIDIA/numba-cuda/actions/runs/14518427298/job/40733627061?pr=155
> 
> ```
> Finding cudart from System
> 	Located at /usr/local/cuda/lib64/libcudart.so.12.8.90
> 	Trying to open library...	ok
> Finding cudadevrt from System
> 	Located at /usr/local/cuda/lib64/libcudadevrt.a
> 	Checking library...	ok
> ```
> 
> The nvidia-cuda-runtime-cu12 wheel also needs installing in the test script, I think.

This has been updated, now I see
```
Finding cudart from NVIDIA NVCC Wheel
	Located at /pyenv/versions/3.13.3/lib/python3.13/site-packages/nvidia/cuda_runtime/lib/libcudart.so.12
```

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -45,22 +81,127 @@ def _nvvm_lib_dir():
...
+                    cu_ver = "11.2" if not IS_WIN32 else "112"
+                elif major == 12:
+                    cu_ver = "12" if not IS_WIN32 else "120"
+                else:
+                    raise NotImplementedError(f"CUDA {major} is not supported")
+
+                return os.path.join(
+                    lib_dir,
+                    f"libnvrtc.so.{cu_ver}"
+                    if not IS_WIN32
```
rwgk, line 165: To make this neat, I'd flip these around, e.g. `cu_ver = "112" if IS_WIN32 else "11.2"`

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -78,26 +219,46 @@ def _cuda_home_static_cudalib_path():
         return ("lib64",)
 
 
+def _get_cudalib_wheel():
+    """Get the cudalib path from the NVCC wheel."""
+    site_paths = [site.getusersitepackages()] + site.getsitepackages()
+    libdir = not IS_WIN32 and "lib" or "bin"
```
rwgk, line 225: This one too (`libdir = "bin" if IS_WIN32 else "lib"`)

(What you have right now is a real brain teaser!)

rwgk (approved):
Just a couple nits.

gmarkall (approved):
As of the latest commit, [dfe25c8](https://github.com/NVIDIA/numba-cuda/pull/155/commits/dfe25c81d9b8d2ff4d4581a3518f26192e789af3), this is still working for me for both Conda packages and pip wheels on Windows and Linux.

I think the only caveat is that cudadevrt.lib and the headers are not pip-installable, but that's not an issue we can solve inside this PR and that shouldn't block it.

numba_cuda/numba/cuda/cuda_paths.py
leofang: I think the logic here is not for `cuver`alone, but for the OS-dependent DSO name.

numba_cuda/numba/cuda/cuda_paths.py
leofang: The original patch was meant to pre-load all `nvrtc64_120_0.dll`'s dependencies. As long as all needed DLLs are loaded, at JIT time Windows is happy. (Very different behavior from Linux, I know.)

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -3,14 +3,48 @@
...
+from numba.misc.findlib import find_lib
 from numba import config
-
+import ctypes
 
 _env_path_tuple = namedtuple("_env_path_tuple", ["by", "info"])
 
+SEARCH_PRIORITY = [
+    "Conda environment",
+    "Conda environment (NVIDIA package)",
```
leofang, line 17: Leaving a note here, no action needed for this PR. 

I understand this is just a refactoring of existing code, but still having this distinction (conda-forge/nvidia) is a bit nerve wrecking, especially after CUDA 12.0 where both channels started unifying the layouts (different from `cudatoolkit` from 11.x and before) and eventually became interchangeable around 12.5. cc @jakirkham for comments.

leofang (approved):

numba_cuda/numba/cuda/cuda_paths.py
gmarkall: This distinction is here because there's a distinction in the packages though - we can't unify it here because it's not unified in the wider landscape of CUDA toolkit conda packaging.

leofang: I assume you're referring to the layout differences between CUDA 11/12?

kkraus14:
@brandon-b-miller looks like there's a couple of comments from @rwgk and then this is ready to merge?

ZzEeKkAa:
I've just run tests on `nvmath-python` by using this branch and removing 
```
    # Patch Numba to support wheels
    _utils.patch_numba_nvvm(nvvm)

    # our device apis only support cuda 12+
    _utils.force_loading_nvrtc("12")
    nvrtc.NVRTC.__new__ = __nvrtc_new__
```

And everything works good so far, thank you!

PR #156, opened 2025-03-13
Don't verify libdevice

gmarkall:
Libdevice should not be subject to the verifier's requirements. To implement this change, we make a small refactor to `CompilationUnit` so that it can be verified separately from compilation.

This also includes a couple of small changes:

- Linker usage tests are skipped if the test binaries are not present. This makes no difference on CI, but saves errors occurring for local users who run the test suite without building the linker test binaries.
- The name of the function that downgrades the LLVM IR is updated to match the currently-used LLVM version (15).

### DIFF
--- a/numba_cuda/numba/cuda/cudadrv/nvvm.py
+++ b/numba_cuda/numba/cuda/cudadrv/nvvm.py
@@ -199,11 +199,51 @@ class NVVM(object):
 
 
 class CompilationUnit(object):
-    def __init__(self):
+    """
+    A CompilationUnit is a set of LLVM modules that are compiled to PTX or
+    LTO-IR with NVVM.
+
+    Compilation options are accepted as a dict mapping option names to values,
+    with the following considerations:
+
+    - Underscores (`_`) in option names are converted to dashes (`-`), to match
+      NVVM's option name format.
+    - Options that take a value will be emitted in the form "-<name>=<value>".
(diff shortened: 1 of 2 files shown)


numba_cuda/numba/cuda/cudadrv/nvvm.py
```diff
@@ -199,12 +199,52 @@ def check_error(self, error, msg, exit=False):
...
+        def stringify_option(k, v):
+            k = k.replace('_', '-')
+
+            if v is None:
+                return f'-{k}'.encode('utf-8')
+
+            if isinstance(v, bool):
+                v = int(v)
+
+            return f'-{k}={v}'.encode('utf-8')
```
kkraus14, line 237: nitpick: no need for this to be redefined each time a `CompilationUnit` is created. Move it out of here to be a freestanding function?

kkraus14 (approved):

numba_cuda/numba/cuda/cudadrv/nvvm.py
gmarkall: Thanks - it's a good point... I'll leave it as-is, because it's no worse than it was before (it was redefined the same number of times when it was part of `compile()`, and I think we expect this area to change a bit when we move to using code.core for the NVVM binding.

Issue #159, opened 2025-03-13
[BUG] Overloaded functions with default value are not callable without all arguments


ZzEeKkAa:
**Describe the bug**
Overloaded functions with default value are not callable without all arguments

**Steps/Code to reproduce bug**
```python
from numba import njit, cuda
from numba.extending import overload
import numpy as np


def bar(x, y=None):
    return x

@overload(bar)
def ol_bar(x, y=None):
    return lambda x, y=None: 2*x

# njit works

@njit
def use_overload(a):
    return bar(a)

print(use_overload(1))

# cuda fails

@cuda.jit
def add_arrays_gpu(x, y, out):
    idx = cuda.grid(1)
    if idx < x.shape[0]:
        out[idx] = x[idx] + bar(y[idx]) # <- signature mismatch, `bar(y[idx], 1)` works

x = np.arange(10, dtype=np.int32)
y = np.arange(10, dtype=np.int32)
out = np.empty_like(x)

add_arrays_gpu[1, x.shape[0]](x, y, out)
```

**Expected behavior**
Compiles and works

**Environment details (please complete the following information):**
 - Environment location: n/a
 - Method of numba-cuda install: github.com/numba/numba , main branch

**Additional context**
Not being able to overload method that have two signatures: `(alpha, a, b, beta, c)` and `(alpha, a, lda, b, ldb, beta, c, ldc)` blocks me to register Matmul from cublasdx as a struct in [nvmath-python](https://github.com/NVIDIA/nvmath-python) without breaking backward compatibility.

Possible fix should probably consider these two bugs: https://github.com/numba/numba/issues/9979 , https://github.com/numba/numba/issues/9980

ZzEeKkAa:
After debugging a little bit, `njit` generates call like this: `(int64, omitted(default=None))` at the `numba.core.untyped_passes.FixupArgs`, while cuda does this `(int64)`

PR #160, opened 2025-03-14
Add support for default overload values

ZzEeKkAa:
Currently cuda does not support overloads with default values.

Fixes: #159
Closes: https://github.com/numba/numba/pull/9982


numba_cuda/numba/cuda/tests/cudapy/test_nested_calls.py
```diff
@@ -0,0 +1,55 @@
+"""
+Test problems in nested calls.
+Usually due to invalid type conversion between function boundaries.
```
gmarkall, line 3: I imagine this is copy/pasted from Numba, and is no longer relevant? I think it would be fine just to remove it.

numba_cuda/numba/cuda/tests/cudapy/test_nested_calls.py
gmarkall, line 1: All the mentions of generated jit and nested calls are not really pertaining to what's being tested here. I think it would be better if it were explicit that this is testing keyword arguments in the CUDA target instead (i.e. rename the test case, change names / comments that talk about nesting or generated jit so they talk about kwargs, etc)

numba_cuda/numba/cuda/tests/cudapy/test_nested_calls.py
ZzEeKkAa: Yes, I just copied it from numba and adapted for kernel run to be consistent

gmarkall (changes requested):
I think the change is basically fine and I've been testing / experimenting locally where it also looks good.

However, I think the test case is a bit misleading to the reader, probably because it's copied from some other Numba tests - it seems a bit nitpicky, but I think it would be confusing for someone trying to understand what the point of those tests is later, so it would be good to fix up the test case so that it's consistent with itself in its code, naming, and any docstrings / comments.

gmarkall (approved):

PR #161, opened 2025-03-14
Switch to real NumPy from mock NumPy

gmarkall:
Now that keyword arguments are supported for overloads, there is no need to use the kwarg-free mock NumPy functions, and the real ones can be used instead.


brandon-b-miller (approved):
🎉 thanks @gmarkall . What level of testing do you think we need before we can start to officially support these features through the cuda target?

isVoid (approved):

isVoid:
Re @brandon-b-miller , I think there are tests in Numba that test Numpy API usage in nopython mode. Like this one:
https://github.com/numba/numba/blob/d650f3057839f99a0ab77262f27e80f4b9ae6739/numba/tests/test_np_functions.py#L4
Just thinking maybe we should migrate these tests over to Numba-CUDA.

PR #165, opened 2025-03-18
Allow Adding External Linkage through Target Context

isVoid:
After a few iterations offline, we discovered that Numbast cannot pass the external information by either overriding typing or lowering behavior. We discovered that the simplest to unblock Numbast from compilation based on inferred usage is by adding them to the target context at lower time.


isVoid:
Offline discussion: we should roll this PR into https://github.com/NVIDIA/numba-cuda/pull/166 and use that PR as a test.

PR #166, opened 2025-03-19
Add Bfloat16 Low++ Bindings

isVoid:
This PR introduces low++ bfloat16 bindings to Numba-CUDA. They are exposed via `numba.cuda.cuda_bf16` module. Available intrinsic functions includes:
- htrunc
- hceil
- hfloor
- hrint
- hsqrt
- hrsqrt
- hrcp
- hlog
- hlog2
- hlog10
- hcos
- hsin
- hexp
- hexp2
- hexp10


numba_cuda/numba/cuda/cuda_bf16.py
```diff
@@ -0,0 +1,8548 @@
+
+# Automatically generated by Numbast Static Binding Generator
```
gmarkall, line 2: It would be good if Numbast could provide a bit more information that makes this more reproducible - e.g.:

- Numbast version
- Input files
- Command line used for execution
- CUDA toolkit version
- etc.

Though I realise this should probably be a Numbast feature request - what do you think?

gmarkall:
Thanks for the warning! It appears that Firefox on Linux is OK to expand `cuda_bf16.py`.

numba_cuda/numba/cuda/cuda_bf16.py
isVoid: Yes I think that's a reasonable ask though. The binding file should contain all information from which the binding itself is generated. Perhaps even including the whole of the config file!

isVoid:
I'm unable to reproduce this issue locally:
https://github.com/NVIDIA/numba-cuda/actions/runs/14092024779/job/39470965278?pr=166

numba_cuda/numba/cuda/cuda_bf16.py
```diff
@@ -0,0 +1,5469 @@
...
+# Generator Information:
+# Ast_canopy version: 0.3.0
+# Numbast version: 0.3.0
+# Generation command: /home/wangm/numbast/numbast/src/numbast/__main__.py --cfg-path configs/cuda_bf16.yml --output-dir numba_cuda/numba/cuda/
+# Static binding generator parameters: {'cfg_path': 'configs/cuda_bf16.yml', 'output_dir': 'numba_cuda/numba/cuda/', 'input_header': None, 'retain': None, 'types': None, 'datamodels': None, 'compute_capability': None, 'run_ruff_format': True}
+# Config file path (relative to the path of the generated binding): ../../../../configs/cuda_bf16.yml
+# Cudatoolkit version: (12, 8)
+# Default CUDA_HOME path: /home/wangm/micromamba/envs/numbast
```
isVoid, line 10: These are the generated bindings @gmarkall

numba_cuda/numba/cuda/cuda_bf16.py
isVoid: Addressed in https://github.com/NVIDIA/numba-cuda/pull/166/commits/079f8978fe48d71c81078339eb0feda1b15d248b

kkraus14:
Created issue #198 to help with tracking. Unclear if this PR is intended to be the totality of the support we can add for bf16 data types or if we'll have follow up work to get them supported as much as possible.

isVoid:
> Created issue #198 to help with tracking. Unclear if this PR is intended to be the totality of the support we can add for bf16 data types or if we'll have follow up work to get them supported as much as possible.

Let's focus on just the low++ bindings for bfloat16 in this PR. The immediate follow up is to add support to higher level API support, such as `math.ceil(bfloat16)` maps to `cuda.cuda_bf16.hceil`.

This translates to use automation to generate a large chunk of wrapper code such as

```python
@overload(math.ceil)
def math_ceil(a, b):
    if isinstance(a, bfloat16) and isinstance(b, bfloat16):
        def impl(a, b):
            return cuda.cuda.bf16.ceil(a, b)
```
But that's a sufficiently separable body of work from this PR.

numba_cuda/numba/cuda/cudadrv/linkable_code.py
```diff
@@ -10,13 +11,19 @@ class LinkableCode:
...
         self._name = name
+        self._data = data
 
     @property
     def name(self):
         return self._name or self.default_name
 
+    @property
+    def data(self):
+        if isinstance(self._data, io.StringIO):
```
gmarkall, line 23: This should be documented so that users are aware they can use a StringIO instance for data to generate the implementation on-the-fly

numba_cuda/numba/cuda/target.py
```diff
@@ -73,6 +73,7 @@ def __init__(self, typingctx, target='cuda'):
         self.data_model_manager = cuda_data_manager.chain(
             datamodel.default_manager
         )
+        self._external_linkage = set()
```
gmarkall, line 76: AI for Graham: look into if there's a way to eliminate this

numba_cuda/numba/cuda/tests/cudapy/test_bfloat16_bindings.py
gmarkall, line 1: It would be good to have a test of using these for shared and local memory (with an xfail if it does not work)

gmarkall: We should maybe (optionally) try it with NRT enabled and creating bfloat16 arrays with `np.zeros(shape, dtype=nv_bfloat16)` in kernels

numba_cuda/numba/cuda/cuda_bf16.py
gmarkall, line 1: I tried replacing the `as_numba_type` with `typeof_impl` with

```diff
diff --git a/numba_cuda/numba/cuda/cuda_bf16.py b/numba_cuda/numba/cuda/cuda_bf16.py
index f9122ac..4863fe2 100644
--- a/numba_cuda/numba/cuda/cuda_bf16.py
+++ b/numba_cuda/numba/cuda/cuda_bf16.py
@@ -31,7 +31,7 @@ from numba.core.extending import register_model
 from numba.types import int16
 import numba
 from numba.core.extending import lower_cast
-from numba.extending import as_numba_type
+from numba.extending import typeof_impl
 from numba.types import Number
 from llvmlite import ir
 from numba.cuda.cudadecl import register_global
@@ -80,7 +80,10 @@ _type_unnamed1401637 = _type_class_unnamed1401637()
 # Make Python API for struct
 unnamed1401637 = type("unnamed1401637", (), {"_nbtype": _type_unnamed1401637})
 
-as_numba_type.register(unnamed1401637, _type_unnamed1401637)
+
+@typeof_impl.register(unnamed1401637)
+def typeof_unnamed1401637(val, c):
+    return _type_unnamed1401637
 
 
 @register_model(_type_class_unnamed1401637)
@@ -128,7 +131,10 @@ _type_unnamed1401746 = _type_class_unnamed1401746()
 # Make Python API for struct
 unnamed1401746 = type("unnamed1401746", (), {"_nbtype": _type_unnamed1401746})
 
-as_numba_type.register(unnamed1401746, _type_unnamed1401746)
+
+@typeof_impl.register(unnamed1401746)
+def typeof_unnamed1401746(val, c):
+    return _type_unnamed1401746
 
 
 @register_model(_type_class_unnamed1401746)
@@ -183,7 +189,10 @@ _type___nv_bfloat16 = _type_class___nv_bfloat16()
 # Make Python API for struct
 __nv_bfloat16 = type("__nv_bfloat16", (), {"_nbtype": _type___nv_bfloat16})
 
-as_numba_type.register(__nv_bfloat16, _type___nv_bfloat16)
+@typeof_impl.register(__nv_bfloat16)
+def typeof___nv_bfloat16(val, c):
+    breakpoint()
+    return _type___nv_bfloat16
 
 
 @register_model(_type_class___nv_bfloat16)
@@ -1543,7 +1552,10 @@ _type___nv_bfloat162 = _type_class___nv_bfloat162()
 # Make Python API for struct
 __nv_bfloat162 = type("__nv_bfloat162", (), {"_nbtype": _type___nv_bfloat162})
 
-as_numba_type.register(__nv_bfloat162, _type___nv_bfloat162)
+
+@typeof_impl.register(__nv_bfloat162)
+def typeof___nv_bfloat162(val, c):
+    return _type___nv_bfloat162
```

and it does not provide the same functionality required for instance checks. `as_numba_type` happens to work, but I think that is a lucky side effect (that I think should be formalised in Numba) rather than an intended use case.

The `typeof_impl` support will eventually be needed if we want to support passing in bfloat16 objects we create in Python, or treating them as global constants.

isVoid: > support passing in bfloat16 objects we create in Python, or treating them as global constants.

Would that mean something like:
```python
@njit
def kernel(dtype):
    x = dtype(42)

kernel[1, 1](bfloat16)
```

numba_cuda/numba/cuda/cuda_bf16.py
```diff
@@ -0,0 +1,5070 @@
...
+from numba.types import Function
+
+
+# Prefixes:
+
+numba.config.CUDA_ENABLE_PYNVJITLINK = True
+
+# Shim Stream:
+
+entry_point = "#include </usr/local/cuda/include/cuda_bf16.h>"
```
isVoid, line 57: We discussed offline that it would be more reliable if we vend the cuda_bf16 header in Numba and generate binding for it. This means:
1. If there's a new CTK major release and Numba supports it, the generated binding whose shims contains implementation from older CTK release is still supported. 
2. When we want to also upgrade the bindings, we would need to re-vend the header and re-generate the bindings as well.

numba_cuda/numba/cuda/cudadrv/linkable_code.py
isVoid: Done in 0c85778

numba_cuda/numba/cuda/cuda_bf16.py
```diff
@@ -0,0 +1,5078 @@
...
+
+
+# Prefixes:
+
+import importlib
+
+if importlib.util.find_spec("pynvjitlink"):
+    numba.config.CUDA_ENABLE_PYNVJITLINK = True
+else:
+    raise RuntimeError("Numbast generated bindings require pynvjitlink.")
```
isVoid, line 59: The bindings should work without pynvjitlink installed using the ctypes linker. Let's remove this part and make it a user duty to enable pynvjitlink if it is needed.

numba_cuda/numba/cuda/cudadrv/linkable_code.py
```diff
@@ -1,10 +1,13 @@
+import io
 from .mappings import FILE_EXTENSION_MAP
 
 
 class LinkableCode:
     """An object that holds code to be linked from memory.
 
-    :param data: A buffer containing the data to link.
+    :param data: A buffer, StringIO or BytesIO containing the data to link.
```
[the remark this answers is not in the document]
isVoid: > In this PR the _KeyedStringIO class seems to handle this, only adding code for each lowering to the shim object once.

Yes - it subclasses the `StringIO` class, providing additional methods that only writes once to the stream for a given key.

If you foresee this class can be useful, can we upstream this class to Numba-CUDA, other Numbast bindings could make use of this too.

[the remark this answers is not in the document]
isVoid: The only thing I'm not sure about using a string stream for the shim function is that I'm not sure when to reset the stream so that a separate compile pass can regenerate the shim functions from a fresh stream.

But I guess that's not any worse than our previous effort - pre-populating all of shim functions in a massive block and compile that.

numba_cuda/numba/cuda/cuda_bf16.py
gmarkall: It's not passing the type in itself, it's instances of the type I was referring to, either as arguments:

```python
@cuda.jit
def kernel(x):
    ...

x = bfloat16(3.14)
kernel[1, 1](x)
```

or as a constant defined outside the kernel, but referenced inside:

```python
x = bfloat16(3.14)

@cuda.jit
def kernel():
    print(x) # Assuming we had a way to print a bfloat16

kernel[1, 1]()
```

numba_cuda/numba/cuda/cudadrv/linkable_code.py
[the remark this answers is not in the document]
gmarkall: I think it's OK to defer figuring out resetting the stream for now. (feel free to resolve this thread if you're happy)

numba_cuda/numba/cuda/cuda_bf16.py
gmarkall: I was able to reproduce this file by installing Numbast from https://github.com/NVIDIA/numbast/pull/106. I had to make a couple of changes to the environment installation suggested in the README:

- Using clangdev 18, not >=18 - this lead to clangdev 20 being picked, with which Numbast wouldn't build
- I installed CUDA 12.8 instead of 12.6 suggested by the environment file, because the generated binding suggests 12.8 was used - I'm not sure this makes a difference given that the required headers are vendored here anyway.

There was a trivial difference in the generated file, probably because something is a tiny bit out of sync:

```diff
@@ -69,13 +69,15 @@ class _KeyedStringIO(io.StringIO):
         self.seek(0)
 
 
+shim_defines = ""
 shim_include = (
     "#include <"
     + os.path.join(os.path.abspath(os.path.dirname(__file__)), "cuda_bf16.h")
     + ">"
 )
+shim_prefix = shim_defines + "\n" + shim_include
 shim_stream = _KeyedStringIO()
-shim_stream.write(shim_include)
+shim_stream.write(shim_prefix)
 shim_obj = CUSource(shim_stream)
 
 # Enums:
```

but I think this shows that generation is suitably reproducible.

isVoid: > Using clangdev 18, not >=18 - this lead to clangdev 20 being picked, with which Numbast wouldn't build

Numbast will pin to 18 with version 0.3.0 . Just waiting to publish a new package as we merge in the feature branch.

isVoid: > There was a trivial difference in the generated file

Correct - for reference, this change is not a features requested by the bfloat16.

numba_cuda/numba/cuda/cuda_bf16.py
isVoid: > until we (plan to) extend Numbast to allow constructing these types outside of kernels for use within as parameters or global constants

I think that will be an interesting use case. 

I created a task to track: https://github.com/NVIDIA/numbast/issues/119

numba_cuda/numba/cuda/tests/cudapy/test_bfloat16_bindings.py
```diff
@@ -0,0 +1,234 @@
+from numba import config
+import numba.cuda as cuda
+from numba.cuda.testing import unittest, CUDATestCase
+from numba.cuda.cuda_bf16 import (
+    nv_bfloat16,
```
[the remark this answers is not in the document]
isVoid: I'll add a feature in Numbast to allow creating custom aliases to existing bindings.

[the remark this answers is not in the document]
gmarkall: Just to check (maybe for discussion when we meet later) is that something for the future? I'm not sure I am clear about whether it makes more sense to keep the name from the header for the low++ bindings, and then add some manual implementation on top for the "Pythonic" API for this that maps `bfloat16` to `nv_bfloat16`.

numba_cuda/numba/cuda/tests/cudapy/test_bfloat16_bindings.py
```diff
@@ -0,0 +1,250 @@
...
+        hrcp,
+        hlog,
+        hlog2,
+        hlog10,
+        hcos,
+        hsin,
+        hexp,
+        hexp2,
+        hexp10,
+    )
```
gmarkall, line 27: We can run tests without pynvjitlink:

```suggestion
if config.CUDA_ENABLE_PYNVJITLINK:
    from numba.cuda.cuda_bf16 import (
        nv_bfloat16,
        htrunc,
        hceil,
        hfloor,
        hrint,
        hsqrt,
        hrsqrt,
        hrcp,
        hlog,
        hlog2,
        hlog10,
        hcos,
        hsin,
        hexp,
        hexp2,
        hexp10,
    )
```

numba_cuda/numba/cuda/tests/cudapy/test_bfloat16_bindings.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+dtypes = [int16, int32, int64, uint16, uint32, uint64, float32]
+
+
+@unittest.skipIf(
+    (cuda.get_current_device().compute_capability < (8, 0))
+    or (not config.CUDA_ENABLE_PYNVJITLINK),
+    "bfloat16 require compute capability 8.0+ or pynvjitlink is not enabled",
```
gmarkall, line 36:
```suggestion
    (cuda.get_current_device().compute_capability < (8, 0)),
    "bfloat16 requires compute capability 8.0+",

```

numba_cuda/numba/cuda/tests/cudapy/test_bfloat16_bindings.py
gmarkall: I tried this and it turns out not to work; I think we should worry about that in a future PR. I used these tests to check (feel free to incorporate them):

```python
    @unittest.expectedFailure
    def test_shared_array(self):
        # bfloat16 is presently non-functional for shared arrays
        def use_shared():
            cuda.shared.array(1, nv_bfloat16)

        sig = tuple()
        cuda.compile_ptx(use_shared, sig)

    @unittest.expectedFailure
    def test_local_array(self):
        # bfloat16 is presently non-functional for local arrays
        def use_local():
            cuda.local.array(1, nv_bfloat16)

        sig = tuple()
        cuda.compile_ptx(use_local, sig)

    @unittest.expectedFailure
    def test_allocate_array(self):
        def allocate_array():
            np.empty(1, dtype=nv_bfloat16)

        with override_config("CUDA_ENABLE_NRT", True):
            sig = tuple()
            cuda.compile_ptx(allocate_array, sig)

    @unittest.expectedFailure
    def test_shared_array_numba_type(self):
        # bfloat16 is presently non-functional for shared arrays, even using
        # the Numba type rather than the constructor function
        def use_shared_with_nbtype():
            cuda.shared.array(1, _type___nv_bfloat16)

        sig = tuple()
        cuda.compile_ptx(use_shared_with_nbtype, sig)
```

(the last one also required `from numba.cuda.bf16 import _type___nv_bfloat16`)

numba_cuda/numba/cuda/cuda_bf16.py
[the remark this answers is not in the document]
gmarkall: For the scope of this PR, I've changed my mind about this - we should not write about the existence of this in the docs until we have the implementation / "user interface" we would hope to present to the user (feel free to resolve this conversation if you're happy with this).

isVoid:
The [previous failure](https://github.com/NVIDIA/numba-cuda/actions/runs/14759424763/job/41436311657) results from that Numbast dynamically resolves the name of the datamodel module in Numba. The latest Numba version changed the module name from `numba.core.datamodel.models` to `numba.core.datamodel.old_models`. Instead, I consistently resolve from `numba.core.datamodel` to get what Numba exposes by default.

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -372,19 +372,25 @@ def compile(src, name, cc, ltoir=False):
...
 
+    nvrtc_version = nvrtc.get_version()
+    nvrtc_ver_major = nvrtc_version[0]
+
     cudadrv_path = os.path.dirname(os.path.abspath(__file__))
     numba_cuda_path = os.path.dirname(cudadrv_path)
-    numba_include = f"-I{numba_cuda_path}"
+
+    numba_include = (
+        f"-I{os.path.join(numba_cuda_path, 'include', str(nvrtc_ver_major))}"
```
gmarkall, line 382: This will reference a non-existent folder as soon as there's a CUDA 13 release - can we just select 11 if the NVRTC version is 11 and 12 otherwise?

PR #167, opened 2025-03-19
Support consuming NRT functions from external libraries

brandon-b-miller:
This PR moves some NRT function declarations to a header that can be found at build time by external libraries. It also fixes a bug where NRT wasn't linked if its declarations were present in the passed user source, but not the half of the PTX numba generates itself.


numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -234,21 +236,18 @@ def link_to_library_functions(library_functions, library_path,
         self.reload_init = []
 
     def maybe_link_nrt(self, link, tgt_ctx, asm):
-        if not tgt_ctx.enable_nrt:
+        if not tgt_ctx.enable_nrt or self.nrt is False:
```
gmarkall, line 239: Do you want the `nrt` kwarg for the JIT decorator to take precedence over the global NRT setting for enabling it? I think this will never link the NRT if it is globally disabled.

gmarkall:
Would this be better broken into two PRs? One for making the NRT functions available externally, and one for the kwarg?

gmarkall:
Also, if we're going to make `nrt.cuh` available for external users, I think we need a documentation section explaining it and outlining what we consider the public API and expected behaviour to be.

numba_cuda/numba/cuda/dispatcher.py
brandon-b-miller: Good observation. I think the best behavior might be 
- kwarg takes precedence, allowing users definitive control over if the NRT code should be added to the link for a particular kernel
- global flag enables the detection mechanism

But I'm not sure if numba's code generation will actually provide increfs/decrefs without `enable_nrt` set to true in the target context, is that correct? In that case we might be adding the refcounting functions to the link but not actually calling them. 

The motivation for this kwarg is twofold: Where it came up was an edge case I observed in cuDF where an NRT managed object is created and immediately returned, such as in `lambda string: return string`. In this case it works out that NRT functions are used, but they only appear in the separately provided PTX file. The ASM that numba inspects when deciding to link NRT or not contains only the declarations for the parent functions that eventually call the NRT ones, so this defeats numba's NRT detection.

Overall I feel that unless the NRT detection is absolutely definitive somehow it might be more reliable to let the user decide when to link NRT. This has the drawback of creating two switches, one to enable the *possiblity* of NRT, and then a second one to actually add it to the link. But even that I think its a little better than somewhat shaky auto detection.

brandon-b-miller: > re you saying that numba-cuda doesn't link the NRT when it should? If so, then I would have seen that as a bug in numba-cuda?

Yes, this is the case. It is the PTX file that is provided by cuDF that is not checked. The PTX from the numba side is checked, but doesn't contain the actual NRT calls - it only contains a call to a shim function that uses NRT functions within its implementation.

Initially I felt that since cuDF was somewhat burying the NRT calls maybe it should also be the one to signal to numba that the NRT implementations are also needed (or the same for any consuming library that does the same thing). 

But I also see that it could be considered a bug in numba-cuda that _all_ the input PTX files are not checked for NRT functions. What do you think?

gmarkall: I think if we support using the NRT functions from external code, then ideally we should be checking for calls to NRT functions in external code rather than asking the user to do additional work - particularly if it requires signalling in the `@cuda.jit` decorator, because the person writing the jitted function can often be someone different to the person who implemented another library.

We've seen this problem with the `link` and `extensions` kwargs already, where libraries want to use these features, but then have to ask the users to remember to use them with no explicit connection between use of the library and the requirement to add the kwargs.

brandon-b-miller: What do you think of the changes in https://github.com/NVIDIA/numba-cuda/pull/167/commits/0b23a097fb391424ea830975ab6bbc06e6721d6e and https://github.com/NVIDIA/numba-cuda/pull/167/commits/dd568e74ded42adf99aed5cdc7f809872d895ce6 ?

brandon-b-miller:
> Also, if we're going to make `nrt.cuh` available for external users, I think we need a documentation section explaining it and outlining what we consider the public API and expected behaviour to be.

I agree. I think that given that the numpy API is available now we're getting to the point where a doc update is needed to describe the available NRT features. What would you say to giving that section of the docs a broad overhaul, either as part of this PR or separately, including what `nrt.cuh` is for and how to use it?

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -237,16 +236,18 @@ def maybe_link_nrt(self, link, tgt_ctx, asm):
...
-
-        nrt_in_asm = re.findall(pattern, asm)
-
-        basedir = os.path.dirname(os.path.abspath(__file__))
-        if nrt_in_asm:
+        nrt_in_asm = lambda asm: any(fn in asm for fn in self.NRT_functions)
+        link_nrt = nrt_in_asm(asm)
+        if not link_nrt:
+            for file in link:
+                if file.endswith('ptx'):
```
isVoid, line 243: Alternatively, there can also be `linkable_code.PTXSource` in the input. Could there be input in other format that also contains NRT calls?

numba_cuda/numba/cuda/tests/nrt/test_nrt.py
```diff
@@ -4,11 +4,52 @@
...
+    sig_ = types.int32()
+    # call the external function, passing the pointer
+    result = context.compile_internal(
+        builder,
+        call_allocate_shim,
+        sig_,
+        (),
+    )
+
+    # return len(&input)
```
isVoid, line 51:
```suggestion
```

numba_cuda/numba/cuda/tests/nrt/test_nrt.py
```diff
@@ -76,6 +117,31 @@ def g(out_ary):
 
         self.assertEqual(out_ary[0], 1)
 
+    def test_nrt_detect_linked_ptx_file(self):
+        src = """;
+                 extern "C" __device__ void* NRT_Allocate(size_t size);
```
isVoid, line 122: Would this be an opportunity to check if `get_include` work as intended?

```suggestion
                 #include <{get_include()}>
```

isVoid: The general idea is that user can depend on Numba's C++ declaration of the NRT APIs, but not to depend on their own, right?

numba_cuda/numba/cuda/dispatcher.py
brandon-b-miller: Thats a reasonable question. This would only currently catch the case where the user supplied a PTX file that utilized NRT functions. I suppose it could also occur with a `.cu` file, which it shouldn't be hard to update things for.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -103,6 +103,12 @@ def make_logger():
     return logger
 
 
+@functools.lru_cache(maxsize=None)
+def read_file(filepath, how='r'):
```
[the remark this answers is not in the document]
brandon-b-miller: Done in https://github.com/NVIDIA/numba-cuda/pull/167/commits/5bd9b7a4dd8ef01d600b4c00d43a3acd55d4869a

numba_cuda/numba/cuda/dispatcher.py
brandon-b-miller: https://github.com/NVIDIA/numba-cuda/pull/167/commits/ff478f462ebf186f59a0104db9762ad396cc276f

numba_cuda/numba/cuda/dispatcher.py
isVoid: Apologies for being late to the discussion.

> particularly if it requires signalling in the @cuda.jit decorator, because the person writing the jitted function can often be someone different to the person who implemented another library.

If it's the library dev duty to specify `nrt` arg, how about adding it to the `LinkableCode` interface? The downside is that this cannot be specified if user would want to directly use a path to the `link` argument. Perhaps we can make that an unsupported case or simply deprecate such use case.

What let me to think of this is that https://github.com/NVIDIA/numba-cuda/commit/0b23a097fb391424ea830975ab6bbc06e6721d6e NRT discovery mechanism may not be able to handle binary format input such as LTOIR (disassembly could work but I think that's a too much price to pay).

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -264,17 +265,24 @@ def maybe_link_nrt(self, link, tgt_ctx, asm):
...
-        nrt_in_asm = re.findall(pattern, asm)
-
-        basedir = os.path.dirname(os.path.abspath(__file__))
-        if nrt_in_asm:
+        nrt_in_asm = lambda asm: any(fn in asm for fn in self.NRT_functions)
+        link_nrt = nrt_in_asm(asm)
+        if not link_nrt:
+            for file in link:
+                if isinstance(file, LinkableCode):
+                    asm = file.data.decode("utf-8")
```
isVoid, line 273: `LinkableCode` can also be binary formats like archives, fatbins or LTOIR. Only text format is applicable here, which includes `CUSource` and `PTXSource`

```suggestion
                if isinstance(file,(CUSource, PTXSource)):
                    asm = file.data.decode("utf-8")
```

When it comes to other kind, we would have to silently not parse them I suppose? Disassembling them also works but would be too high a price to pay IMO.

numba_cuda/numba/cuda/dispatcher.py
gmarkall: > If it's the library dev duty to specify `nrt` arg, how about adding it to the `LinkableCode` interface?

I think this sounds like a good idea.

> The downside is that this cannot be specified if user would want to directly use a path to the `link` argument.

I think that's not too significant a downside. I see the provision of a path to link as a bit of a legacy anyway, because it's not very convenient - it's just what happened to be implemented first, a long time ago, perhaps before the API design had a lot of consideration.

> Perhaps we can make that an unsupported case or simply deprecate passing in a raw path (and reintroduce it as a `Path` subclass to `LinkableCode`).

I think introducing a `Path` subclass might not fit exactly right, because the linker needs to know what type of object is being passed in. A helper function that determines the type based on the extension, reads it in, and then returns the appropriate `LinkableCode` subclass instance would probably be OK though. (Perhaps a class method `LinkableCode.from_path()`, maybe?).

isVoid: >  (Perhaps a class method LinkableCode.from_path(), maybe?).

I think that's a good design.

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -265,17 +266,28 @@ def maybe_link_nrt(self, link, tgt_ctx, asm):
...
+        nrt_in_asm = lambda asm: any(fn in asm for fn in self.NRT_functions)
+        link_nrt = nrt_in_asm(asm)
+        if not link_nrt:
+            for file in link:
+                if isinstance(file, LinkableCode):
+                    if file.nrt:
+                        link_nrt = True
+                        break
+                    elif isinstance(file, (CUSource, PTXSource)):
+                        asm = file.data.decode("utf-8")
```
isVoid, line 278: Note that user can still specify non-binary input which can lead to `asm` object undefined here.

```suggestion
                if isinstance(file, LinkableCode):
                    if file.nrt:
                        link_nrt = True
                        break
                    elif isinstance(file, (CUSource, PTXSource)):
                        asm = file.data.decode("utf-8")
                    else:
                        asm = ""
```

numba_cuda/numba/cuda/runtime/nrt.py
```diff
@@ -32,6 +32,11 @@
     config.CUDA_ENABLE_NRT = ENABLE_NRT
 
 
+def get_include():
```
isVoid, line 35: Just a nit: I think a name like `get_nrt_include_path` is more specific to what this does.

numba_cuda/numba/cuda/tests/nrt/test_nrt.py
```diff
@@ -76,6 +149,76 @@ def g(out_ary):
 
         self.assertEqual(out_ary[0], 1)
 
+    def test_nrt_detect_linked_ptx_file(self):
+        src = f"#include <{get_include()}/nrt.cuh>"
+        src += """;
+                 extern "C" __device__ int extern_func(int* nb_retval){
+                     NRT_Allocate(1);
```
isVoid, line 156: Hmm, this test case leaks memory right? Even though it's 1 byte it may not pass the memory leak tests.

Makes me feel that we should also make a binding for `NRT_Deallocate` and use that following `allocate_shim`.

numba_cuda/numba/cuda/runtime/nrt.py
brandon-b-miller: This is made to mirror similar APIs like [numpy get_include()](https://numpy.org/devdocs/reference/generated/numpy.get_include.html) or [pyarrow get_include()](https://arrow.apache.org/docs/python/generated/pyarrow.get_include.html)

isVoid: Ah I was not aware. It looks good then!

numba_cuda/numba/cuda/tests/nrt/test_nrt.py
```diff
@@ -4,11 +4,89 @@
...
+
+
+@cuda_decl_registry.register_global(allocate_deallocate_handle)
+class AllocateShimImpl(AbstractTemplate):
+    def generic(self, args, kws):
+        return signature(types.void)
+
+
+# let there be a __device__ function expecting
+# a string_view* and returning size_type
```
isVoid, line 67: This comment is going stale.

isVoid (approved):

ci/test_wheel_deps_wheels.sh
```diff
@@ -25,16 +25,18 @@ test_dir = root + \"numba/cuda/tests/test_binary_generation/\"
 print(test_dir)
 "
 
-NUMBA_CUDA_TEST_BIN_DIR=$(python -c "$PY_SCRIPT")
```
gmarkall, line 28: This moves the building of the tests but not the logging step associated with it.

numba_cuda/numba/cuda/cudadrv/linkable_code.py
```diff
@@ -16,14 +16,20 @@ class LinkableCode:
     """
 
     def __init__(
-        self, data, name=None, setup_callback=None, teardown_callback=None
+        self,
+        data,
+        name=None,
+        setup_callback=None,
+        teardown_callback=None,
+        nrt=False,
```
gmarkall, line 24: The docstring should be updated for this new parameter.

numba_cuda/numba/cuda/tests/test_binary_generation/Makefile
```diff
@@ -52,6 +54,13 @@ all:
 	nvcc $(NVCC_FLAGS) $(OBJECT_FLAGS) -o $(OUTPUT_DIR)/test_device_functions.o test_device_functions.cu
 	nvcc $(NVCC_FLAGS) $(LIBRARY_FLAGS) -o $(OUTPUT_DIR)/test_device_functions.a test_device_functions.cu
 
+	nvcc $(NVCC_FLAGS) $(CUBIN_FLAGS) -o $(OUTPUT_DIR)/nrt_extern.cubin nrt_extern.cu -I$(NRT_INCLUDE_DIR)
+	nvcc $(NVCC_FLAGS) $(FATBIN_FLAGS) -o $(OUTPUT_DIR)/nrt_extern.fatbin nrt_extern.cu -I$(NRT_INCLUDE_DIR)
+	nvcc $(NVCC_FLAGS) $(MULTI_FATBIN_FLAGS) -o $(OUTPUT_DIR)/nrt_extern_multi.fatbin nrt_extern.cu -I$(NRT_INCLUDE_DIR)
+	nvcc $(NVCC_FLAGS) $(PTX_FLAGS) -o $(OUTPUT_DIR)/nrt_extern.ptx nrt_extern.cu -I$(NRT_INCLUDE_DIR)
+	nvcc $(NVCC_FLAGS) $(OBJECT_FLAGS) -o $(OUTPUT_DIR)/nrt_extern.o nrt_extern.cu -I$(NRT_INCLUDE_DIR)
+	nvcc $(NVCC_FLAGS) $(LIBRARY_FLAGS) -o $(OUTPUT_DIR)/nrt_extern.a nrt_extern.cu -I$(NRT_INCLUDE_DIR)
+
```
gmarkall, line 63: Should we also be generating LTO-IR for `nrt_extern`?

numba_cuda/numba/cuda/tests/nrt/test_nrt.py
```diff
@@ -77,6 +155,55 @@ def g(out_ary):
...
 
 
+class TestNrtLinking(CUDATestCase):
+    def run(self, result=None):
+        with override_config("CUDA_ENABLE_NRT", True):
+            super(TestNrtLinking, self).run(result)
+
+    def test_nrt_detect_linked_ptx_file(self):
+        src = f"#include <{get_include()}/nrt.cuh>"
+        src += """;
```
gmarkall, line 165: The superfluous semicolon causes a warning.

```suggestion
        src += """
```

gmarkall (changes requested):
The implementation of `add_data()` (#235) should probably be fixed within this PR in ctypes and cuda-python linkers, because without that fix the additional tests fail, e.g.:

```
ERROR: test_nrt_detect_linkable_code (numba.cuda.tests.nrt.test_nrt.TestNrtLinking.test_nrt_detect_linkable_code) (code=<numba.cuda.cudadrv.linkable_code.Cubin object at 0x706fbc9763d0>)
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/tests/nrt/test_nrt.py", line 204, in test_nrt_detect_linkable_code
    kernel[1, 1]()
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 737, in __call__
    return self.dispatcher.call(
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 883, in call
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 891, in _compile_for_args
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba/numba/core/compiler_lock.py", line 35, in _acquire_compile_lock
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 1159, in compile
    kernel.bind()
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 375, in bind
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/codegen.py", line 260, in get_cufunc
    cubin = self.get_cubin(cc=device.compute_capability)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/codegen.py", line 237, in get_cubin
    self._link_all(linker, cc, ignore_nonlto=False)
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/codegen.py", line 204, in _link_all
    linker.add_file_guess_ext(path, ignore_nonlto)
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/cudadrv/driver.py", line 2886, in add_file_guess_ext
    self.add_data(
    ^^^^^^^^^^^^^
AttributeError: 'CtypesLinker' object has no attribute 'add_data'
```

I haven't looked into why this is not also failing on CI, but I think it ought to also be checked on CI.

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -265,19 +266,33 @@ def maybe_link_nrt(self, link, tgt_ctx, asm):
         if not tgt_ctx.enable_nrt:
```
gmarkall, line 266: It's hard to check the logic of `maybe_link_nrt()` without being certain what conditions the NRT is supposed to be linked for; I feel like it needn't be this complicated, but that's hard to verify. Can you add some documentation in the form of comments spelling out what conditions are being checked for / why they're being checked?

gmarkall: For example - I thought we concluded that we'd require `LinkableCode` instances to explicitly have `nrt` set, so there would be no need to scan the data - the user creating a `LinkableCode` should know whether it's using NRT or not.

gmarkall:
I just merged `main` so this branch can pick up the recent CI fixes.

numba_cuda/numba/cuda/tests/test_binary_generation/generate_raw_ltoir.py
```diff
@@ -160,7 +163,9 @@ def main(sourcepath, outputpath, arch):
         help="compute arch to target (e.g. sm_87). Defaults to sm_50.",
         default="sm_50",
     )
-
+    parser.add_argument(
+        "--nrt", type=lambda x: x.lower() == "true", default=False
+    )
```
gmarkall, line 168: Could this be:
```suggestion
    parser.add_argument("--nrt", action="store_true")
```

to keep it simple? I'm not sure I understand what is going on with the lambda here.

numba_cuda/numba/cuda/tests/test_binary_generation/Makefile
```diff
@@ -52,7 +54,16 @@ all:
...
 	# Generate LTO-IR wrapped in a fatbin
 	nvcc $(NVCC_FLAGS) $(LTOIR_FLAGS) -o $(OUTPUT_DIR)/test_device_functions.ltoir.o test_device_functions.cu
+	nvcc $(NVCC_FLAGS) $(LTOIR_FLAGS) -o $(OUTPUT_DIR)/nrt_extern.ltoir.o nrt_extern.cu -I$(NRT_INCLUDE_DIR)
 	# Generate LTO-IR in a "raw" LTO-IR container
 	python generate_raw_ltoir.py --arch sm_$(GPU_CC) -o $(OUTPUT_DIR)/test_device_functions.ltoir test_device_functions.cu
+	python generate_raw_ltoir.py --arch sm_$(GPU_CC) -o $(OUTPUT_DIR)/nrt_extern.ltoir nrt_extern.cu --nrt=True
```
gmarkall, line 69: Note this would just be
```suggestion
	python generate_raw_ltoir.py --arch sm_$(GPU_CC) -o $(OUTPUT_DIR)/nrt_extern.ltoir nrt_extern.cu --nrt
```

if the suggestion below (https://github.com/NVIDIA/numba-cuda/pull/167/files#r2072385599) were accepted.

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -262,22 +264,53 @@ def link_to_library_functions(
         self.reload_init = []
 
     def maybe_link_nrt(self, link, tgt_ctx, asm):
+        """
+        Add the NRT source code to the link if the neccesary conditions are met:
+        If NRT is enabled for the CUDATargetContext, return True if we either
+        detect .extern .func decls of NRT functions in any file in the link or
+        any of the passed LinkableCode objects have NRT enabled. In the special
+        case of PTXSource or CUSource objects, the source code will be inspected
+        even if the NRT flag is not set.
```
gmarkall, line 273: Why are we inspecting the source of PTX and CUDA C/C++ source files? Shouldn't the user be expected to explicitly state that they require NRT by setting the `LinkableCode` `nrt` property? There's a couple of reasons I think it would be better to require this:

* I think a while back we concluded it would be hard to perfectly determine whether arbitrary linkable code used the NRT (potential false positives and false negatives), so we shouldn't try.
* A user might pass in an object as PTX or CUDA source, and rely on NRT getting linked in automatically. If they then decide to switch to LTO-IR for better performance, the automatic linking will no longer work, and their kernels will appear to suddenly have unresolved NRT symbols.

brandon-b-miller: I am quite +1 on requiring the user the be explicit about when they need NRT and less autodetection. Does the following seem right then? We'd link NRT if:

- it is globally enabled

**And** either:
- NRT is autodetected in the main asm (this is the only thing we try and autodetect NRT symbols in)
- a LinkableCode object with NRT specifically enabled is passed

If a user passes a `.cu` or `.ptx` file (in any form, either as a path or as a `LinkableCode` instance) _without_ NRT explicitly set, it may fail to link if the code does contain NRT symbols.

brandon-b-miller: Should be addressed in https://github.com/NVIDIA/numba-cuda/pull/167/commits/ff12e16cc0ede0d35e5f71b3afbcd2fc3db52ee4

numba_cuda/numba/cuda/tests/test_binary_generation/generate_raw_ltoir.py
brandon-b-miller: Just some leftover string validation from a previous iteration here I believe.

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -262,22 +263,38 @@ def link_to_library_functions(
...
         if not tgt_ctx.enable_nrt:
             return
 
-        all_nrt = "|".join(self.NRT_functions)
-        pattern = (
-            r"\.extern\s+\.func\s+(?:\s*\(.+\)\s*)?("
-            + all_nrt
-            + r")\s*\([^)]*\)\s*;"
-        )
+        def nrt_in_asm(asm):
```
gmarkall, line 277: A minor nit: the declaration of `nrt_in_asm` seems superfluous now that it's called exactly once, and the original code would seem to have been sufficient with on the addition of the checks of linkable code below.

numba_cuda/numba/cuda/dispatcher.py
gmarkall: I think this is addressed from a technical standpoint (modulo the minor comment below) but the docstring here is now out of date, suggesting that PTX and CU sources will be inspected.

gmarkall (changes requested):
I think this is generally looking good - just a couple of minor comments on the diff.

I tried merging in #241 locally and it got the tests passing for me.

Once #241 is in and the comments are addressed, I think this will be good to merge.

gmarkall (approved):

PR #168, opened 2025-03-19
Fix NRT stats for cuda python

gmarkall:
The NRT stats implementation did not take the difference in ctypes pointers between the ctypes and cuda-python bindings into account. This PR changes the implementation to handle pointer representations in both sets of bindings.

A test for single stats is also added - this functionality was previously not exercised by the test suite.


brandon-b-miller (approved):
LGTM

PR #170, opened 2025-03-21
Feature: pre commit improvement

ZzEeKkAa:
# Description

- Replace `flake8` with `ruff` code style to automatically format code. `ruff` has same set of features as `flake8`, so having both linters is redundant. `ruff` also adds auto formatting features like `black`. Config was ported from `.flake8` to `.pre-commit-config.yaml` with these differences:
    - remove ignores for `W503` and `W504` (line break after binary operator).
- Add pre-commit hooks that removes trailing spaces, ensures line brakes in the end of the file, formats yaml, toml, json, prevents big files to be added to history etc
- Run pre-commit in separate commit and add it to git blame ignore.

This PR must be merged without squash, to separate config changes (that we want in history) from formatter changes (that we do not want). If rebased, please update `.git-blame-ignore-revs` commit hash.

## Old description

- Add black code style to automatically format code.
- Run pre-commit in separate commit and add it to git blame ignore.

This PR must be merged without squash!


ZzEeKkAa:
I realized the pre-commit is already run, so I'll force push to this branch to remove github action that I added.

This PR must be merged without squash!

isVoid:
@ZzEeKkAa what do you think of `ruff` v.s. `black`? I'm using `ruff` in Numbast and it's faster and usually run into fewer non-auto fixable errors than black IMO

ZzEeKkAa:
I don't have anything against ruff) Just did not use it before

isVoid:
> I don't have anything against ruff) Just did not use it before

I suggest using `ruff` - Numbast will deliver static bindings formatted with `ruff` and I think it should run into fewer issue if Numba also uses a ruff formatter upon re-formatting to Numba's code standard.

gmarkall:
> This PR must be merged without squash!

Is this because of the addition of `.git-blame-ignore-revs`? I think even if I rebase and merge, commit hashes end up different due to something Github does. So I think to make this work correctly, I might need to "rebase and merge" then just add the `.git-blame-ignore-revs` in a follow-up PR afterwards.



> > I don't have anything against ruff) Just did not use it before
> 
> I suggest using `ruff`

I'm in favour of this PR (I just never had the experience of setting something up like this before or the time to figure it out) and I'm happy with either of `black` or `ruff`, whatever you (@isVoid and @ZzEeKkAa) agree on.

leofang:
Let's use Ruff as it's used by all other CUDA Python projects.

ZzEeKkAa:
> > This PR must be merged without squash!
> 
> Is this because of the addition of `.git-blame-ignore-revs`?

Yes

> I think even if I rebase and merge, commit hashes end up different due to something Github does. So I think to make this work correctly, I might need to "rebase and merge" then just add the `.git-blame-ignore-revs` in a follow-up PR afterwards.

Idea is to keep history for the changes to `pyproject.toml`/`.pre-commit-config.yaml` and do not show commit where I run `pre-commit run -a` because it does not have logical sense.

So yeah, if commit will be different after rebase (not sure why it is needed for branch merge, we need to update commit hash there.

ZzEeKkAa:
Just rebased the branch to work with ruff. Should be good to go)

ZzEeKkAa:
> Numbast will deliver static bindings formatted with `ruff` and I think it should run into fewer issue if Numba also uses a ruff formatter upon re-formatting to Numba's code standard.

I guess it is generally a bad idea to run any linters/formatters on the generated code. But I'm all in for `ruff`.

numba_cuda/numba/cuda/cudadrv/dummyarray.py
```diff
@@ -301,25 +304,28 @@ def reshape(self, *newdims, **kws):
...
+                )
             else:
-                newdims = newdims[0:unknownidx] \
-                    + (self.size // knownsize,) \
-                    + newdims[unknownidx + 1:]
+                newdims = (
+                    newdims[0:unknownidx]
+                    + (self.size // knownsize,)
+                    + newdims[unknownidx + 1 :]
+                )
```
isVoid, line 315: We even have backslash line breaks! 🤣

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -449,46 +492,51 @@ def launch(self, args, griddim, blockdim, stream=0, sharedmem=0):
...
             driver.device_to_host(ctypes.addressof(excval), excmem, excsz)
             if excval.value != 0:
                 # An error occurred
                 def load_symbol(name):
-                    mem, sz = cufunc.module.get_global_symbol("%s__%s__" %
-                                                              (cufunc.name,
-                                                               name))
+                    mem, sz = cufunc.module.get_global_symbol(
+                        "%s__%s__" % (cufunc.name, name)
+                    )
```
isVoid, line 512: This looks so much nicer, but it would be even better if ruff can auto format into f-strings.

numba_cuda/numba/cuda/simulator/kernel.py
```diff
@@ -181,27 +189,26 @@ def debug_wrapper(*args, **kwargs):
...
-            ctaid = 'ctaid=%s' % list(self.blockIdx)
-            if str(e) == '':
-                msg = '%s %s' % (tid, ctaid)
+            tid = "tid=%s" % list(self.threadIdx)
+            ctaid = "ctaid=%s" % list(self.blockIdx)
+            if str(e) == "":
+                msg = "%s %s" % (tid, ctaid)
             else:
-                msg = '%s %s: %s' % (tid, ctaid, e)
+                msg = "%s %s: %s" % (tid, ctaid, e)
```
isVoid, line 205: Same, f-strings would be nice.

numba_cuda/numba/cuda/simulator/kernelapi.py
```diff
@@ -16,19 +16,20 @@
...
         self.y = y
         self.z = z
 
     def __str__(self):
-        return '(%s, %s, %s)' % (self.x, self.y, self.z)
+        return "(%s, %s, %s)" % (self.x, self.y, self.z)
 
     def __repr__(self):
-        return 'Dim3(%s, %s, %s)' % (self.x, self.y, self.z)
+        return "Dim3(%s, %s, %s)" % (self.x, self.y, self.z)
```
isVoid, line 32: ditto

isVoid (approved):
I managed to give a rough pass to the code which mostly looks good to me. I went over the new ruff config v.s. the flake8 config and discovered a nice small discussion around W503 and W504 - apparently it was controversial and ruff didn't implement them:
https://github.com/astral-sh/ruff/issues/4125

As for the f-string dicussion below - I realized they are unsafe fixes. They might be a nice code style improvement in a future PR if we can scrutinize into each of the fstring use case deeper. But they shouldn't block this PR.

@ZzEeKkAa What do you think the current developer should do to merge this PR into current working branches?

ZzEeKkAa:
> I managed to give a rough pass to the code which mostly looks good to me. I went over the new ruff config v.s. the flake8 config and discovered a nice small discussion around W503 and W504 - apparently it was controversial and ruff didn't implement them: [astral-sh/ruff#4125](https://github.com/astral-sh/ruff/issues/4125)

Yes, and in `flake8` those where in the ignore list, so no changes for `numba-cuda` by simply removing them.

> As for the f-string dicussion below - I realized they are unsafe fixes. They might be a nice code style improvement in a future PR if we can scrutinize into each of the fstring use case deeper. But they shouldn't block this PR.

Yes, lets target it in another PR.

> @ZzEeKkAa What do you think the current developer should do to merge this PR into current working branches?

Simple merge + resolve conflicts generally should work, otherwise it will be nice to get `.pre-commit-config.yaml` and `pyproject.toml` from this MR (or git checkout first commit from this PR), run pre-commit on changed files and merge after this.

ZzEeKkAa:
@gmarkall could we merge this PR?

gmarkall:
I just merged this, but I had to do it outside the normal PR flow because even "rebase and merge" changes the commit hashes, which would mess up the `.git-blame-ignore-revs` file. As such, this isn't detected as merged, and I'll close this PR.

PR #176, opened 2025-03-25
Add col/row-major layout to shared and local memory

ZzEeKkAa:
Add support for col/row-major layout for local and shared memory:

```python
cuda.shared.array((10,), dtype=np.int32, layout='F')
```

This will greatly improve user experience for those who work with libraries like cublasdx and function expects col-major inputs:

https://github.com/NVIDIA/nvmath-python/blob/main/examples/device/cublasdx_simple_gemm_fp32.py#L37-L41

```python
# cuBLASDx requires column-major arrays but cuda.shared.array creates row-major
# arrays (only) so we emulate a column-major array by flipping dimensions
smem_a = cuda.shared.array(shape=(a_dim[1], a_dim[0]), dtype=np.float32)
smem_b = cuda.shared.array(shape=(b_dim[1], b_dim[0]), dtype=np.float32)
smem_c = cuda.shared.array(shape=(c_dim[1], c_dim[0]), dtype=np.float32)
...
MM(alpha, smem_a, smem_b, beta, smem_c)
```

I have question about name. Do we want have `layout` or do we want `order` to be compatible with numpy?

### DIFF
--- a/numba_cuda/numba/cuda/cudadecl.py
+++ b/numba_cuda/numba/cuda/cudadecl.py
@@ -25,7 +25,7 @@ register_number_classes(register_global)
 
 class Cuda_array_decl(CallableTemplate):
     def generic(self):
-        def typer(shape, dtype):
+        def typer(shape, dtype, layout=None):
 
             # Only integer literals and tuples of integer literals are valid
             # shapes
@@ -38,11 +38,19 @@ class Cuda_array_decl(CallableTemplate):
                     return None
             else:
                 return None
+            
+            # Only string literals are valid layouts
+            if layout in (None, types.none):
+                layout_str = 'C'
+            else:
+                if not isinstance(layout, types.StringLiteral):
+                    return None
+                layout_str = layout.literal_value
 
             ndim = parse_shape(shape)
             nb_dtype = parse_dtype(dtype)
             if nb_dtype is not None and ndim is not None:
-                return types.Array(dtype=nb_dtype, ndim=ndim, layout='C')
+                return types.Array(dtype=nb_dtype, ndim=ndim, layout=layout_str)
 
         return typer
 
--- a/numba_cuda/numba/cuda/cudaimpl.py
+++ b/numba_cuda/numba/cuda/cudaimpl.py
@@ -92,45 +92,55 @@ def _get_unique_smem_id(name):
 
 
 @lower(cuda.shared.array, types.IntegerLiteral, types.Any)
+@lower(cuda.shared.array, types.IntegerLiteral, types.Any, types.StringLiteral)
 def cuda_shared_array_integer(context, builder, sig, args):
     length = sig.args[0].literal_value
     dtype = parse_dtype(sig.args[1])
+    layout = sig.return_type.layout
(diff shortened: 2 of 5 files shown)


samaid:
> Do we want have `layout` or do we want `order` to be compatible with numpy?

My few cents. In nvmath-python terminology it is a layout. But your changes are in the array structure, which mimics numpy . For this reason I think the `order` is better.

leofang:
Following np is good 👍

numba_cuda/numba/cuda/cudadecl.py
```diff
@@ -38,11 +38,19 @@ def typer(shape, dtype):
                     return None
             else:
                 return None
+            
+            # Only string literals are valid layouts
+            if layout in (None, types.none):
+                layout_str = 'C'
+            else:
+                if not isinstance(layout, types.StringLiteral):
+                    return None
```
gmarkall, line 47: Instead of returning `None` and the user getting a generic message, we can force some more specific information to be exposed by requiring a string literal:

```diff
diff --git a/numba_cuda/numba/cuda/cudadecl.py b/numba_cuda/numba/cuda/cudadecl.py
index 9105041..602f38d 100644
--- a/numba_cuda/numba/cuda/cudadecl.py
+++ b/numba_cuda/numba/cuda/cudadecl.py
@@ -11,6 +11,7 @@ from numba.core.typing.templates import (AttributeTemplate, ConcreteTemplate,
                                          AbstractTemplate, CallableTemplate,
                                          signature, Registry)
 from numba.cuda.types import dim3
+from numba.core.errors import RequireLiteralValue
 from numba.core.typeconv import Conversion
 from numba import cuda
 from numba.cuda.compiler import declare_device_function_template
@@ -44,7 +45,8 @@ class Cuda_array_decl(CallableTemplate):
                 layout_str = 'C'
             else:
                 if not isinstance(layout, types.StringLiteral):
-                    return None
+                    msg = "layout must be a literal string)"
+                    raise RequireLiteralValue(msg)
                 layout_str = layout.literal_value
 
             ndim = parse_shape(shape)
```

then the message presented to the user will include:

```
   Rejected as the implementation raised a specific error:
     RequireLiteralValue: layout must be a literal string
```

gmarkall (changes requested):
I think the code looks good; I agree with previous reviewers that `order` to match NumPy would be good. (I guess it's just an anomaly that Numba internally refers to it as `layout`, and there's no need to expose that outside).

I think the documentation will need some mention of the new kwarg.

gmarkall (changes requested):
All changes so far look good. Unfortunately the docs for shared and local arrays are not generated from the docstrings, so they will still omit the order kwarg. They are in: https://nvidia.github.io/numba-cuda/reference/kernel.html#memory-management

So before merge, could you please:

- Update the user-facing documentation
- (Optionally) add a test when passing an invalid type for the order - e.g. a non-literal or non-string type (the case where a valid type is used but an invalid value, e.g. "D", is handled a bit deeper in Numba and produces a meaningful error message "NumbaValueError: Invalid layout 'D'", so I don't think that needs checking here).

Many thanks!

Issue #177, opened 2025-03-26
[FEA] CI should test the simulator


gmarkall:
**Is your feature request related to a problem? Please describe.**

CI does not test the CUDA simulator, like gpuCI in the main Numba repo used to. The simulator is important because it is used as a debugging facility by various users, and is also used as a way to have CI on users' CUDA code when they have no CUDA device available in CI.

**Describe the solution you'd like**

There should be a CI configuration that ensures that the CUDA simulator can run and pass the test suite.

**Describe alternatives you've considered**

There aren't really acceptable alternatives - manual testing of the simulator is not a proper solution.

**Additional context**

None.

kkraus14:
Fixed by #178

PR #178, opened 2025-03-26
Fix the CUDA simulator

gmarkall:
Over the last few months, various changes and additions have been made that have not been reflected in the CUDA simulator, and as a result the test suite is no longer even runnable. Ideally CI would test the simulator (#177) but a first and immediately necessary step is to get the simulator working again.

There were a couple of items necessary on top of just updating the simulator:

- Fixing aliasing of the `Runtime` object and the `numba.cuda.runtime` package by renaming it to `memory_management`.
- Removing some invalid `reshape` tests. This is required because the simulator runs these functions using the NumPy implementation, which rightly complains that the inputs are invalid.


gmarkall:
Looks like the CUDA simulator also needs fixing to work with NumPy 2.2.

.github/workflows/conda-python-simulator.yaml
gmarkall, line 1: I have ended up with two different yaml files for the simulator tests in my confusion, and there should only be one.

.github/workflows/conda-python-simulator.yaml
```diff
@@ -0,0 +1,107 @@
+on:
```
brandon-b-miller, line 1: This file looks to be very similar to `conda-python-build.yaml`

```
20c20
<         default: "ci/build_python.sh"
---
>         default: "ci/test_conda_simulator.sh"
27,29d26
<       upload_to_anaconda:
<         type: boolean
<         required: false
56c53
<       - name: Compute Build Matrix
---
>       - name: Compute Simulator Test Matrix
69c66
<   build:
---
>   test:
```

I wonder if some of this can be factored out to somewhere common, I'll fiddle with it a bit locally and see if I can provide a patch.

numba_cuda/numba/cuda/simulator/__init__.py
```diff
@@ -38,11 +38,20 @@
...
 
+    sys.modules["numba.cuda.bf16"] = bf16
     sys.modules["numba.cuda.compiler"] = compiler
+    sys.modules["numba.cuda._internal"] = _internal
+    sys.modules["numba.cuda._internal.cuda_bf16"] = _internal.cuda_bf16
+
+    from numba.cuda.simulator import memory_management
+
+    sys.modules["numba.cuda.memory_management"] = memory_management
+    sys.modules["numba.cuda.memory_management.nrt"] = memory_management.nrt
```
brandon-b-miller, line 57: there's something about this that's making me scratch my head. Here we're adding modules in a way that seems like it's making them available for the CUDA simulator, but later on it looks like the CUDA simulator disables NRT globally. Are these just here so everything imports correctly, even if they're not used?

gmarkall: They're here so that users of these modules / packages can import them without error (e.g. tests, user code, etc.). It's important that anything that might get imported has a dummy entry in `sys.modules` so that the real module never gets imported later (e.g. if we didn't populate `sys.modules` with dummy ones) because then we'd end up with bits of simulator imported and bits of the real target, and that very quickly leads to a mess that crashes with a surprising exception.

.github/workflows/conda-python-simulator.yaml
gmarkall: Additionally - the simulator-test.yaml file is also very similar, but I don't expect it to remain that way. I feel like I need to get this PR over the line before making further modifications, though, as every small churn on `main` needs a corresponding update here. 

Since I hope to evolve the simulator testing, I'd rather not do / undo factoring things together with the build yaml files across two (or more PRs).

PR #179, opened 2025-03-26
Add CUDA C ABI arg packer ensuring compatibility of argument types with CUDA C ABI 

### LINKED ISSUE #175: [BUG] C ABI passes struct values incorrectly when compiling with `compile()`
**Describe the bug**

As observed in https://github.com/NVIDIA/cccl/issues/4248: 

Consider an extern "C" device function accepting two structs as inputs:

```c
struct MyStruct {
    char a;
    char b;
    char c;
};

extern "C" __device__ void foo(MyStruct a, MyStruct b) {

}
```

Here's the corresponding PTX:

```assembly
.visible .func foo(
        .param .align 1 .b8 foo_param_0[3],
        .param .align 1 .b8 foo_param_1[3]
)
{

        ret;

}
```

Now, consider defining this device function using numba.cuda, using `@gpu_struct` to define the type of the arguments, and keeping in mind that gpu_struct uses [StructModel](https://github.com/numba/numba/blob/c21aa9273ef4298392695b1f4613d29456b53e5c/numba/core/datamodel/old_models.py#L514) to define the underlying numba type:

```python
import numba
from numba import cuda
import numpy as np

from cuda.parallel.experimental.struct import gpu_struct

@gpu_struct
class MyStruct:
    a: np.int8
    b: np.int8
    c: np.int8

def op(a, b):
    pass

tp = numba.typeof(MyStruct(1, 2, 3))
ptx, _ = cuda.compile(op, (tp, tp))

print(ptx)
```

Here's the output of the above script:

```assembly
.visible .func  (.param .b64 func_retval0) op(
        .param .b32 op_param_0,
        .param .b32 op_param_1,
        .param .b32 op_param_2,
        .param .b32 op_param_3,
        .param .b32 op_param_4,
        .param .b32 op_param_5
)
{
        .reg .b64       %rd<2>;


        mov.u64         %rd1, 0;
        st.param.b64    [func_retval0+0], %rd1;
        ret;

}
```

Comparing the two PTXs, we see an ABI difference: on the C++ side, struct arguments are of type `.b8[N]` where `N` is the size of the struct. On the numba side however, we note that structs have been decomposed into their members, and the function accepts a `.b32` for each member. ([reason for promotion to 32-bit](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html?highlight=abi#function-declarations-and-definitions)).

This is because the C ABI calling convention is packing struct value arguments according to the Numba calling convention, not the C one. It just uses the default argument packer:

https://github.com/NVIDIA/numba-cuda/blob/fbbc0402ca14b931fd00f2484c0af34730058bfe/numba_cuda/numba/cuda/target.py#L306

Reproducer in code above.

**Expected behavior**

The C ABI calling convention should pack struct values according to the C ABI calling convention, yielding the same function prototypes as CUDA C/C++ code compiled with NVCC.

This will probably require the implementation of a new arg packer for the C calling convention.

All environments.

**Additional context**

No other context.

shwina:
Closes #175.

This PR introduces `CUDACABIArgPacker` - a subclass of [`ArgPacker`](https://github.com/numba/numba/blob/c21aa9273ef4298392695b1f4613d29456b53e5c/numba/core/datamodel/packer.py#L60) that ensures arguments are packed according to the C ABI calling convention. Largely, it does this simply by invoking the data model's `get_data_type()` as opposed to `get_value_type()`.

### DIFF
--- a/numba_cuda/numba/cuda/target.py
+++ b/numba_cuda/numba/cuda/target.py
@@ -302,7 +302,7 @@ class CUDATargetContext(BaseContext):
     def get_arg_packer(self, fe_args):
         return CUDACABIArgPacker(self.data_model_manager, fe_args)
 
-    
+
 class CUDACallConv(MinimalCallConv):
     pass
 
--- a/numba_cuda/numba/cuda/target.py
+++ b/numba_cuda/numba/cuda/target.py
@@ -9,6 +9,7 @@ from numba.core.base import BaseContext
 from numba.core.callconv import BaseCallConv, MinimalCallConv
 from numba.core.typing import cmathdecl
 from numba.core import datamodel
+from numba.core.datamodel.packer import ArgPacker, _Unflattener, _flatten
 
 from .cudadrv import nvvm
 from numba.cuda import codegen, ufuncs
@@ -59,6 +60,61 @@ class CUDATypingContext(typing.BaseContext):
 
 # -----------------------------------------------------------------------------
 # Implementation
+
+
+class CUDACABIArgPacker(ArgPacker):
+    """
+    Compute the position for each high-level typed argument.
+    It flattens every composite argument into primitive types.
+    It maintains a position map for unflattening the arguments.
+
+    Since struct (esp. nested struct) have specific ABI requirements (e.g.
+    alignment, pointer address-space, ...) in different architecture (e.g.
+    OpenCL, CUDA), flattening composite argument types simplifes the call
+    setup from the Python side.  Functions are receiving simple primitive
+    types and there are only a handful of these.
+    """
+
+    def __init__(self, dmm, fe_args):
+        self._dmm = dmm
+        self._fe_args = fe_args
+        self._nargs = len(fe_args)
+
+        self._dm_args = []
+        argtys = []
+        for ty in fe_args:
+            dm = self._dmm.lookup(ty)
+            self._dm_args.append(dm)
+            argtys.append(dm.get_data_type())
+        self._unflattener = _Unflattener(argtys)
+        self._be_args = list(_flatten(argtys))
+
+    def as_arguments(self, builder, values):
+        """Flatten all argument values
+        """
+        if len(values) != self._nargs:
+            raise TypeError("invalid number of args: expected %d, got %d"
+                            % (self._nargs, len(values)))
+
+        if not values:
+            return ()
+
+        args = [dm.as_data(builder, val)
+                for dm, val in zip(self._dm_args, values)
+                ]
+
+        args = tuple(_flatten(args))
+        return args
+
+    def from_arguments(self, builder, args):
+        """Unflatten all argument values
+        """
+        valtree = self._unflattener.unflatten(args)
+        values = [dm.from_data(builder, val)
+                  for dm, val in zip(self._dm_args, valtree)
+                  ]
+
+        return values
 
 
 VALID_CHARS = re.compile(r'[^a-z0-9]', re.I)
@@ -243,7 +299,10 @@ class CUDATargetContext(BaseContext):
     def get_ufunc_info(self, ufunc_key):
         return ufuncs.get_ufunc_info(ufunc_key)
 
+    def get_arg_packer(self, fe_args):
+        return CUDACABIArgPacker(self.data_model_manager, fe_args)
 
+    
 class CUDACallConv(MinimalCallConv):
     pass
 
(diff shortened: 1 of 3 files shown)


numba_cuda/numba/cuda/target.py
```diff
@@ -61,6 +62,61 @@ def resolve_value_type(self, val):
 # Implementation
 
 
+class CUDACABIArgPacker(ArgPacker):
```
shwina, line 65: This repeats a significant amount of logic in [`ArgPacker`](https://github.com/numba/numba/blob/c21aa9273ef4298392695b1f4613d29456b53e5c/numba/core/datamodel/packer.py#L60).

Perhaps if we could have the base class encapsulate the calls to `dm.get_arguement_type()`, `dm.as_argument()`, and `dm.from_argument()` in overridable methods, that would eliminate the redundancy, but I'm not sure if that really makes sense or if numba would be receptive to that change.

numba_cuda/numba/cuda/tests/cudadrv/test_packer.py
shwina, line 1: Is this the right place for this test?

numba_cuda/numba/cuda/target.py
```diff
@@ -61,6 +62,61 @@ def resolve_value_type(self, val):
...
+    Compute the position for each high-level typed argument.
+    It flattens every composite argument into primitive types.
+    It maintains a position map for unflattening the arguments.
+
+    Since struct (esp. nested struct) have specific ABI requirements (e.g.
+    alignment, pointer address-space, ...) in different architecture (e.g.
+    OpenCL, CUDA), flattening composite argument types simplifes the call
+    setup from the Python side.  Functions are receiving simple primitive
+    types and there are only a handful of these.
+    """
```
shwina, line 76: This docstring should go away

numba_cuda/numba/cuda/target.py
```diff
@@ -61,6 +62,54 @@ def resolve_value_type(self, val):
 # Implementation
 
 
+class CUDACABIArgPacker(ArgPacker):
+    """
+    Subclass of ArgPacker which packs arguments according to the
+    CUDA C ABI calling convention.
+    """
+
+    def __init__(self, dmm, fe_args):
```
leofang, line 71: Dumb question from someone who has little knowledge in Numba internal: As you said the subclass here looks (almost?) identical to `ArgPacker`. So is the C ABI information fully encoded in the `dmm` argument?

numba_cuda/numba/cuda/tests/cudadrv/test_packer.py
gmarkall: Yes.

gmarkall: Actually nevermind, I realised this is in `cudadrv`, whereas it should be in `cudapy`.

gmarkall:
I spent a little more time looking at this today. I think we really want to be verifying that we generate LLVM IR and PTX that will have the correct argument types.

For example, I ran:

```python
from numba import cuda
from numba.cuda.typing.typeof import typeof

import itertools

def f(a, b):
    pass

unituple_type = typeof((1, 2, 3))
tuple_type = typeof((1, 2.0, 3))
nested_tuple_type = typeof((1, (2.0, 3.0)))

types = (unituple_type, tuple_type, nested_tuple_type)

for args in itertools.product(types, types):
    print(args)
    ptx, res_ty = cuda.compile(f, args, device="True", abi="c", output="ptx")
    print(ptx)
```

with `NUMBA_DUMP_LLVM=1` to get a sense of what we should be looking at. 

For example, when `args` is `(UniTuple(int64, 3), Tuple((int64, float64, int64)))`, we should look at the arguments to the function in LLVM IR and make sure it looks like:

```llvm
define i8* @"f"([3 x i64] %".1", {i64, double, i64} %".2")
```

as opposed to what we got before this PR, which was:

```llvm
define i8* @"f"(i64 %".1", i64 %".2", i64 %".3", i64 %".4", double %".5", i64 %".6")
```

and for the PTX, do the arguments match the PTX of a CUDA C++ device function with struct arguments defined like:

```cuda
struct X {
  int64_t a;
  int64_t b;
  int64_t c;
};

struct Y {
  int64_t a;
  double b;
  int64_t c;
};
```

For this example I'm seeing

```asm
.visible .func  (.param .b64 func_retval0) f(
	.param .align 8 .b8 f_param_0[24],
	.param .align 8 .b8 f_param_1[24]
)
```

which probably fits. But I'm not sure if at the NVVM IR level, `[3 x i64]` is correct for a three-element struct - we would have to check, and ensure there is a alignment so that if we produce LTO-IR to link with NVRTC-generated LTO-IR, we won't confuse the linker by not having types that are exactly the same structure.

PR #189, opened 2025-04-02
Ensure RTC Bindings Clamp to the Maximum Supported CC

### LINKED ISSUE #182: [BUG] Tests fail with CUDA 11.4
**Describe the bug**
In #150, I tried to update the numba-cuda CI matrix to match that of RAPIDS, with a wider support matrix. I saw that tests failed in CUDA 11.4 with the following error:

```
FAIL: test_linking_cu_error (numba.cuda.tests.cudadrv.test_linker.TestLinker)
  File "/opt/conda/envs/test/lib/python3.9/site-packages/numba_cuda/numba/cuda/tests/cudadrv/test_linker.py", line 197, in test_linking_cu_error
    self.assertIn('NVRTC Compilation failure', msg)
AssertionError: 'NVRTC Compilation failure' not found in 'Failed to call nvrtcCompileProgram: NVRTC_ERROR_INVALID_OPTION'

```

See https://github.com/NVIDIA/numba-cuda/actions/runs/13724549728/job/38387960456?pr=150#step:9:3850.

bdice:
Resolves #182.

This PR adds check in nvrtc bindings to make sure that the device compute capability is not exceeding the supported target of the installed RTC. If the device compute capability exceeds, RTC binding will fall back to choosing the maximum supported CC of the underlying compiler.


isVoid:
Root cause:
NVRTC 11.4 supports `-arch` option up to `compute_80`. `L4` is `compute_89`.

bdice:
Great. We could change this job to use a V100 (x86_64) or A100 (ARM) runner, but @isVoid pointed out to me that we don't produce `rapidsai/citestwheel` images for CUDA 11.4. https://github.com/rapidsai/ci-imgs/blob/4b6b16f9de8028a8dd2c98329f269f158040adcf/matrix.yaml#L62-L64

I am not sure if there's anything _preventing_ us from making CUDA 11.4 citestwheel images. We could try adding them if needed. Product question: how badly do we need/want to support CUDA 11.4 in the numba-cuda CI?

kkraus14:
> I am not sure if there's anything _preventing_ us from making CUDA 11.4 citestwheel images. We could try adding them if needed. Product question: how badly do we need/want to support CUDA 11.4 in the numba-cuda CI?

We shouldn't prioritize this for `numba-cuda`.

bdice:
@kkraus14 Great. I am happy with that. Shall we close this PR and issue #182, then?

bdice:
Perhaps we should modify this PR to just remove the TODO and not change the test matrix.

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -279,6 +283,7 @@ def compile(src, name, cc, ltoir=False):
         options += ["-std=c++17"]
 
     # Compile the program
+    print(options)
```
kkraus14, line 286:
```suggestion
```

kkraus14:
We need the fix from https://github.com/NVIDIA/numba-cuda/pull/189/commits/8150534bd3d522d9de797bfcf3fae07000450a73, but we can probably just test for CUDA 11.8.

.github/actions/compute-matrix/action.yaml
```diff
@@ -20,15 +20,13 @@ runs:
         "
 
         export TEST_MATRIX="
-          # TODO: Use CUDA 11.4.3 below.
-          - { CUDA_VER: '11.8.0', ARCH: 'amd64', PY_VER: '3.9',  LINUX_VER: 'rockylinux8', GPU: 'l4',   DRIVER: 'latest' }
+          - { CUDA_VER: '11.4.3', ARCH: 'amd64', PY_VER: '3.9',  LINUX_VER: 'rockylinux8', GPU: 'l4',   DRIVER: 'latest' }
```
bdice, line 23:
```suggestion
          - { CUDA_VER: '11.8.0', ARCH: 'amd64', PY_VER: '3.9',  LINUX_VER: 'rockylinux8', GPU: 'l4',   DRIVER: 'latest' }
```

.github/workflows/conda-python-tests.yaml
```diff
@@ -71,23 +71,22 @@ jobs:
           export MATRICES="
             pull-request:
               # amd64
-              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', GPU: 'l4',   DRIVER: 'earliest', DEPENDENCIES: 'oldest' }
+              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.4.3', LINUX_VER: 'rockylinux8', GPU: 'l4',   DRIVER: 'earliest', DEPENDENCIES: 'oldest' }
```
bdice, line 74:
```suggestion
              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', GPU: 'l4',   DRIVER: 'earliest', DEPENDENCIES: 'oldest' }
```

.github/workflows/conda-python-tests.yaml
```diff
@@ -71,23 +71,22 @@ jobs:
...
+              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', GPU: 'l4',   DRIVER: 'earliest', DEPENDENCIES: 'oldest' }
               # TODO: Use CUDA 12.8.0 and h100 below.
               - { ARCH: 'amd64', PY_VER: '3.13', CUDA_VER: '12.5.1', LINUX_VER: 'ubuntu24.04', GPU: 'l4',   DRIVER: 'latest',   DEPENDENCIES: 'latest' }
               # arm64
               - { ARCH: 'arm64', PY_VER: '3.11', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', GPU: 'a100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
             nightly:
               # amd64
-              # TODO: Use CUDA 11.4.3 below.
-              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', GPU: 'l4',   DRIVER: 'earliest', DEPENDENCIES: 'oldest' }
+              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.4.3', LINUX_VER: 'rockylinux8', GPU: 'l4',   DRIVER: 'earliest', DEPENDENCIES: 'oldest' }
```
bdice, line 82:
```suggestion
              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', GPU: 'l4',   DRIVER: 'earliest', DEPENDENCIES: 'oldest' }
```

.github/workflows/conda-python-tests.yaml
```diff
@@ -71,23 +71,22 @@ jobs:
...
               - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu20.04', GPU: 'l4',   DRIVER: 'latest',   DEPENDENCIES: 'latest' }
               - { ARCH: 'amd64', PY_VER: '3.11', CUDA_VER: '12.0.1', LINUX_VER: 'rockylinux8', GPU: 'l4',   DRIVER: 'latest',   DEPENDENCIES: 'latest' }
               - { ARCH: 'amd64', PY_VER: '3.12', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', GPU: 'l4',   DRIVER: 'latest',   DEPENDENCIES: 'latest' }
               # TODO: Use CUDA 12.8.0 and h100 below.
               - { ARCH: 'amd64', PY_VER: '3.13', CUDA_VER: '12.5.1', LINUX_VER: 'ubuntu24.04', GPU: 'l4',   DRIVER: 'latest',   DEPENDENCIES: 'latest' }
               # arm64
-              # TODO: Use CUDA 11.4.3 below.
-              - { ARCH: 'arm64', PY_VER: '3.9',  CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu20.04', GPU: 'a100', DRIVER: 'latest',   DEPENDENCIES: 'oldest' }
+              - { ARCH: 'arm64', PY_VER: '3.9',  CUDA_VER: '11.4.3', LINUX_VER: 'ubuntu20.04', GPU: 'a100', DRIVER: 'latest',   DEPENDENCIES: 'oldest' }
```
bdice, line 89:
```suggestion
              - { ARCH: 'arm64', PY_VER: '3.9',  CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu20.04', GPU: 'a100', DRIVER: 'latest',   DEPENDENCIES: 'oldest' }
```

.github/workflows/wheels-test.yaml
```diff
@@ -79,7 +79,7 @@ jobs:
           #
           export MATRICES="
             pull-request:
-              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', GPU: 'l4',   DRIVER: 'latest', DEPENDENCIES: 'oldest' }
+              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.4.3', LINUX_VER: 'rockylinux8', GPU: 'l4',   DRIVER: 'latest', DEPENDENCIES: 'oldest' }
```
bdice, line 82:
```suggestion
              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', GPU: 'l4',   DRIVER: 'latest', DEPENDENCIES: 'oldest' }
```

bdice (commented):
Let's go back to testing 11.8.0. Then this should be mergeable.

.github/actions/compute-matrix/action.yaml
```diff
@@ -20,15 +20,13 @@ runs:
...
-          # TODO: Use CUDA 11.4.3 below.
           - { CUDA_VER: '11.8.0', ARCH: 'amd64', PY_VER: '3.9',  LINUX_VER: 'rockylinux8', GPU: 'l4',   DRIVER: 'latest' }
           - { CUDA_VER: '11.8.0', ARCH: 'amd64', PY_VER: '3.10', LINUX_VER: 'ubuntu20.04', GPU: 'l4',   DRIVER: 'latest' }
           - { CUDA_VER: '12.0.1', ARCH: 'amd64', PY_VER: '3.11', LINUX_VER: 'ubuntu22.04', GPU: 'l4',   DRIVER: 'latest' }
           - { CUDA_VER: '12.2.2', ARCH: 'amd64', PY_VER: '3.12', LINUX_VER: 'ubuntu22.04', GPU: 'l4',   DRIVER: 'latest' }
           # TODO: Use CUDA 12.8.0 below.
           - { CUDA_VER: '12.5.1', ARCH: 'amd64', PY_VER: '3.13', LINUX_VER: 'ubuntu24.04', GPU: 'l4',   DRIVER: 'latest' }
-          # TODO: Use CUDA 11.4.3 below.
-          - { CUDA_VER: '11.8.0', ARCH: 'arm64', PY_VER: '3.9',  LINUX_VER: 'rockylinux8', GPU: 'a100', DRIVER: 'latest' }
+          - { CUDA_VER: '11.4.3', ARCH: 'arm64', PY_VER: '3.9',  LINUX_VER: 'rockylinux8', GPU: 'a100', DRIVER: 'latest' }
```
bdice, line 29:
```suggestion
          - { CUDA_VER: '11.8.0', ARCH: 'arm64', PY_VER: '3.9',  LINUX_VER: 'rockylinux8', GPU: 'a100', DRIVER: 'latest' }
```

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -251,6 +251,10 @@ def compile(src, name, cc, ltoir=False):
     nvrtc = NVRTC()
     program = nvrtc.create_program(src, name)
 
+    if nvrtc.get_version() < (12, 0):
+        # maximum supported target architecture for 11.x is 8.0
+        cc = min(cc, (8, 0))
```
bdice, line 256: CUDA 11.8 supports Hopper, arch 9.0.
```suggestion
        # Maximum supported target architecture for 11.x is cc 9.0 in CUDA 11.8
        cc = min(cc, (9, 0))
```

bdice: Do we need a full range of support here? Do we need to account for CUDA 12.8 introducing Blackwell support?

isVoid: > Do we need a full matrix of arch/CUDA version support here? Do we need to account for CUDA 12.8 introducing Blackwell support?

You can target older architecture and the compiled binary would still work on newer hardware. In CTK 11's case this means you can only target at maximum for Hopper albeit you are running on Blackwell hardware. Although a warning saying that you may not fully exploit the machine's hardware capability could be helpful?

bdice: > But nvrtc is still missing "sm_90" and "compute_90" option?
That seems like a bug in the docs. CUDA 11.8 should have full support for Hopper (sm90).

bdice: > Do we need a full matrix of arch/CUDA version support here?

Sorry, my intent wasn't clear. What I meant was, do we need to spell out arch support for every CUDA version? Or is there some other reason why 11.8 is special?

```python
if nvrtc.get_version() < (11, 0):
    raise RuntimeError("Unsupported CUDA version. CUDA 11.0 or higher is required.")
elif nvrtc.get_version() < (11, 1):
    # Maximum supported target architecture for CUDA 11.0 is cc 8.0
    cc = min(cc, (8, 0))
elif nvrtc.get_version() < (11, 5):
    # Maximum supported target architecture for CUDA 11.1-11.4 is cc 8.6
    cc = min(cc, (8, 6))
elif nvrtc.get_version() < (11, 8):
    # Maximum supported target architecture for CUDA 11.5-11.7.1 is cc 8.7
    cc = min(cc, (8, 7))
elif nvrtc.get_version() < (12, 8):
    # Maximum supported target architecture for CUDA 11.8-12.6 is cc 9.0
    cc = min(cc, (9, 0))
```

But really, we already have this whole table written out in numba-cuda here, and we should use that instead of hardcoding something new: https://github.com/NVIDIA/numba-cuda/blob/fb7c1a2e2c0542327b251128b81f4c2b432ac7c2/numba_cuda/numba/cuda/cudadrv/nvvm.py#L337-L354

bdice:
@isVoid Can you rewrite the PR title and description? I think you'll do a better job of summarizing the bugfix you have included here.

numba_cuda/numba/cuda/cudadrv/nvrtc.py
isVoid: > do we need to spell out arch support for every CUDA version? Or is there some other reason why 11.8 is special?
A full matrix would be good. The patch I provided was only poking at the CI error. And we can reuse the table you proposed. I was also hoping `cuda.core` rtc model can capture something like this nicely. cc @leofang

Q: What happens if the device CC is lower than the minimum supported CC by the CTK? Should we just err?

bdice: > Q: What happens if the device CC is lower than the minimum supported CC by the CTK? Should we just err?

Yes. I don't know the error classes that might exist in numba/numba-cuda, but perhaps there's a good one that indicates "unsupported." It should probably inherit from `RuntimeError`, or maybe we just use `RuntimeError.`

isVoid:
> @isVoid Can you rewrite the PR title and description? I think you'll do a better job of summarizing the bugfix you have included here.

~~Done, can you turn the PR to ready for review state? I pushed the suggested fixes above.~~
NM, I found out that I can hit the switch myself.

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -3,6 +3,9 @@
 from numba.cuda.cudadrv.error import (NvrtcError, NvrtcCompilationError,
                                       NvrtcSupportError)
 from numba.cuda.cuda_paths import get_cuda_paths
+
+from .info import CTK_SUPPORTED
```
bdice, line 7: It looks like this is not aligned with the rest of the file. I think we're preferring full import paths like `from numba.cuda ...` instead of importing from the local module structure.

However, I see that `nvvm.py` has a different style. I guess I would recommend being consistent within each file.

```suggestion
from numba.cuda.cudadrv.info import CTK_SUPPORTED
```

numba_cuda/numba/cuda/cudadrv/info.py
bdice, line 1: Should we rename this to `_info` if it's meant to be private? Or at least rename the dict of data to start with an underscore.

numba_cuda/numba/cuda/cudadrv/info.py
```diff
@@ -0,0 +1,19 @@
+# Maps CTK version -> (min supported cc, max supported cc) inclusive
+CTK_SUPPORTED = {
```
bdice, line 2: Perhaps this can be renamed to `_CUDA_CC_MIN_MAX_SUPPORT` or a similarly descriptive (and private) name.
```suggestion
_CUDA_CC_MIN_MAX_SUPPORT = {
```

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -251,6 +254,31 @@ def compile(src, name, cc, ltoir=False):
     nvrtc = NVRTC()
     program = nvrtc.create_program(src, name)
 
+    version = nvrtc.get_version()
+    if version < (11, 0):
+        # Maximum supported target architecture for 11.x is cc 9.0 in CUDA 11.8
```
bdice, line 259: This comment doesn't apply here anymore.
```suggestion
```

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -251,6 +254,31 @@ def compile(src, name, cc, ltoir=False):
...
+        raise RuntimeError(
+            "Unsupported CUDA version. CUDA 11.0 or higher is required."
+        )
+    else:
+        min_cc, max_cc = CTK_SUPPORTED[version]
```
bdice, line 264: Don't assume that `version` is in `CTK_SUPPORTED`, or else new CUDA versions won't be testable without raising a `KeyError`. We should use `CTK_SUPPORTED.get(version, None)` or `try: ... except KeyError: ...`. We should use the "latest CUDA" key if no exact match is found, perhaps with an additional warning.

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -251,6 +254,31 @@ def compile(src, name, cc, ltoir=False):
...
+        if version < min_cc:
+            raise RuntimeError(
+                f"Device Compute Capability ({cc}) is lower "
+                f"than minimum supported CC ({min_cc}) with installed "
+                "cudatoolkit. Consider downgrading cudatoolkit to supported "
+                "version."
+            )
```
bdice, line 271: Please replace `cudatoolkit` with something more specific that shows the NVRTC version that was detected. Also, I don't think it is very helpful to recommend downgrading CUDA here. It isn't very actionable without more information, and I don't think it should be numba-cuda's job to recommend which NVRTC/CUDA Toolkit/driver versions support which devices.
```suggestion
            raise RuntimeError(
                f"Device Compute Capability ({cc}) is lower than the "
                f"minimum supported Compute Capability ({min_cc}) with "
                f"NVRTC {".".join(version)}."
            )
```

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -251,6 +254,31 @@ def compile(src, name, cc, ltoir=False):
...
+        elif version > max_cc:
+            warnings.warn(
+                f"Device Compute Capability ({cc}) is higher "
+                f"than maximum supported CC ({max_cc}) with installed "
+                "cudatoolkit. Numba will limit the compilation to the maximum"
+                "supported CC. This may hinder performance, consider upgrading"
+                "your cudatoolkit to supported version."
+            )
```
bdice, line 279: Similar to the above. In this case, I think upgrading NVRTC is an appropriate recommendation in the warning.
```suggestion
            warnings.warn(
                f"Device Compute Capability ({cc}) is higher than the "
                f"maximum supported Compute Capability ({max_cc}) with "
                f"NVRTC {".".join(version)}. Numba will limit the compilation "
                f"to the maximum supported Compute Capability ({max_cc}). "
                "This may hinder performance, consider upgrading NVRTC."
            )
```

bdice (commented):
@isVoid Some suggestions for improvement.

numba_cuda/numba/cuda/cudadrv/nvrtc.py
isVoid: I agree that we should stay consistent with module import patterns.

numba_cuda/numba/cuda/cudadrv/nvrtc.py
isVoid: > We should use the "latest CUDA" key if no exact match is found, perhaps with an additional warning.

This feels more robust to me. Ideally we should have all CTK documented here, or is that a too much a ask?

bdice: > Ideally we should have all supported CTK documented here, or is that a too much a ask?

I think all the supported CTK versions are already in the dict? Maybe I am not understanding correctly.

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -251,6 +253,40 @@ def compile(src, name, cc, ltoir=False):
...
+    if version < (11, 0):
+        raise RuntimeError(
+            "Unsupported CUDA version. CUDA 11.0 or higher is required."
+        )
+    else:
+        try:
+            min_cc, max_cc = _CUDA_CC_MIN_MAX_SUPPORT[version]
+        except KeyError:
+            min_cc, max_cc = _CUDA_CC_MIN_MAX_SUPPORT.values()[-1]
+            max_rtc_ver = _CUDA_CC_MIN_MAX_SUPPORT.keys()[-1]
```
bdice, line 268: Don't assume how this is sorted.
```suggestion
            max_rtc_ver, (min_cc, max_cc) = sorted(
                _CUDA_CC_MIN_MAX_SUPPORT.items(), reverse=True
            )[0]
```

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -251,6 +253,40 @@ def compile(src, name, cc, ltoir=False):
...
+                f"minimum supported Compute Capability ({cc_str(min_cc)}) "
+                f"with NVRTC {ver_str(version)}."
+            )
+        elif version > max_cc:
+            warnings.warn(
+                f"Device Compute Capability ({cc_str(cc)}) is higher than the "
+                f"maximum supported Compute Capability ({cc_str(max_cc)}) with "
+                f"NVRTC {ver_str(version)}. Numba will limit the compilation "
+                f"to the maximum supported Compute Capability {cc_str(max_cc)}"
+                ". This may hinder performance, consider upgrading NVRTC."
```
bdice, line 286:
```suggestion
                f"Device Compute Capability ({ver_str(cc)}) is higher than the "
                f"maximum supported Compute Capability ({ver_str(max_cc)}) with "
                f"NVRTC {ver_str(version)}. Numba will limit the compilation to "
                f"the maximum supported Compute Capability {ver_str(max_cc)}. "
                "This may hinder performance, consider upgrading NVRTC."
```

numba_cuda/numba/cuda/cudadrv/_info.py
```diff
@@ -0,0 +1,19 @@
+# Maps CTK version -> (min supported cc, max supported cc) inclusive
+_CUDA_CC_MIN_MAX_SUPPORT = {
+    (11, 1): ((3, 5), (8, 0)),
+    (11, 2): ((3, 5), (8, 6)),
+    (11, 3): ((3, 5), (8, 6)),
+    (11, 4): ((3, 5), (8, 7)),
```
bdice, line 6: Wikipedia says that 8.7 wasn't added until CUDA 11.5 but I cannot find a definitive reference in the CUDA docs for this. 😕

numba_cuda/numba/cuda/cudadrv/_info.py
```diff
@@ -0,0 +1,19 @@
+# Maps CTK version -> (min supported cc, max supported cc) inclusive
+_CUDA_CC_MIN_MAX_SUPPORT = {
```
leofang, line 2: A drive-by comment, feel free to ignore. Hard-coding this list is not needed because it is queryable via NVRTC `nvrtcGetSupportedArchs` API: https://docs.nvidia.com/cuda/nvrtc/#group__query_1ga04c729958964c5e7b889a644f086ad4a, which is exposed in cuda-bindings.

isVoid: This API is added after 11.2: https://docs.nvidia.com/cuda/archive/11.2.2/cuda-toolkit-release-notes/index.html#cuda-general-new-features. 

I think it's a good idea to depend on this - feels more robust than hard coding as documentation can have bugs (like [this](https://github.com/NVIDIA/numba-cuda/pull/189#discussion_r2032076810)). This already covers a majority of the the CTK enlisted here, except from 11.1, which we manually added. cc @bdice

isVoid: Unrelated note - I think it would be great to add supported arch query API for libnvvm as well.

numba_cuda/numba/cuda/cudadrv/_info.py
```diff
@@ -0,0 +1,19 @@
...
+    (11, 1): ((3, 5), (8, 0)),
```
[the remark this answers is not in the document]
isVoid: This is from the NVRTC doc though:
https://docs.nvidia.com/cuda/archive/11.1.0/nvrtc/index.html#group__options

Also if you look at 11.0 release, which adds support for 80, but RTC supports up to 75. I think this flag was being actively maintained at the time during releases.

kkraus14 (approved):

numba_cuda/numba/cuda/cudadrv/_info.py
gmarkall: I'm late to this PR, but it's OK to add this query at any point, because numba-cuda requires CUDA 11.2: https://nvidia.github.io/numba-cuda/user/overview.html#software

(it doesn't work with 11.1 and older because it has the older LLVM 3.4-based NVVM)

PR #193, opened 2025-04-09
Feature: cuda specific make_attribute_wrapper

ZzEeKkAa:
Vendor numba's `make_attribute_wrapper` until it supports `cuda` target. Vendored function was changed to use cuda specific data model manager chained with a default data model manager. That makes it possible to wrap array attributes that are not supported by `numba`, but supported by `numba-cuda` (like `fp16` arrays).

### DIFF
--- a/numba_cuda/numba/cuda/extending.py
+++ b/numba_cuda/numba/cuda/extending.py
@@ -5,7 +5,6 @@ Added for symmetry with the core API
 from numba.core.extending import intrinsic as _intrinsic
 from numba.cuda.models import register_model  # noqa: F401
 from numba.cuda import models  # noqa: F401
-from numba.cuda.cudaimpl import lower  # noqa: F401
 
 intrinsic = _intrinsic(target="cuda")
 
--- a/numba_cuda/numba/cuda/extending.py
+++ b/numba_cuda/numba/cuda/extending.py
@@ -38,7 +38,7 @@ def make_attribute_wrapper(typeclass, struct_attr, python_attr):
         model = data_model_manager.lookup(typ)
         if not isinstance(model, StructModel):
             raise TypeError(
-                f"make_struct_attribute_wrapper() needs a type with a StructModel, but got {model}"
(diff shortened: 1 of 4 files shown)


numba_cuda/numba/cuda/extending.py
```diff
@@ -3,5 +3,56 @@
 """
 
 from numba.core.extending import intrinsic as _intrinsic
+from numba.cuda.models import register_model  # noqa: F401
+from numba.cuda import models  # noqa: F401
 
 intrinsic = _intrinsic(target="cuda")
+
+
+def make_attribute_wrapper(typeclass, struct_attr, python_attr):
```
gmarkall, line 12: This could do with a comment mentioning that it's vendored from `numba.core.extending` with a change to consider the CUDA data model manager, for the benefit of future readers of the code outside of the context of the PR.

numba_cuda/numba/cuda/tests/cudapy/test_extending.py
```diff
@@ -35,15 +35,17 @@ def sum_intervals(i, j):
...
         models,
-        register_model,
         type_callable,
         typeof_impl,
     )
     from numba.core.typing.templates import AttributeTemplate
     from numba.cuda.cudadecl import registry as cuda_registry
     from numba.cuda.cudaimpl import lower_attr as cuda_lower_attr
+    from numba.cuda.extending import (
+        register_model,
```
ZzEeKkAa, line 46: @gmarkall I've update this test by importing `register_model` from `numba.cuda.extending`, so later in this file `make_attribute_wrapper` works on the model registered only in `numba.cuda`. However I agree, that we need to add a test that `make_attribute_wrapper` is called on the struct that uses types both from cuda dmm and default dmm

ZzEeKkAa: Okay, it has `types.float64`, that uses default dmm. So no need for extra tests.

gmarkall (approved):

PR #195, opened 2025-04-11
NRT Doc update

brandon-b-miller:
Starting a PR to make NRT features public through the docs given that some of the numpy API works now. Much to do here.


isVoid:
It comes to me that https://github.com/NVIDIA/numba-cuda/blob/main/docs/source/reference/envvars.rst hasn't document the two new envrionment variable that controls NRT: `NUMBA_CUDA_ENABLE_NRT` and `NUMBA_CUDA_NRT_STATS`.

PR #196, opened 2025-04-11
Report NVRTC builtin operation failures to the user

gmarkall:
If NVRTC could not load its builtins library, then the user would get an error message like:

```
numba.cuda.cudadrv.error.NvrtcError:
  Failed to call nvrtcCompileProgram: NVRTC_ERROR_BUILTIN_OPERATION_FAILURE
```

With this change, we specifically transform this error value into a corresponding exception, and provide the user with the log from NVRTC so they can more easily diagnose the issue.

They will now see an error like:

```
numba.cuda.cudadrv.error.NvrtcError: NVRTC Compilation failure whilst compiling memsys.cu:

nvrtc: error: failed to open libnvrtc-builtins.so.12.8.
  Make sure that libnvrtc-builtins.so.12.8 is installed correctly.
```

which clearly explains the issue.

### DIFF
--- a/numba_cuda/numba/cuda/cudadrv/error.py
+++ b/numba_cuda/numba/cuda/cudadrv/error.py
@@ -32,5 +32,9 @@ class NvrtcCompilationError(NvrtcError):
     pass
 
 
+class NvrtcBuiltinOperationFailure(NvrtcError):
+    pass
+
+
 class NvrtcSupportError(ImportError):
     pass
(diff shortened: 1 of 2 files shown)


gmarkall:
cc @jiel-nv - this should make it clear if you encounter this issue in future.

gmarkall:
I can't think of an easy way to test this. I tested it manually by renaming the `libnvrtc-builtins.so.12.8` file. I can't think of a test of this that doesn't also involve mocking / modifying the `NvrtcProgram` class so much I'm not sure we'd be meaningfully testing what's there.

PR #222, opened 2025-04-16
Local variable debug info deduplication

jiel-nv:
This PR improves the debug info for local variables.

1. Add a CUDA native lowering pass to the lowering pipeline, so that we can customize the numba cuda IR w/o affecting upstream numba.
2. Switch the llvm intrinsic call "llvm.dbg.declare" to "llvm.dbg.value" in order to provide more accurate value tracking.
3. Eliminate the user variable aliases (the .$1 variations) in DILocalVariable.
4. Merge DILocalVariable nodes with same name and type but different lines by using cache.
5. Merge the BooleanLiteral type into boolean type and merge the IntegerLiteral type into integer type.
6. Add three unit tests.

Fixes nvbug#5027648, nvbug#5009771, nvbug#5120628.


gmarkall:
@jiel-nv Many thanks for the PR! I'll add it to my queue to review, but note I'm OOO until next Tuesday, so will have to pick it up after then.

numba_cuda/numba/cuda/lowering.py
```diff
@@ -0,0 +1,41 @@
...
+
+        # Emit llvm.dbg.value instead of llvm.dbg.declare for local scalar
+        # variables immediately after a store instruction.
+        if (
+            self.context.enable_debuginfo
+            and (
+                name not in self._singly_assigned_vars
+                or self._disable_sroa_like_opt
+            )
+            and not name.startswith("$")
```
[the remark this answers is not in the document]
gmarkall: Ah, I see now that the first two conditions are used in the superclass `storevar()` and suppresses storing the variable when they are true. 

I think the last is because we don't emit debuginfo for internal names (those beginning with a `$`).

gmarkall, line 20: Perhaps the following clarifies these for other readers:
```suggestion
            # Conditions used to elide stores in parent method
            and (
                name not in self._singly_assigned_vars
                or self._disable_sroa_like_opt
            )
            # No emission of debuginfo for internal names
            and not name.startswith("$")
```

gmarkall (commented):
I think this generally looks good - there is one suggestion on the diff which I think will save time for future readers of the code (it took me quite a while to work out).

numba_cuda/numba/cuda/lowering.py
jiel-nv: Yes, I will apply the suggested comments.

gmarkall (approved):
Thanks for the update! I think this is good to merge now but will wait until after the CI fixes (#238) are done.

PR #227, opened 2025-04-18
Add Compiler timer to Dispatcher.compile and Dispatcher.compile_device

isVoid:
This PR adds a compile timers for cuda dispatcher compile paths.

### DIFF
--- a/numba_cuda/numba/cuda/compiler.py
+++ b/numba_cuda/numba/cuda/compiler.py
@@ -118,6 +118,10 @@ class CUDABackend(LoweringPass):
         """
         lowered = state["cr"]
         signature = typing.signature(state.return_type, *state.args)
+        if "metadata" in state:
+            md = state["metadata"]
+        else:
+            md = None
 
         state.cr = cuda_compile_result(
             typing_context=state.typingctx,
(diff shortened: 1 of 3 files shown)


numba_cuda/numba/cuda/compiler.py
```diff
@@ -118,6 +118,10 @@ def run_pass(self, state):
         """
         lowered = state["cr"]
         signature = typing.signature(state.return_type, *state.args)
+        if "metadata" in state:
```
gmarkall, line 121: Are there situations where `metadata` is not in `state`, or is this "just in case"?

isVoid: It's just in case, I reran the tests by accessing `.metadata` unconditionally and all passes.

isVoid: So I removed the check in [ff7852c](https://github.com/NVIDIA/numba-cuda/pull/227/commits/ff7852c87bbf3477c4ea4e5f3f5d5946a15ab92e)

numba_cuda/numba/cuda/tests/test_numba_events.py
gmarkall, line 1: I think this should be in the `cudapy` subdir as it's a test of things related to compilation. (there are no other test modules in the `numba.cuda.tests` module)

PR #230, opened 2025-04-22
Fix Invalid NVVM IR emitted when lowering shfl_sync APIs

### LINKED ISSUE #228: [BUG] Warp operations (`shfl_sync`, etc.) must use a constant int for the mode
**Describe the bug**

Our use of warp intrinsics presently provides non-constant `mode` arguments to the warp operation intrinsics. The `mode` parameter is required to be a constant int - see https://docs.nvidia.com/cuda/nvvm-ir-spec/index.html?highlight=shfl%2520sync#warp-level-operations

Instead, it is used like:

```ll
%".6" = trunc i64 0 to i32
; ...
%".12" = call {i32, i1} @"llvm.nvvm.shfl.sync.i32"(i32 %".7", i32 %".6", i32 %".9", i32 %"arg.src_lane", i32 %".8")
```

Where the `mode` parameter in the call is `%".6" - it is a constant truncated, not a constant int.

This happens when all warp operations are used (those in `intrinsic_wrapper.py`).

**Expected behavior**

We need to use a constant int for the mode, like

```ll
%".12" = call {i32, i1} @"llvm.nvvm.shfl.sync.i32"(i32 %".7", i32 0, i32 %".9", i32 %"arg.src_lane", i32 %".8")
```

All environments.

isVoid:
Per nvvm documentation, shuffle APIs takes a IR constant for the `mode` parameter. In current Numba implementation, it is a variable. This could crash the NVVM lowering when the ConstantFolds are not applied to IR (-opt=3 e.g.). This PR fixes #228

### DIFF
--- a/numba_cuda/numba/cuda/cudadecl.py
+++ b/numba_cuda/numba/cuda/cudadecl.py
@@ -100,43 +100,43 @@ class Cuda_syncwarp(ConcreteTemplate):
     cases = [signature(types.none), signature(types.none, types.i4)]
 
 
-@register
-class Cuda_shfl_sync_intrinsic(ConcreteTemplate):
-    key = cuda.shfl_sync_intrinsic
-    cases = [
-        signature(
-            types.Tuple((types.i4, types.b1)),
-            types.i4,
-            types.i4,
-            types.i4,
-            types.i4,
-            types.i4,
-        ),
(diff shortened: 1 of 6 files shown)


isVoid:
Superceded by #231

PR #231, opened 2025-04-22
Fix Invalid NVVM IR emitted when lowering shfl_sync APIs

isVoid:
Per nvvm documentation, shuffle APIs takes a IR constant for the mode parameter. In current Numba implementation, it is a variable. This could crash the NVVM lowering when the ConstantFolds are not applied to IR (-opt=3 e.g.). This PR fixes https://github.com/NVIDIA/numba-cuda/issues/228


gmarkall (changes requested):
Thanks for moving this PR forward and fixing the correctness issues in it.

When I tried to follow through the implementation to understand what's going on and correlate it with the NVVM IR and CUDA C/C++ programming documentation, I found it very hard to follow what was going on, because the implementation diverges from both in a lot of small ways (because it always did in Numba, not as a result of this PR).

There was also some redundant work with creating structs and extracting values from them that was not quite redundant in the old implementation. Now that the implementation is all combined into a single intrinsic instead of the old way of having typing, lowering, a stub, and jitted functions, the extra packing and unpacking is un-needed.

To get to a point where I could follow through what was going on I found I had to iterate through modifying the code to align it with the docs and C/C++ implementation, and get rid of the extra packing and unpacking - this is in https://github.com/gmarkall/numba-cuda/commit/931f865948d8315a91bca2ab37c2b320a648f3f0

A couple of other notes from whilst I was doing that:
- I kept making mistakes that involved `test_shfl_sync_types` test failing, so it was helpful to use subTests for it: https://github.com/gmarkall/numba-cuda/commit/de76f697f3a87adaef55e1d14303a77d71e4b187
- The docstrings for the intrinsics were missing, so they wouldn't have been visible in an IDE or shown up in the documentation.

I suspect we will also have a similar issue for the vote sync intrinsic, which is documented as requiring a constant `mode` as well: https://docs.nvidia.com/cuda/nvvm-ir-spec/index.html#vote - we should get this PR in and then repeat the process for that intrinsic (and maybe check if there are other intrinsics we use where we make this mistake too).

numba_cuda/numba/cuda/tests/cudapy/test_warp_ops.py
```diff
@@ -144,6 +146,33 @@ def test_shfl_sync_xor(self):
         compiled[1, nelem](ary, xor)
         self.assertTrue(np.all(ary == exp))
 
+    def test_shfl_sync_const_mode_val(self):
```
gmarkall, line 149: We should also have a variant of this test that uses `compile_ptx()` with `cc=(10, 0)` to ensure that all is well when going through the path that uses LLVM 18.x IR.

numba_cuda/numba/cuda/tests/cudapy/test_warp_ops.py
```diff
@@ -144,6 +146,33 @@ def test_shfl_sync_xor(self):
...
+            (use_shfl_sync_up, 4),
+            (use_shfl_sync_down, 4),
+            (use_shfl_sync_xor, 16),
+        ]
+
+        args_re = r"\((.*)\)"
+        m = re.compile(args_re)
+
+        for func, value in subtest:
+            compiled = cuda.jit("void(int32[:], int32)")(func)
```
gmarkall, line 163: Perhaps we could use `self.subTest()` here too?

numba_cuda/numba/cuda/intrinsics.py
```diff
@@ -205,3 +205,134 @@ def syncthreads_or(typingctx, predicate):
...
+            rv2 = builder.extract_value(ret2, 0)
+            pred = builder.extract_value(ret1, 1)
+            rv1_64 = builder.zext(rv1, ir.IntType(64))
+            rv2_64 = builder.zext(rv2, ir.IntType(64))
+            rv_shl = builder.shl(rv2_64, context.get_constant(types.i8, 32))
+            rv = builder.or_(rv_shl, rv1_64)
+            if value_type == types.float64:
+                rv = builder.bitcast(rv, ir.DoubleType())
+            ret = cgutils.make_anonymous_struct(builder, (rv, pred))
+        return builder.extract_value(ret, 0)
```
gmarkall, line 336: This is the creation / unpacking of a struct that is redundant that I was referring to in the main text of the review.

numba_cuda/numba/cuda/intrinsics.py
```diff
@@ -205,3 +205,134 @@ def syncthreads_or(typingctx, predicate):
...
+            value2 = builder.trunc(value_lshr, ir.IntType(32))
+            ret1 = builder.call(func, (mask, mode, value1, index, clamp))
+            ret2 = builder.call(func, (mask, mode, value2, index, clamp))
+            rv1 = builder.extract_value(ret1, 0)
+            rv2 = builder.extract_value(ret2, 0)
+            pred = builder.extract_value(ret1, 1)
+            rv1_64 = builder.zext(rv1, ir.IntType(64))
+            rv2_64 = builder.zext(rv2, ir.IntType(64))
+            rv_shl = builder.shl(rv2_64, context.get_constant(types.i8, 32))
+            rv = builder.or_(rv_shl, rv1_64)
```
gmarkall, line 332: What we call `rv` here maps to the `d` value in the pseudocode in the NVVM IR specification in https://docs.nvidia.com/cuda/nvvm-ir-spec/index.html#data-movement

numba_cuda/numba/cuda/intrinsics.py
```diff
@@ -205,3 +205,134 @@ def syncthreads_or(typingctx, predicate):
...
+            if value_type == types.float64:
+                rv = builder.bitcast(rv, ir.DoubleType())
```
gmarkall, line 334: I think this check is not necessary and we can always just bitcast to the expected return type.

gmarkall: (which is done in one of the commits linked to in the main review text)

isVoid:
> The docstrings for the intrinsics were missing, so they wouldn't have been visible in an IDE or shown up in the documentation.

Can you elaborate? The code is directly documented here: https://github.com/NVIDIA/numba-cuda/blob/main/docs/source/reference/kernel.rst#warp-intrinsics. It should show up when the API is accessible from `numba.cuda`?

numba_cuda/numba/cuda/tests/cudapy/test_warp_ops.py
isVoid: [b359f9a](https://github.com/NVIDIA/numba-cuda/pull/231/commits/b359f9a78cba7d8ef00512920f56d5d74f0343a6)

numba_cuda/numba/cuda/intrinsics.py
```diff
@@ -205,3 +205,174 @@ def syncthreads_or(typingctx, predicate):
...
+    return shfl_sync_intrinsic(
+        typingctx, membermask_type, mode_value, a_type, b_type, c_value
+    )
+
+
+def shfl_sync_intrinsic(
+    typingctx,
+    membermask_type,
+    mode_value,
+    a_type,
```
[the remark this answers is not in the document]
gmarkall: Good point - there's no need to get it again from the signature, so I think that definition later of it can be removed - see below. What do you think?

numba_cuda/numba/cuda/tests/cudapy/test_warp_ops.py
```diff
@@ -144,6 +147,47 @@ def test_shfl_sync_xor(self):
...
+        subtest = [
+            use_shfl_sync_idx,
+            use_shfl_sync_up,
+            use_shfl_sync_down,
+            use_shfl_sync_xor,
+        ]
+
+        for func in subtest:
+            with self.subTest(func=func.__name__):
+                compile_ptx(func, (int32[:], int32), cc=(10, 0))
```
gmarkall, line 189: I'm a bit surprised this passed on CI on toolkits that pre-date CC 10.0. I'm looking into what's going on here now, but do you know why this didn't fail on older toolkits?

gmarkall: OK, I now recall it's a long-standing behaviour that if we ask for a higher CC than supported, we just use the highest supported CC (because PTX for an older CC can be compiled for hardware of a newer CC). 

https://github.com/NVIDIA/numba-cuda/blob/9c727b0cb66e0a4b907cc27fe0da206e745436ef/numba_cuda/numba/cuda/cudadrv/nvvm.py#L478-L479

gmarkall (approved):
I'm happy with this if you are - whether you want to remove the second assignment of `a_type` marked on the diff or to proceed with it as-is.

numba_cuda/numba/cuda/intrinsics.py
```diff
@@ -205,3 +205,174 @@ def syncthreads_or(typingctx, predicate):
...
+        """
+        The NVVM shfl_sync intrinsic only supports i32, but the CUDA C/C++
+        intrinsic supports both 32- and 64-bit ints and floats, so for feature
+        parity, i32, i64, f32, and f64 are implemented. Floats by way of
+        bitcasting the float to an int, then shuffling, then bitcasting
+        back."""
+        membermask, a, b = args
+
+        # Types
+        a_type = sig.args[1]
```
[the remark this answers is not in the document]
gmarkall: Further to our discussion earlier where you suggested not doing this so that it's not captured from the outer function - I'll not commit this suggestion, and merge the PR as-is.

Issue #233, opened 2025-04-28
[BUG] TypeError on numba.cuda.is_available() with no CUDA installed


kenblu24:
**Describe the bug**
`numba.cuda.is_available()` gives `TypeError` with no CUDA drivers installed or paths set

**Steps/Code to reproduce bug**
1. Create a virtualenv and activate it
2. Install `uv pip install numba-cuda==0.9.0`
3. Run `python -c "import numba.cuda; print(numba.cuda.is_available())"`

If system does not have cuda available, I get this traceback:

Machine 1 with no CUDA
---
```pycon
Python 3.11.0 (main, Oct 24 2022, 18:26:48) [MSC v.1933 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from superneuromat import SNN
  File "<stdin>", line 1, in <module>
  File "C:\Users\kenblu24\snm\Lib\site-packages\superneuromat\__init__.py", line 1, in <module>
    from .neuromorphicmodel import SNN
  File "C:\Users\kenblu24\snm\Lib\site-packages\superneuromat\neuromorphicmodel.py", line 24, in <module>
    GPU_AVAILABLE = numba and cuda.is_available()
                              ^^^^^^^^^^^^^^^^^^^
  File "C:\Users\kenblu24\snm\Lib\site-packages\numba_cuda\numba\cuda\device_init.py", line 115, in is_available
    return driver_is_available and nvvm.is_available()
                                   ^^^^^^^^^^^^^^^^^^^
  File "C:\Users\kenblu24\snm\Lib\site-packages\numba_cuda\numba\cuda\cudadrv\nvvm.py", line 69, in is_available
    NVVM()
  File "C:\Users\kenblu24\snm\Lib\site-packages\numba_cuda\numba\cuda\cudadrv\nvvm.py", line 158, in __new__
    inst.driver = open_cudalib("nvvm")
                  ^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\kenblu24\snm\Lib\site-packages\numba_cuda\numba\cuda\cudadrv\libs.py", line 83, in open_cudalib
    path = get_cudalib(lib)
           ^^^^^^^^^^^^^^^^
  File "C:\Users\kenblu24\snm\Lib\site-packages\numba_cuda\numba\cuda\cudadrv\libs.py", line 54, in get_cudalib
    return get_cuda_paths()[lib].info or _dllnamepattern % lib
[...]

PR #234, opened 2025-04-29
Fix `cuda.is_available` when cuda libs aren't present

brandon-b-miller:
Closes https://github.com/NVIDIA/numba-cuda/issues/233.

This could use some kind of test, perhaps one that patches out the finder functions to return None just for the duration of that test.

### DIFF
--- a/numba_cuda/numba/cuda/cuda_paths.py
+++ b/numba_cuda/numba/cuda/cuda_paths.py
@@ -201,6 +201,8 @@ def _get_nvrtc_wheel():
 
 def _get_libdevice_paths():
     by, libdir = _get_libdevice_path_decision()
+    if not libdir:
+        return _env_path_tuple(by, None)
     out = os.path.join(libdir, "libdevice.10.bc")
     return _env_path_tuple(by, out)
 


kkraus14 (approved):

leofang:
cc @rwgk for vis

kkraus14:
> This could use some kind of test, perhaps one that patches out the finder functions to return None just for the duration of that test.

Going to merge this without the test. This will be replaced by the `path_finder` work @rwgk is doing anyway in the near-ish future and that will be more rigorously documented and tested.

brandon-b-miller:
> > This could use some kind of test, perhaps one that patches out the finder functions to return None just for the duration of that test.
> 
> Going to merge this without the test. This will be replaced by the `path_finder` work @rwgk is doing anyway in the near-ish future and that will be more rigorously documented and tested.

Got it- thanks, @kkraus14 .

rwgk:
> This could use some kind of test, perhaps one that patches out the finder functions to return None just for the duration of that test.

I'd use mocking only as a last resort in this context. I've seen too many ways that can go horribly wrong. (One way I think about it: often creates about as many problems as you're hoping to avoid, if you're lucky.)

Systematic testing against real environments is much better, and I think it's feasible (for the path_finder). (Hopefully not famous last words.)

PR #236, opened 2025-05-01
Add CUDAArray type and implementation

ZzEeKkAa:

### DIFF
--- a/numba_cuda/numba/cuda/types.py
+++ b/numba_cuda/numba/cuda/types.py
@@ -1,4 +1,5 @@
 from numba.core import types
+from numba.cuda.cudadrv import nvvm
 
 
 class Dim3(types.Type):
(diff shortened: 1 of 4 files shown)


numba_cuda/numba/cuda/types.py
```diff
@@ -38,3 +39,83 @@ class CUDADispatcher(types.Dispatcher):
...
+        super().__init__(dtype, ndim, layout, readonly, name, aligned)
+        self.addrspace = addrspace
+
+        if name is None:
+            type_name = "array"
+            if not self.mutable:
+                type_name = "readonly " + type_name
+            if not self.aligned:
+                type_name = "unaligned " + type_name
+            self.name = "%s(%s, %sd, %s, addrspace(%d))" % (
```
gmarkall, line 64: As the address spaces in nvvm.py are just integers, might it be worth converting them to be an enum class so that it's easier to get the array type to print like

```python
array(int64, 1, 'C', SHARED)
```

instead of

```python
array(int64, 1, 'C', addrspace(3))
```
?

It might make interactive debug / development a little easier without having to mentally translate the address space numbers to names - there aren't to many uses of the address spaces so I would hope that updating the uses (if necessary) wouldn't be too burdensome.

numba_cuda/numba/cuda/types.py
```diff
@@ -38,3 +39,83 @@ class CUDADispatcher(types.Dispatcher):
     # is still probably a good idea to have a separate type for CUDA
     # dispatchers, and this type might get other differentiation from the CPU
     # dispatcher type in future.
+
+
+class CUDAArray(types.Array):
```
gmarkall, line 44: Might the `mangling_args` property need implementing as well? Two methods that differ only in the address space of an array could end up mangling to the same name and potentially creating a symbol clash.

For example:

```python
from numba.core.itanium_mangler import mangle_type
from numba.cuda.types import CUDAArray
from numba import types

shared_array = CUDAArray(types.int64, 1, 'C', addrspace=3)
generic_array = CUDAArray(types.int64, 1, 'C', addrspace=0)

shared_mangled = mangle_type(shared_array)
generic_mangled = mangle_type(generic_array)

print(shared_mangled)
print(generic_mangled)
assert shared_mangled != generic_mangled
```

gives

```
9CUDAArrayIxLi1E1C7mutable7alignedE
9CUDAArrayIxLi1E1C7mutable7alignedE
  File "/home/gmarkall/numbadev/issues/numba-cuda-236/mangle_test.py", line 13, in <module>
    assert shared_mangled != generic_mangled
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -848,7 +848,12 @@ def typeof_pyval(self, val):
         # Based on _DispatcherBase.typeof_pyval, but differs from it to support
         # the CUDA Array Interface.
         try:
-            return typeof(val, Purpose.argument)
+            tp = typeof(val, Purpose.argument)
```
gmarkall, line 851: I'm concerned this could have a non-trivial impact on kernel launch time. Can you do a microbenchmark to check how much this impacts the latency of launches with various numbers of array arguments?

numba_cuda/numba/cuda/types.py
gmarkall: Similarly, what about the unification (the `unify()` method) and conversion (`can_convert_to()`)? If `unify()` is not implemented, then all CUDA arrays will end up unifying to `Array` types instead, even if the set of types to unify were all in the same address space.

Conversions will also lose address space information, or perhaps even allow invalid conversions - I think we should not allow conversion from shared to local address space, for example, but conversions to the generic address space should always be OK.

gmarkall (commented):
I haven't fully reviewed this yet but want to discuss thoughts so far - some of them are marked on the diff.

I think the main design concern I have at the moment is with trying to ensure that all array types in kernels are a `CUDAArray` type instead of an `Array` type - I think this might impact launch latency and have a lot of edge cases we need to find. Is an alternative path to keep `Array` types coexisting with `CUDAArray` types in kernels, but treat `Array` types as being in the generic address space? The idea here is to leave the decorator and dispatcher logic unchanged so we don't have to try and make `CUDAArray` types in the critical path of a launch.

gmarkall:
> all arrays inside cuda.jit are now using CUDAArray instead of types.Array. That breaks some api like requesting implementation with general purpose array signature. Thats why many tests had to be updated;

I think this will break user code too - if we can find a way to avoid doing that I'd strongly prefer to.

ZzEeKkAa:
> > all arrays inside cuda.jit are now using CUDAArray instead of types.Array. That breaks some api like requesting implementation with general purpose array signature. Thats why many tests had to be updated;
> 
> I think this will break user code too - if we can find a way to avoid doing that I'd strongly prefer to.

Yes, that's what I'm worry about. Still thinking about potential ways to avoid it

PR #238, opened 2025-05-02
Fix package installation for wheels CI 

gmarkall:
Normal wheel tests should not install any CUDA pip packages because they should use the toolkit from the container.

The test depending on the toolkit wheels should install the specific toolkit version used for the test (as opposed to just using the latest version, which they were set up to do before).


gmarkall (commented):
As the CI is passing again, and test-wheel-deps-wheels is now getting the toolkit version specified from the environment, I'm going to merge this to proceed with the release.

PR #240, opened 2025-05-05
Allow External Code to Use Coorperative Group

isVoid:
Currently, Numba-CUDA detects the use of cudaCGIntrinsicHandle in PTX and auto launch the kernel with `cuLaunchKernelCooperative`. This only works when the Numba kernel itself uses the internal typing extensions which gaurentees the use of this call. For external device functions that invokes CG intrinsics, they don't have a way to pass in info to inform Numba to cooperative launch kernel. This PR adds that support.

- [x] Add tests.


isVoid:
Note to self: see https://github.com/NVIDIA/numba-cuda/pull/243#discussion_r2080402679, the `CUDACodelibrary.use_coorperative` flag should be carried over from one codelib to another.

numba_cuda/numba/cuda/codegen.py
gmarkall, line 1: I think the `_reduce_states` and `_rebuild` method should also preserve the value of `self.use_cooperative`. It seems not to affect anything for this PR, but it would be good to keep it in line with the added property so that code using CGs can be correctly pickled. (I'm a little surprised the existing test of caching functions with CGs did not pick up on this)

numba_cuda/numba/cuda/tests/cudapy/test_cooperative_groups.py
```diff
@@ -149,6 +157,32 @@ def test_max_cooperative_grid_blocks(self):
         self.assertEqual(blocks1d, blocks2d)
         self.assertEqual(blocks1d, blocks3d)
 
+    @skip_unless_cc_60
+    def test_external_coorperative_func(self):
```
gmarkall, line 161:
```suggestion
    def test_external_cooperative_func(self):
```

numba_cuda/numba/cuda/tests/cudapy/test_cooperative_groups.py
```diff
@@ -149,6 +157,32 @@ def test_max_cooperative_grid_blocks(self):
...
+        cta_barrier = cuda.declare_device(
+            "cta_barrier", sig=sig, use_cooperative=True
+        )
+
+        cudapy_test_path = os.path.dirname(__file__)
+        tests_path = os.path.dirname(cudapy_test_path)
+        data_path = os.path.join(tests_path, "data")
+        src = os.path.join(data_path, "cta_barrier.cu")
+
+        @cuda.jit(link=[src])
```
gmarkall, line 174: Would it be more representative of the way we hope APIs will be used to put the `link` data in the `cta_barrier` declaration above?

numba_cuda/numba/cuda/tests/cudapy/test_cooperative_groups.py
```diff
@@ -149,6 +157,32 @@ def test_max_cooperative_grid_blocks(self):
         self.assertEqual(blocks1d, blocks2d)
         self.assertEqual(blocks1d, blocks3d)
 
+    @skip_unless_cc_60
+    def test_external_coorperative_func(self):
+        sig = signature(
+            CPointer(int32),
+        )
+        cta_barrier = cuda.declare_device(
+            "cta_barrier", sig=sig, use_cooperative=True
```
gmarkall, line 166: The general idea I'm getting at with the comment below (will not have correct formatting or be in the right place to use the definition of `src`, I suspect):

```suggestion
            "cta_barrier", sig=sig, link=[src], use_cooperative=True
```

gmarkall (changes requested):
The overall picture looks good - some minor comments on the diff.

numba_cuda/numba/cuda/codegen.py
```diff
@@ -70,6 +70,8 @@ def __init__(self, codegen, name):
         self._setup_functions = []
         self._teardown_functions = []
 
+        self.use_coorperative = False
```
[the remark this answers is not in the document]
isVoid: How-come this doesn't fail any tests?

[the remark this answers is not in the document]
isVoid: Ah - when we construct an `ExternalFunction` object, we don't have a test that test the default option of this flag is set to `False`, instead, we depend on the argument of `declare_device`.

numba_cuda/numba/cuda/tests/cudapy/test_cooperative_groups.py
isVoid: Yup, will address together with the comment below.

isVoid: Done in 2efcc47

numba_cuda/numba/cuda/codegen.py
isVoid: I think unfortunately there's a bug in cache test... 
`CachingTestUsesCases` rely on `_TestModule.check_module` to execute in a separate process to verify the on disk caching. However this method is really out of sync with the main `test_caching` tests. Thinking of a fix.

isVoid: Spent more time digging into the caching module, I think the module can take some more love in improvement. For example `test_cache_cg_clean_run` is vending code from `run_in_separate_process`, which can effectively be rewritten as a separate test class with a unique set of test use case file and `modname`. Additionally, some tests was not invoked with test in separate process.

Setting that aside, I think the reason why this wasn't caught in existing caching test is that the `_Kernel` module caches the `cooperative` attribute, which is what governs the final launch method. There maybe a hidden incoherence between codelib and the `_Kernel` object, but was not tracked in any tests.

isVoid: Addressed in [8e9c0f7](https://github.com/NVIDIA/numba-cuda/pull/240/commits/8e9c0f7ac7338d96150042099c72d253118117a0)

Also in [fab0cb7](https://github.com/NVIDIA/numba-cuda/pull/240/commits/fab0cb701bf10cc83fd53d03d51b389cebb1acd0), I partially optimized the cg caching tests since they are the most relavant improvement here. Punting the rest to a future PR.

numba_cuda/numba/cuda/tests/cudapy/test_caching.py
```diff
@@ -275,6 +224,37 @@ def f():
                 pass
 
 
+@skip_on_cudasim("Simulator does not implement caching")
+class CUDACooperativeGroupTest(SerialMixin, DispatcherCacheUsecasesTest):
```
gmarkall, line 228: I think we need to preserve the comment about the original issue - I can imagine myself looking at this in future and not realising that we're specifically ensuring that caching works when function using CG is loaded from the cache prior to any other compilation:
```suggestion
class CUDACooperativeGroupTest(SerialMixin, DispatcherCacheUsecasesTest):
    # See Issue #9432: https://github.com/numba/numba/issues/9432
    # If a cached function using CG sync was the first thing to compile,
    # the compile would fail.
```

gmarkall (approved):

PR #241, opened 2025-05-06
Fixes for adding data to link

gmarkall:
- The kind of LTO-IR should have been based on the file extension map, not the string `"ltoir"`.
- The `add-data()` method was unimplemented for the Ctypes and CUDA Python linkers. I have left it unimplemented in the `MVCLinker` as the functionality is not needed immediately in this linkers, and it will eventually be removed.
- Add a test of adding data to the link via the active code library for all kinds of linkable code.
- Test binaries are now built for the conda package and wheel tests, because they are also used in the testsuite even when pynvjitlink is not in use.

### DIFF
--- a/ci/test_wheel.sh
+++ b/ci/test_wheel.sh
@@ -8,7 +8,7 @@ rapids-logger "Install testing dependencies"
 python -m pip install \
     psutil \
     cffi \
-    "cuda-python=11.8.*" \
+    "cuda-python==11.8.*" \
     pytest
 
 
--- a/ci/test_wheel.sh
+++ b/ci/test_wheel.sh
@@ -10,7 +10,7 @@ rapids-logger "Install testing dependencies"
 python -m pip install \
     psutil \
     cffi \
-    "cuda-python==${CUDA_VER_MAJOR_MINOR}.*" \
+    "cuda-python==${CUDA_VER_MAJOR_MINOR%.*}.*" \
     pytest
 
 
--- a/ci/test_wheel.sh
+++ b/ci/test_wheel.sh
@@ -3,12 +3,14 @@
 
 set -euo pipefail
 
+CUDA_VER_MAJOR_MINOR=${CUDA_VER%.*}
+
 rapids-logger "Install testing dependencies"
 # TODO: Replace with rapids-dependency-file-generator
 python -m pip install \
     psutil \
     cffi \
-    "cuda-python==11.8.*" \
+    "cuda-python==${CUDA_VER_MAJOR_MINOR}.*" \
     pytest
 
 
@@ -19,7 +21,6 @@ test_dir = root + \"numba/cuda/tests/test_binary_generation/\"
 print(test_dir)
 "
 
-CUDA_VER_MAJOR_MINOR=${CUDA_VER%.*}
 if [ "${CUDA_VER_MAJOR_MINOR%.*}" == "11" ]
 then
   rapids-logger "Skipping test build for CUDA 11"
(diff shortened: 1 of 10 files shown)


numba_cuda/numba/cuda/tests/cudapy/test_extending.py
```diff
@@ -160,5 +162,89 @@ def f(r, x):
...
+                        return types.uint32
+
+                return typer
+
+            @lower_builtin(external_add, types.uint32, types.uint32)
+            def lower_external_add(context, builder, sig, args):
+                context.active_code_library.add_linking_file(code_object)
+                i32 = ir.IntType(32)
+                fnty = ir.FunctionType(i32, [i32, i32])
+                fn = cgutils.get_or_insert_function(
```
brandon-b-miller, line 231: TIL

brandon-b-miller:
A couple of the cuda 11 jobs are failing with 
```
RuntimeError: Failed to dlopen libnvrtc.so.12
```
This could be related to the fact that the wrong version of the cuda bindings is coming in somehow:

```
Collecting cuda-bindings~=12.8.0 (from cuda-python)
```

numba_cuda/numba/cuda/tests/cudapy/test_extending.py
```diff
@@ -160,5 +162,89 @@ def f(r, x):
...
+    def test_extension_adds_linkable_code(self):
+        files = (
+            (test_device_functions_a, cuda.Archive),
+            (test_device_functions_cubin, cuda.Cubin),
+            (test_device_functions_cu, cuda.CUSource),
+            (test_device_functions_fatbin, cuda.Fatbin),
+            (test_device_functions_o, cuda.Object),
+            (test_device_functions_ptx, cuda.PTXSource),
+            (test_device_functions_ltoir, cuda.LTOIR),
+        )
```
ZzEeKkAa, line 203: Side question, do we want to add `pytest` to `numba-cuda`?

gmarkall: Yes - I haven't had time to do anything about it though.

ZzEeKkAa: I can help with that next week

ci/test_wheel.sh
```diff
@@ -8,9 +8,31 @@ rapids-logger "Install testing dependencies"
 python -m pip install \
     psutil \
     cffi \
-    cuda-python \
+    "cuda-python=11.8.*" \
```
brandon-b-miller, line 11:
```suggestion
    "cuda-python==11.8.*" \
```

ci/test_wheel.sh
```diff
@@ -3,14 +3,37 @@
...
 
+CUDA_VER_MAJOR_MINOR=${CUDA_VER%.*}
+
 rapids-logger "Install testing dependencies"
 # TODO: Replace with rapids-dependency-file-generator
 python -m pip install \
     psutil \
     cffi \
-    cuda-python \
+    "cuda-python==${CUDA_VER_MAJOR_MINOR}.*" \
```
brandon-b-miller, line 13: I think this may have to be just the major version or parsed such that we get `12.*`
```
ERROR: No matching distribution found for cuda-python==12.2.*
```

Would something like this work?

```suggestion
    "cuda-python==${CUDA_VER%%.*}.*" \
```

gmarkall: I thought the problem was that there aren't old cuda-python versions for these newer Python versions, so I was looking at redistributing the test matrices so that the version constraints can be satisfied.

bdice: Please use `12.*` or equivalent here. It is recommended to use the newest cuda-python even with older CTK versions. There is only a major (not minor) `cuda-version` constraint for `cuda-python`.

brandon-b-miller (approved):

PR #242, opened 2025-05-06
Improve debug info for kernel arguments

jiel-nv:
In order to leverage more accurate debug info tracking from the nvvm compiler, kernel argument name need to be exact matching the name from the metadata, which is using source level user symbol.

(1) Change the CUDACallConv on CUDA target to not prefix "arg." on argument name;
(2) Call CUDA DIBuilder to build up customized DISubroutineType which is different than the parent method;
(3) Add a debug info unit test to guard this change.

This fixes nvbug#5027369.

### DIFF
--- a/numba_cuda/numba/cuda/debuginfo.py
+++ b/numba_cuda/numba/cuda/debuginfo.py
@@ -81,9 +81,9 @@ class CUDADIBuilder(DIBuilder):
 
         return self.module.add_debug_info(
             "DISubroutineType",
-             {
+            {
                 "types": self.module.add_metadata(md),
-             },
+            },
         )
 
     def mark_variable(
--- a/numba_cuda/numba/cuda/debuginfo.py
+++ b/numba_cuda/numba/cuda/debuginfo.py
@@ -59,6 +59,30 @@ class CUDADIBuilder(DIBuilder):
         # For other cases, use upstream Numba implementation
         return super()._var_type(lltype, size, datamodel=datamodel)
 
+    def _di_subroutine_type(self, line, function, argmap):
+        # The function call conv needs encoding.
+        llfunc = function
+        md = []
+
+        # Create metadata type for return value
+        if len(llfunc.args) > 0:
+            lltype = llfunc.args[0].type
+            size = self.cgctx.get_abi_sizeof(lltype)
+            mdtype = self._var_type(lltype, size, datamodel=None)
+            md.append(mdtype)
+
+        # Create metadata type for arguments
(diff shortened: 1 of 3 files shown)


numba_cuda/numba/cuda/target.py
```diff
@@ -290,7 +290,16 @@ def get_ufunc_info(self, ufunc_key):
...
 
 class CUDACallConv(MinimalCallConv):
-    pass
+    def decorate_function(self, fn, args, fe_argtypes, noalias=False):
+        """
+        Set names and attributes of function arguments.
+        """
+        assert not noalias
+        arginfo = self._get_arg_packer(fe_argtypes)
+        # Do not prefix "arg." on argument name, so that nvvm compiler
```
gmarkall, line 299: I *think* that it's OK not to prefix `arg.` onto the name - as far as I can tell it's a convention that was added that has not become load-bearing in any way - I don't think a duplicate name will appear inside a function body.

numba_cuda/numba/cuda/debuginfo.py
```diff
@@ -59,6 +59,33 @@ def _var_type(self, lltype, size, datamodel=None):
...
         return super()._var_type(lltype, size, datamodel=datamodel)
 
+    def _di_subroutine_type(self, line, function, argmap):
+        # The function call conv needs encoding.
+        llfunc = function
+        md = []
+
+        # Create metadata type for return value
+        if len(llfunc.args) > 0:
+            lltype = llfunc.args[0].type
```
gmarkall, line 69: Does omitting the replacement of `.` with `$` compared to the original implementation not cause issues for GDB? I note the comment in the original version:

```python
name = llarg.name.replace('.', '$')    # for gdb to work correctly
```

I thought that a dot in a name might be problematic for it? (c.f. the same syntax for traversing the members of a struct or union)

gmarkall (commented):
I think this looks good in general - I do have one question on the diff, and I'd like to understand if / why it's OK.

numba_cuda/numba/cuda/debuginfo.py
jiel-nv: That `.` versioned name used to be written to the metadata node as the value of 'name' of DILocalVariable. But currently the `.` versioned variable nodes are merged to the unversioned / user symbol named metadata node. Therefore, they are no longer to be present in DWARF entry.
And when represented as member of a struct or union, their names in the metadata are also unversioned names with corresponding types, so there won't be issues omitting this original replacement.
In short, neither `.` nor `$` will show up in debugger.

gmarkall (approved):

PR #243, opened 2025-05-08
Fixes for linking external code

gmarkall:
The issue fixed by this PR is that in general, when a called function makes use of external code (e.g. through a `LinkableCode` object), the required external code may not have been added to the link. The mechanism for constructing the list of items to link, which was to generate all the code then try to figure out what to link after the fact in the `Dispatcher` (using the `get_cres_link_objects()` function) was broken, and possibly fundamentally flawed.

The approach here is to instead keep track of all linked objects through code libraries - whenever code libraries are linked, the files they link are added to the link for the current code library.

Some refactoring and documentation of  `declare_device_function()` is added. This is aimed at making it more understandable; prior to this PR I didn't really quite know how or why typing and lowering for external functions worked.

Additional tests for calling external code through a device function and an overload are added.

More details are in individual commit messages (the commits are incremental / self-contained) - see them for further details.

### DIFF
--- a/numba_cuda/numba/cuda/codegen.py
+++ b/numba_cuda/numba/cuda/codegen.py
@@ -58,6 +58,57 @@ def disassemble_cubin_for_cfg(cubin):
     return run_nvdisasm(cubin, flags)
 
 
+class ExternalCodeLibrary(CodeLibrary):
+    """Holds code produced externally, for linking with generated code."""
+
+    def __init__(self, codegen, name):
+        super().__init__(codegen, name)
+        # Files to link
+        self._linking_files = set()
+        # Setup and teardown functions for the module.
+        # The order is determined by the order they are added to the codelib.
+        self._setup_functions = []
+        self._teardown_functions = []
+
+    @property
+    def modules(self):
+        # There are no LLVM IR modules in an ExternalCodeLibrary
+        return set()
+
+    def add_linking_file(self, path_or_obj):
+        # Adding new files after finalization is prohibited, in case the list
+        # of libraries has already been added to another code library; the
+        # newly-added files would be omitted from their linking process.
+        self._raise_if_finalized()
+
+        if isinstance(path_or_obj, LinkableCode):
+            if path_or_obj.setup_callback:
+                self._setup_functions.append(path_or_obj.setup_callback)
+            if path_or_obj.teardown_callback:
+                self._teardown_functions.append(path_or_obj.teardown_callback)
+
+        self._linking_files.add(path_or_obj)
+
+    def add_ir_module(self, module):
+        raise NotImplementedError("Cannot add LLVM IR to external code")
+
+    def add_linking_library(self, library):
+        raise NotImplementedError("Cannot add libraries to external code")
+
+    def finalize(self):
+        self._raise_if_finalized()
+        self._finalized = True
+
+    def get_asm_str(self):
+        raise NotImplementedError("No assembly for external code")
+
+    def get_llvm_str(self):
+        raise NotImplementedError("No LLVM IR for external code")
+
+    def get_function(self, name):
+        raise NotImplementedError("Cannot get function from external code")
+
+
 class CUDACodeLibrary(serialize.ReduceMixin, CodeLibrary):
     """
     The CUDACodeLibrary generates PTX, SASS, cubins for multiple different
@@ -299,6 +350,8 @@ class CUDACodeLibrary(serialize.ReduceMixin, CodeLibrary):
 
         self._linking_libraries.add(library)
         self._linking_files.update(library._linking_files)
+        self._setup_functions.extend(library._setup_functions)
+        self._teardown_functions.extend(library._teardown_functions)
 
     def add_linking_file(self, path_or_obj):
         if isinstance(path_or_obj, LinkableCode):
--- a/numba_cuda/numba/cuda/codegen.py
+++ b/numba_cuda/numba/cuda/codegen.py
@@ -298,6 +298,7 @@ class CUDACodeLibrary(serialize.ReduceMixin, CodeLibrary):
         self._raise_if_finalized()
 
         self._linking_libraries.add(library)
+        self._linking_files.update(library._linking_files)
 
     def add_linking_file(self, path_or_obj):
         if isinstance(path_or_obj, LinkableCode):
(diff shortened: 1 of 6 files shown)


numba_cuda/numba/cuda/codegen.py
```diff
@@ -298,6 +349,9 @@ def add_linking_library(self, library):
         self._raise_if_finalized()
 
         self._linking_libraries.add(library)
+        self._linking_files.update(library._linking_files)
+        self._setup_functions.extend(library._setup_functions)
+        self._teardown_functions.extend(library._teardown_functions)
```
isVoid, line 354: I feel like this is somewhat underrated in the PR description. Code libraries from one nested dispatcher is added into another code library that belongs to top level dispatcher via `add_linking_library` call. Without this patch, much information from lower level library is lost.

numba_cuda/numba/cuda/codegen.py
```diff
@@ -58,6 +58,57 @@ def disassemble_cubin_for_cfg(cubin):
...
+        # newly-added files would be omitted from their linking process.
+        self._raise_if_finalized()
+
+        if isinstance(path_or_obj, LinkableCode):
+            if path_or_obj.setup_callback:
+                self._setup_functions.append(path_or_obj.setup_callback)
+            if path_or_obj.teardown_callback:
+                self._teardown_functions.append(path_or_obj.teardown_callback)
+
+        self._linking_files.add(path_or_obj)
```
isVoid, line 90: I feel the only nit I have for this PR is that this code is replicated in both `CUDACodelibrary` and `ExternalCodeLibrary`. I also see they share quite a few things in common and wonder if we could reduce duplication by sharing some logic in a common base class?

gmarkall: It is a good point - I'll see if I can clear things up a bit.

numba_cuda/numba/cuda/codegen.py
gmarkall: I think you're saying that the idea is that whenever we link to a code library, we now make sure to capture all the information from the linked code library - is my understanding correct?

The general idea of this PR was that we should always be keeping track of the required information for linking correctly, and not require any special treatment in the dispatcher. This does make me think that as a follow on, there should be an improvement to the `cuda.compile()` API to allow the linking dependencies to be determined - otherwise you can compile code that you have to know how to link. (I don't want to try to tackle this straight away, because this PR addresses a concrete deficiency that exists right now).

numba_cuda/numba/cuda/codegen.py
gmarkall: If I use `ExternalCodeLibrary` as a common base class, then the diff looks like this:

```diff
diff --git a/numba_cuda/numba/cuda/codegen.py b/numba_cuda/numba/cuda/codegen.py
index cb72c0d..12c4a95 100644
--- a/numba_cuda/numba/cuda/codegen.py
+++ b/numba_cuda/numba/cuda/codegen.py
@@ -109,11 +109,13 @@ class ExternalCodeLibrary(CodeLibrary):
         raise NotImplementedError("Cannot get function from external code")
 
 
-class CUDACodeLibrary(serialize.ReduceMixin, CodeLibrary):
+class CUDACodeLibrary(serialize.ReduceMixin, ExternalCodeLibrary):
     """
     The CUDACodeLibrary generates PTX, SASS, cubins for multiple different
     compute capabilities. It also loads cubins to multiple devices (via
     get_cufunc), which may be of different compute capabilities.
+
+    CUDACodeLibrary may also be linked to external code.
     """
 
     def __init__(
@@ -149,15 +151,6 @@ class CUDACodeLibrary(serialize.ReduceMixin, CodeLibrary):
         # IR from this module - in that sense they are "linked" by NVVM at PTX
         # generation time, rather than at link time.
         self._linking_libraries = set()
-        # Files to link with the generated PTX. These are linked using the
-        # Driver API at link time.
-        self._linking_files = set()
-        # List of setup functions to the loaded module
-        # the order is determined by the order they are added to the codelib.
-        self._setup_functions = []
-        # List of teardown functions to the loaded module
-        # the order is determined by the order they are added to the codelib.
-        self._teardown_functions = []
         # Should we link libcudadevrt?
         self.needs_cudadevrt = False
 
@@ -353,15 +346,6 @@ class CUDACodeLibrary(serialize.ReduceMixin, CodeLibrary):
         self._setup_functions.extend(library._setup_functions)
         self._teardown_functions.extend(library._teardown_functions)
 
-    def add_linking_file(self, path_or_obj):
-        if isinstance(path_or_obj, LinkableCode):
-            if path_or_obj.setup_callback:
-                self._setup_functions.append(path_or_obj.setup_callback)
-            if path_or_obj.teardown_callback:
-                self._teardown_functions.append(path_or_obj.teardown_callback)
-
-        self._linking_files.add(path_or_obj)
-
     def get_function(self, name):
         for fn in self._module.functions:
             if fn.name == name:
```

So they get to share the `add_linking_file()` method and three attributes (`_linking_files`, `_setup_functions()`, and `_teardown_functions()`). The `add_linking_file()` methods weren't quite exactly the same - `CUDACodeLibrary`'s version would allow linking files after finalization. I **think** that it should be OK to prevent that, but I'm not 100% sure if it would expose some latent bug or cause some other regression.

Given the above - do you still think it's worth it to try to reduce the level of duplication (given that it seems to not be a great deal, and most of it is in a method that does differ semantically between the two)?

gmarkall: Ah, nevermind - running the test suite locally hits:

```
FAIL: test_cannot_cache_linking_libraries (numba.cuda.tests.cudapy.test_caching.CUDACachingTest.test_cannot_cache_linking_libraries)
RuntimeError: operation impossible on finalized object <Library 'CUDACachingTest.test_cannot_cache_linking_libraries.<locals>.f' at 0x7004d6ec6f50>

During handling of the above exception, another exception occurred:

  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_caching.py", line 271, in test_cannot_cache_linking_libraries
    with self.assertRaisesRegex(RuntimeError, msg):
AssertionError: "Cannot pickle CUDACodeLibrary with linking files" does not match "operation impossible on finalized object <Library 'CUDACachingTest.test_cannot_cache_linking_libraries.<locals>.f' at 0x7004d6ec6f50>"
```

and tens of other issues with slightly different consequences, but the same root cause. So all they can really share is three lists, without much more substantial changes.

gmarkall:
@isVoid Thanks for the review - I've made a couple of responses to the comments - I am not sure I can see any changes I can make at present, but open to further suggestions.

isVoid (approved):

PR #244, opened 2025-05-08
Add Test to Use Bf16 Bindings inside Device Function

isVoid:
Currently there isn't a test for bfloat16 bindings inside device function, and its failing because LinkableCode from nested function calls are not carried into the final compile result. https://github.com/NVIDIA/numba-cuda/pull/243 is supposed to fix this feature gap.

Depends on https://github.com/NVIDIA/numba-cuda/pull/243


gmarkall:
#243 is merged - hopefully a merge from `main` should turn this green now?

gmarkall (approved):

PR #245, opened 2025-05-09
Add Bfloat16 High Level API, Documentation

isVoid:
Follow up to #166 , this PR adds high level binding to bfloat16 type and exposes them in documentations.


gmarkall:
Was this also waiting for #243?

isVoid:
> Was this also waiting for #243?

No it shouldn't be waiting.

numba_cuda/numba/cuda/api.py
```diff
@@ -95,6 +95,14 @@ def is_float16_supported():
     return True
 
 
+def is_bfloat16_supported():
+    """Whether bfloat16 are supported.
+
+    bfloat16 are only supported on devices with compute capability >= 8.0
+    """
+    return current_context().device.supports_bfloat16
```
gmarkall, line 103: Do we want to check for CUDA 12 here as well?

isVoid: Does it look a bit strange that `Device.supports_bfloat16` and `api.is_bfloat16_supported` has different semantics?

numba_cuda/numba/cuda/api.py
```diff
@@ -10,9 +10,11 @@
 from .cudadrv import devicearray, devices, driver
 from numba.core import config
 from numba.cuda.api_util import prepare_shape_strides_dtype
+from numba.cuda.cudadrv.runtime import get_version
 
 # NDarray device helper
 
+cuda_version = get_version()
```
isVoid, line 17: `get_version` initializes device and should not be done at import time.
```suggestion
```

numba_cuda/numba/cuda/api.py
```diff
@@ -95,6 +97,17 @@ def is_float16_supported():
     return True
 
 
+def is_bfloat16_supported():
+    """Whether bfloat16 are supported.
+
+    bfloat16 are only supported on devices with compute capability >= 8.0 and cuda version >= 12.0
+    """
+    return current_context().device.supports_bfloat16 and cuda_version >= (
```
isVoid, line 105: Instead, we should get cuda version from here:
```suggestion
    """
    cuda_version = get_version()
    return current_context().device.supports_bfloat16 and cuda_version >= (
```

docs/source/reference/types.rst
```diff
@@ -54,3 +54,134 @@ mutating vector components will be added in a future release.
...
+Narrow Data Types
+~~~~~~~~~~~~~~~~~
+
+Bfloat16
+--------
+
+.. note::
+
+    Bfloat16 is only supported with CUDA version 12.0+, and only supported on
+    device compute capability with 8.0+.
```
gmarkall, line 68:
```suggestion
    devices with compute capability 8.0 or above.
```

docs/source/reference/types.rst
```diff
@@ -54,3 +54,134 @@ mutating vector components will be added in a future release.
...
+
+To determine whether Numba supports compiling code that uses the ``bfloat16``
+type in the current configuration, use:
```
gmarkall, line 71:
```suggestion
To determine whether ``bfloat16`` is supported in the current configuration,
use:
```

docs/source/reference/types.rst
```diff
@@ -54,3 +54,134 @@ mutating vector components will be added in a future release.
...
+
+.. function:: numba.cuda.is_bfloat16_supported()
+
+    Returns ``True`` if current device supports bfloat16. ``False`` otherwise.
```
gmarkall, line 75:
```suggestion
    Returns ``True`` if the current device and toolkit support bfloat16.
    ``False`` otherwise.
```

docs/source/reference/types.rst
```diff
@@ -54,3 +54,134 @@ mutating vector components will be added in a future release.
...
+
+Data Movement and Casts
+***********************
+
+Construction a single instance of bfloat16 object takes 1 API:
```
gmarkall, line 80:
```suggestion
Construction of a single instance of a ``bfloat16`` object:
```

docs/source/reference/types.rst
```diff
@@ -54,3 +54,134 @@ mutating vector components will be added in a future release.
...
+.. function:: numba.cuda.bf16.bfloat16(a)
+
+    Constructs a bfloat16 data from existing device scalar. Supported scalar types:
+    - float64
+    - float32
+    - int64
+    - int32
+    - uint64
+    - uint32
+    On cuda version 12.0+ machines, supports construction from ``float16``
```
gmarkall, line 91: I think you need an extra space in rst before a nested bulleted list otherwise it renders as:

> Supported scalar types: - float64 - float32 - int64 - int32 - uint64 - uint32 On cuda version ...

We might also rename the parameter `scalar` and refer to it in the description, as well as marking the types as code. Since we only support CUDA 12 for blfoat16, I think we don't need to call it out explicitly:

```suggestion
    Constructs a ``bfloat16`` from existing device `scalar`. Supported scalar
    types:

    - ``float64``
    - ``float32``
    - ``float16``
    - ``int64``
    - ``int32``
    - ``uint64``
    - ``uint32``
```

docs/source/reference/types.rst
```diff
@@ -54,3 +54,134 @@ mutating vector components will be added in a future release.
...
+
+Conversely, ``bfloat16`` data can be cast back to existing native data type via
+``dtype(h)``, where ``dtype`` is one of the data types above (except float16),
```
gmarkall, line 94: Is it the case that you can't cast to a `float16` even on CUDA 12?

docs/source/reference/types.rst
```diff
@@ -54,3 +54,134 @@ mutating vector components will be added in a future release.
...
+and ``X`` is a bfloat16 object.
```
gmarkall, line 95: Should `X` be `h`?

docs/source/reference/types.rst
```diff
@@ -54,3 +54,134 @@ mutating vector components will be added in a future release.
...
+
+Arithmatics
+************
```
gmarkall, line 98:
```suggestion
Arithmetic
**********
```

docs/source/reference/types.rst
```diff
@@ -54,3 +54,134 @@ mutating vector components will be added in a future release.
...
+
+A ``bfloat16`` data can be computed with another ``bfloat16`` data with many
+supported aritheatic operators. The list of supported operations:
```
gmarkall, line 101: I think the list includes unary and binary operations, so we should be general about how many operands there are:
```suggestion
Supported arithmetic operations on ``bfloat`16`` operands are:
```

docs/source/reference/types.rst
```diff
@@ -54,3 +54,134 @@ mutating vector components will be added in a future release.
...
+
+- Arithmatics (``+, -, *, /``)
+- Arithmatic assignment oeprators (``+=, -=, *=, /=``)
+- Logical operators (``==, !=, >, <, >=, <=``)
+- Unary arithmatics (``+, -``)
```
gmarkall, line 106: I think this renders a little more nicely if each operator is in its own code quotes:

```suggestion
- Arithmetic (``+``, ``-``, ``*``, ``/``)
- Arithmetic assignment operators (``+=``, ``-=``, ``*=``, ``/=``)
- Logical operators (``==``, ``!=``, ``>``, ``<``, ``>=``, ``<=``)
- Unary arithmetic (``+``, ``-``)
```

docs/source/reference/types.rst
```diff
@@ -54,3 +54,134 @@ mutating vector components will be added in a future release.
...
+
+Math Intrinsics
+***************
+
```
gmarkall, line 110: In this section there's some inconsistency in argument naming - sometimes it's `h` and sometimes `a` - I imagine `h` might have been followed from a convention for `float16` docs somewhere (i.e. `h` for half). 

Would it be better to always use `a`, or perhaps `x` for a generic argument name, or `b` for a `bfloat16`?

docs/source/reference/types.rst
```diff
@@ -54,3 +54,134 @@ mutating vector components will be added in a future release.
...
+.. function:: numba.cuda.bf16.hceil(h)
+    Compute the smallest integer value not less than ``h``.
+
+.. function:: numba.cuda.bf16.hfloor(h)
+    Calculate the largest integer value which is less than or equal to ``h``.
+
+.. function:: numba.cuda.bf16.hrint(h)
+    Round ``h`` to the nearest integer value in nv_bfloat16 floating-point format,
+    with bfloat16way cases rounded to the nearest even integer value.
```
gmarkall, line 125: I think find / replace "half" -> "bfloat16" has caused some mischief here!

numba_cuda/numba/cuda/tests/cudapy/test_bfloat16_bindings.py
```diff
@@ -22,18 +22,17 @@
...
-cuda_version = get_version()
+from numba.cuda import is_bfloat16_supported
 
 dtypes = [int16, int32, int64, uint16, uint32, uint64, float32]
 
 
 @unittest.skipIf(
-    (cuda.get_current_device().compute_capability < (8, 0)),
-    "bfloat16 requires compute capability 8.0+",
+    not is_bfloat16_supported(),
```
gmarkall, line 34: I think this is going to call `runtime.get_version()` at test discovery time - it might be working OK but I think it would be safer to do:

```python
if not is_bfloat16_supported():
    self.skipTest()
```

inside the test body. (It's not an ideal pattern, but there've been other instances where I've had to do this in the test suite).

gmarkall: (I just noticed this was already not a good pattern because of calling `cuda.get_current_device()` at test dicsovery time - I missed that in previous reviews.

numba_cuda/numba/cuda/tests/cudapy/test_bfloat16.py
```diff
@@ -0,0 +1,50 @@
+from numba import cuda, float32
+from numba.cuda.bf16 import bfloat16
+from numba.cuda.testing import unittest, CUDATestCase
+
+import math
+
+
+@unittest.skipIf(
+    not cuda.is_bfloat16_supported(),
```
gmarkall, line 9: (similar note to the one on the `skipIf()` in `test_bfloat16_bindings.py` below)

numba_cuda/numba/cuda/bf16.py
```diff
@@ -0,0 +1,112 @@
...
+    return _make_unary(a, hexp)
+
+
+try:
+    from math import exp2
+
+    @overload(exp2, target="cuda")
+    def exp2_ol(a):
+        return _make_unary(a, hexp2)
+except ImportError:
```
gmarkall, line 89: I'm amazed this was only added in 3.11! :-)

gmarkall (changes requested):
Thanks for the PR! I made quite a few comments on the diff but most are just suggestions on the documentation.

There are a couple of things I think we should change to make it safer to avoid CUDA APIs during test discovery, and a couple of other notes as well.

docs/source/reference/types.rst
isVoid: No, you can convert bfloat16 into float16, the logic exists in cuda_fp16.h, not in cuda_bf16.h, which is why the functionality isn't implemented yet.

docs/source/reference/types.rst
isVoid: Yes.

docs/source/reference/types.rst
isVoid:
```suggestion
    with half way cases rounded to the nearest even integer value.
```

docs/source/reference/types.rst
isVoid: I think the inconsistency comes from my simple mimic of https://docs.nvidia.com/cuda/cuda-math-api/cuda_math_api/group__CUDA__MATH____BFLOAT16__FUNCTIONS.html, which also has some of the inconsistencies.

I kinda like the idea of `b` for `bfloat16`, if that's ok with you, let's just adopt this name convention.

numba_cuda/numba/cuda/tests/cudapy/test_bfloat16_bindings.py
isVoid: Ah - would you say the implicit rule is that the device should be initialized right before the first time we actually invoke a kernel?

docs/source/reference/types.rst
isVoid: Done [709ecf9](https://github.com/NVIDIA/numba-cuda/pull/245/commits/709ecf95757564cbab2dad8995fe9f08d06fd81b)

numba_cuda/numba/cuda/tests/cudapy/test_bfloat16.py
isVoid: Done [144497e](https://github.com/NVIDIA/numba-cuda/pull/245/commits/144497e63447c055e1a39b5907c1aef6a790e31a)

numba_cuda/numba/cuda/tests/cudapy/test_bfloat16_bindings.py
gmarkall: It should be initialized right before the first time we need to interact with the device - it needn't be a kernel launch - it could be an allocation, an enumeration of the devices in the machine (i.e. `cuda.detect()`), etc.

I think I was thinking of the rule as "not at import time, and not at test discovery time", but your rule probably puts a closer bound on it.

gmarkall (approved):

PR #181, opened 2025-03-31
Fix cuda.jit decorator inline

ZzEeKkAa:
In the `@cuda.jit` decorator `inline` is being ignored and unused. Because it is explicitly declared in the function signature it is not part of the `kws` and is not passed further to the dispatcher. This PR fixes it by removing it from function signature.

Here is an affected example:

```python
from numba import cuda
import numpy as np

@cuda.jit(inline="always")
def set_zero(a):
    a[0]=0

@cuda.jit
def f(a):
    set_zero(a)


a = np.ones(2)

f[1, 2](a)

print(a)
```

optimized LLVM code before the change:
```llvm
LLVM DUMP: Post kernel fixup _ZN8__main__1fB2v1B92cw51cXTLSUwv1sDUaKthoaNgqamjgBQYtuBwBgY3WszbglIhRBiSyJAEUBMOTALJ3bbV6rDgVbdSgEaTOlASyDOu1QQAE5ArrayIdLi1E1C7mutable7alignedE
; ModuleID = "f$1"
target triple = "nvptx64-nvidia-cuda"
target datalayout = "e-p:64:64:64-i1:8:8-i8:8:8-i16:16:16-i32:32:32-i64:64:64-i128:128:128-f32:32:32-f64:64:64-v16:16:16-v32:32:32-v64:64:64-v128:128:128-n16:32:64"

@"_ZN08NumbaEnv8__main__1fB2v1B92cw51cXTLSUwv1sDUaKthoaNgqamjgBQYtuBwBgY3WszbglIhRBiSyJAEUBMOTALJ3bbV6rDgVbdSgEaTOlASyDOu1QQAE5ArrayIdLi1E1C7mutable7alignedE" = common global i8* null
define void @"_ZN8__main__1fB2v1B92cw51cXTLSUwv1sDUaKthoaNgqamjgBQYtuBwBgY3WszbglIhRBiSyJAEUBMOTALJ3bbV6rDgVbdSgEaTOlASyDOu1QQAE5ArrayIdLi1E1C7mutable7alignedE"(i8* %"arg.a.0", i8* %"arg.a.1", i64 %"arg.a.2", i64 %"arg.a.3", double* %"arg.a.4", i64 %"arg.a.5.0", i64 %"arg.a.6.0")
{
entry:
  %"inserted.meminfo" = insertvalue {i8*, i8*, i64, i64, double*, [1 x i64], [1 x i64]} undef, i8* %"arg.a.0", 0
[...]

PR #246, opened 2025-05-09
Fix inlining behaviour at the NVVM IR level

gmarkall:
PR #181 aimed to align the behaviour of the `inline` kwarg with that of upstream Numba, in that it now forces inlining at the Numba IR level. It turns out that this kwarg in Numba-CUDA already had the prior effect of enabling inlining at the NVVM IR level.

Because the default value of `inline` is `"never"`, this was interpreted by the `compile_cuda()` function as a `True`ish value and every device function got marked with the `alwaysinline` function attribute. This is a minor problem in that it probably forces a lot of inlining that we don't want, but also a major problem in that it triggers an NVVM bug that was only resolved in CUDA 12.3 that causes a hang in `nvvmCompileProgram()`.

To rectify these issues, we add the `forceinline` kwarg to the `@cuda.jit` decorator and the `cuda.compile[_*]()` functions. Now, `compile_cuda()` will only enable inlining at the NVVM IR level for `forceinline` and not `inline`. This is aligned with the behaviour of upstream Numba (see numba/numba#10068). We now document the `inline` and `forceinline` kwargs to clarify the intent and behaviour for users.

For clarity: the behaviour is now:

- The `inline` kwarg enables inlining only at the Numba IR level.
- The `forceinline` kwarg enables inlining only at the NVVM IR level.

### DIFF
--- a/numba_cuda/numba/cuda/decorators.py
+++ b/numba_cuda/numba/cuda/decorators.py
@@ -17,6 +17,7 @@ def jit(
     func_or_sig=None,
     device=False,
     inline="never",
+    forceinline=False,
     link=[],
     debug=None,
     opt=None,
@@ -39,6 +40,14 @@ def jit(
        .. note:: A kernel cannot have any return value.
     :param device: Indicates whether this is a device function.
     :type device: bool
+    :param inline: Enables inlining at the Numba IR level when set to
+       ``"always"``. See `Notes on Inlining
+       <https://numba.readthedocs.io/en/stable/developer/inlining.html>`_.
+    :type inline: str
+    :param forceinline: Enables inlining at the NVVM IR level when set to
+       ``True``. This is accomplished by adding the ``alwaysinline`` function
+       attribute to the function definition.
+    :type forceinline: bool
     :param link: A list of files containing PTX or CUDA C/C++ source to link
        with the function
     :type link: list
@@ -85,7 +94,9 @@ def jit(
         DeprecationWarning(
             "Passing bool to inline argument is deprecated, please refer to "
             "Numba's documentation on inlining: "
-            "https://numba.readthedocs.io/en/stable/developer/inlining.html"
+            "https://numba.readthedocs.io/en/stable/developer/inlining.html. "
+            "You may have wanted the forceinline argument instead, to force "
+            "inlining at the NVVM IR level."
         )
 
         inline = "always" if inline else "never"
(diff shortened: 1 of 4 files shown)


brandon-b-miller (approved):

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -183,7 +184,7 @@ def __init__(
             self.argtypes,
             debug=self.debug,
             lineinfo=lineinfo,
-            inline=inline,
```
ZzEeKkAa, line 186: Just double checking, since it is happening on LLVM IR level there is no need to pass `inline`, because it affects only Numba IR?

numba_cuda/numba/cuda/decorators.py
```diff
@@ -39,6 +40,14 @@ def jit(
        .. note:: A kernel cannot have any return value.
     :param device: Indicates whether this is a device function.
     :type device: bool
+    :param inline: Enables inlining at the Numba IR level when set to
+       ``"always"``. See `Notes on Inlining
+       <https://numba.readthedocs.io/en/stable/developer/inlining.html>`_.
+    :type inline: str
+    :param forceinline: Enables inlining at the NVVM IR level when set to
```
ZzEeKkAa, line 47: Do we want it to be bool, or we want to expose llvm IR attributes directly:
https://llvm.org/docs/LangRef.html#function-attributes

gmarkall: I want it to be `bool` so it's consistent with Numba. There is a plan for Numba to expose LLVM attributes directly, so I'll aim to align with that in future.

gmarkall:
> There was luck of docs that explains that behavior...

Not only that, but I didn't even know `inline` was working in any sense at all... I thought it was one of those things that was perpetually (or for a really long time) broken / a no-op.

numba_cuda/numba/cuda/dispatcher.py
gmarkall: Yeah, exactly - `inline` affects online Numba IR, `forceinline` affects only LLVM IR (which is how it should be when all is working correctly)

gmarkall:
@ZzEeKkAa Many thanks for the review - I've responded to the comments - let me know if I should follow up any further!

gmarkall:
I'll merge this a little later today, as it seems to have been the resolution to RAPIDS hanging (in https://github.com/rapidsai/cudf/pull/18688) in the v0.10.1 release - assuming there are no objections?

PR #251, opened 2025-05-14
Use specific `cuCtxGetCurrent` API version

gmarkall:
May be required in some setups to ensure the correct API version is selected.


kkraus14:
@gmarkall I checked and I don't believe this is needed. Closing this.

PR #252, opened 2025-05-14
Change CI to only be manually triggered to save on CI runs

kkraus14:
Changes CI to require manually invoking it to save on CI runs. Previously, it would run on every commit on PRs regardless of if they were in draft or ready for review. We've used this configuration in the `cuda-python` repo and it hasn't been particularly painful and has allowed for more free flowing commits without burning CI resources unnecessarily.

### DIFF
--- a/.github/copy-pr-bot.yaml
+++ b/.github/copy-pr-bot.yaml
@@ -4,3 +4,6 @@
 # https://docs.gha-runners.nvidia.com/apps/copy-pr-bot/
 
 enabled: true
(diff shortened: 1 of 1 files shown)


cryos (approved):

gmarkall:
Do we need to write the "ok to test" comment every time we want a CI run after this is merged?

kkraus14:
> Do we need to write the "ok to test" comment every time we want a CI run after this is merged?

Yes

PR #253, opened 2025-05-14
Add LTO Support with BF16

isVoid:
In https://github.com/NVIDIA/numba-cuda/pull/245, we added bfloat16 API bindings. And it turns out that we missed testing the usage of the bindings with `lto=True`. This PR adds that.


isVoid:
@gmarkall it comes to me that whether we should adopt the similar usage here: https://github.com/NVIDIA/numba-cuda/pull/240 and go out to enable `lto=True` by default when we use Numbast generated bindings? I occurred to me that the overhead of non-LTO FFI is quite high.

gmarkall:
> @gmarkall it comes to me that whether we should adopt the similar usage here: #240 and go out to enable `lto=True` by default when we use Numbast generated bindings? I occurred to me that the overhead of non-LTO FFI is quite high.

I think we should have a separate PR to enable LTO by default in general as long as pynvjitlink is available and it is new enough for the current GPU. There might be some caveats / nuance to this behaviour, but I think the general approach is that we should be doing LTO as much as possible because it's such a performance win with any external code.

isVoid:
Closing: LTO is on by default. This means bfloat16 are tested under LTO mode currently. This PR now adds little value.

PR #254, opened 2025-05-16
Allow Numba NVRTC Binding Search Additional Paths

isVoid:
Currently, the nvrtc binding searches a few hardcoded internal paths when looking for headers. This limits the foreign function usage to only depend on standard CUDA libraries. This PR adds `CUDA_RTC_EXTRA_SEARCH_PATHS` entry, a colon separated path list, which defines additional search paths when compiling external functions.

Note that these search paths are placed *after* the standard cudatookit and numba-cuda internal search paths.

closes https://github.com/NVIDIA/numba-cuda/issues/46

### DIFF
--- a/docs/source/user/cuda_ffi.rst
+++ b/docs/source/user/cuda_ffi.rst
@@ -239,7 +239,7 @@ of NVRTC subject to the following considerations:
    separated.
 
 Extra Search Path Example
-*************************
+^^^^^^^^^^^^^^^^^^^^^^^^^
 
 This example demonstrates calling a foreign function which includes additional
 headers that does not exist in default Numba-CUDA search paths.
--- a/docs/source/user/cuda_ffi.rst
+++ b/docs/source/user/cuda_ffi.rst
@@ -273,7 +273,7 @@ joined with ``:`` before setting ``config.CUDA_NVRTC_EXTRA_SEARCH_PATHS``:
    :caption: from ``test_ex_extra_includes`` in ``numba/cuda/tests/doc_examples/test_ffi.py``
    :start-after: magictoken.ex_extra_search_paths.begin
    :end-before: magictoken.ex_extra_search_paths.end
-   :dedent: 8
+   :dedent: 12
    :linenos:
 
 Next, use ``saxpy`` as intended.
@@ -283,7 +283,7 @@ Next, use ``saxpy`` as intended.
    :caption: from ``test_ex_extra_includes`` in ``numba/cuda/tests/doc_examples/test_ffi.py``
    :start-after: magictoken.ex_extra_search_paths_kernel.begin
    :end-before: magictoken.ex_extra_search_paths_kernel.end
-   :dedent: 8
+   :dedent: 12
    :linenos:
 
 
--- a/docs/source/user/cuda_ffi.rst
+++ b/docs/source/user/cuda_ffi.rst
@@ -238,8 +238,8 @@ of NVRTC subject to the following considerations:
    :envvar:`NUMBA_CUDA_RTC_EXTRA_SEARCH_PATHS`. Multiple paths should be colon
    separated.
 
-Extra Search Path Example
-^^^^^^^^^^^^^^^^^^^^^^^^^
+Extra Search Paths Example
+~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 This example demonstrates calling a foreign function which includes additional
 headers that does not exist in default Numba-CUDA search paths.
--- a/docs/source/user/cuda_ffi.rst
+++ b/docs/source/user/cuda_ffi.rst
@@ -233,7 +233,58 @@ of NVRTC subject to the following considerations:
   on Linux and ``$env:CUDA_PATH\include`` on Windows. It can be modified using
   the environment variable :envvar:`NUMBA_CUDA_INCLUDE_PATH`.
 - The CUDA include directory will be made available to NVRTC on the include
-  path; additional includes are not supported.
+  path.
+- Additional search paths can be set to the environment variable
+   :envvar:`NUMBA_CUDA_RTC_EXTRA_SEARCH_PATHS`. Multiple paths should be colon
+   separated.
+
+Extra Search Path Example
+*************************
+
+This example demonstrates calling a foreign function which includes additional
+headers that does not exist in default Numba-CUDA search paths.
+
+The definition of the C++ template APIs are defined in two different locations:
+
+.. literalinclude:: ../../../numba_cuda/numba/cuda/tests/doc_examples/ffi/include/mul.cuh
+   :language: C
+   :caption: ``numba/cuda/tests/doc_examples/ffi/include/mul.cuh``
+   :linenos:
+
+.. literalinclude:: ../../../numba_cuda/numba/cuda/tests/data/include/add.cuh
+   :language: C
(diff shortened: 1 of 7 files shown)


numba_cuda/numba/cuda/tests/doc_examples/test_ffi.py
```diff
@@ -85,6 +85,56 @@ def reduction_caller(result, array):
...
+        import os
+
+        basedir = os.path.dirname(os.path.abspath(__file__))
+        mul_dir = os.path.join(basedir, "ffi", "include")
+        saxpy_cu = os.path.join(basedir, "ffi", "saxpy.cu")
+
+        testdir = os.path.dirname(basedir)
+        add_dir = os.path.join(testdir, "data", "include")
+
+        old_setting = config.CUDA_RTC_EXTRA_SEARCH_PATHS
```
gmarkall, line 100:
```suggestion
        old_setting = config.CUDA_NVRTC_EXTRA_SEARCH_PATHS
```

numba_cuda/numba/cuda/tests/doc_examples/test_ffi.py
```diff
@@ -85,6 +85,56 @@ def reduction_caller(result, array):
...
+        block_size = 32
+        num_blocks = (size // block_size) + 1
+
+        vector_saxpy[num_blocks, block_size](a, X, Y, R)
+
+        expected = a * X + Y
+        np.testing.assert_equal(R, expected)
+
+        # Reset config setting
+        config.CUDA_RTC_EXTRA_SEARCH_PATHS = old_setting
```
gmarkall, line 136:
```suggestion
        config.CUDA_NVRTC_EXTRA_SEARCH_PATHS = old_setting
```

numba_cuda/numba/cuda/tests/doc_examples/test_ffi.py
gmarkall: The `old_setting` should probably be managed by the `override_config` context manager - e.g. 

https://github.com/NVIDIA/numba-cuda/blob/66a7b6fb807418715b9977b26095822a9bad5d8f/numba_cuda/numba/cuda/tests/nrt/test_nrt.py#L91-L92

so that if an error occurs during the test the original setting is restored for the rest of the test suite.

docs/source/user/cuda_ffi.rst
```diff
@@ -233,7 +233,58 @@ of NVRTC subject to the following considerations:
...
+  path.
+- Additional search paths can be set to the environment variable
+   :envvar:`NUMBA_CUDA_RTC_EXTRA_SEARCH_PATHS`. Multiple paths should be colon
+   separated.
+
+Extra Search Paths Example
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This example demonstrates calling a foreign function which includes additional
+headers that does not exist in default Numba-CUDA search paths.
```
gmarkall, line 245:
```suggestion
This example demonstrates calling a foreign function that includes additional
headers not in the default Numba-CUDA search paths.
```

docs/source/user/cuda_ffi.rst
```diff
@@ -233,7 +233,58 @@ of NVRTC subject to the following considerations:
...
+
+The definition of the C++ template APIs are defined in two different locations:
```
gmarkall, line 247:
```suggestion
The definitions of the C++ template APIs are in two different locations:
```

docs/source/user/cuda_ffi.rst
```diff
@@ -233,7 +233,58 @@ of NVRTC subject to the following considerations:
...
+   :language: C
+   :caption: ``numba/cuda/tests/doc_examples/ffi/include/mul.cuh``
+   :linenos:
+
+.. literalinclude:: ../../../numba_cuda/numba/cuda/tests/data/include/add.cuh
+   :language: C
+   :caption: ``numba/cuda/tests/data/include/add.cuh``
+   :linenos:
+
+Neither header exists in the default search paths of Numba-CUDA, however, the
```
gmarkall, line 259:
```suggestion
Neither of the headers are in the default search paths of Numba-CUDA, but the
```

docs/source/user/cuda_ffi.rst
```diff
@@ -233,7 +233,58 @@ of NVRTC subject to the following considerations:
...
+foreign device function ``saxpy`` depends on them:
+
+.. literalinclude:: ../../../numba_cuda/numba/cuda/tests/doc_examples/ffi/saxpy.cu
+   :language: C
+   :caption: ``numba/cuda/tests/data/doc_examples/ffi/saxpy.cu``
+   :linenos:
+
+In the Python Code, assume ``mul_dir`` and ``add_dir`` are set to the folder that contains
+``mul.cuh`` and ``add.cuh`` respectively. Use ``:`` to join them to a single string and
+set to ``CUDA_RTC_EXTRA_SEARCH_PATHS`` in config.
```
gmarkall, line 269:
```suggestion
In the Python code, assume that ``mul_dir`` and ``add_dir`` are set to the
paths that contain ``mul.cuh`` and ``add.cuh`` respectively. The paths are
joined with ``:`` before setting ``config.CUDA_NVRTC_EXTRA_SEARCH_PATHS``:
```

docs/source/user/cuda_ffi.rst
```diff
@@ -233,7 +233,58 @@ of NVRTC subject to the following considerations:
...
+
+.. literalinclude:: ../../../numba_cuda/numba/cuda/tests/doc_examples/test_ffi.py
+   :language: python
+   :caption: from ``test_ex_extra_includes`` in ``numba/cuda/tests/doc_examples/test_ffi.py``
+   :start-after: magictoken.ex_extra_search_paths.begin
+   :end-before: magictoken.ex_extra_search_paths.end
+   :dedent: 8
+   :linenos:
+
+Next, use ``saxpy`` as intended.
```
gmarkall, line 279:
```suggestion
Next, use ``saxpy`` as intended:
```

gmarkall (changes requested):
I think the feature and implementation looks good. I have a couple of items on the diff:

- Can we use "NVRTC" rather than "RTC" in the config variable name to avoid ambiguity? I have tried to make this all as suggestions that can be accepted to avoid manually going and editing the code
- I suggested some clarifications for readability in the docs - let me know if you prefer another form of any of these.
- The `override_config` context manager is normally used to handle changing configuration during a test, so that the original configuration is restored if the test does not run to completion.

Following resolution of these I think this is good to merge.

numba_cuda/numba/cuda/tests/doc_examples/test_ffi.py
isVoid: I thought of this as well - except I was hoping that the code can demonstrate setting to the `numba.config` code since it's the suggested way to show using it in user code. It's probably ok to wrap the code section with `override config` while still keeping this line.

isVoid: Done in [e1b7c93](https://github.com/NVIDIA/numba-cuda/pull/254/commits/e1b7c93716ee606789a4bf57afe60393a0f100fa), I also verified that afterwards the config is indeed reset to empty string.

gmarkall: That makes sense :+1:

gmarkall (approved):

PR #265, opened 2025-05-29
Use Numbast generated float16 bindings

### LINKED ISSUE #199: [FEA] Support fp16 data type

lakshayg:
This PR roughly follows the same approach as #166 and #245 to support float16 using the CUDA headers instead of a manual implementation.

The main factors driving changes in this PR are:
- float16 is already a type supported within numba (unlike bfloat16). We need to ensure that the bindings generated using Numbast support all the functionality that was previously supported. Some notable factors are:
  - cu11 does not support all the operations (example: __half2char) that were provided by the original numba implementation. We therefore use a custom implementation of casting functions using inline assembly (which is supported by both cu11 and cu12).
  - Numbast does not generate any type-promotion rules. This leads to undesired type casting when mixing float16 with other numeric types. To work around this, we re-use the type promotion rules built into numba by using custom lowering logic.
- float16 support in original numba implementation did not require the user to import any modules. It was available just by importing numba-cuda. To enable this, we create typing/target registries for fp16 and register them in CUDATypingContext and CUDATargetContext.
- The original implementation of many float16 operations used inline assembly. The new implementation call wrapper functions instead. This makes the assembly calls not directly accessible from the compiled kernel. We use LTO in a number of tests so all the indirection goes away as we can access the PTX.

This PR depends on the following issues/PRs:

- [x] [Allow Inspection of Link-Time Optimized PTX](https://github.com/NVIDIA/numba-cuda/pull/326)
- [x] [Itanium Name Mangler to resolve shim function name conflicts](https://github.com/NVIDIA/numbast/pull/139)
- [x] [[FEA] Generate @lower from non-explicit constructors](https://github.com/NVIDIA/numbast/issues/137)
- [x] [[FEA] Allow configuring skip_prefix](https://github.com/NVIDIA/numbast/issues/140)
- [x] [[FEA] Support excluding typedefs in generated bindings](https://github.com/NVIDIA/numbast/issues/141)
- [x] [Allow creation of custom typing/target registry](https://github.com/NVIDIA/numbast/pull/161)

Closes #199

### DIFF
--- a/docs/source/reference/kernel.rst
+++ b/docs/source/reference/kernel.rst
@@ -507,6 +507,8 @@ precision parts of the CUDA Toolkit documentation.
 16-bit Floating Point Intrinsics
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+.. warning:: Starting numba 0.18, LTO is required for performant float16 operations.
+
 The functions in the ``cuda.fp16`` module are used to operate on 16-bit
 floating point operands. These functions return a 16-bit floating point result.
 
--- a/numba_cuda/numba/cuda/_internal/cuda_fp16.py
+++ b/numba_cuda/numba/cuda/_internal/cuda_fp16.py
@@ -16,7 +16,9 @@ import operator
 import numba
 from llvmlite import ir
 from numba import types
+from numba.cuda.cudadrv.driver import _have_nvjitlink
 from numba.core.datamodel import PrimitiveModel, StructModel
+from numba.core.errors import NumbaPerformanceWarning
 from numba.core.extending import (
     make_attribute_wrapper,
     register_model,
@@ -51,6 +53,7 @@ from numba.types import (
     uint64,
     void,
 )
+from warnings import warn
 
 typing_registry = TypingRegistry()
 register = typing_registry.register
@@ -66,6 +69,17 @@ float32x2 = vector_types["float32x2"]
 # Setups:
 
 # Shim Stream:
+
+lto_warning_raised = False
+
+
+def lto_warning_callback(_):
+    global lto_warning_raised
+    if not _have_nvjitlink():
+        if not lto_warning_raised:
+            lto_warning_raised = True
+            msg = "float16 relies on LTO for performance. LTO requires nvjitlink, which is not available or not sufficiently recent (>=12.3)"
+            warn(NumbaPerformanceWarning(msg))
 
 
 class _KeyedStringIO(io.StringIO):
@@ -89,7 +103,7 @@ shim_include = "#include <" + "cuda_fp16.h" + ">"
 shim_prefix = shim_defines + "\n" + shim_include
 shim_stream = _KeyedStringIO()
 shim_stream.write(shim_prefix)
-shim_obj = CUSource(shim_stream)
+shim_obj = CUSource(shim_stream, setup_callback=lto_warning_callback)
 
 # Enums:
 
--- a/numba_cuda/numba/cuda/tests/cudapy/test_warning.py
+++ b/numba_cuda/numba/cuda/tests/cudapy/test_warning.py
@@ -1,7 +1,12 @@
 import numpy as np
 from numba import cuda
-from numba.cuda.testing import unittest, CUDATestCase, skip_on_cudasim
-from numba.tests.support import linux_only, override_config
+from numba.cuda.cudadrv import driver
+from numba.cuda.testing import (
+    unittest,
+    CUDATestCase,
+    skip_on_cudasim,
+)
+from numba.tests.support import linux_only, override_config, run_in_subprocess
 from numba.core.errors import NumbaPerformanceWarning
 from numba.core import config
 import warnings
@@ -9,6 +14,26 @@ import warnings
 
 @skip_on_cudasim("cudasim does not raise performance warnings")
 class TestWarnings(CUDATestCase):
+    def test_float16_warn_if_lto_missing(self):
+        fp16_kernel_invocation = """
+import math
+from numba import cuda, core
+
+@cuda.jit
+def kernel():
+    x = core.types.float16(1.0)
+    y = math.sin(x)
+
+kernel[1,1]()
+kernel[1,1]()
+"""
+        performance_warning = "float16 relies on LTO for performance"
+        expected_warning_count = 0 if driver._have_nvjitlink() else 1
+        _, err = run_in_subprocess(fp16_kernel_invocation)
+        self.assertEqual(
+            err.decode().count(performance_warning), expected_warning_count
+        )
+
     def test_inefficient_launch_configuration(self):
         @cuda.jit
         def kernel():


gmarkall:
> Migrate unit tests to the new implementation

Since `float16` is a public API, the existing unit tests must continue to pass, otherwise we'd be breaking things for existing users.

lakshayg:
I have been thinking about how to properly support float16 but there are some challenges in doing that without breaking some existing functionality or adding custom code to support non-library functionality.

CUDA supports fp16, not as a compiler recognized type, but as a library type. The way it is currently implemented in numba and supported in numba-cuda, almost makes it feel like a native type. For example, numba accepts this code without complaining. `hsin` is a library type that's only defined for `__half` but here it accepts a f32.

```
@numba.cuda.jit
def f():
  return numba.cuda.fp16.hsin(np.float32(1.0))
```

If we try this with `bfloat16` the compiler emits an error

```
There are 2 candidate implementations:
  - Of which 2 did not match due to:
  Type Restricted Function in function 'hsin': File: unknown: Line unknown.
    With argument(s): '(float32)':
   No match for registered cases:
    * (__nv_bfloat16,) -> __nv_bfloat16
```

There are unit tests explicitly checking for functionality that the library would not support: numba_cuda/numba/cuda/tests/cudapy/test_casting.py:82

I think we would probably want to go the backwards compatibility route and implement overloads/codegen as necessary. My next steps will be to eliminate the need for `cpp_function_wrappers.cu`. Would like to get your thoughts on this.

lakshayg:
I tried a couple of things over the last few days:

1. Used llvmlite.ir.Half() as the be_type for float16 only to learn that the generated LLVM-IR is not straightforward to translate to NVVM-IR
2. Tried to get numba to "promote" `numba.core.types.float16` to `numba.cuda.fp16.half` whenever needed but I couldn't find a straightforward way of doing this. This mostly just involved studying the numba and numba-cuda codebase
3. Talked to @isVoid and he suggested trying to make Numbast generated bindings work with the existing `numba.core.types.float16` type. I did this by making generated types aliases to the existing float16 type and removed most of the existing logic needed to support float16 but I'm running into a number of errors when running tests. This seems to be the most promising approach so far. I'm posting some of the errors here:
    - Unable to convert any type to float16. This happened because I removed the lower_cast that defined conversions to and from float16 thinking it won't be needed as the generated python file already defines various casts and they work as expected when called in a jit compiled kernel. However, they don't work when called outside of the jitted function. I suspect they behave as __device__ functions and therefore cannot be used. I therefore ended up restoring the original lower_casts that used inline ptx.
    - `numba.cuda.cudadrv.driver.CudaAPIError: [700] Call to cuModuleUnload results in CUDA_ERROR_ILLEGAL_ADDRESS`
        + Not sure why this happens.
        + I suspected this happened because the numbast generated type was not identical to the pre-existing float16. The generated one defines an `alignof_` property and makes use of it at various places whereas the existing one does not have this information. I worked around this by hardcoding this value in the generated code but that does not resolve the error
    - RuntimeError: Cannot pickle CUDACodeLibrary with linking files
        + Not sure why this happens.
    - There are more errors but they seem to be simpler to resolve and mostly have to do with overload selection

isVoid:
> If we try this with bfloat16 the compiler emits an error

I think, the reason why this errs is that the typing doesn't have the rules defined for Numba-CUDA's bfloat16 type:
https://github.com/numba/numba/blob/6cc96d049d9a486fb03eef1f2479a792596eb4e6/numba/core/typeconv/rules.py

I think you can override this by adding the `can_convert_to` function for bfloat16 type (`_type___nv_bfloat16`) in cuda_bf16.py.

In addition Numba-CUDA implemented the lower cast implementation from any Float type to float16, 
https://github.com/NVIDIA/numba-cuda/blob/0587e400a28ae4ff00fd0fa660abfa5656e86996/numba_cuda/numba/cuda/cudaimpl.py#L371-L380

We don't have the equivalent for bfloat16 -> float16 at this point. Looking at `cuda_bf16.h` it seems like efficient conversion from bf16 to fp16 isn't added till sm_90, and there isn't an intrinsic that does this. I think this is out of hands of Numbast generated bindings and require manual implementation for Numba-CUDA to implement this conversion.

https://github.com/NVIDIA/numba-cuda/blob/0587e400a28ae4ff00fd0fa660abfa5656e86996/numba_cuda/numba/cuda/include/12/cuda_bf16.h#L5060-L5067

> My next steps will be to eliminate the need for cpp_function_wrappers.cu

Yes - I think its function can be replaced by Numbast shim functions now.

isVoid:
> Used llvmlite.ir.Half() as the be_type for float16 only to learn that the generated LLVM-IR is not straightforward to translate to NVVM-IR

Can you post the error here? Interested to learn.

> Tried to get numba to "promote" numba.core.types.float16 to numba.cuda.fp16.half whenever needed but I couldn't find a straightforward way of doing this.

It's worth noting that `numba.core.types.float16` and ` numba.cuda.fp16.half` are two different things. `float16` is a type object and `half` is merely a handle to the constructor of the this type. There is some way to unify these together, but I haven't spent enough time on this yet.

> Unable to convert any type to float16.

I think there are some lower casts generated from the CUDA C++ decls: https://github.com/NVIDIA/numba-cuda/blob/749672f6747f114fe103da8d3d4ba5187561100c/numba_cuda/numba/cuda/_internal/cuda_fp16.py#L1187 Are they imported when you run the tests?

> numba.cuda.cudadrv.driver.CudaAPIError: [700] Call to cuModuleUnload results in CUDA_ERROR_ILLEGAL_ADDRESS

Can you try launching the kernel that tests with `cuda.jit(debug=True)`, and run with cuda-gdb / compute sanitizer?

> RuntimeError: Cannot pickle CUDACodeLibrary with linking files

This is known issue. I think there are some limitation with serializing external files. But probably we can lift the restriction to serializing in-memory code sources (the shim function that Numbast generates).

lakshayg:
Thanks for the comments @isVoid. I'll be able to look into this earliest tomorrow night. Will update as soon as I have had a chance.

lakshayg:
I am down to just 1 failing test! When running tests with `python -m numba.runtests numba.cuda.tests`, `numba.cuda.tests.cudapy.test_bfloat16_bindings.Bfloat16Test.test_ctor` fails with this error:

```
numba.cuda.cudadrv.driver.LinkerError: [300] Call to cuLinkAddData results in CUDA_ERROR_INVALID_SOURCE
error   : Multiple definition of 'operator_lt_1' in '<unnamed-cu>.ptx', first defined in '<unnamed-cu>.ptx'
```

I suspect this is due to bindings for fp16 and bf16 conflicting. Numbast should probably provide a namespace feature so that bindings don't conflict.

lakshayg:
I have cleaned up the history a little and pushed the latest version

numba_cuda/numba/cuda/tests/cudapy/test_casting.py
```diff
@@ -129,6 +131,7 @@ def test_float_to_int(self):
                     self.assertEqual(cfunc(-12.3), int(-12.3))
 
     @skip_on_cudasim("Compilation unsupported in the simulator")
+    @unittest.skip("Bindings generated using Numbast")
```
lakshayg: @isVoid There are a number of tests that check for specific ptxas instructions. This is probably because previously the implementation used inline assembly. Now that we use Numbast generated bindings, could you please confirm that it is okay to remove these tests altogether or let me know if you have ideas for an alternative test.

numba_cuda/numba/cuda/tests/nocuda/test_function_resolution.py
```diff
@@ -8,6 +8,7 @@
 @skip_on_cudasim("Skip on simulator due to use of cuda_target")
 class TestFunctionResolution(unittest.TestCase):
     def test_fp16_binary_operators(self):
+        import numba.cuda.fp16
```
lakshayg: @isVoid A number of tests now need numba.cuda.fp16 to be imported manually to be able to use the bindings. This would break existing code. Do you have suggestions on how to work around this?

lakshayg:
> I am down to just 1 failing test! When running tests with `python -m numba.runtests numba.cuda.tests`, `numba.cuda.tests.cudapy.test_bfloat16_bindings.Bfloat16Test.test_ctor` fails with this error:
> 
> ```
> numba.cuda.cudadrv.driver.LinkerError: [300] Call to cuLinkAddData results in CUDA_ERROR_INVALID_SOURCE
> error   : Multiple definition of 'operator_lt_1' in '<unnamed-cu>.ptx', first defined in '<unnamed-cu>.ptx'
> ```
> 
> I suspect this is due to bindings for fp16 and bf16 conflicting. Numbast should probably provide a namespace feature so that bindings don't conflict.

https://github.com/NVIDIA/numbast/pull/139 fixes this issue

gmarkall:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/265/commits/cae3704901950ee3392f34d74bfc801307a22561

numba_cuda/numba/cuda/tests/cudapy/test_casting.py
isVoid: I tried a few things and didn't have a strightforward way to check if the intrinsic ptx is generated in the final binary. Because Numbast binding uses the kernel -> shim_function -> intrinsic indirection, you only get the first level call in the kernel PTX. To collapse everything in to the kernel, we need to enable `lto=True` and look at the SASS. This is the closest I got:
```python
        for pyfunc, size in zip(pyfuncs, sizes):
            pyfunc_compiled = cuda.jit(device=True)(pyfunc)
            @cuda.jit(lto=True)
            def cast(x):
                pyfunc_compiled(x)

            kernel = cast.compile((f2,))
            sass = cast.inspect_sass()
```
However, it's generally not advised to check against SASS because there's no guarantee of this over architecture change.

One idea of an alternate test is that, if you write a simple cuda C++ kernel, that looks like:
```c++
char __device__ __forceinline__ cast(half input) {
    return char(input);
}

void __global__ test_kernel(half input) {
    auto x = cast(input);
}
```
And compile this with NVRTC into sass, try to compare the result of this against the native cuda C++ output and see if they are identical?

isVoid: > you only get the first level call in the kernel PTX

I think it would be nice if `Dispatcher.inspect_asm` can not only print out result of the top level PTX, but also all compiled PTX underlying its code library. It would be more helpful in debugging situations like this.

cc @gmarkall

numba_cuda/numba/cuda/tests/nocuda/test_function_resolution.py
isVoid: Kudos to @gmarkall giving some hints on this. So the reason why we are leveraging the `import numba.cuda.fp16` to enable all these features, are that we are effectively inserting the typing and lowering into `cudadecl` and `cudaimpl` registry eagerly. There are APIs in typingctx and targetctx that adds them into the registry at compile time:
https://github.com/NVIDIA/numba-cuda/blob/main/numba_cuda/numba/cuda/target.py#L24

So to look into our bindings retrospectively, it seems like `cuda_fp16.py` should maintain its own registry object, and override the typing and lowering functions.
```
from numba.core.typing.templates import Registry as TypingRegistry
from numba.core.imputils import Registry as TargetRegistry, lower_cast

typing_registry = TypingRegistry()
register = typing_registry.register
register_attr = typing_registry.register_attr
register_global = typing_registry.register_global
target_registry = TargetRegistry()
lower = target_registry.lower
lower_attr = target_registry.lower_getattr
lower_constant = target_registry.lower_constant
```

Then, in `load_additional_registries`, you can use `self.install_registry` in the similar way to the rest of the installations.
```
import numba.cuda.fp16
self.install_registry(numba.cuda.fp16.typing_registry)
```

lakshayg: This worked, thanks!

isVoid:
@lakshayg Let's try not to force push to a PR in the future, instead, pull from the upstream `main` and use merge to resolve conflicts. Force pushes erase previous suggestion history, and makes them harder to track.

numba_cuda/numba/cuda/tests/cudapy/test_intrinsics.py
```diff
@@ -1083,6 +1089,7 @@ def test_round_f8(self):
 
     def test_round_to_f4(self):
         compiled = cuda.jit("void(float32[:], float32, int32)")(simple_round_to)
+        return
```
isVoid: Any reason why we are skipping this test?

isVoid (changes requested):

numba_cuda/numba/cuda/tests/cudapy/test_intrinsics.py
lakshayg: This is currently failing in CI and I'm talking with Graham on how to resolve it. This seems like a platform specific issue. Forgot that I had disabled this test until you pointed in out. Thanks!

lakshayg:
@isVoid I am trying to resolve `<unnamed-cu>(41): error: class "__half" has no member "operator char"` by eliminating calls to `operator char` with `static_cast<char>`.

I suspect that this has to do with cuda 11 vs 12. The cuda_fp16.h header at `numba_cuda/numba/cuda/include/11/cuda_fp16.h` does not have `operator char` but the one at `numba_cuda/numba/cuda/include/12/cuda_fp16.h` does.

I don't have high hopes that this will fix the issue but don't have have other ideas either since I can't repro the error locally. Let's see if this resolves the issue. In case it works, this is the diff for numbast

```diff
diff --git a/numbast/src/numbast/utils.py b/numbast/src/numbast/utils.py
index 5f98ad0..3d8fa42 100644
--- a/numbast/src/numbast/utils.py
+++ b/numbast/src/numbast/utils.py
@@ -247,7 +247,7 @@ def make_struct_conversion_operator_shim(
     conv_op_shim = """{includes}
 extern "C" __device__ int
 {shim_name}({return_type} &retval, {struct_name} *self) {{
-    retval = self->{method_name}();
+    retval = static_cast<{return_type}>(*self);
     return 0;
 }}
     """
```

lakshayg:
Was able to reproduce the issue locally and with the static_cast version, this is the new error:

```
<unnamed-cu>(41): error: more than one conversion function from "__half" to "char" applies:
            function "__half::operator float() const"
/home/lakshayg/proj/numba-cuda/numba_cuda/numba/cuda/include/11/cuda_fp16.hpp(204): here
            function "__half::operator short() const"
/home/lakshayg/proj/numba-cuda/numba_cuda/numba/cuda/include/11/cuda_fp16.hpp(222): here
            function "__half::operator unsigned short() const"
/home/lakshayg/proj/numba-cuda/numba_cuda/numba/cuda/include/11/cuda_fp16.hpp(225): here
            function "__half::operator int() const"
/home/lakshayg/proj/numba-cuda/numba_cuda/numba/cuda/include/11/cuda_fp16.hpp(228): here
            function "__half::operator unsigned int() const"
/home/lakshayg/proj/numba-cuda/numba_cuda/numba/cuda/include/11/cuda_fp16.hpp(231): here
            function "__half::operator long long() const"
/home/lakshayg/proj/numba-cuda/numba_cuda/numba/cuda/include/11/cuda_fp16.hpp(234): here
            function "__half::operator unsigned long long() const"
/home/lakshayg/proj/numba-cuda/numba_cuda/numba/cuda/include/11/cuda_fp16.hpp(237): here
            function "__half::operator __nv_bool() const"
/home/lakshayg/proj/numba-cuda/numba_cuda/numba/cuda/include/11/cuda_fp16.hpp(241): here
```

Since there's no direct conversion to `char`, it has multiple options for choosing an intermediate type.

numba_cuda/numba/cuda/tests/cudapy/test_casting.py
isVoid: Inside the CUDA-Python projects, there are pre-existing practice in checking against SASS, for example in CCCL:

https://github.com/NVIDIA/cccl/blob/39bd2ec64e28a9c6f5f42dce23caf30610dc072f/python/cuda_cccl/cuda/cccl/parallel/experimental/_cccl_interop.py#L332-L333

numba_cuda/numba/cuda/tests/cudapy/test_intrinsics.py
```diff
@@ -561,13 +560,13 @@ def foo(out):
 
     def test_popc_u1(self):
         compiled = cuda.jit("void(int32[:], uint8)")(simple_popc)
-        ary = np.zeros(1, dtype=np.int8)
+        ary = np.zeros(1, dtype=np.int32)
```
lakshayg, line 564: This bug was detected using `compute-sanitizer`

numba_cuda/numba/cuda/tests/cudapy/test_operator.py
```diff
@@ -426,12 +431,13 @@ def test_fp16_int8_comparison_ptx(self):
             operator.lt: "setp.lt.f16",
             operator.le: "setp.le.f16",
             operator.eq: "setp.eq.f16",
-            operator.ne: "setp.ne.f16",
+            operator.ne: "setp.neu.f16",
```
lakshayg, line 442: The C++ implementation uses `setp.neu.fp16`. The original implementation was incorrect.

numba_cuda/numba/cuda/tests/cudapy/test_casting.py
lakshayg: The LTO PTX feature resolved this issue.

PR #272, opened 2025-06-02
Add support for launch bounds

gmarkall:
Adds support for launch bounds, specified as the `launch_bounds` kwarg of the `@cuda.jit` decorator or `cuda.compile()` function. The API design is intended to follow that of the [Launch Bounds API in CUDA C/C++](https://docs.nvidia.com/cuda/archive/12.9.0/cuda-c-programming-guide/index.html#launch-bounds).

Bounds may be specified as a scalar or a tuple of between one and three items. Tuple items provide:

- The maximum number of threads per block,
- The minimum number of blocks per SM,
- The maximum number of blocks per cluster.

If a scalar is provided instead, it is used as the maximum number of threads per block.

A sketch of some possile uses of the API in the `@cuda.jit` decorator:

```python
# Specify max threads per block only:
@cuda.jit(launch_bounds=128)
def kernel(...):
    ...

# Specify max threads per block and minimum blocks per SM:
@cuda.jit(launch_bounds=(max_threads_per_block, min_blocks_per_sm))
def kernel(...):
    ...

# Specify all three kinds of bounds:
@cuda.jit(launch_bounds=(max_threads_per_block, min_blocks_per_sm, max_blocks_per_cluster))
def kernel(...):
    ...
```

cc @ZzEeKkAa.

### DIFF
--- a/numba_cuda/numba/cuda/compiler.py
+++ b/numba_cuda/numba/cuda/compiler.py
@@ -630,7 +630,7 @@ def compile(
 
                           If a scalar is provided, it is used as the maximum
                           number of threads per block.
-    :type launch_bounds: (int, tuple)
+    :type launch_bounds: int or tuple[int]
     :return: (code, resty): The compiled code and inferred return type
     :rtype: tuple
     """
--- a/numba_cuda/numba/cuda/compiler.py
+++ b/numba_cuda/numba/cuda/compiler.py
@@ -630,7 +630,7 @@ def compile(
 
                           If a scalar is provided, it is used as the maximum
                           number of threads per block.
-    :type launch_bounds: int or tuple[int]
+    :type launch_bounds: int | tuple[int]
     :return: (code, resty): The compiled code and inferred return type
     :rtype: tuple
     """
--- a/numba_cuda/numba/cuda/compiler.py
+++ b/numba_cuda/numba/cuda/compiler.py
@@ -575,6 +575,7 @@ def compile(
     abi_info=None,
     output="ptx",
     forceinline=False,
+    launch_bounds=None,
 ):
     """Compile a Python function to PTX or LTO-IR for a given set of argument
     types.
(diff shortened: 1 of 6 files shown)


numba_cuda/numba/cuda/compiler.py
```diff
@@ -620,6 +621,16 @@ def compile(
...
+    :param launch_bounds: Kernel launch bounds, specified as a scalar or a tuple
+                          of between one and three items. Tuple items provide:
+
+                          - The maximum number of threads per block,
+                          - The minimum number of blocks per SM,
+                          - The maximum number of blocks per cluster.
+
+                          If a scalar is provided, it is used as the maximum
+                          number of threads per block.
+    :type launch_bounds: (int, tuple)
```
gmarkall: I'm not quite sure if this is the correct way to document that `launch_bounds` should be an `int` or `tuple` - does this suggest it's a tuple of `int` and a `tuple`?

kkraus14: I think Python typing would normally represent it as `Union[int, tuple[int]]` but otherwise I think you can just freeform it as `int or tuple[int]`?

gmarkall: Thanks - `int or tuple[int]` seems to have played well with the Sphinx docstring style used here.

ZzEeKkAa: maybe `int|tuple[int]`? This is how you define typing in signature

gmarkall: Thanks @ZzEeKkAa , that also works and looks more like a typing signature - I've changed to that.

numba_cuda/numba/cuda/cudadrv/nvvm.py
```diff
@@ -857,6 +857,48 @@ def set_cuda_kernel(function):
...
+    properties = (
+        # Max threads per block
+        "maxntidx",
+        # Min blocks per multiprocessor
+        "minctasm",
+        # Max blocks per cluster
+        "cluster_max_blocks",
+    )
+
+    for prop, bound in zip(properties, launch_bounds):
```
kkraus14, line 954: If someone passes a launch_bounds of greater length than 3, this zip will hide it maybe unexpectedly? Should we check that and error in that case?

kkraus14 (approved):

numba_cuda/numba/cuda/cudadrv/nvvm.py
gmarkall: We should - I've added a check and relevant tests.

kkraus14 (approved):

kkraus14:
@gmarkall anything else you wanted to do here? Otherwise this seems good to merge!

gmarkall:
> @gmarkall anything else you wanted to do here? Otherwise this seems good to merge!

I had no further plans - many thanks!

PR #273, opened 2025-06-04
Add CUDA 12.9 and new architectures

jakirkham:
Now that [CUDA 12.9 is out]( https://docs.nvidia.com/cuda/archive/12.9.0/cuda-toolkit-release-notes/index.html ), add it to the list. Also add [the new architectures]( https://docs.nvidia.com/cuda/archive/12.9.0/cuda-toolkit-release-notes/index.html#general-cuda ) for 10.3 and 12.1


kkraus14 (approved):

numba_cuda/numba/cuda/cudadrv/nvvm.py
```diff
@@ -369,37 +369,40 @@ def get_log(self):
     (9, 0),
     (10, 0),
     (10, 1),
+    (10, 3),
```
jakirkham, line 372: So 10.3 is only available on CUDA 12.9 (not 12.8). As a result it appears the CUDA 12.8 tests failed [on CI](  https://github.com/NVIDIA/numba-cuda/actions/runs/15431269306/job/43429522715?pr=273#step:9:2712 ):

```python
libnvvm : error: -arch=compute_103 is an unsupported option
NVVM_ERROR_INVALID_OPTION
```

So we need to exclude when 10.3 is used

numba_cuda/numba/cuda/cudadrv/nvvm.py
```diff
@@ -369,37 +369,40 @@ def get_log(self):
...
+    (11, 7): [((3, 5), (8, 7)),],
+    (11, 8): [((3, 5), (9, 0)),],
+    (12, 0): [((5, 0), (9, 0)),],
+    (12, 1): [((5, 0), (9, 0)),],
+    (12, 2): [((5, 0), (9, 0)),],
+    (12, 3): [((5, 0), (9, 0)),],
+    (12, 4): [((5, 0), (9, 0)),],
+    (12, 5): [((5, 0), (9, 0)),],
+    (12, 6): [((5, 0), (9, 0)),],
+    (12, 8): [((5, 0), (10, 2)), ((12, 0), (12, 0)),],
```
jakirkham: Have restructured how this mapping works to include a range of ranges. That way CUDA 12.8 is composed of two ranges:

1. A range from 5.0-10.2
2. A range from 12.0-12.0

As ranges are inclusive in the logic already, this picks up all CUDA compute capabilities for CUDA 12.8. While ignoring 10.3, which is only added in CUDA 12.9

numba_cuda/numba/cuda/cudadrv/nvvm.py
```diff
@@ -369,37 +369,40 @@ def get_log(self):
...
 
 def ccs_supported_by_ctk(ctk_version):
     try:
         # For supported versions, we look up the range of supported CCs
-        min_cc, max_cc = _CUDA_CC_MIN_MAX_SUPPORT[ctk_version]
         return tuple(
-            [cc for cc in COMPUTE_CAPABILITIES if min_cc <= cc <= max_cc]
+            [cc for cc in COMPUTE_CAPABILITIES if min_cc <= cc <= max_cc
+             for min_cc, max_cc in _CUDA_CC_MIN_MAX_SUPPORT[ctk_version]]
         )
```
jakirkham: To accomplish the list of ranges approach, only a slight adjustment in the `list`-comprehension gives us the desired behavior

jakirkham (commented):
CUDA 12.8 ran into CI failures as it was also picking up compute capability 10.3, which it doesn't support

Have rewritten the logic below to exclude 10.3 on CUDA 12.8. Included a few PR review comments below to explain how this was implemented

jakirkham:
Thanks for reviewing Keith! 🙏

Have fixed the CI failures as noted above

This is ready for another run/review when you are ready

numba_cuda/numba/cuda/cudadrv/nvvm.py
jakirkham: JFYI needed to flip the `for`-loops below: https://github.com/NVIDIA/numba-cuda/pull/273/commits/6211ce5b38ce53ea5464ce1322cdbcb82a20d1e5

Otherwise the idea is the same

numba_cuda/numba/cuda/cudadrv/nvvm.py
gmarkall: In case there is a CC 10.2 in the future, I think we should change the 12.8 supported toolkits not to include it:

```suggestion
    (12, 8): [((5, 0), (10, 1)), ((12, 0), (12, 0)),],
```

jakirkham: Oops yes. Must have been thinking of an exclusive upper bound

jakirkham: Fixed in commit: https://github.com/NVIDIA/numba-cuda/pull/273/commits/a791cca38f555d1ad647db4d47e9997037943cc4

jakirkham:
Thanks Graham! 🙏

Included the review feedback and pushed a fix for the style check

This is ready for another CI run

gmarkall (approved):
Looks good to me! (I'm planning to merge this once our CI is working again, since there are specifically changes to the CUDA 12.8 toolkit support, and that's the version that's presently broken in CI)

jakirkham:
Thanks Graham and Keith! 🙏

PR #275, opened 2025-06-04
Make 7.5 the default compute capability

gmarkall:
The [CUDA Programming Guide, v12.9, section 1.5.4](https://docs.nvidia.com/cuda/archive/12.9.0/cuda-toolkit-release-notes/index.html#deprecated-architectures), states:

> Maxwell, Pascal, and Volta architectures are now feature-complete with no further enhancements planned. While CUDA Toolkit 12.x series will continue to support building applications for these architectures, offline compilation and library support will be removed in the next major CUDA Toolkit version release. Users should plan migration to newer architectures, as future toolkits will be unable to target Maxwell, Pascal, and Volta GPUs.

This PR updates Numba-CUDA to use 7.5 as the default, for compatibility with future major versions. This is presently safe to do as all supported toolkits support CC 7.5 - when the default was set at 5.2, some supported toolkit versions only supported up to 7.0 or 7.2.

### DIFF
--- a/numba_cuda/numba/cuda/tests/cudadrv/test_nvvm_driver.py
+++ b/numba_cuda/numba/cuda/tests/cudadrv/test_nvvm_driver.py
@@ -30,7 +30,8 @@ class TestNvvmDriver(unittest.TestCase):
             self.skipTest("-gen-lto unavailable in this toolkit version")
 
         nvvmir = self.get_nvvmir()
-        ltoir = nvvm.compile_ir(nvvmir, opt=3, gen_lto=None, arch="compute_75")
+        arch = "compute_%d%d" % nvvm.LOWEST_CURRENT_CC
+        ltoir = nvvm.compile_ir(nvvmir, opt=3, gen_lto=None, arch=arch)
 
         # Verify we correctly passed the option by checking if we got LTOIR
         # from NVVM (by looking for the expected magic number for LTOIR)


gmarkall:
This will need updating to accommodate #273 once it is in.

numba_cuda/numba/cuda/cudadrv/nvvm.py
```diff
@@ -431,6 +428,21 @@ def get_log(self):
...
+#   further enhancements planned. While CUDA Toolkit 12.x series will continue
+#   to support building applications for these architectures, offline
+#   compilation and library support will be removed in the next major CUDA
+#   Toolkit version release. Users should plan migration to newer
+#   architectures, as future toolkits will be unable to target Maxwell, Pascal,
+#   and Volta GPUs."
+#
+# In order to maintain compatibility with future toolkits, we use Turing (7.5)
+# as the default CC if it is not otherwise specified.
+LOWEST_CURRENT_CC = (7, 5)
```
kkraus14, line 444: It would be great if we could query this information somehow in the future instead of having to hardcode it. Probably gets easier once we depend on `cuda.bindings`?

numba_cuda/numba/cuda/tests/cudadrv/test_nvvm_driver.py
```diff
@@ -30,7 +30,7 @@ def test_nvvm_compile_nullary_option(self):
             self.skipTest("-gen-lto unavailable in this toolkit version")
 
         nvvmir = self.get_nvvmir()
-        ltoir = nvvm.compile_ir(nvvmir, opt=3, gen_lto=None, arch="compute_52")
+        ltoir = nvvm.compile_ir(nvvmir, opt=3, gen_lto=None, arch="compute_75")
```
kkraus14: Should we use `LOWEST_CURRENT_CC` here instead of hardcoding to `compute_75`?

numba_cuda/numba/cuda/tests/cudadrv/test_nvvm_driver.py
```diff
@@ -138,9 +138,9 @@ def test_nvvm_warning(self):
 class TestArchOption(unittest.TestCase):
     def test_get_arch_option(self):
         # Test returning the nearest lowest arch.
-        self.assertEqual(nvvm.get_arch_option(5, 3), "compute_53")
         self.assertEqual(nvvm.get_arch_option(7, 5), "compute_75")
         self.assertEqual(nvvm.get_arch_option(7, 7), "compute_75")
```
kkraus14, line 143: +1 should we make these more programmatic to guard against future failures?

kkraus14 (approved):

numba_cuda/numba/cuda/cudadrv/nvvm.py
gmarkall: It would be great - I'm finding it hard to see how to query it though, because this is the knowledge used as a fallback when we encounter an unknown future toolkit - so I'm not sure how easy it would be to get the right answer here in future, even with `cuda.bindings`.

numba_cuda/numba/cuda/tests/cudadrv/test_nvvm_driver.py
gmarkall: Yeah, I think that would be safe / future-proof.

numba_cuda/numba/cuda/tests/cudadrv/test_nvvm_driver.py
gmarkall: That could be done, but as it's changed once in 5 years I'm not sure it's worth the effort to programmatically figure out valid cases for these tests. (I'm also not sure this will continue to exist once we're using `cuda.core` / `cuda.bindings` more heavily.)

numba_cuda/numba/cuda/cudadrv/nvvm.py
leofang: If NVRTC is a hard dependency, perhaps try the first arch reported by `nvrtc.nvrtcGetSupportedArchs()`?

numba_cuda/numba/cuda/tests/cudadrv/test_nvvm_driver.py
gmarkall: I just noticed, this will only start to fail in conjunction with code changes in `nvvm.py`, because it's just testing logic in there. So it will naturally only need updating at the time where `nvvm.py` changes (if it exists / remains in future).

numba_cuda/numba/cuda/tests/cudadrv/test_nvvm_driver.py
gmarkall: (now done)

numba_cuda/numba/cuda/cudadrv/nvvm.py
kkraus14: This is good to go so I'm going to merge this now and open this as a follow up issue.

PR #278, opened 2025-06-05
Add optional dependencies 

### LINKED ISSUE #277: [BUG] numba-cuda does not support pure CTK wheel users
#155 was only half-done for numba-cuda's wheel support story. During an offline meeting, we agreed to add optional dependencies so that users can do, e.g.
```shell
$ pip install numba-cuda[cu12]
```
to get a self-consistent CUDA 12 environment set up. This was not done.

This issue is compounded by the custom steps in setting up the wheel CI:
https://github.com/NVIDIA/numba-cuda/blob/0587e400a28ae4ff00fd0fa660abfa5656e86996/ci/test_wheel_deps_wheels.sh#L13-L21
which is not documented in the installation guide and does not mimic end user experience.

leofang:
Close #277.

### DIFF
--- a/ci/test_wheel_deps_wheels.sh
+++ b/ci/test_wheel_deps_wheels.sh
@@ -9,25 +9,11 @@ declare -A CTK_CURAND_VMAP=( ["12.8"]="10.3.9" ["12.9"]="10.3.10")
 CUDA_VER_MAJOR_MINOR=${CUDA_VER%.*}
 CURAND_VER="${CTK_CURAND_VMAP[${CUDA_VER_MAJOR_MINOR}]}"
 
-rapids-logger "Install testing dependencies"
-# TODO: Replace with rapids-dependency-file-generator
-python -m pip install \
-    psutil \
-    cffi \
-    "cuda-python==${CUDA_VER_MAJOR_MINOR}.*" \
-    "nvidia-cuda-runtime-cu12==${CUDA_VER_MAJOR_MINOR}.*" \
-    "nvidia-curand-cu12==${CURAND_VER}.*" \
-    "nvidia-cuda-nvcc-cu12==${CUDA_VER_MAJOR_MINOR}.*" \
-    "nvidia-cuda-nvrtc-cu12==${CUDA_VER_MAJOR_MINOR}.*" \
-    pynvjitlink-cu12 \
-    pytest
-
-
-rapids-logger "Install wheel"
+rapids-logger "Install wheel with test dependencies"
 package=$(realpath wheel/numba_cuda*.whl)
 echo "Package path: $package"
-python -m pip install $package
-
+# TODO: control version pinning to honor TEST_MATRIX once the cuda-toolkit metapackage is up
+python -m pip install $package[cu12,test]
 
 rapids-logger "Build tests"
 PY_SCRIPT="
--- a/ci/test_wheel_deps_wheels.sh
+++ b/ci/test_wheel_deps_wheels.sh
@@ -7,13 +7,14 @@ set -euo pipefail
 # appropriate cuRAND versions
 declare -A CTK_CURAND_VMAP=( ["12.8"]="10.3.9" ["12.9"]="10.3.10")
 CUDA_VER_MAJOR_MINOR=${CUDA_VER%.*}
+CUDA_VER_MAJOR=${CUDA_VER%.*.*}
 CURAND_VER="${CTK_CURAND_VMAP[${CUDA_VER_MAJOR_MINOR}]}"
 
 rapids-logger "Install wheel with test dependencies"
 package=$(realpath wheel/numba_cuda*.whl)
 echo "Package path: ${package}"
-# TODO: control version pinning to honor TEST_MATRIX once the cuda-toolkit metapackage is up
-python -m pip install "${package}"[cu12,test]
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -20,6 +20,21 @@ license = { text = "BSD 2-clause" }
...
+    "cuda-python==12.9.*",  # supports all CTK 12.x
+    "nvidia-cuda-nvcc-cu12",  # for libNVVM
+    "nvidia-cuda-runtime-cu12",
+    "nvidia-cuda-nvrtc-cu12",
+]
+test = [
+    "psutil",
+    "cffi",
+    "nvidia-curand-cu12",
+    "pynvjitlink-cu12",
```
kkraus14, line 48: What if someone wants to test cu11?

leofang: My understanding is that currently the test-wheels-deps-wheels CI (where this script is executed) only runs with CUDA 12.8, due to this filter: 
https://github.com/NVIDIA/numba-cuda/blob/0587e400a28ae4ff00fd0fa660abfa5656e86996/.github/workflows/pr.yaml#L137

ci/test_wheel_deps_wheels.sh
```diff
@@ -9,25 +9,11 @@ declare -A CTK_CURAND_VMAP=( ["12.8"]="10.3.9" ["12.9"]="10.3.10")
...
-    "nvidia-cuda-nvcc-cu12==${CUDA_VER_MAJOR_MINOR}.*" \
-    "nvidia-cuda-nvrtc-cu12==${CUDA_VER_MAJOR_MINOR}.*" \
-    pynvjitlink-cu12 \
-    pytest
-
-
-rapids-logger "Install wheel"
+rapids-logger "Install wheel with test dependencies"
 package=$(realpath wheel/numba_cuda*.whl)
 echo "Package path: $package"
```
bdice:
```suggestion
echo "Package path: ${package}"
```

ci/test_wheel_deps_wheels.sh
```diff
@@ -9,25 +9,11 @@ declare -A CTK_CURAND_VMAP=( ["12.8"]="10.3.9" ["12.9"]="10.3.10")
...
-python -m pip install $package
-
+# TODO: control version pinning to honor TEST_MATRIX once the cuda-toolkit metapackage is up
+python -m pip install $package[cu12,test]
```
bdice:
```suggestion
python -m pip install "${package}"[cu12,test]
```

pyproject.toml
kkraus14: If we're introducing an optional dependency, that's somewhat inherently user facing, where we should probably follow suite with what we did for `cuda.core`: https://github.com/NVIDIA/cuda-python/blob/main/cuda_core/pyproject.toml#L51-L53?

leofang: Good idea. I split out `test-cu12` from `test` in commit 4dad6c25. This does make the CI scripts slightly more general.

leofang:
It seems the conda CI failures are related to the cuda-compat issues?

leofang:
It seems the CI is fixed!

pyproject.toml
```diff
@@ -20,6 +20,23 @@ license = { text = "BSD 2-clause" }
...
+]
+test = [
+    "psutil",
+    "cffi",
+    "pytest",
+]
+test-cu12 = [
+    "nvidia-curand-cu12",
+    "pynvjitlink-cu12",
+]
```
kkraus14: I guess we can always do `cu11` in a follow up, but it feels really weird we're only doing this for `cu12` and not `cu11`.

gmarkall (approved):

pyproject.toml
leofang: added in commit b14dbe1

docs/source/user/overview.rst
```diff
@@ -84,7 +88,7 @@ CUDA Bindings
 ~~~~~~~~~~~~~
 
 Numba supports interacting with the CUDA Driver API via the `NVIDIA CUDA Python
-bindings <https://nvidia.github.io/cuda-python/>`_ and its own ctypes-based
+bindings <https://nvidia.github.io/cuda-python/>`_ and its own Cython-based
```
kkraus14: I think this should read "or its own ctypes-based". I think this is saying it can use either `cuda-bindings` or its own ctypes layer as of now.

pyproject.toml
```diff
@@ -20,6 +20,32 @@ license = { text = "BSD 2-clause" }
...
+    "cffi",
+    "pytest",
+]
+test-cu11 = [
+    "nvidia-curand-cu11",
+]
+test-cu12 = [
+    "nvidia-curand-cu12",
+    "pynvjitlink-cu12",
+]
```
kkraus14: I think we want to add `numba-cuda[test]` to both of these so that someone doesn't have to do `numba-cuda[test,test-cuXX]`?

docs/source/user/overview.rst
leofang: Ah, ok I parsed this sentence differently.. Let me clarify

leofang: see commit 0d5d314a

pyproject.toml
leofang: see commit 170bb71c6

PR #279, opened 2025-06-05
Drop `apt-get update` calls

jakirkham:
These are pulling in newer versions of `cuda-compat`, which appear to be missing some files. So simply drop these calls so that update doesn't occur. This should keep the `cuda-compat` version that is known to work.


jakirkham:
Superseded by PR: https://github.com/rapidsai/ci-imgs/pull/279

bdice:
I see this is already closed but I wanted to post here for clarity.

The `apt-get update` commands may be necessary to fetch package lists before we can remove the CUDA packages in subsequent lines, since `/var/lib/apt/lists/*` is typically deleted after running Docker commands.

In other words, `apt-get update` isn't the culprit here (it would have to call `apt-get upgrade` to actually get newer package versions). We should keep this as-is, and merge https://github.com/rapidsai/ci-imgs/pull/279 instead.

PR #284, opened 2025-06-09
Enable NV bindings by default and add them as a dependency

brandon-b-miller:
This PR sets `NUMBA_CUDA_USE_NVIDIA_BINDING` to 1 by default and adds `cuda-core` and `cuda-bindings` as a dependency.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,14 +22,14 @@ dependencies = ["numba>=0.59.1"]
 
 [project.optional-dependencies]
 cu11 = [
-    "cuda-bindings[cu11]",
+    "cuda-bindings==11.8.*",
     "cuda-python==11.8.*",  # supports all CTK 11.x
     "nvidia-cuda-nvcc-cu11",  # for libNVVM
     "nvidia-cuda-runtime-cu11",
     "nvidia-cuda-nvrtc-cu11",
 ]
 cu12 = [
-    "cuda-bindings[cu12]",
+    "cuda-bindings==12.9.*",
     "cuda-python==12.9.*",  # supports all CTK 12.x
     "nvidia-cuda-nvcc-cu12",  # for libNVVM
     "nvidia-cuda-runtime-cu12",


brandon-b-miller:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/284/commits/f4ec9917b2d76a6ecb7335bb2607313de69fc243

gmarkall:
I don't quite know how the dependencies work in `meta.yaml` and `pyproject.toml`, but it seems the changes here are causing all the CUDA toolkit packages to get upgraded to 12.9 when Numba-CUDA is installed for testing.

pyproject.toml
```diff
@@ -18,7 +18,7 @@ authors = [
 ]
 license = { text = "BSD 2-clause" }
 requires-python = ">=3.9"
-dependencies = ["numba>=0.59.1"]
+dependencies = ["numba>=0.59.1", "cuda-bindings>=12.9.0", "cuda-core==0.2.0"]
```
leofang: Please don't do this. This will force everyone to use CTK wheels. Dependency management is additive not subtractive, so we enable wheels users through the `cuXX` optional dependencies listed below, while keeping the "base" dependencies minimal for local CTK users.

leofang: xref: https://github.com/NVIDIA/numba-cuda/issues/277

leofang: Let's add `cuda-core[cu12]` to `cu12` below, xref: https://github.com/NVIDIA/numba-cuda/pull/133#discussion_r2130191438. (Same for `cu11`.) It'd handle the wheel dependencies right.

brandon-b-miller:
Looks like just a couple of failures related to some more recent code through the new binding. Looking into it.

leofang (approved):
Thanks, Brandon!

conda/recipes/numba-cuda/meta.yaml
```diff
@@ -28,6 +28,7 @@ requirements:
   run:
     - python
     - numba >=0.59.1
+    - cuda-core
```
jakirkham: Should we either make this optional (like wheels) or should we also add `cuda-python` (as wheels do with their CUDA 11/12)?

jakirkham (commented):
Thanks Brandon! 🙏

Had a question below

pyproject.toml
```diff
@@ -22,12 +22,14 @@ dependencies = ["numba>=0.59.1"]
 
 [project.optional-dependencies]
 cu11 = [
+    "cuda-core[cu11]",
```
kkraus14: We should pin these to a specific version for the time being. We don't have any API guarantees in `cuda-core` yet and it's very likely we will break things out from underneath you.

gmarkall: Is this PR using `cuda-core` at all? What's the reason for adding it in this PR, instead of `cuda-python` (as @jakirkham mentions above in https://github.com/NVIDIA/numba-cuda/pull/284/files#r2136485402)

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -82,25 +73,43 @@
...
-elif _pynvjitlink_enabled_in_cfg is not None:
-    ENABLE_PYNVJITLINK = _pynvjitlink_enabled_in_cfg
-else:
-    ENABLE_PYNVJITLINK = importlib.util.find_spec("pynvjitlink") is not None
-
 if not hasattr(config, "CUDA_ENABLE_PYNVJITLINK"):
     config.CUDA_ENABLE_PYNVJITLINK = ENABLE_PYNVJITLINK
 
 
+USE_NV_BINDING = _get_option("CUDA_USE_NVIDIA_BINDING", fallback=False)
```
gmarkall: If the fallback is `False`, doesn't this make the NVIDIA binding disabled by default?

brandon-b-miller: I was going to blame this on a cosmic ray before I realized the whole logic is actually bad. Numba upstream sets a default value of zero for `NUMBA_CUDA_USE_NVIDIA_BINDING` but not for `NUMBA_CUDA_ENABLE_PYNVJITLINK`. I put back the old logic for the latter env var and invented new, hopefully correct this time logic for the former.

pyproject.toml
brandon-b-miller: It should have been `cuda-bindings`, sorry. Linker PR changes creeping in.

kkraus14: As far as I know it is not, but https://github.com/NVIDIA/numba-cuda/pull/133 will pull it in. Right now there isn't a reason for use `cuda-core` vs `cuda-python`, but in the near future we intend to have more frequent releases of `cuda-core` vs `cuda-python`, which would allow us to pull in new versions sooner to get bugfixes / new features as needed.

brandon-b-miller:
Something's not quite right with a few of the tests here still, apologies for the churn. Will circle back here later today.

gmarkall:
> Something's not quite right with a few of the tests here still, apologies for the churn. Will circle back here later today.

I think we've not really been testing with the NVIDIA bindings, so it's quite plausible that some of it broke that may need fixing up - you may be looking at genuine issues here.

brandon-b-miller:
The issue was that we really need to set all our config options in the main init, otherwise things that check the config on import see the value before it gets set.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -174,7 +157,7 @@ def locate_driver_and_loader():
             )
         if not os.path.isfile(envpath):
             raise ValueError(
-                "NUMBA_CUDA_DRIVER %s is not a valid file "
+                "NUMBA_CUDA_DR IVER %s is not a valid file "
```
gmarkall: An errant space:

```suggestion
                "NUMBA_CUDA_DRIVER %s is not a valid file "
```

gmarkall (approved):
Looks good to me (modulo the typo that was introduced in `NUMBA_CUDA_DRIVER`).

pyproject.toml
```diff
@@ -22,12 +22,14 @@ dependencies = ["numba>=0.59.1"]
 
 [project.optional-dependencies]
 cu11 = [
+    "cuda-bindings[cu11]",
```
kkraus14: `cu11` doesn't exist as an optional dependency of the `cuda-bindings` package. There's the `all` optional dependency you can use that includes `nvcc`, `nvrtc`, and `nvjitlink` if that's what you're looking for? Otherwise you can use the same `all` optional dependency on the `cuda-python` package and it will pull `cuda-bindings[all]`.

kkraus14 (approved):

conda/recipes/numba-cuda/meta.yaml
```diff
@@ -28,6 +28,7 @@ requirements:
   run:
     - python
     - numba >=0.59.1
+    - cuda-bindings
```
kkraus14, line 31: Just to confirm, for conda we're making this a required dependency whereas for `pip` we're making it an optional dependency. Is this intended?

brandon-b-miller: Since we're using the bindings by default, it doesn't feel perfect to me that they be optional anywhere - but based on https://github.com/NVIDIA/numba-cuda/pull/284#discussion_r2135880815 the goal IIUC is to be sensitive to what packages we pull in for a minimal installation.

numba_cuda/numba/cuda/__init__.py
```diff
@@ -1,5 +1,36 @@
...
+# explicitly disabled, do not use pynvjitlink, even if present in the env.
+_pynvjitlink_enabled_in_env = _readenv(
+    "NUMBA_CUDA_ENABLE_PYNVJITLINK", bool, None
+)
+_pynvjitlink_enabled_in_cfg = getattr(config, "CUDA_ENABLE_PYNVJITLINK", None)
+
+if _pynvjitlink_enabled_in_env is not None:
+    ENABLE_PYNVJITLINK = _pynvjitlink_enabled_in_env
+elif _pynvjitlink_enabled_in_cfg is not None:
+    ENABLE_PYNVJITLINK = _pynvjitlink_enabled_in_cfg
```
bdice, line 17: The ordering of this logic makes it sound like environment variables can override the config, but the check below (`if not hasattr(config, "CUDA_ENABLE_PYNVJITLINK")` indicates that it does _not_ override the config. Can we reorder this for clarity? Better yet, can we avoid setting the config again if the config is already set (since that is treated as the highest override)?

kkraus14 (approved):

numba_cuda/numba/cuda/__init__.py
kkraus14: Merging this where we can address this as a follow up.

PR #287, opened 2025-06-10
Fix lineinfo generation when compile_internal used (#271)

### LINKED ISSUE #271: [BUG] Setting lineinfo=true for a kernel also enable debug building.
**Describe the bug**
If I add the `lineinfo=true` parameter to a `@numba.cuda.jit()` decorator, then the kernel is generated usinga a debug build instead of an optimized build, resulting in different performance. 

1. Create the following python program using `numpy` and `numba-cuda`.
```python
import numpy as np
import numba.cuda

@numba.cuda.jit(opt=True)
def compute_mag(inp: np.ndarray, out: np.ndarray) -> None:
    idx = numba.cuda.blockIdx.x * numba.cuda.blockDim.x + numba.cuda.threadIdx.x

    out[idx] = abs(inp[idx])


arr1 = np.random.random((256000))
arr2 = np.random.random((256000))
cplx = arr1 + arr2 * 1j

result = np.zeros_like(arr1, dtype=np.float32)


compute_mag[1000, 256](cplx, result)
```

2. Profile the script using Nsight compute.

3. Modify the above program by adding the `lineinfo=True` option to the decorator on the kernel function so it looks like this:
```python
import numpy as np
import numba.cuda


@numba.cuda.jit(opt=True, lineinfo=True)
def compute_mag(inp: np.ndarray, out: np.ndarray) -> None:
    idx = numba.cuda.blockIdx.x * numba.cuda.blockDim.x + numba.cuda.threadIdx.x

    out[idx] = abs(inp[idx])


arr1 = np.random.random((256000))
arr2 = np.random.random((256000))
cplx = arr1 + arr2 * 1j

result = np.zeros_like(arr1, dtype=np.float32)


compute_mag[1000, 256](cplx, result)
```

4. Profile the modified script in Nsight compute.
5. Compare the generated SASS between the two runs, as well as the kernel runtime.

**Expected behavior**
* The generated SASS for the two runs should be identical.
* The run times should be similar.
* The only major difference is that in the second run is able to also show python source mapping for the kernel

**Actual Behavior**
The second run is much slower, and has many more generated SASS instructions, due to being built in debug mode.

<details>
<summary> SASS source comparison in Nsight compute</summary>

Image

</details>

 - Environment location: Running on a Windows 10 computer.
 - GPU: RTX5070
 - Python version: 3.13.3 (With GIL)
 - Method of numba-cuda install: `pip install numba-cuda`

**Additional context**
In CUDA C++ via Visual Studio, I am able to build a program where lineinfo is enabled, but the build is still optimized, unlike what is happening here.

gmarkall:
Lineinfo generation is broken by function implementations that generate code via `context.compile_internal()`. The root cause is that the implementation eventually reaches upstream Numba's `BaseContext._compile_subroutine_no_cache()` method, which ignores the flags in the context stack and creates new ones. The outcome of this is that the debug info kind is forgotten, leading to a default debug info kind of `"FullDebug"` being emitted, which then enables the PTX `debug` target, leading to deoptimized code.

This change works around the issue (pending a fix upstream) by overriding the `_compile_subroutine_no_cache()` implementation to use flags from the context stack when they are otherwise not provided. The fix to upstream will look like a similar modification of the method.

The `CUDAFlags` class is moved to its own module to avoid circular import dependencies between `compiler.py` and `target.py`.

Fixes #271.

### DIFF
--- a/numba_cuda/numba/cuda/compiler.py
+++ b/numba_cuda/numba/cuda/compiler.py
@@ -14,8 +14,6 @@ from numba.core.compiler import (
     sanitize_compile_result_entries,
     CompilerBase,
     DefaultPassBuilder,
-    Flags,
-    Option,
     CompileResult,
 )
 from numba.core.compiler_lock import global_compiler_lock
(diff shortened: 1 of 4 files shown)


gmarkall:
This is confirmed to resolve the original issue reported in the user's environment: https://github.com/NVIDIA/numba-cuda/issues/271#issuecomment-2961333462

Issue #280, opened 2025-06-06
[FEA] Plumb kernel launch configuration such that it's accessible from rewriting and lowering phases


tpn:
Consider:

```python
threads_per_block = 32
items_per_thread = 4

@cuda.jit
def kernel1(d_in, d_out):
    thread_data = cuda.local.array(items_per_thread, dtype=dtype)
    cuda.block.load(
        d_in,
        thread_data,
        threads_per_block,
        items_per_thread,
    )
    cuda.block.store(
        d_out,
        thread_data,
        threads_per_block,
        items_per_thread,
    )
```

This is a common, canonical example for interacting with `cuda.coop` today.  Unfortunately, having the block dimensions `threads_per_block` (`blockDim`) be derived from a global/non-local/freevar prevents a user from writing a generic kernel that can be shared with other consumers (i.e. if they have a library which ships pre-written Numba CUDA kernels).

A workaround would be to plumb these items as kernel arguments, e.g.:

```
@cuda.jit
def kernel1(d_a, d_b, threads_per_block, items_per_thread):
    ...
```

That's fine for `items_per_thread`, but it means you'd have to duplicate `threads_per_block`, like this:

```python
kernel1[1, threads_per_block](d_in, d_out, threads_per_block, items_per_thread)
```

The ideal arrangement would look like this:

```python

@cuda.jit
def kernel(d_in, d_out, items_per_thread):
     thread_data = cuda.local.array(items_per_thread, dtype=d_in.dtype)
     cuda.block.load(
        d_in,
        thread_data,
        items_per_thread,
    )
    cuda.block.store(
        d_out,
        thread_data,
        items_per_thread,
    )

k = kernel1[1, threads_per_block]
# XXX ^ k = _LaunchConfig(), has all the juicy stuff like dims & smem.
k(d_a, d_b, items_per_thread=4)
```

At the `XXX` point above, there exists a `_LaunchConfiguration()` object which has all of the information we'd want downstream during rewriting and lowering (`blockDim`, smem, etc.).  Thus, it would be beneficial to have an official way of accessing that information from downstream (rewriting, lowering) contexts.

I am happy to take on the development for this work.

tpn:
PR: https://github.com/NVIDIA/numba-cuda/pull/288

PR #288, opened 2025-06-10
Implement a thread-local means to access kernel launch config.

tpn:
This allows downstream passes, such as rewriting, to access information about the kernel launch for which they have been enlisted to participate.

Posting this as a PR now to get feedback on the overall approach.  Assuming this solution is acceptable, I'll follow up with tests and docs.

### DIFF
--- a/numba_cuda/numba/cuda/dispatcher.py
+++ b/numba_cuda/numba/cuda/dispatcher.py
@@ -26,6 +26,7 @@ from numba.cuda.errors import (
     missing_launch_config_msg,
     normalize_kernel_dimensions,
 )
+from numba.cuda.launchconfig import launch_config_ctx
 from numba.cuda import types as cuda_types
 from numba.cuda.locks import module_init_lock
 from numba.cuda.memory_management.nrt import rtsys, NRT_LIBRARY
@@ -830,10 +831,16 @@ class CUDADispatcher(Dispatcher, serialize.ReduceMixin):
         """
         Compile if necessary and invoke this kernel with *args*.
         """
-        if self.specialized:
-            kernel = next(iter(self.overloads.values()))
-        else:
-            kernel = _dispatcher.Dispatcher._cuda_call(self, *args)
+        with launch_config_ctx(
+            griddim=griddim,
+            blockdim=blockdim,
+            stream=stream,
+            sharedmem=sharedmem,
+        ):
+            if self.specialized:
+                kernel = next(iter(self.overloads.values()))
+            else:
+                kernel = _dispatcher.Dispatcher._cuda_call(self, *args)
 
         kernel.launch(args, griddim, blockdim, stream, sharedmem)
 
--- a/numba_cuda/numba/cuda/launchconfig.py
+++ b/numba_cuda/numba/cuda/launchconfig.py
@@ -1,0 +1,62 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from contextvars import ContextVar
+from contextlib import contextmanager
+from typing import Any, Tuple, Optional
+
+
+@dataclass(frozen=True, slots=True)
+class LaunchConfig:
+    """
+    Helper class used to encapsulate kernel launch configuration for storing
+    and retrieving from a thread-local ContextVar.
+    """
+
+    griddim: Tuple[int, int, int]
+    blockdim: Tuple[int, int, int]
+    stream: Any
+    sharedmem: int
+
+    def __str__(self) -> str:
+        g = "×".join(map(str, self.griddim))
+        b = "×".join(map(str, self.blockdim))
+        return (
+            f"<LaunchConfig grid={g}, block={b}, "
+            f"stream={self.stream}, smem={self.sharedmem}B>"
+        )
+
+
+_launch_config_var: ContextVar[Optional[LaunchConfig]] = ContextVar(
+    "_launch_config_var",
+    default=None,
+)
+
+
+def current_launch_config() -> Optional[LaunchConfig]:
+    """
+    Read the launch config visible in *this* thread/asyncio task.
+    Returns None if no launch config is set.
+    """
+    return _launch_config_var.get()
+
+
+@contextmanager
+def launch_config_ctx(
+    *,
+    griddim: Tuple[int, int, int],
+    blockdim: Tuple[int, int, int],
+    stream: Any,
+    sharedmem: int,
+):
+    """
+    Install a LaunchConfig for the dynamic extent of the with-block.
+    The previous value (if any) is restored automatically.
+    """
+    token = _launch_config_var.set(
+        LaunchConfig(griddim, blockdim, stream, sharedmem)
+    )
+    try:
+        yield
+    finally:
+        _launch_config_var.reset(token)


tpn:
Will close #280.

gmarkall:
Using the simple benchmark from https://github.com/numba/numba/issues/3003#issuecomment-627872661:

```python
from numba import cuda
import numpy as np


@cuda.jit('void()')
def some_kernel_1():
    return

@cuda.jit('void(float32[:])')
def some_kernel_2(arr1):
    return

@cuda.jit('void(float32[:],float32[:])')
def some_kernel_3(arr1,arr2):
    return

@cuda.jit('void(float32[:],float32[:],float32[:])')
def some_kernel_4(arr1,arr2,arr3):
    return

@cuda.jit('void(float32[:],float32[:],float32[:],float32[:])')
def some_kernel_5(arr1,arr2,arr3,arr4):
    return

arr = cuda.device_array(10000, dtype=np.float32)

%timeit some_kernel_1[1, 1]()
%timeit some_kernel_2[1, 1](arr)
%timeit some_kernel_3[1, 1](arr,arr)
%timeit some_kernel_4[1, 1](arr,arr,arr)
%timeit some_kernel_5[1, 1](arr,arr,arr,arr)
```

On `main`:

```
6.6 μs ± 11.3 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
10.6 μs ± 79.9 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
14.1 μs ± 65.8 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
17.4 μs ± 236 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
20.3 μs ± 38.5 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

On this branch:

```
8.6 μs ± 70.7 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
13 μs ± 41.3 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
17.2 μs ± 152 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
20.3 μs ± 35.9 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
24.4 μs ± 44.4 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```
From this crude benchmark, it appears to add 2-4μs per launch, or 17-23% overhead. I don't yet know how we should consider this (or whether the benchmark is appropriate) but I think it's a consideration to keep in mind when thinking about this approach.

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -830,12 +831,20 @@ def call(self, args, griddim, blockdim, stream, sharedmem):
...
-            kernel = _dispatcher.Dispatcher._cuda_call(self, *args)
+        with launch_config_ctx(
+            dispatcher=self,
+            args=args,
+            griddim=griddim,
+            blockdim=blockdim,
+            stream=stream,
+            sharedmem=sharedmem,
+        ):
+            if self.specialized:
```
gmarkall, line 1622: Specialized kernels cannot be recompiled, so a new launch configuration would not be able to affect the compilation of a new version - so this check could be kept outside the context manager.

numba_cuda/numba/cuda/launchconfig.py
```diff
@@ -0,0 +1,70 @@
...
+from contextvars import ContextVar
+from contextlib import contextmanager
+from typing import Any, Tuple, Optional, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from numba.cuda.dispatcher import CUDADispatcher
+
+
+@dataclass(frozen=True, slots=True)
+class LaunchConfig:
```
gmarkall, line 20: There seems to be quite some overlap with `dispatcher._LaunchConfiguration` in this class (an observation at this point - I don't know whether it makes sense to combine them)

kkraus14:
> Using the simple benchmark from [numba/numba#3003 (comment)](https://github.com/numba/numba/issues/3003#issuecomment-627872661):
> 
> ```python
> from numba import cuda
> import numpy as np
> 
> 
> @cuda.jit('void()')
> def some_kernel_1():
>     return
> 
> @cuda.jit('void(float32[:])')
> def some_kernel_2(arr1):
>     return
> 
> @cuda.jit('void(float32[:],float32[:])')
> def some_kernel_3(arr1,arr2):
>     return
> 
> @cuda.jit('void(float32[:],float32[:],float32[:])')
> def some_kernel_4(arr1,arr2,arr3):
>     return
> 
> @cuda.jit('void(float32[:],float32[:],float32[:],float32[:])')
> def some_kernel_5(arr1,arr2,arr3,arr4):
>     return
> 
> arr = cuda.device_array(10000, dtype=np.float32)
> 
> %timeit some_kernel_1[1, 1]()
> %timeit some_kernel_2[1, 1](arr)
> %timeit some_kernel_3[1, 1](arr,arr)
> %timeit some_kernel_4[1, 1](arr,arr,arr)
> %timeit some_kernel_5[1, 1](arr,arr,arr,arr)
> ```
> 
> On `main`:
> 
> ```
> 6.6 μs ± 11.3 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
> 10.6 μs ± 79.9 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
> 14.1 μs ± 65.8 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
> 17.4 μs ± 236 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
> 20.3 μs ± 38.5 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
> ```
> 
> On this branch:
> 
> ```
> 8.6 μs ± 70.7 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
> 13 μs ± 41.3 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
> 17.2 μs ± 152 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
> 20.3 μs ± 35.9 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
> 24.4 μs ± 44.4 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
> ```
> 
> From this crude benchmark, it appears to add 2-4μs per launch, or 17-23% overhead. I don't yet know how we should consider this (or whether the benchmark is appropriate) but I think it's a consideration to keep in mind when thinking about this approach.

Kernel launch latency is something we need to care about moving forward and is becoming more and more of a bottleneck in important workloads. Adding 2-4μs per launch for this is probably unacceptable, but that being said our launch latency right now is much higher than we'd like in general where we probably need to rework the entire launch path at some point in the not too distant future.

gmarkall:
> we probably need to rework the entire launch path at some point in the not too distant future.

I think that's another thing that concerns me - if we implement something like this, then it constrains how we can rework the launch path in future if we have to go on supporting it.

tpn:
Made obsolete by PR #804.

PR #292, opened 2025-06-12
Port change and test from numba/numba#9595 to Numba-CUDA

gmarkall:
This change and test will need to be reflected in Numba-CUDA to ensure that it continues to work well with upstream Numba once https://github.com/numba/numba/pull/9595 is in.


gmarkall:
I think this will need numba/numba#9595 to be merged and available in a release for it to work. Keeping this parked for now.

PR #293, opened 2025-06-12
[DOCS] Improve the introduction and installation section of the documentation

gmarkall:
The scope of these changes is to make it so that:

*  A user landing on the index page of the documentation gets a coherent, quick summary of Numba, and links to different parts of the document.
* Proceeding through the docs to the first page, they get installation instructions that are both relevant and correct
* Following that, the documentation continues with the "usual" introduction to writing CUDA kernels.

The Installation section here is mostly what was called "Overview" before - anything from that section that wasn't to do with installation was pushed into the next page that explains how to write kernels - I have not put much effort into improving that page here.

The Installation section should give users a good reference on how to install numba-cuda and have the required bindings installed and usable, or they should be able to disable the NVIDIA bindings if they don't want to use them.

### DIFF
--- a/docs/source/user/installation.rst
+++ b/docs/source/user/installation.rst
@@ -1,57 +1,34 @@
-========
-Overview
-========
+.. _numba-cuda-installation:
 
-Numba-CUDA supports programming NVIDIA CUDA GPUs by directly compiling a
-restricted subset of Python code into CUDA kernels and device functions
-following the CUDA execution model.
+============
+Installation
+============
 
-
-Terminology
-===========
-
-Several important terms in the topic of CUDA programming are listed here:
-
-- *host*: the CPU
-- *device*: the GPU
-- *host memory*: the system main memory
-- *device memory*: onboard memory on a GPU card
-- *kernels*: a GPU function launched by the host and executed on the device
-- *device function*: a GPU function executed on the device which can only be
-  called from the device (i.e. from a kernel or another device function)
-
-
-Programming model
-=================
-
-Most CUDA programming facilities exposed by Numba map directly to the CUDA
-C language offered by NVidia.  Therefore, it is recommended you read the
-official `CUDA C programming guide <http://docs.nvidia.com/cuda/cuda-c-programming-guide>`_.
-
+Requirements
+============
 
 Supported GPUs
-==============
+--------------
 
 Numba supports all NVIDIA GPUs that are supported by the CUDA Toolkit it uses.
 Presently for CUDA 11 this ranges from Compute Capabilities 3.5 to 9.0, and for
 CUDA 12 this ranges from 5.0 to 12.1, depending on the exact installed version.
 
 
-.. _numba-cuda-installation:
-
-Installation
-============
-
-CUDA Toolkits
--------------
+Supported CUDA Toolkits
+-----------------------
 
 Numba-CUDA aims to support all minor versions of the two most recent CUDA
 Toolkit releases. Presently 11 and 12 are supported; CUDA 11.2 is the minimum
 required, because older releases (11.0 and 11.1) have a version of NVVM based on
-previous and incompatible LLVM versions.
+a previous and incompatible LLVM version.
 
-For further information of version compatibility between toolkit and driver
+For further information about version compatibility between toolkit and driver
 versions, refer to :ref:`minor-version-compatibility`.
+
+
+Installation with a Python package manager
+==========================================
 
 Conda users can install the CUDA Toolkit into a conda environment.
 
@@ -71,24 +48,22 @@ or::
 
     $ pip install numba-cuda[cu11]
 
-If you are not using Conda or if you want to use a different version of CUDA
-toolkit, the following describes how Numba searches for a CUDA toolkit
-installation.
+If you are not using Conda/pip or if you want to use a different version of CUDA
+toolkit, :ref:`cudatoolkit-lookup` describes how Numba searches for a CUDA toolkit.
 
+
+Configuration
+=============
 
 .. _cuda-bindings:
 
 CUDA Bindings
-~~~~~~~~~~~~~
+-------------
 
 Numba supports interacting with the CUDA Driver API via either the `NVIDIA CUDA
 Python bindings <https://nvidia.github.io/cuda-python/>`_ or its own ctypes-based
 bindings. Functionality is equivalent between the two binding choices. The
 NVIDIA bindings are the default, and the ctypes bindings are now deprecated.
-
-ctypes-based bindings are presently the default, but the NVIDIA bindings will
-be used by default (if they are available in the environment) in a future Numba
-release.
 
 If the NVIDIA bindings are not present in your environment, you can install them
 with::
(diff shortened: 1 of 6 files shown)


docs/source/user/installation.rst
```diff
@@ -0,0 +1,111 @@
...
+
+
+Installation with a Python package manager
+==========================================
+
+Conda users can install the CUDA Toolkit into a conda environment.
+
+For CUDA 12, ``cuda-nvcc`` and ``cuda-nvrtc`` are required::
+
+    $ conda install -c conda-forge numba-cuda cuda-nvcc cuda-nvrtc "cuda-version>=12.0"
```
kkraus14: Should we consider just adding these to the conda recipe?

docs/source/user/installation.rst
```diff
@@ -0,0 +1,111 @@
...
+can be enabled by setting the environment variable
+:envvar:`NUMBA_CUDA_USE_NVIDIA_BINDING` to ``"0"``.
+
+
+.. _cudatoolkit-lookup:
+
+Setting CUDA Installation Path
+------------------------------
+
+Numba searches for a CUDA toolkit installation in the following order:
```
kkraus14: I think this order is for the ctypes based bindings and doesn't reflect the `cuda-bindings` based search order. For `cuda-bindings` it will follow the behavior described here: https://github.com/NVIDIA/cuda-python/tree/main/cuda_bindings/cuda/bindings/_path_finder#library-loading-search-priority

docs/source/user/kernels.rst
```diff
@@ -24,6 +30,28 @@ consider how to use and access memory in order to minimize bandwidth
...
+- *kernels*: a GPU function launched by the host and executed on the device
+- *device function*: a GPU function executed on the device which can only be
+  called from the device (i.e. from a kernel or another device function)
+
+
+Programming model
+=================
+
+Most CUDA programming facilities exposed by Numba map directly to the CUDA
+C language offered by NVidia.  Therefore, it is recommended you read the
```
kkraus14:
```suggestion
C language offered by NVIDIA.  Therefore, it is recommended you read the
```

docs/source/user/installation.rst
kkraus14: To add, once we depend on `cuda-bindings` in the conda recipe, `cuda-nvrtc` is already a dependency. What is `cuda-nvcc` needed for?

docs/source/user/installation.rst
```diff
@@ -0,0 +1,111 @@
...
+NVIDIA bindings are the default, and the ctypes bindings are now deprecated.
+
+If the NVIDIA bindings are not present in your environment, you can install them
+with::
+
+   $ conda install -c conda-forge cuda-bindings
+
+if you are using Conda, or::
+
+   $ pip install cuda-bindings[cu11]
```
brandon-b-miller: Is `cu11` an extra for `cuda-bindings`? Same with cu12 below

kkraus14: No it is not. We should be doing `cuda-bindings[all]==11` / `cuda-bindings[all]==12`. That being said, I think we should just advertise the `numba-cuda[cuXX]` path instead?

docs/source/user/installation.rst
gmarkall: I'm stretching my memory a bit here, but I think it might have been the only way to get libdevice installed (I'm not sure if that's still current).

docs/source/user/installation.rst
gmarkall: Thanks - I didn't realise this until our sync meeting today, after I made this PR. Will update.

docs/source/user/installation.rst
gmarkall: I would be happy to change this but I'm not clear what the exact correct thing to write should be. Can you make a suggestion please?

docs/source/user/installation.rst
kkraus14: If it's only for `libdevice` then that comes from the `cuda-nvvm-tools` conda package these days that we should depend on instead.

docs/source/user/installation.rst
leofang: I believe we can just keep the text but remove all conda/pip install commands from this section. In the previous section, we already taught users how to install things, and `cuda-bindings` will be installed there. So, here we just need to teach users how to fall back to ctypes.

docs/source/user/installation.rst
leofang: `cuda-nvcc` was probably used to provide libnvvm. I need to look it up and check which 12.x did we split it from nvcc-impl.

docs/source/user/installation.rst
leofang: Might be better to add a note explaining the search order might change in the future?

leofang:
@gmarkall the root README also has installation instructions that need to be updated 😅

docs/source/user/installation.rst
gmarkall: I just remembered - if you want to run the full test suite, you need to have nvcc installed to build the test binaries. I don't know whether this means it makes sense to have it installed for end-users though.

gmarkall:
> @gmarkall the root README also has installation instructions that need to be updated 😅

Thanks for the reminder - I want to handle that in a separate PR when I can get round to it. My thinking here was that a little improvement is better than no improvement because I didn't have time to do both.

docs/source/user/installation.rst
gmarkall: > Should we consider just adding these to the conda recipe?

If I run

```
$ mamba create -n test-install-numba-cuda-015 python=3.12 nvidia::numba-cuda
```

(using the NVIDIA channel only because 0.15.0 is not yet on conda-forge right now)

then it will install (amongst others)

```
  Package                     Version  Build                 Channel           Size
─────────────────────────────────────────────────────────────────────────────────────
  Install:
─────────────────────────────────────────────────────────────────────────────────────

 
  + cuda-version                 12.9  h4f385c5_3            conda-forge       22kB
  + cuda-nvvm-dev_linux-64    12.9.86  ha770c72_1            conda-forge       27kB
  + cuda-nvvm-tools           12.9.86  he02047a_1            conda-forge       24MB
  + cuda-nvvm-impl            12.9.86  he02047a_1            conda-forge       21MB
  + libnvjitlink              12.9.86  h5888daf_0            conda-forge       31MB
  + cuda-nvrtc                12.9.86  h5888daf_0            conda-forge       67MB
  + cuda-nvvm                 12.9.86  h69a702a_1            conda-forge       25kB
  + cuda-bindings              12.9.0  py312h77ce6f0_0       conda-forge     Cached
  + numba                      0.61.2  py312h7bcfee6_1       conda-forge     Cached
  + numba-cuda                 0.15.0  py_0                  nvidia          Cached
```

so it looks like `mamba install numba-cuda` is going to be sufficient to pull in all required dependencies for running Numba-CUDA - I'll simplify the docs here.

docs/source/user/installation.rst
gmarkall: I'll make it clear that the behaviour using the NVIDIA bindings follows path-finder's logic, and that the ctypes behaviour is for the deprecated path.

docs/source/user/installation.rst
gmarkall: Thanks Leo - I've now removed these, and it does read better.

gmarkall:
Thanks all - I believe I've now addressed all comments (and it does read better as a result!).

docs/source/user/installation.rst
```diff
@@ -0,0 +1,105 @@
...
+
+    $ conda install -c conda-forge numba-cuda "cuda-version>=12.0"
+
+Alternatively, you can install all CUDA 12 dependencies from PyPI via ``pip``::
+
+    $ pip install numba-cuda[cu12]
+
+For CUDA 11, ``cudatoolkit`` is required::
+
+    $ conda install -c conda-forge numba-cuda cudatoolkit "cuda-version>=11.2,<12.0"
```
kkraus14: `cudatoolkit` is already implicitly added from the `cuda-bindings` dependency we added.
```suggestion
    $ conda install -c conda-forge numba-cuda "cuda-version>=11.2,<12.0"
```

kkraus14 (approved):

leofang:
> > @gmarkall the root README also has installation instructions that need to be updated 😅
> 
> Thanks for the reminder - I want to handle that in a separate PR when I can get round to it. My thinking here was that a little improvement is better than no improvement because I didn't have time to do both.

Let me take care of this!

PR #294, opened 2025-06-12
Raise better error when `cuda.bindings` are not available

brandon-b-miller:
Raises a better error with instructions of what to do if the base install is used and no user provided cuda bindings are present in the environment.

### DIFF
--- a/numba_cuda/numba/cuda/__init__.py
+++ b/numba_cuda/numba/cuda/__init__.py
@@ -35,12 +35,12 @@ if config.CUDA_USE_NVIDIA_BINDING:
     if not importlib.util.find_spec("cuda.bindings"):
         raise ImportError(
             "CUDA bindings not found. Please install through "
-            "the cuda-bindings[cuXY] package where XY is the "
-            "required CUDA version. Alternatively, install "
-            "numba-cuda-cuXY to install the required binding "
-            "automatically. If no CUDA bindings are desired, "
-            "set NUMBA_CUDA_USE_NVIDIA_BINDING=0 to enable "
-            "ctypes bindings."
+            "the cuda-bindings package. Alternatively, install "
+            "numba-cuda[cuXY], where XY is the required CUDA "
+            "version, to install the binding automatically. "
+            "If no CUDA bindings are desired, set the env var "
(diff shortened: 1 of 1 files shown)


numba_cuda/numba/cuda/__init__.py
```diff
@@ -31,6 +31,17 @@
 else:
     USE_NV_BINDING = True
     config.CUDA_USE_NVIDIA_BINDING = USE_NV_BINDING
+if config.CUDA_USE_NVIDIA_BINDING:
+    if not importlib.util.find_spec("cuda.bindings"):
+        raise ImportError(
+            "CUDA bindings not found. Please install through "
+            "the cuda-bindings[cuXY] package where XY is the "
+            "required CUDA version. Alternatively, install "
+            "numba-cuda-cuXY to install the required binding "
```
kkraus14: `numba-cuda-cuXY` doesn't exist. For pip there is `numba-cuda[cuXY]`. For conda, we should just make it all work out of the box where someone would need to forcibly break their environment to hit this error.

brandon-b-miller: For conda I think it should just work right? It should be a required run dependency now.

kkraus14: We need to update the upstream conda-forge recipe, but in the recipe published here, yes.

numba_cuda/numba/cuda/__init__.py
```diff
@@ -35,12 +35,12 @@
...
             "CUDA bindings not found. Please install through "
-            "the cuda-bindings[cuXY] package where XY is the "
-            "required CUDA version. Alternatively, install "
-            "numba-cuda-cuXY to install the required binding "
-            "automatically. If no CUDA bindings are desired, "
-            "set NUMBA_CUDA_USE_NVIDIA_BINDING=0 to enable "
-            "ctypes bindings."
+            "the cuda-bindings package. Alternatively, install "
+            "numba-cuda[cuXY], where XY is the required CUDA "
+            "version, to install the binding automatically. "
```
kkraus14: Should we reference `pip` here?

PR #295, opened 2025-06-13
Fix `DeviceNDArray.device_ctypes_pointer` to always be ctypes

gmarkall:
When the NVIDIA bindings are in use, `device_ctypes_pointer` should always be a ctypes pointer, otherwise the interface changes depending on which set of bindings are in use. This has been a long-standing inconsistency, but it needs to be resolved.

The fundamental change is only to `DeviceNDArray.device_ctypes_pointer`. The associated changes are all fixups to maintain internal consistency.

This likely covers 95% of broken use cases, with the remaining small number being related to streams and events, that are both more involved to fix and less likely to be directly affected by users (because users will mostly not look directly at the stream / event handle).

### DIFF
--- a/numba_cuda/numba/cuda/memory_management/nrt.py
+++ b/numba_cuda/numba/cuda/memory_management/nrt.py
@@ -113,10 +113,13 @@ class _Runtime:
             self._compile_memsys_module()
 
         # Allocate space for NRT_MemSys
-        ptr, nbytes = self._memsys_module.get_global_symbol("memsys_size")
         memsys_size = ctypes.c_uint64()
+        ptr, nbytes = self._memsys_module.get_global_symbol("memsys_size")
+        device_memsys_size = ptr.device_ctypes_pointer
+        if USE_NV_BINDING:
+            device_memsys_size = device_memsys_size.value
         driver.cuMemcpyDtoH(
-            ctypes.addressof(memsys_size), ptr.device_ctypes_pointer, nbytes
+            ctypes.addressof(memsys_size), device_memsys_size, nbytes
         )
         self._memsys = device_array(
             (memsys_size.value,), dtype="i1", stream=stream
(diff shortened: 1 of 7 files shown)


numba_cuda/numba/cuda/memory_management/nrt.py
```diff
@@ -113,10 +113,13 @@ def allocate(self, stream=None):
             self._compile_memsys_module()
 
         # Allocate space for NRT_MemSys
-        ptr, nbytes = self._memsys_module.get_global_symbol("memsys_size")
         memsys_size = ctypes.c_uint64()
+        ptr, nbytes = self._memsys_module.get_global_symbol("memsys_size")
+        device_memsys_size = ptr.device_ctypes_pointer
+        if USE_NV_BINDING:
+            device_memsys_size = device_memsys_size.value
```
kkraus14, line 120: `device_memsys_size` would have previously been a `CUdeviceptr` object, but now will be an integer after this. We currently handle this in `cuda.bindings.driver` but I'm not sure if this is documented as behavior we guarantee. cc @leofang

kkraus14 (approved):

numba_cuda/numba/cuda/memory_management/nrt.py
leofang: We currently accept the device pointer as a Python int (through a seemingly inefficient boxing-unboxing)
```cython
    cdef cydriver.CUdeviceptr cysrcDevice
    if srcDevice is None:
        psrcDevice = 0
    elif isinstance(srcDevice, (CUdeviceptr,)):
        psrcDevice = int(srcDevice)
    else:
        psrcDevice = int(CUdeviceptr(srcDevice))
    cysrcDevice = <cydriver.CUdeviceptr><void_ptr>psrcDevice
```
Before we make significant change to the codegen, it'd likely stay intact.

leofang:
LGTM, thx Graham/Keith!

PR #303, opened 2025-06-16
Clarify pynvjitlink option precedence.

### LINKED ISSUE #290: Clean up `CUDA_ENABLE_PYNVJITLINK` logic
The ordering of this logic makes it sound like environment variables can override the config, but the check below (`if not hasattr(config, "CUDA_ENABLE_PYNVJITLINK")` indicates that it does _not_ override the config. Can we reorder this for clarity? Better yet, can we avoid setting the config again if the config is already set (since that is treated as the highest override)?

_Originally posted by @bdice in https://github.com/NVIDIA/numba-cuda/pull/284#discussion_r2141233209_

bdice:
I asked a question on #284 that we agreed to improve in a later PR (#290). Here is a proposal for a simpler implementation of option precedence for enabling pynvjitlink.

Closes #290.

### DIFF
--- a/numba_cuda/numba/cuda/__init__.py
+++ b/numba_cuda/numba/cuda/__init__.py
@@ -7,7 +7,7 @@ from .utils import _readenv
 # 1. Config setting "CUDA_ENABLE_PYNVJITLINK" (highest priority)
 # 2. Environment variable "NUMBA_CUDA_ENABLE_PYNVJITLINK"
 # 3. Auto-detection of pynvjitlink module (lowest priority)
-if not hasattr(config, "CUDA_ENABLE_PYNVJITLINK"):
+if getattr(config, "CUDA_ENABLE_PYNVJITLINK", None) is None:
     if (
         _pynvjitlink_enabled_in_env := _readenv(
             "NUMBA_CUDA_ENABLE_PYNVJITLINK", bool, None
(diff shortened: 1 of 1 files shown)


numba_cuda/numba/cuda/__init__.py
```diff
@@ -3,23 +3,21 @@
...
-elif _pynvjitlink_enabled_in_cfg is not None:
-    ENABLE_PYNVJITLINK = _pynvjitlink_enabled_in_cfg
-else:
-    ENABLE_PYNVJITLINK = importlib.util.find_spec("pynvjitlink") is not None
-
+# Enable pynvjitlink based on the following precedence:
+# 1. Config setting "CUDA_ENABLE_PYNVJITLINK" (highest priority)
+# 2. Environment variable "NUMBA_CUDA_ENABLE_PYNVJITLINK"
+# 3. Auto-detection of pynvjitlink module (lowest priority)
 if not hasattr(config, "CUDA_ENABLE_PYNVJITLINK"):
```
kkraus14: Is there ever a situation where `config.CUDA_ENABLE_PYNVJITLINK` is set to `None` that we need to then check the other options?

Maybe change this to `if getattr(config, "CUDA_ENABLE_PYNVJITLINK", None) is None`?

bdice:
```suggestion
if getattr(config, "CUDA_ENABLE_PYNVJITLINK", None) is None:
```

kkraus14 (approved):

PR #305, opened 2025-06-23
Improve the debug mode code generation for dynamic types.

jiel-nv:
The polymorphic variables do not need zero finalization after the usage of each variation, since they share the same memory in the same artificial debug union. The cleanup work is done in the last basic block of the function.

Also, do not use the cached temporaries as the pointer for the variations, they need to be regenerated and type casted from the debug union on each load/store.

### DIFF
--- a/numba_cuda/numba/cuda/lowering.py
+++ b/numba_cuda/numba/cuda/lowering.py
@@ -2,7 +2,7 @@ from numba.core.lowering import Lower
 from llvmlite import ir
 from numba.core import ir as numba_ir
 from numba.core import types
-
+from llvmlite.ir import Constant
 
 class CUDALower(Lower):
     def storevar(self, value, name, argidx=None):
@@ -54,6 +54,9 @@ class CUDALower(Lower):
 
         self.poly_var_typ_map = {}
         self.poly_var_loc_map = {}
+        self.poly_var_set = set()
+        self.poly_cleaned = False
+        self.lastblk = max(self.blocks.keys())
 
         # When debug info is enabled, walk through function body and mark
         # variables with polymorphic types.
@@ -90,9 +93,10 @@ class CUDALower(Lower):
             index = name.find(".")
             src_name = name[:index] if index > 0 else name
             if src_name in self.poly_var_typ_map:
-                dtype = types.UnionType(self.poly_var_typ_map[src_name])
-                datamodel = self.context.data_model_manager[dtype]
+                self.poly_var_set.add(name)
                 if src_name not in self.poly_var_loc_map:
+                    dtype = types.UnionType(self.poly_var_typ_map[src_name])
+                    datamodel = self.context.data_model_manager[dtype]
                     # UnionType has sorted set of types, max at last index
                     maxsizetype = dtype.types[-1]
                     # Create a single element aggregate type
@@ -101,13 +105,7 @@ class CUDALower(Lower):
                     ptr = self.alloca_lltype(src_name, lltype, datamodel)
                     # save the location of the union type for polymorphic var
                     self.poly_var_loc_map[src_name] = ptr
-                # Any member of this union type shoud type cast ptr to fetype
-                lltype = self.context.get_value_type(fetype)
-                castptr = self.builder.bitcast(
-                    self.poly_var_loc_map[src_name], ir.PointerType(lltype)
-                )
-                # Remember the pointer
-                self.varmap[name] = castptr
+                return
 
         super()._alloca_var(name, fetype)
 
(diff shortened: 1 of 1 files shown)


jiel-nv:
Soft ping.

numba_cuda/numba/cuda/lowering.py
```diff
@@ -120,3 +119,38 @@ def store_var_needed(self, name):
...
+        """
+        if name in self.poly_var_set:
+            if (
+                self._cur_ir_block == self.blocks[self.lastblk]
+                and not self.poly_cleaned
+            ):
+                # Decref, zero-fill the debug union for polymorphic only
+                # at the last block
+                for k, v in self.poly_var_loc_map.items():
+                    self.decref(self.typeof(k), self.builder.load(v))
```
gmarkall: I'm concerned that calling `decref` for all the members of the union during the last block will break the invocation of destructor functions. There's not much in the way of tests for this in Numba-CUDA as the functionality is still being worked on slowly (and we're not going to see an issue for scalars, only device-allocated data like NumPy arrays and UDF strings in RAPIDS), but I want to avoid a design choice here for debug support that makes reference counting and lifetime management trickier, or even incorrect.

To recap my understanding, the union exists because only one of the members of the union is live at a time - is that right? If that is the case, then if the decrefs are pushed to the last block then the state of an object used by the destructor might no longer exist by the time we're in the last block. Perhaps we can make this safe by changing the logic so that decrefs are still produced at the time `delvar` is called, and only push back the zeroing to a single point in the last block. Something like:

```python
if name in self.poly_var_set:
    fetype = self.typeof(name)
    ptr = self.poly_var_loc_map[name]
    self.decref(fetype, self.builder.load(ptr))
    if (self._cur_ir_block == self.blocks[self.lastblk] and not self.poly_cleaned):
        # ... rest of the logic to zero-fill the debug union for polymorphic only at the last block
    return

super().delvar(name)
```
I'm not sure I got the `fetype` and `ptr` code 100% correct, but I hope this illustrates the idea. Does the potential issue and solution make sense to you?

numba_cuda/numba/cuda/lowering.py
```diff
@@ -120,3 +119,38 @@ def store_var_needed(self, name):
...
+
+        super().delvar(name)
+
+    def getvar(self, name):
+        """
+        Get a pointer to the given variable's slot.
+        """
+        if name in self.poly_var_set:
+            index = name.find(".")
+            src_name = name[:index] if index > 0 else name
```
gmarkall: A tiny nit, maybe a simpler way to do this in Python:

```suggestion
            src_name = name.split(".")[0]
```

jiel-nv:
Thank you for the review. I will address all the review comments in the next commit.

jiel-nv:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/305/commits/7cc07eb520df96792b9ca46a53a49283b37e0b08

numba_cuda/numba/cuda/lowering.py
jiel-nv: Resolved in https://github.com/NVIDIA/numba-cuda/pull/305/commits/ea455212d51eb8feb988e95a36bde1192869ea5d

numba_cuda/numba/cuda/lowering.py
jiel-nv: Resolved in https://github.com/NVIDIA/numba-cuda/pull/305/commits/ea455212d51eb8feb988e95a36bde1192869ea5d

jiel-nv:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/305/commits/36b56cdead1352ae2c2c568f6bd2bbdd12d2dff0

jiel-nv:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/305/commits/a5f6690760e66fe876f37d36b28078935ed3be67

gmarkall:
@jiel-nv Thanks for the updates - will review as soon as I can.

On the labels - could you please use "waiting on reviewer" rather than "waiting on CI" when there are new changes? "waiting on CI" is something that would be merged as soon as its CI has passed, but I should have another look at this to understand the changes first.

jiel-nv:
@gmarkall Sorry for the misuse of label. Yes, I will remember the correct usage next time. Thank you!

gmarkall:
> @gmarkall Sorry for the misuse of label. Yes, I will remember the correct usage next time. Thank you!

No worries, I think it's not clearly documented anywhere, but just a convention I was using to handle items as part of my "system"... thanks for your kind understanding :slightly_smiling_face:

gmarkall:
Not sure what's going on with the CI here, but will re-run the failed checks once the remaining running ones have finished.

gmarkall:
> Not sure what's going on with the CI here, but will re-run the failed checks once the remaining running ones have finished.

I'm re-running the jobs, it seems there was a transient outage.

gmarkall:
Many thanks for the updates!

Issue #302, opened 2025-06-16
Deprecate/remove `cuda_paths.py` entirely


leofang:
It is the implicit assumption that by the completion of #128 we'll rely on the Path Finder from `cuda.bindings` to find/load CTK components. It means this file will no longer be used:
https://github.com/NVIDIA/numba-cuda/blob/main/numba_cuda/numba/cuda/cuda_paths.py
We probably should add deprecation warnings asap.

rwgk:
xref: https://github.com/NVIDIA/cuda-python/issues/1036

I'm using that as a focal point to track the work on the cuda-python side.

brandon-b-miller:
Blocked by https://github.com/NVIDIA/cuda-python/issues/1157 currently

cpcloud:
Closing in favor of https://github.com/NVIDIA/cuda-python/issues/1036

PR #308, opened 2025-06-24
Use `pathfinder` for dynamic libraries

brandon-b-miller:
Towards https://github.com/NVIDIA/numba-cuda/issues/302

This PR switches to using `pathfinder` to locate `nvvm` and `nvrtc` within `numba-cuda`.

### DIFF
--- a/numba_cuda/numba/cuda/cuda_paths.py
+++ b/numba_cuda/numba/cuda/cuda_paths.py
@@ -32,7 +32,7 @@ def temporary_env_var(key, value):
         yield
     finally:
         if old_value is None:
-            del os.environ[key]
+            os.environ.pop(key, None)
         else:
             os.environ[key] = old_value
 
@@ -559,33 +559,49 @@ def _get_nvvm():
     # 2. If CUDA_HOME/CUDA_PATH are set, pathfinder would have found it - give up
     # 3. Use nvrtc's location to infer CUDA installation root
     # 4. Temporarily set CUDA_HOME and retry pathfinder
-
     # First, try pathfinder directly
     try:
         return pathfinder.load_nvidia_dynamic_lib("nvvm")
-    except pathfinder.DynamicLibNotFoundError:
-        pass
+    except pathfinder.DynamicLibNotFoundError as e:
+        nvvm_exc = e
+
+    def _raise_original(reason: str) -> None:
+        raise pathfinder.DynamicLibNotFoundError(
+            f"{reason}; original nvvm error: {nvvm_exc}"
+        ) from nvvm_exc
 
     # If CUDA_HOME or CUDA_PATH is set, pathfinder would have found libnvvm
     # based on the environment variable(s) - nothing more we can do
     if os.environ.get("CUDA_HOME") or os.environ.get("CUDA_PATH"):
-        raise pathfinder.DynamicLibNotFoundError
-
+        _raise_original("nvvm not found and CUDA_HOME/CUDA_PATH is set")
     # Try to locate nvrtc - this library is almost certainly needed if nvvm is needed (in the context of numba-cuda)
-    loaded_nvrtc = _get_nvrtc()
-
+    try:
+        loaded_nvrtc = _get_nvrtc()
+    except Exception as nvrtc_exc:
+        raise pathfinder.DynamicLibNotFoundError(
+            f"nvrtc load failed while inferring CUDA_HOME; original nvvm error: {nvvm_exc}"
+        ) from nvrtc_exc
     # If nvrtc was not found via system-search, we can't reliably determine
     # the CUDA installation structure
     if loaded_nvrtc.found_via != "system-search":
-        raise pathfinder.DynamicLibNotFoundError
+        _raise_original(
+            f"nvrtc found via {loaded_nvrtc.found_via}, cannot infer CUDA_HOME"
+        )
     # Search backward from nvrtc's location to find a directory with "nvvm" subdirectory
     cuda_home = _find_cuda_home_from_lib_path(loaded_nvrtc.abs_path)
     if cuda_home is None:
-        raise pathfinder.DynamicLibNotFoundError
-
+        _raise_original(
+            f"nvrtc path did not map to CUDA_HOME ({loaded_nvrtc.abs_path})"
+        )
     # Temporarily set CUDA_HOME and retry pathfinder
     with temporary_env_var("CUDA_HOME", cuda_home):
-        library = pathfinder.load_nvidia_dynamic_lib("nvvm")
+        try:
+            library = pathfinder.load_nvidia_dynamic_lib("nvvm")
+        except pathfinder.DynamicLibNotFoundError as exc:
+            raise pathfinder.DynamicLibNotFoundError(
+                f"nvvm not found after inferring CUDA_HOME={cuda_home}; "
+                f"original nvvm error: {nvvm_exc}"
+            ) from exc
         library.found_via = "system-search"
         return library
 


numba_cuda/numba/cuda/cudadrv/libs.py
```diff
@@ -90,6 +93,7 @@ def check_static_lib(path):
 
 
 def _get_source_variable(lib, static=False):
+    # remove? only used in test()
```
brandon-b-miller: Should we remove this function and `test()` below? `cuda_paths` also supports saying what the source of the loaded library is, such as if it's a wheel vs conda lib etc. Path finder treats this as an implementation detail.

We could always change it to just report the path without the source but its possible this function has proliferated a bit and is used for debugging out in the wild.

numba_cuda/numba/cuda/cudadrv/libs.py
```diff
@@ -67,7 +70,7 @@ def get_cuda_include_dir():
     Note that this does not list the `CUDA_INCLUDE_PATH` entry in user
     configuration.
     """
-
+    # CUDA headers
```
brandon-b-miller: These aren't a library but are still needed by numba, is this something `pathfinder` will support?

leofang:
@rwgk could you help review? (I can't add you as a reviewer for some reason)

numba_cuda/numba/cuda/cudadrv/libs.py
rwgk: Yes, that's the plan.

numba_cuda/numba/cuda/cudadrv/libs.py
rwgk: > Path finder treats this as an implementation detail.

A hint could be added to the `LoadedDL` type if that's useful. Do you think it will be?

`LoadedDL` reliably includes the absolute path, even if the library was loaded outside of path finder. The path is more conclusive and usually it's very obvious if it's from a wheel, conda, or full CTK installation.

rwgk (approved):
I don't have a lot of numba-cuda specific context, but what I see in this PR looks good to me.

numba_cuda/numba/cuda/cudadrv/libs.py
brandon-b-miller: Thanks, that's helpful context. I think the _best_ solution then would be one that maintains the current function, but decodes what the library source is on the numba side. I can take a whack at this and see how it looks.

numba_cuda/numba/cuda/cudadrv/libs.py
```diff
@@ -19,6 +19,7 @@
 from numba.cuda.cudadrv.driver import locate_driver_and_loader, load_driver
 from numba.cuda.cudadrv.error import CudaSupportError
 from numba.core import config
+from cuda.bindings import path_finder
```
leofang: @rwgk wouldn't this break once we make `pathfinder` a standalone module? (https://github.com/NVIDIA/cuda-python/pull/723)

rwgk: I was thinking not, because my intention is to keep the backward compatibility code for a `cuda_bindings` release or two.

I just saw @kkraus14 asked about adding a deprecation warning in that code (on NVIDIA/cuda-python#723): I'd don't think there is a rush to get rid of the few lines of backward compatibility code, therefore I'd lean towards making it easy for the world, by staging: one `cuda_bindings` release without deprecation warning, the next one or two `cuda_bindings` releases with warning, then remove the backward compatibility code.

(This is very similar to how features are deprecated in CPython.)

kkraus14: I'm good with the approach of phasing in the deprecation.

numba_cuda/numba/cuda/cudadrv/libs.py
gmarkall: If it's possible to keep the `test()` functionality I'd prefer that - it has been quite important for debugging in the past. For a lot of misconfigurations on peoples' systems, Numba-CUDA has not been the root cause of an issue but it is often the first place where a problem manifests, so it's been invaluable for showing people what the issue is and demonstrating that it's not just that "Numba-CUDA is broken". It's OK if the format of the output changes, but whatever information we can provide here from the CUDA pathfinder that provides parity with existing functionality will be helpful.

Also note that Numba will attempt to call it when producing sysinfo, so if it is removed entirely then it'll break numba sysinfo's (running `python -m numba -s` or `numba -s`) ability to show info about CUDA: https://github.com/numba/numba/blob/0c633b896526b039171c0f800df70320e1c61a53/numba/misc/numba_sysinfo.py#L381

rwgk: > but whatever information we can provide here from the CUDA pathfinder that provides parity with existing functionality will be helpful.

Ok, tx, I'll work on adding that in.

numba_cuda/numba/cuda/cudadrv/libs.py
ZzEeKkAa: +1 for this feature in `pathfinder`

brandon-b-miller:
Blocked on https://github.com/NVIDIA/cuda-python/issues/1157 for now

brandon-b-miller:
@kkraus14 @rwgk @leofang could you take another look at this when you get a chance? This moves us onto pathfinder for nvrtc (100% of the time) and nvvm (99% of the time, excepting system search which we fall back to)

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -689,3 +514,49 @@ def _get_include_dir():
...
+            if nvvm.exists():
+                dl = pathfinder._dynamic_libs.load_nvidia_dynamic_lib.load_with_abs_path(
+                    "nvvm", nvvm, "system-search"
+                )
+                return dl
+        raise pathfinder.DynamicLibNotFoundError("nvvm not found")
+
+
+def _get_nvrtc():
+    return pathfinder.load_nvidia_dynamic_lib("nvrtc")
```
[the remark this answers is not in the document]
brandon-b-miller: Thanks, pathfinder doesn't need a custom system fallback for nvrtc, only nvvm.

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -689,3 +514,50 @@ def _get_include_dir():
...
+    return _env_path_tuple(nvrtc.found_via, nvrtc.abs_path)
+
+
+def _get_nvvm_path():
+    nvvm = _get_nvvm()
+    return _env_path_tuple(nvvm.found_via, nvvm.abs_path)
+
+
+def _get_nvvm_system_path():
+    nvvm_lib_dir = get_system_ctk("nvvm")
```
rwgk: If `IS_WIN32` is true, this will always be `None` (see line 177 in this file; current state of this PR):

```python
def get_system_ctk(*subdirs):
    """Return path to system-wide cudatoolkit; or, None if it doesn't exist."""
    # Linux?
    if not IS_WIN32:
        # Is cuda alias to /usr/local/cuda?
        # We are intentionally not getting versioned cuda installation.
        result = os.path.join("/usr/local/cuda", *subdirs)
        if os.path.exists(result):
            return result
        return None
    return None
```

So the `IS_WIN32` tests in this function further down here will certainly by false.

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -689,3 +514,50 @@ def _get_include_dir():
...
+        path_or_none = _get_nvvm_system_path()
+        if path_or_none is not None:
+            nvvm = pathlib.Path(path_or_none)
+            if nvvm.exists():
+                # TODO: Expose private cuda-python APIs
+                dl = pathfinder._dynamic_libs.load_nvidia_dynamic_lib.load_with_abs_path(
+                    "nvvm", nvvm, "system-search"
+                )
+                return dl
+        raise pathfinder.DynamicLibNotFoundError("nvvm not found") from e
```
rwgk: Here is an alternative implementation that avoids `get_system_ctk()`. It comes complete with a (Cursor-generated) unit test. I tested it interactively on Linux and Windows. Please copy out the meat, I believe you can plug it in here unchanged.

```python
"""
Alternative implementation for _get_nvvm() that avoids using get_system_ctk("nvvm").

This uses nvrtc's location to infer the CUDA installation root and temporarily
sets CUDA_HOME to help pathfinder find nvvm.
"""

import os
import pathlib
from contextlib import contextmanager

# These would be imported from cuda.bindings in the real implementation
try:
    from cuda import pathfinder
except ImportError:
    pathfinder = None
    print("Warning: cuda.bindings.pathfinder not available, using mock")


@contextmanager
def temporary_env_var(key, value):
    """Context manager to temporarily set an environment variable."""
    old_value = os.environ.get(key)
    os.environ[key] = value
    try:
        yield
    finally:
        if old_value is None:
            del os.environ[key]
        else:
            os.environ[key] = old_value


def _find_cuda_home_from_lib_path(lib_path):
    """
    Walk up from a library path to find a directory containing 'nvvm' subdirectory.

    For example, given /usr/local/cuda/lib64/libnvrtc.so.12,
    this would find /usr/local/cuda (which contains nvvm/).

    Returns the path if found, None otherwise.
    """
    current = pathlib.Path(lib_path).resolve()

    # Walk up the directory tree
    for parent in current.parents:
        nvvm_subdir = parent / "nvvm"
        if nvvm_subdir.is_dir():
            return str(parent)

    return None


def _get_nvrtc():
    """Get nvrtc library info via pathfinder."""
    if pathfinder is None:
        return None
    return pathfinder.load_nvidia_dynamic_lib("nvrtc")


def _get_nvvm():
    # Strategy:
    # 1. Try pathfinder directly
    # 2. If CUDA_HOME/CUDA_PATH are set, pathfinder would have found it - give up
    # 3. Use nvrtc's location to infer CUDA installation root
    # 4. Temporarily set CUDA_HOME and retry pathfinder

    # First, try pathfinder directly
    try:
        return pathfinder.load_nvidia_dynamic_lib("nvvm")
    except pathfinder.DynamicLibNotFoundError:
        pass

    # If CUDA_HOME or CUDA_PATH is set, pathfinder would have found libnvvm
    # based on the environment variable(s) - nothing more we can do
    if os.environ.get("CUDA_HOME") or os.environ.get("CUDA_PATH"):
        return None

    # Try to locate nvrtc - this library is almost certainly needed if nvvm is needed (in the context of numba-cuda)
    try:
        loaded_nvrtc = _get_nvrtc()
    except pathfinder.DynamicLibNotFoundError:
        return None

    # If nvrtc was not found via system-search, we can't reliably determine
    # the CUDA installation structure
    if loaded_nvrtc.found_via != "system-search":
        return None

    # Search backward from nvrtc's location to find a directory with "nvvm" subdirectory
    cuda_home = _find_cuda_home_from_lib_path(loaded_nvrtc.abs_path)
    if cuda_home is None:
        return None

    # Temporarily set CUDA_HOME and retry pathfinder
    with temporary_env_var("CUDA_HOME", cuda_home):
        try:
            return pathfinder.load_nvidia_dynamic_lib("nvvm")
        except pathfinder.DynamicLibNotFoundError:
            return None


# For testing without cuda.bindings
class MockDynamicLib:
    def __init__(self, name, abs_path, found_via):
        self.name = name
        self.abs_path = abs_path
        self.found_via = found_via


def test_find_cuda_home():
    """Test the directory walking logic."""
    import tempfile

    # Create a mock CUDA installation structure
    with tempfile.TemporaryDirectory() as tmpdir:
        cuda_root = pathlib.Path(tmpdir) / "cuda"
        lib64 = cuda_root / "lib64"
        nvvm = cuda_root / "nvvm"
        nvvm_lib64 = nvvm / "lib64"

        lib64.mkdir(parents=True)
        nvvm_lib64.mkdir(parents=True)

        # Create mock library files
        nvrtc_path = lib64 / "libnvrtc.so.12"
        nvrtc_path.touch()

        nvvm_lib = nvvm_lib64 / "libnvvm.so.4"
        nvvm_lib.touch()

        # Test finding CUDA_HOME from nvrtc path
        found_cuda_home = _find_cuda_home_from_lib_path(str(nvrtc_path))

        # Compare resolved paths to handle Windows short path names (e.g., RGROSS~1)
        expected = str(cuda_root.resolve())
        assert found_cuda_home == expected, (
            f"Expected {expected}, got {found_cuda_home}"
        )
        print(f"SUCCESS: Found CUDA_HOME={found_cuda_home} from nvrtc at {nvrtc_path}")

        # Test that the nvvm directory exists at the found location
        assert (pathlib.Path(found_cuda_home) / "nvvm").is_dir()
        print(f"SUCCESS: nvvm/ subdirectory exists at {found_cuda_home}/nvvm")


if __name__ == "__main__":
    print("Testing directory walking logic...")
    test_find_cuda_home()
    print("\nAll tests passed!")

    print("\n" + "=" * 60)
    print("Attempting to use actual pathfinder (if available)...")
    print("=" * 60)

    if pathfinder is not None:
        result = _get_nvvm()
        if result:
            print(f"SUCCESS: Found nvvm at {result.abs_path}")
            print(f"         Found via: {result.found_via}")
        else:
            print("Could not load nvvm")
    else:
        print("Skipping (pathfinder not available)")
```

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -689,3 +529,93 @@ def _get_include_dir():
...
+    # Temporarily set CUDA_HOME and retry pathfinder
+    with temporary_env_var("CUDA_HOME", cuda_home):
+        try:
+            return pathfinder.load_nvidia_dynamic_lib("nvvm")
+        except pathfinder.DynamicLibNotFoundError:
+            return None
+
+
+def _get_nvrtc():
+    return pathfinder.load_nvidia_dynamic_lib("nvrtc")
```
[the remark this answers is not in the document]
brandon-b-miller: Pathfinder should cover all cases for `nvrtc`, thus if it can't find it, we want to propagate the error. This includes the system path, so, there shouldn't be any need to fall back to the inbuilt mechanism for invoking a system search.

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -528,43 +400,11 @@ def get_cuda_home_include():
...
-    return _env_path_tuple(by, path)
-
-
 def get_cuda_paths():
     """Returns a dictionary mapping component names to a 2-tuple
     of (source_variable, info).
 
     The returned dictionary will have the following keys and infos:
-    - "nvvm": file_path
-    - "nvrtc": file_path
```
kkraus14, line 567: Did you mean to swap these and instead removed them? It seems like they're still in the dict

pyproject.toml
```diff
@@ -27,12 +27,14 @@ dependencies = ["numba>=0.60.0", "cuda-bindings>=12.9.1,<14.0.0", "cuda-core>=0.
 cu12 = [
     "cuda-bindings>=12.9.1,<13.0.0",
     "cuda-core>=0.3.0,<1.0.0",
+    "cuda-pathfinder>=1.3.1",
```
kkraus14: Restrict to `<2.0` as we may want to break the API in the future

pyproject.toml
```diff
@@ -27,12 +27,14 @@ dependencies = ["numba>=0.60.0", "cuda-bindings>=12.9.1,<14.0.0", "cuda-core>=0.
...
     # install nvcc for libNVVM
     "cuda-toolkit[cudart,nvcc,nvrtc,nvjitlink,cccl]==12.*",
 ]
 cu13 = [
     "cuda-bindings==13.*",
     "cuda-core>=0.3.2,<1.0.0",
+    "cuda-pathfinder>=1.3.1",
```
kkraus14: +1

kkraus14 (approved):

numba_cuda/numba/cuda/cuda_paths.py
brandon-b-miller: From a previous iteration - nice catch. Reverted.

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -20,6 +23,20 @@
...
+@contextmanager
+def temporary_env_var(key, value):
+    """Context manager to temporarily set an environment variable."""
+    old_value = os.environ.get(key)
+    os.environ[key] = value
+    try:
+        yield
+    finally:
+        if old_value is None:
+            del os.environ[key]
```
rwgk: Minor, be very easy hardening:
```
            os.environ.pop(key, None)
```

When `temporary_env_var()` exits, it deletes the key if it didn't exist originally. If the code inside the context already removed the key, `del os.environ[key]` will raise `KeyError`. Using `pop(key, None)` is safer and preserves the intended behavior.

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -689,3 +531,76 @@ def _get_include_dir():
...
+    # First, try pathfinder directly
+    try:
+        return pathfinder.load_nvidia_dynamic_lib("nvvm")
+    except pathfinder.DynamicLibNotFoundError:
+        pass
+
+    # If CUDA_HOME or CUDA_PATH is set, pathfinder would have found libnvvm
+    # based on the environment variable(s) - nothing more we can do
+    if os.environ.get("CUDA_HOME") or os.environ.get("CUDA_PATH"):
+        raise pathfinder.DynamicLibNotFoundError
```
rwgk: Oh, sorry, I forgot to replace the bare `raise` here when I worked on my previous suggestion.

The "happy path" is fine as is, but the error paths will produce potentially confusing exceptions.

I switched Cursor models to codex high, which gave me a deluxe version:

___

### Suggestion: preserve original `DynamicLibNotFoundError` with context

Right now the fallback branches raise a bare `DynamicLibNotFoundError`, which
drops the original context and makes debugging harder. A small refactor can
preserve the original `nvvm` error for “benign” exits while still using
`raise-from` for unexpected errors (e.g., failures while loading `nvrtc`).

```suggestion
def _get_nvvm():
    # Strategy:
    # 1. Try pathfinder directly
    # 2. If CUDA_HOME/CUDA_PATH are set, pathfinder would have found it - give up
    # 3. Use nvrtc's location to infer CUDA installation root
    # 4. Temporarily set CUDA_HOME and retry pathfinder

    # First, try pathfinder directly
    try:
        return pathfinder.load_nvidia_dynamic_lib("nvvm")
    except pathfinder.DynamicLibNotFoundError as nvvm_exc:
        pass

    def _raise_original(reason: str) -> None:
        raise pathfinder.DynamicLibNotFoundError(
            f"{reason}; original nvvm error: {nvvm_exc}"
        ) from nvvm_exc

    # If CUDA_HOME or CUDA_PATH is set, pathfinder would have found libnvvm
    # based on the environment variable(s) - nothing more we can do
    if os.environ.get("CUDA_HOME") or os.environ.get("CUDA_PATH"):
        _raise_original("nvvm not found and CUDA_HOME/CUDA_PATH is set")

    # Try to locate nvrtc - this library is almost certainly needed if nvvm is needed (in the context of numba-cuda)
    try:
        loaded_nvrtc = _get_nvrtc()
    except Exception as nvrtc_exc:
        raise pathfinder.DynamicLibNotFoundError(
            f"nvrtc load failed while inferring CUDA_HOME; original nvvm error: {nvvm_exc}"
        ) from nvrtc_exc

    # If nvrtc was not found via system-search, we can't reliably determine
    # the CUDA installation structure
    if loaded_nvrtc.found_via != "system-search":
        _raise_original(
            f"nvrtc found via {loaded_nvrtc.found_via}, cannot infer CUDA_HOME"
        )
    # Search backward from nvrtc's location to find a directory with "nvvm" subdirectory
    cuda_home = _find_cuda_home_from_lib_path(loaded_nvrtc.abs_path)
    if cuda_home is None:
        _raise_original(
            f"nvrtc path did not map to CUDA_HOME ({loaded_nvrtc.abs_path})"
        )

    # Temporarily set CUDA_HOME and retry pathfinder
    with temporary_env_var("CUDA_HOME", cuda_home):
        try:
            library = pathfinder.load_nvidia_dynamic_lib("nvvm")
        except pathfinder.DynamicLibNotFoundError as exc:
            raise pathfinder.DynamicLibNotFoundError(
                f"nvvm not found after inferring CUDA_HOME={cuda_home}; "
                f"original nvvm error: {nvvm_exc}"
            ) from exc
        library.found_via = "system-search"
        return library
```

That's a little bit on the overkill side. In the controlled/benign cases we could also simply `raise nvvm_exc`, to save a few lines for building the custom messages for each situation. Please pick and choose based on your judgment. The main idea is that we don't want to mask bugs, but also surface the original `nvvm_exc`.

brandon-b-miller: done in [b87fab2](https://github.com/NVIDIA/numba-cuda/pull/308/commits/b87fab22843a3864476c09cc700020b80ecdd873)

rwgk (approved):
Looks good to me!

PR #310, opened 2025-06-25
Enable LTO by default when pynvjitlink is available

gmarkall:
Enabling LTO by default when pynvjitlink is available should:

- Provide a general improvement in performance for various use cases, particularly those linking external code. This ought to be benchmarked, but I'm making an assumption that it helps for now based on prior anecdotal / informal experience.
- Make the case where users link LTO-IR to kernels or as part of device function declarations "just work" as long as pynvjitlink is installed.

A further improvement would still be to error out when a users tries to link LTO-IR when pynvjitlink is not installed - that is left to be done in a future PR.

### DIFF
--- a/numba_cuda/numba/cuda/tests/cudapy/test_errors.py
+++ b/numba_cuda/numba/cuda/tests/cudapy/test_errors.py
@@ -90,6 +90,7 @@ class TestJitErrors(CUDATestCase):
         self.assertIn("resolving callee type: type(CUDADispatcher", excstr)
         self.assertIn("NameError: name 'floor' is not defined", excstr)
 
+    @skip_on_cudasim("Simulator does not use pynvjitlink")
     def test_lto_without_pynvjitlink_error(self):
         with self.assertRaisesRegex(RuntimeError, "LTO requires pynvjitlink"):
             with override_config("CUDA_ENABLE_PYNVJITLINK", False):
--- a/numba_cuda/numba/cuda/tests/cudadrv/test_nvjitlink.py
+++ b/numba_cuda/numba/cuda/tests/cudadrv/test_nvjitlink.py
@@ -304,7 +304,7 @@ class TestLinkerUsage(CUDATestCase):
         env = os.environ.copy()
         env["NUMBA_CUDA_ENABLE_PYNVJITLINK"] = "0"
         with self.assertRaisesRegex(
-            AssertionError, "LTO and additional flags require PyNvJitLinker"
+            AssertionError, "LTO requires pynvjitlink, which is not enabled"
         ):
             # Actual error raised is `ValueError`, but `run_in_subprocess`
             # reraises as AssertionError.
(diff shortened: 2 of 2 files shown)


tpn:
This looks like it'll fix the recent cuda.coop issues I ran into, so, +1 from me.

brandon-b-miller (approved):
LGTM

brandon-b-miller:
Looking into the failures here

numba_cuda/numba/cuda/decorators.py
```diff
@@ -136,6 +141,13 @@ def jit(
     if device and kws.get("link"):
         raise ValueError("link keyword invalid for device function")
 
+    if lto is None:
+        # Default to using LTO if pynvjitlink is available and we're not debugging
+        lto = config.CUDA_ENABLE_PYNVJITLINK and not debug
+    else:
+        if lto and not config.CUDA_ENABLE_PYNVJITLINK:
+            raise RuntimeError("LTO requires pynvjitlink, which is not enabled")
```
brandon-b-miller, line 149: A few tests pop because the error raised here is not exactly the one that the tests are expecting. We should update `test_nvjitlink.py` to expect the new error

```
diff --git a/numba_cuda/numba/cuda/tests/cudadrv/test_nvjitlink.py b/numba_cuda/numba/cuda/tests/cudadrv/test_nvjitlink.py
index 8cd33bb..0eb12d9 100644
--- a/numba_cuda/numba/cuda/tests/cudadrv/test_nvjitlink.py
+++ b/numba_cuda/numba/cuda/tests/cudadrv/test_nvjitlink.py
@@ -323,7 +323,7 @@ class TestLinkerUsage(CUDATestCase):
         env.pop("NUMBA_CUDA_ENABLE_PYNVJITLINK", None)
         with override_config("CUDA_ENABLE_PYNVJITLINK", False):
             with self.assertRaisesRegex(
-                AssertionError, "LTO and additional flags require PyNvJitLinker"
+                AssertionError, "LTO requires pynvjitlink, which is not enabled"
             ):
                 run_in_subprocess(
                     self.src.format(
```

brandon-b-miller:
To fix the simulator test we need to add `lto=None` to [numba_cuda/numba/cuda/simulator/api.py](https://github.com/NVIDIA/numba-cuda/blob/307443402a4b69608c931988a6c264748f26674a/numba_cuda/numba/cuda/simulator/api.py#L118) but locally for me there's more needed to fix the failing test, still tracking down exactly what.

numba_cuda/numba/cuda/tests/cudapy/test_errors.py
```diff
@@ -89,6 +90,16 @@ def kernel_func():
         self.assertIn("resolving callee type: type(CUDADispatcher", excstr)
         self.assertIn("NameError: name 'floor' is not defined", excstr)
 
+    def test_lto_without_pynvjitlink_error(self):
```
brandon-b-miller, line 94: Are we just missing a `@skip_on_cudasim` here?

brandon-b-miller:
> To fix the simulator test we need to add `lto=None` to [numba_cuda/numba/cuda/simulator/api.py](https://github.com/NVIDIA/numba-cuda/blob/307443402a4b69608c931988a6c264748f26674a/numba_cuda/numba/cuda/simulator/api.py#L118) but locally for me there's more needed to fix the failing test, still tracking down exactly what.

Actually I think it's just https://github.com/NVIDIA/numba-cuda/pull/310/files#r2167468144

PR #314, opened 2025-07-02
Fix nvrtc resolution when CUDA_HOME env is set

ZzEeKkAa:
`libnvrtc.so` is located in the `lib64` folder not in the separate folder. I guess it is a typo based on the `nvvm` location. Not sure about Windows path, but was following same logic as `nvvm` does, but without subfolder.

UPD: probably won't be needed after #308

### DIFF
--- a/numba_cuda/numba/cuda/cuda_paths.py
+++ b/numba_cuda/numba/cuda/cuda_paths.py
@@ -97,10 +97,17 @@ def _get_nvrtc_system_ctk():
         return max(candidates)
 
 
+def _nvrtc_lib_dir():
+    if IS_WIN32:
+        return ("bin",)
+    else:
+        return ("lib64",)
+
+
 def _get_nvrtc_path_decision():
     options = _build_options(
         [
-            ("CUDA_HOME", lambda: get_cuda_home("nvrtc")),
+            ("CUDA_HOME", lambda: get_cuda_home(*_nvrtc_lib_dir())),
             ("Conda environment", get_conda_ctk),
             ("Conda environment (NVIDIA package)", get_nvidia_cudalib_ctk),
             ("NVIDIA NVCC Wheel", _get_nvrtc_wheel),
--- a/numba_cuda/numba/cuda/cuda_paths.py
+++ b/numba_cuda/numba/cuda/cuda_paths.py
@@ -97,17 +97,10 @@ def _get_nvrtc_system_ctk():
         return max(candidates)
 
 
-def _nvrtc_lib_dir():
-    if IS_WIN32:
-        return ("bin",)
-    else:
-        return ("lib64",)
-
-
 def _get_nvrtc_path_decision():
     options = _build_options(
         [
-            ("CUDA_HOME", lambda: get_cuda_home(*_nvrtc_lib_dir())),
+            ("CUDA_HOME", lambda: get_cuda_home(_cudalib_path())),
             ("Conda environment", get_conda_ctk),
             ("Conda environment (NVIDIA package)", get_nvidia_cudalib_ctk),
             ("NVIDIA NVCC Wheel", _get_nvrtc_wheel),


brandon-b-miller:
Thanks @ZzEeKkAa , is there a container or something similar I can use to reproduce the failure state?

ZzEeKkAa:
> Thanks @ZzEeKkAa , is there a container or something similar I can use to reproduce the failure state?

I was using just system install of ctk at `/usr/local/cuda-X.Y`. I would expect that any cuda container, for example https://hub.docker.com/r/rapidsai/devcontainers will work.

I was following this instructions to setup system ctk:
```
export CUDA_HOME=/usr/local/cuda-X.Y
export LIBRARY_PATH=$CUDA_HOME/lib64:$LIBRARY_PATH
```

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -97,10 +97,17 @@ def _get_nvrtc_system_ctk():
         return max(candidates)
 
 
+def _nvrtc_lib_dir():
+    if IS_WIN32:
+        return ("bin",)
+    else:
+        return ("lib64",)
```
kkraus14: I think this is the same as the `_cudalib_path` function so maybe we can just use that?

gmarkall (approved):
On `main`, when I run:

```
CUDA_HOME=/usr/local/cuda-12.8 NUMBA_CUDA_USE_NVIDIA_BINDING=0 gdb --args python -c "from numba import cuda; cuda.cudadrv.libs.test(); input()"
```

I get

```
	libcuda.so
	libcuda.so.1
	/usr/lib/libcuda.so
	/usr/lib/libcuda.so.1
	/usr/lib64/libcuda.so
	/usr/lib64/libcuda.so.1
Using loader <class 'ctypes.CDLL'>
	Trying to load driver...	ok
		Loaded from libcuda.so
	Mapped libcuda.so paths:
		/usr/lib/x86_64-linux-gnu/libcuda.so.570.133.20
Finding nvvm from CUDA_HOME
	Located at /usr/local/cuda-12.8/nvvm/lib64/libnvvm.so.4.0.0
	Trying to open library...	ok
Finding nvrtc from CUDA_HOME
	Located at libnvrtc.so
	Trying to open library...	ok
Finding cudadevrt from CUDA_HOME
	Located at /usr/local/cuda-12.8/lib64/libcudadevrt.a
	Checking library...	ok
Finding libdevice from CUDA_HOME
	Located at /usr/local/cuda-12.8/nvvm/libdevice/libdevice.10.bc
	Checking library...	ok
Include directory configuration variable:
	CUDA_INCLUDE_PATH=/usr/local/cuda/include
Finding include directory from CUDA_INCLUDE_PATH Config Entry
	Located at /usr/local/cuda/include
	Checking include directory...	ok
```

The nvrtc path was not located correctly - gdb `info shared` shows it found another `nvrtc.so` on the linker path:

```
0x00007ffef6352e10  0x00007ffef9654342  Yes (*)     /usr/local/cuda-12.8/nvvm/lib64/libnvvm.so.4.0.0
0x00007ffef2643e80  0x00007ffef4c9fc92  Yes (*)     /usr/local/cuda-12.6/targets/x86_64-linux/lib/libnvrtc.so
```

(`nvvm.so` also shown for comparison).

With this branch, the correct nvrtc library is loaded:

```
0x00007ffef6352e10  0x00007ffef9654342  Yes (*)     /usr/local/cuda-12.8/nvvm/lib64/libnvvm.so.4.0.0
0x00007ffeefe1f170  0x00007ffef46b2892  Yes (*)     /usr/local/cuda-12.8/lib64/libnvrtc.so.12.8.93
```

and the output is also as expected:

```
Finding nvrtc from CUDA_HOME
	Located at /usr/local/cuda-12.8/lib64/libnvrtc.so.12.8.93
	Trying to open library...	ok
```

PR #323, opened 2025-07-16
Add status check github action

kkraus14:
Adds an overall status check GitHub action that we can use for branch protection rules.


kkraus14:
This is actually handled in the rapids shared workflow: https://github.com/rapidsai/shared-workflows/blob/branch-25.08/.github/workflows/pr-builder.yaml

PR #326, opened 2025-07-16
Allow Inspection of Link-Time Optimized PTX

isVoid:
Adds kernel method `inspect_lto_ptx`, which allows inspection of link time optimized PTX. This method is only supported when `lto=True` is specified for the kernel.


numba_cuda/numba/cuda/testing.py
```diff
@@ -189,6 +189,10 @@ def skip_if_cudadevrt_missing(fn):
     return unittest.skipIf(cudadevrt_missing(), "cudadevrt missing")(fn)
 
 
+def skip_if_nvjitlink_missing(reason):
```
gmarkall, line 192: Not to be changed for this PR, but there are a few other test files that do something like this that should probably be changed to use this version of it.

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -1169,6 +1175,30 @@ def inspect_asm(self, signature=None):
                     for sig, overload in self.overloads.items()
                 }
 
+    def inspect_lto_ptx(self, signature=None):
+        """
+        Return link-time optimized PTX code for the given signature.
```
gmarkall, line 1180: The docstring here is a bit inconsistent with that of `inspect_asm()` - e.g. missing parameter / return descriptions, and that the code is for the device in the current context.

numba_cuda/numba/cuda/tests/cudapy/test_inspect.py
```diff
@@ -108,6 +115,55 @@ def _test_inspect_sass(self, kernel, name, sass):
...
+        allasms = k.inspect_asm()
+        asm = next(iter(allasms.values()))
+
+        regex = re.compile(r"call(.|\n)*add_f2_f2")
+        self.assertRegex(asm, regex)
+
+        all_ext_asms = k.inspect_lto_ptx()
+        lto_asm = next(iter(all_ext_asms.values()))
+
+        assert "add.f16" in str(lto_asm)
```
gmarkall: The assertions usually follow this form: 

```suggestion
        self.assertIn("add.f16", str(lto_asm))
```

(I think the pattern in the PR is more normal with pytest?)

gmarkall: Additionally, is it worth asserting that we don't have a `call` in the LTO PTX? (It seems reasonable to expect inlining here)

numba_cuda/numba/cuda/tests/cudapy/test_inspect.py
```diff
@@ -108,6 +115,55 @@ def _test_inspect_sass(self, kernel, name, sass):
         self.assertIn("BRA", sass)  # Branch
         self.assertIn("EXIT", sass)  # Exit program
 
+    @skip_on_cudasim("Simulator does not generate code to be inspected")
+    @skip_if_nvjitlink_missing("nvJitLink is required for LTO")
+    def test_inspect_lto_asm(self):
+        ffi = cffi.FFI()
+        temp_add_cu = tempfile.NamedTemporaryFile(suffix=".cu")
```
gmarkall: Is this a case in which a `CUSource` can't be used? (Asking out of curiosity rather than suggesting a change)

numba_cuda/numba/cuda/tests/cudapy/test_inspect.py
```diff
@@ -108,6 +115,55 @@ def _test_inspect_sass(self, kernel, name, sass):
...
+            local_arr2[0] = 2
+
+            ptr = ffi.from_buffer(local_arr)
+            ptr2 = ffi.from_buffer(local_arr2)
+
+            arr[0] = add(ptr, ptr2)
+
+        arr = np.array([0], dtype=np.float16)
+
+        k[1, 1](arr)
```
gmarkall, line 151: If we're going to the trouble of launching the kernel, can we check it's output is correct? E.g.
```suggestion
        k[1, 1](arr)
        np.testing.assert_allclose(arr[0], np.float16(1) + np.float16(2))
```

gmarkall (changes requested):
Many thanks - this looks good... Just a couple of minor nits / questions on the diff.

numba_cuda/numba/cuda/tests/cudapy/test_inspect.py
isVoid: No - I think in a separate experiment I actually used `CUSource`.

numba_cuda/numba/cuda/testing.py
isVoid: Yes, for example: 
https://github.com/NVIDIA/numba-cuda/blob/489045fe3e4b9d6de8d42bbde03e5a23dc416776/numba_cuda/numba/cuda/tests/cudadrv/test_nvjitlink.py#L44-L49

numba_cuda/numba/cuda/dispatcher.py
isVoid: Done in [e08d945](https://github.com/NVIDIA/numba-cuda/pull/326/commits/e08d9455d1297e836dd37154d13c5ac080f8c09b)

numba_cuda/numba/cuda/tests/cudapy/test_inspect.py
isVoid: Done in [7798108](https://github.com/NVIDIA/numba-cuda/pull/326/commits/7798108f106ccff61f0f9a189feda9af1bfb2844), [04eb522](https://github.com/NVIDIA/numba-cuda/pull/326/commits/04eb5228a387f1bd1690ed461dc16a21b5a74b36)

numba_cuda/numba/cuda/tests/cudapy/test_inspect.py
isVoid: Changed to CUSource: [c4fb345](https://github.com/NVIDIA/numba-cuda/pull/326/commits/c4fb345fd39304c6c187e0eb4f889a4ff03018ef)

numba_cuda/numba/cuda/tests/cudapy/test_inspect.py
isVoid: Done in [7798108](https://github.com/NVIDIA/numba-cuda/pull/326/commits/7798108f106ccff61f0f9a189feda9af1bfb2844)

gmarkall (approved):

PR #330, opened 2025-07-18
Vendor in PassManager, typed and untyped pass, and related helpers for CUDA-specific customization

atmnp:
This change vendors in the Pass Manager to ease future CUDA-specific optimizations. This requires also vendoring in all of the passes in the default and Numba CUDA pipelines to maintain the pass registry, as well as the compiler lock machinery.

### DIFF
--- a/numba_cuda/numba/cuda/compiler.py
+++ b/numba_cuda/numba/cuda/compiler.py
@@ -10,14 +10,8 @@ from numba.core import (
     sigutils,
     utils,
 )
-from numba.core.compiler import (
-    sanitize_compile_result_entries,
-    CompilerBase,
-    DefaultPassBuilder,
-    CompileResult,
-)
 from numba.cuda.core.compiler_lock import global_compiler_lock
-from numba.core.compiler_machinery import (
+from numba.cuda.core.compiler_machinery import (
     FunctionPass,
     LoweringPass,
     PassManager,
@@ -25,8 +19,8 @@ from numba.core.compiler_machinery import (
 )
 from numba.core.interpreter import Interpreter
 from numba.core.errors import NumbaInvalidConfigWarning
-from numba.core.untyped_passes import TranslateByteCode
-from numba.core.typed_passes import (
+from numba.cuda.core.untyped_passes import TranslateByteCode
+from numba.cuda.core.typed_passes import (
     IRLegalization,
     NativeLowering,
     AnnotateTypes,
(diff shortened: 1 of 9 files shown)


atmnp:
failing tests fixed + rebased. /ok to test 95c08999e2138c8f88d48887b6674c2b7b537454

atmnp:
Sorry for spamming the CI, ruff kept removing an import that it thought was unused, I didn't catch it originally.

numba_cuda/numba/cuda/compiler.py
```diff
@@ -6,39 +6,672 @@
     types,
     funcdesc,
     config,
-    compiler,
-)
-from numba.core.compiler import (
-    sanitize_compile_result_entries,
-    DefaultPassBuilder,
 )
+from numba.cuda import compiler
```
gmarkall: This appears to now be importing `numba.cuda.compiler` from itself.

numba_cuda/numba/cuda/compiler.py
```diff
@@ -6,39 +6,672 @@
...
+    "environment",
+    "metadata",
+    # List of functions to call to initialize on unserialization
+    # (i.e cache load).
+    "reload_init",
+    "referenced_envs",
+]
+
+
+class CompileResult(namedtuple("_CompileResult", CR_FIELDS)):
```
gmarkall: This duplicates the code that was already vendored in to `CUDACompileResult`. I **think** we can get rid of this, the `compile_result` function, and the `NoPythonBackend` class in `typed_passes.py` , because they shouldn't be used in the CUDA target - it instead has the `CUDABackend` pass, that creates a `CUDACompileResult` with `cuda_compile_result`.

numba_cuda/numba/cuda/compiler.py
```diff
@@ -6,39 +6,672 @@
...
+        inline_pass = InlineClosureCallPass(
+            func_ir, cpu.ParallelOptions(False), {}, False
+        )
+        inline_pass.run()
+    post_proc = postproc.PostProcessor(func_ir)
+    post_proc.run(emit_dels)
+    return func_ir
+
+
+class _CompileStatus(object):
```
gmarkall: This class is unused here and duplicates the one in `numba.cuda.core.compiler._CompileStatus` - I think it can be removed.

numba_cuda/numba/cuda/compiler.py
```diff
@@ -6,39 +6,672 @@
...
+        self.can_fallback = can_fallback
+
+    def __repr__(self):
+        vals = []
+        for k in self.__slots__:
+            vals.append("{k}={v}".format(k=k, v=getattr(self, k)))
+        return ", ".join(vals)
+
+
+class _EarlyPipelineCompletion(Exception):
```
gmarkall: This also duplicates the class in `numba.cuda.core.compiler` and can probably be removed.

numba_cuda/numba/cuda/compiler.py
```diff
@@ -6,39 +6,672 @@
...
+    """
+    Raised to indicate that a pipeline has completed early
+    """
+
+    def __init__(self, result):
+        self.result = result
+
+
+class StateDict(dict):
```
gmarkall: I think this is another duplicate class.

numba_cuda/numba/cuda/compiler.py
```diff
@@ -6,39 +6,672 @@
...
+        try:
+            return self[attr]
+        except KeyError:
+            raise AttributeError(attr)
+
+    def __setattr__(self, attr, value):
+        self[attr] = value
+
+
+def _make_subtarget(targetctx, flags):
```
gmarkall: I think this duplicates the same function in `numba.cuda.core.compiler`

gmarkall (commented):
I've looked through compiler.py so far and I'm submitting the review now as I think there are some items to address there that have happened as a result of other PRs getting in first and duplicating the items in there.

Will continue looking at the other files in the PR.

numba_cuda/numba/cuda/compiler.py
```diff
@@ -1,48 +1,309 @@
...
-    IRLegalization,
-    AnnotateTypes,
-)
-from warnings import warn
-from numba.cuda import nvvmutils
+from numba.core.utils import PYVERSION
+from numba.core.errors import NumbaWarning, NumbaInvalidConfigWarning
+
+if PYVERSION < (3, 10):
+    from numba.core.interpreter import Interpreter as interpreter
```
gmarkall: This is quite confusing! It makes the class look like a module. Is there a reason that

```suggestion
    from numba.core.interpreter import Interpreter
```

wasn't as simple to use as it looks?

numba_cuda/numba/cuda/compiler.py
```diff
@@ -1,48 +1,309 @@
...
+    """
+    Run the compiler frontend over the given Python function, and return
+    the function's canonical Numba IR.
+
+    If inline_closures is Truthy then closure inlining will be run
+    If emit_dels is Truthy the ir.Del nodes will be emitted appropriately
+    """
+    # XXX make this a dedicated Pipeline?
+    func_id = bytecode.FunctionIdentity.from_function(func)
+    interp = interpreter.Interpreter(func_id)
```
gmarkall: I think this should be (assuming you don't import `Interpreter as interpreter`:

```suggestion
    interp = Interpreter(func_id)
```

I think that this doesn't cause an issue right now suggests that the vendored `run_frontend()` is not being used, but I need to check this.

numba_cuda/numba/cuda/compiler.py
```diff
@@ -246,7 +507,7 @@ def lowering_class(self):
         return lowering.CUDALower
 
 
-class CUDABytecodeInterpreter(Interpreter):
+class CUDABytecodeInterpreter(interpreter):
```
gmarkall: As per questions / suggestions above:

```suggestion
class CUDABytecodeInterpreter(Interpreter):
```

numba_cuda/numba/cuda/compiler.py
gmarkall: OK, it isn't being used because we don't have a test that does inlining with a cost model.

gmarkall:
This is a test you need to expose the issue with the `interpreter` / `Interpreter` confusion:

```diff
diff --git a/numba_cuda/numba/cuda/tests/cudapy/test_inline.py b/numba_cuda/numba/cuda/tests/cudapy/test_inline.py
index 2708b2c..96e9283 100644
--- a/numba_cuda/numba/cuda/tests/cudapy/test_inline.py
+++ b/numba_cuda/numba/cuda/tests/cudapy/test_inline.py
@@ -10,7 +10,7 @@ from numba.cuda.testing import (
 
 @skip_on_cudasim("Cudasim does not support inline and forceinline")
 class TestCudaInline(CUDATestCase):
-    def _test_call_inline(self, inline):
+    def _test_call_inline(self, inline, inline_expected):
         """Test @cuda.jit(inline=...)"""
         a = np.ones(2, dtype=np.int32)
 
@@ -33,12 +33,10 @@ class TestCudaInline(CUDATestCase):
         pat = r"call [a-zA-Z0-9]* @"
         match = re.compile(pat).search(llvm_ir)
 
-        if inline == "always" or inline is True:
+        if inline_expected:
             # check that call was inlined
             self.assertIsNone(match, msg=llvm_ir)
         else:
-            assert inline == "never" or inline is False
-
             # check that call was not inlined
             self.assertIsNotNone(match, msg=llvm_ir)
 
@@ -46,16 +44,28 @@ class TestCudaInline(CUDATestCase):
         self.assertNotIn("alwaysinline", llvm_ir)
 
     def test_call_inline_always(self):
-        self._test_call_inline("always")
+        self._test_call_inline("always", True)
 
     def test_call_inline_never(self):
-        self._test_call_inline("never")
+        self._test_call_inline("never", False)
 
     def test_call_inline_true(self):
-        self._test_call_inline(True)
+        self._test_call_inline(True, True)
 
     def test_call_inline_false(self):
-        self._test_call_inline(False)
+        self._test_call_inline(False, False)
+
+    def test_call_inline_costmodel_false(self):
+        def cost_model(expr, caller_info, callee_info):
+            return False
+
+        self._test_call_inline(cost_model, False)
+
+    def test_call_inline_costmodel_true(self):
+        def cost_model(expr, caller_info, callee_info):
+            return True
+
+        self._test_call_inline(cost_model, True)
 
     def _test_call_forceinline(self, forceinline):
         """Test @cuda.jit(forceinline=...)"""
```

numba_cuda/numba/cuda/core/object_mode_passes.py
gmarkall, line 1: I think this can be deleted now?

numba_cuda/numba/cuda/compiler.py
```diff
@@ -1,48 +1,309 @@
...
+    InlineOverloads,
+    PreLowerStripPhis,
+    NativeParforLowering,
+    NoPythonSupportedFeatureValidation,
+    ParforFusionPass,
+    ParforPreLoweringPass,
+)
+
+
+CR_FIELDS = [
```
gmarkall: We still have a redefinition of `CR_FIELDS` in this file (there is another one later)

numba_cuda/numba/cuda/testing.py
```diff
@@ -3,7 +3,8 @@
 import shutil
 import pytest
 from datetime import datetime
-from numba.core.utils import PYVERSION
+from numba.cuda.utils import PYVERSION
+from numba.tests.support import SerialMixin
```
gmarkall: This is no longer used / needed
```suggestion
```

numba_cuda/numba/cuda/core/untyped_passes.py
```diff
@@ -0,0 +1,1989 @@
...
+    errors,
+    types,
+    ir,
+    bytecode,
+    postproc,
+    rewrites,
+    config,
+    transforms,
+    consts,
+    interpreter,
```
gmarkall: I think this should come from `numba.cuda` now.

gmarkall (changes requested):
A few more notes on the diff - a couple of fixes still required, but this is getting closer.

PR #332, opened 2025-07-19
[NFC] Vendor in DIBuilder used by CUDADIBuilder

ashermancinelli:
Vendor in DIBuilder which is used by CUDADIBuilder from numba core into numba-cuda to allow for future CUDA-specific changes/refactoring. I don't believe additional tests are needed.

Please let me know if you think this is a big enough component to warrant its own module (eg `numba_cuda/numba/cuda/core/debuginfo.py`) or it it's sufficient to drop it right into the existing debuginfo module. Thanks in advance!

### DIFF
--- a/numba_cuda/numba/cuda/debuginfo.py
+++ b/numba_cuda/numba/cuda/debuginfo.py
@@ -1,11 +1,384 @@
+import os
+
 from llvmlite import ir
-from numba.core import types, cgutils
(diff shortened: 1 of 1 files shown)


ashermancinelli:
/ok to test [2638b7e](https://github.com/NVIDIA/numba-cuda/commit/2638b7e60d075c347aa87e0df6f6c4ba2dce5101)

Issue #221, opened 2025-04-16
[FEA] Relax return type restrictions for nopython functions


ZzEeKkAa:
**Is your feature request related to a problem? Please describe.**

I want to have a function to tile array, something like that:

```python
from numba import cuda
from numba.extending import overload
import numpy as np


def slice_array(a, x_id, x_size, y_id, y_size):
    pass


@overload(slice_array, inline="always", target="cuda")
def ol_slice_array(a, x_id, x_size, y_id, y_size):
    def impl(a, x_id, x_size, y_id, y_size):
        return a[
            x_id * x_size : (x_id + 1) * x_size : 1,
            y_id * y_size : (y_id + 1) * y_size : 1,
        ]

    return impl


@cuda.jit
def kernel(a):
    a_slice = slice_array(a, 1, 2, 2, 2)
    a_slice[0, 0] = 1


a = np.zeros((4, 6), dtype=np.int32)

kernel[1, 1](a)

assert a[2, 4] == 1
```

**Describe the solution you'd like**

Above code is working)

**Describe alternatives you've considered**

If I don't wrap tiling in a function it works. However I have to use tiling in my code more than ones and it increases chance of error.

Another option is to create intrinsic with llvm codegen, but it feels a bit overkill for the trivial task...

**Additional context**

I guess it comes from numa's [type check](https://github.com/numba/numba/blob/1ba9c54e395e611bbe8d2dbe7a726488f5a75fbd/numba/core/typed_passes.py#L157-L158) and the idea there is to prevent returning array from nopython to python, however this restriction is also applied to functions that work only inside nopython...

ZzEeKkAa:
Shorter reproducer:
```python
from numba import cuda
import numpy as np


@cuda.jit(device=True, forceinline=True)
def slice_array(a, x_id, x_size, y_id, y_size):
    return a[
        x_id * x_size : (x_id + 1) * x_size : 1,
        y_id * y_size : (y_id + 1) * y_size : 1,
    ]


@cuda.jit
def kernel(a):
    a_slice = slice_array(a, 1, 2, 2, 2)
    a_slice[0, 0] = 1


a = np.zeros((4, 6), dtype=np.int32)

kernel[1, 1](a)

assert a[2, 4] == 1
```

PR #335, opened 2025-07-21
[WIP] Feature: allow return array

ZzEeKkAa:
Vendor `NopythonTypeInference` pass and modify it to allow array mutations:

```
@cuda.jit(device=True, forceinline=True)
def slice_array(a, x_id, x_size, y_id, y_size):
    return a[
        x_id * x_size : (x_id + 1) * x_size : 1,
        y_id * y_size : (y_id + 1) * y_size : 1,
    ]
```

TODO: add tests

Fixes: https://github.com/NVIDIA/numba-cuda/issues/221

### DIFF
--- a/numba_cuda/numba/cuda/compiler.py
+++ b/numba_cuda/numba/cuda/compiler.py
@@ -300,6 +300,51 @@ class CUDANopythonTypeInference(NopythonTypeInference):
     def __init__(self):
         NopythonTypeInference.__init__(self)
 
+    def _legalize_array_return_type(self, return_type, interp, targetctx):
+        # Walk IR to discover all arguments and all return statements
+        retstmts = []
+        caststmts = {}
+        argvars = set()
+        for bid, blk in interp.blocks.items():
+            for inst in blk.body:
+                if isinstance(inst, numba_ir.Return):
+                    retstmts.append(inst.value.name)
+                elif isinstance(inst, numba_ir.Assign):
+                    if (
+                        isinstance(inst.value, numba_ir.Expr)
+                        and inst.value.op == "cast"
+                    ):
+                        caststmts[inst.target.name] = inst.value
+                    elif isinstance(inst.value, numba_ir.Arg):
+                        argvars.add(inst.target.name)
+
+        assert retstmts, "No return statements?"
+
+        for var in retstmts:
+            cast = caststmts.get(var)
+            if cast is None or cast.value.name not in argvars:
+                if self._raise_errors:
+                    msg = (
+                        "Only accept returning of array passed into "
+                        "the function as argument"
+                    )
+                    raise errors.NumbaTypeError(msg)
+
+    def _legalize_return_type(self, return_type, interp, targetctx):
+        """
+        Only accept array return type iff it is passed into the function.
+        Reject function object return types if in nopython mode.
+        """
+        if not targetctx.enable_nrt and isinstance(return_type, types.Array):
+            self._legalize_array_return_type(return_type, interp, targetctx)
+
+        elif isinstance(return_type, types.Function) or isinstance(
+            return_type, types.Phantom
+        ):
+            if self._raise_errors:
+                msg = "Can't return function object ({}) in nopython mode"
+                raise errors.NumbaTypeError(msg.format(return_type))
+
     def run_pass(self, state):
         """
         Type inference and legalization. Vendored from numba.core.typed_passes.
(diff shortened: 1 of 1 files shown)


numba_cuda/numba/cuda/compiler.py
```diff
@@ -291,6 +295,89 @@ def run_pass(self, state):
...
+        # Walk IR to discover all arguments and all return statements
+        retstmts = []
+        whitelist_vars = set()
+        for bid, blk in interp.blocks.items():
+            for inst in blk.body:
+                if isinstance(inst, numba_ir.Return):
+                    retstmts.append(inst.value.name)
+                elif isinstance(inst, numba_ir.Assign):
+                    if isinstance(inst.value, numba_ir.Expr):
+                        if inst.value.op not in {"cast", "getitem"}:
```
gmarkall: Thinking about the situations where allowing the result of a cast to be used helps. I presume the case is something like casting `Array(int32, 1, 'C')` to `Array(int32, 1, 'A')` - i.e. changing to a less constrained layout (or swapping a 1D array between C and F layout). Are there any other cases where allowing casts helps?

numba_cuda/numba/cuda/compiler.py
```diff
@@ -291,6 +295,89 @@ def run_pass(self, state):
...
+                            continue
+                        if inst.value.value.name in whitelist_vars:
+                            whitelist_vars.add(inst.target.name)
```
gmarkall: Is there a danger that this misses transitively allowing variable where the blocks aren't visited in the correct order? Does propagation of allowed variables need to iterate to a fixpoint instead? I'm thinking of a case like

```python
if cond:
    b = a[:, 1]
c = b
return c
```

If the block after the `if` is traversed first, is there a risk that returning `c` is disallowed?

gmarkall (changes requested):
I think this is a good idea in principle. I have a couple of questions on the diff (and await the addition of tests).

numba_cuda/numba/cuda/compiler.py
ZzEeKkAa: Yeap, it is a valid point. My thoughts was that blocks are properly ordered. And you can reference variable only if it is above the usage of the same variable. Is it a thing that blocks could be not ordered/nested?

numba_cuda/numba/cuda/compiler.py
ZzEeKkAa: `cast` was there before and apparently it is used in more cases than what you mentioned. In the example from the description it is also generated:
```
$56build_slice.22 = call $56build_slice.21($binop_mul36.14, $binop_mul50.19, $const54.20.1, func=$56build_slice.21, args=(Var($binop_mul36.14, test2.py:15), Var($binop_mul50.19, test2.py:15), Var($const54.20.1, test2.py:15)), kws=(), vararg=None, varkwarg=None, target=None)
$58build_tuple.23 = build_tuple(items=[Var($30build_slice.11, test2.py:14), Var($56build_slice.22, test2.py:15)])
$60binary_subscr.24 = getitem(value=a, index=$58build_tuple.23, fn=<built-in function getitem>)
$70return_value.25 = cast(value=$60binary_subscr.24)
return $70return_value.25
```

numba_cuda/numba/cuda/tests/cudapy/test_array_return.py
```diff
@@ -0,0 +1,82 @@
...
+
+def array_slice_2d(a, x_id, x_size, y_id, y_size):
+    return a[
+        x_id * x_size : (x_id + 1) * x_size : 1,
+        y_id * y_size : (y_id + 1) * y_size : 1,
+    ]
+
+
+def array_local(shape, dtype):
+    return cuda.local.array(shape, dtype=dtype)
```
ZzEeKkAa: @gmarkall I know this is conceptually wrong, since we are trying to return pointer to the stack memory. However if we set it to `forceinline` it should turn into a valid code, but as far as I know it is against llvm design to generate invalid code that turns into valid only because of force inline. Do you know any idea how it potentially could be achieved? I have one use case in `nvmath` that will benefit from it.

ZzEeKkAa:
Is it me messing up the test, or it is out of scope of this MR:
```
Compilation is falling back to object mode WITHOUT looplifting enabled because Function "init_xoroshiro128p_states_cpu" failed type inference due to: Invalid use of type(CPUDispatcher(<function init_xoroshiro128p_state at 0x777b44828400>)) with parameters (array(Record(s0[type=uint64;offset=0],s1[type=uint64;offset=8];16;True), 1d, C), Literal[int](0), uint64)
```

gmarkall:
> Is it me messing up the test, or it is out of scope of this MR:

That's not you, the simulator always does that. It's a bit hard to fix and not really critical so it's never got to the top of the priority list.

numba_cuda/numba/cuda/compiler.py
gmarkall: I was a bit apprehensive because I'm not certain what the ordering is, or is guaranteed to be. I'm also wondering whether phi nodes will be a problem.

ZzEeKkAa: I've updated to use forest of trees to eliminate any issues with block ordering

numba_cuda/numba/cuda/tests/cudapy/test_array_return.py
```diff
@@ -0,0 +1,241 @@
...
+                if len(alias) > 1:
+                    return alias
+            return b
+
+        self._test_array_return(array_return_alias_loop)
+
+    def test_array_return_from_tuple(self):
+        def array_return_from_tuple(a, b):
+            t = (a, b)
+            return t[0]
```
gmarkall, line 123: This is the simplest failing test case (arrays not found through tuples).

numba_cuda/numba/cuda/tests/cudapy/test_array_return.py
```diff
@@ -0,0 +1,256 @@
...
+            x[0], x[1] = 1.0, 2.0
+
+            r[0] = x[0] + x[1]
+
+        r = np.zeros(1, dtype=np.float32)
+
+        kernel[1, 1](r)
+
+        np.testing.assert_equal(r, [3.0])
+
```
gmarkall, line 253: I don't think we should have this xfailing test - it contradicts the `test_array_local_illegal` test above. It's not clear to me how we could have a valid way to return a local array.
```suggestion
```

numba_cuda/numba/cuda/tests/cudapy/test_array_return.py
```diff
@@ -0,0 +1,256 @@
...
+        def array_return_alias_loop(a, b):
+            c = cuda.shared.array(shape=(0,), dtype=np.int32)
+            for arr in (a, b, c):
+                alias = arr
+                if len(alias) > 1:
+                    return alias
+            return b
+
+        # c in the loop is a local array
+        # TODO: do we want to support local and shared arrays?
```
gmarkall, line 116: I don't think we want to support local and shared arrays being returned from a device function that declares them. Local arrays seem like a case we shouldn't support, but I'm less sure about shared arrays - does CUDA C++ allow you to return a shared array that a device function created?

numba_cuda/numba/cuda/tests/cudapy/test_array_return.py
ZzEeKkAa: In C++ it is possible with constexpr

numba_cuda/numba/cuda/tests/cudapy/test_array_return.py
ZzEeKkAa: In C++ it is possible with constexpr

PR #337, opened 2025-07-21
[NFC] Vendor in `interpret` method of bytecode interpreter

ashermancinelli:
Vendor in the interpret method of the numba bytecode interpreter to allow for customizing it outside of numba core. The method has been copied as-is, so I do not believe additional tests are needed.

### DIFF
--- a/numba_cuda/numba/cuda/compiler.py
+++ b/numba_cuda/numba/cuda/compiler.py
@@ -254,7 +254,6 @@ class CUDANativeLowering(NativeLowering):
 
 
 class CUDABytecodeInterpreter(Interpreter):
-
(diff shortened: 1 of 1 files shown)


ashermancinelli:
The tests are failing because the top of tree llvmlite and numba packages do not support Python 3.9 anymore. This means that vendoring in the numba bytecode interpreter will raise exceptions when numba-cuda runs with 3.9. We will need to consider which changes have been made since the dependency was dropped in numba and llvmlite, and if we are able to vendor in some of these changes as they are in numba today.

ashermancinelli:
Now that we have filecheck available I will close this and open a pr with the entire class and filecheck tests.

PR #338, opened 2025-07-22
[NFC] Vendor in Dispatcher as CUDADispatcher to allow for future CUDA-specific customization

VijayKandiah:
This PR vendors in some necessary pieces of `Dispatcher` class into `CUDADispatcher` for future CUDA-specific changes, making the latter inherit directly from `_DispatcherBase`. This change also vendors in supporting classes like `_MemoMixin` and `_FunctionCompiler` for future customizations. This is an NFC, so no additional unit tests are needed.


numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -732,7 +737,137 @@ def load_overload(self, sig, target_context):
             return super().load_overload(sig, target_context)
 
 
-class CUDADispatcher(Dispatcher, serialize.ReduceMixin):
+class _MemoMixin:
```
gmarkall, line 736: I don't think this is used / relevant to the CUDA target - if it is removed, does everything still appear to work?

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -751,10 +886,46 @@ class CUDADispatcher(Dispatcher, serialize.ReduceMixin):
...
 
     def __init__(self, py_func, targetoptions, pipeline_class=CUDACompiler):
-        super().__init__(
-            py_func, targetoptions=targetoptions, pipeline_class=pipeline_class
+        """
+        Parameters
+        ----------
+        py_func: function object to be compiled
+        locals: dict, optional
+            Mapping of local variable names to Numba types.  Used to override
```
gmarkall: There is no `locals` parameter.

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -751,10 +886,46 @@ class CUDADispatcher(Dispatcher, serialize.ReduceMixin):
...
+            py_func,
+            pysig,
+            can_fallback,
+            exact_match_required=False,
         )
 
+        functools.update_wrapper(self, py_func)
+
+        self.targetoptions = targetoptions
+        self.locals = locals
```
gmarkall: It took me a while to figure out how this could possibly be working, then I realised this is assigning the built-in `locals` to `self.locals`. Although it doesn't seem to have disrupted the tests, this seems dangerous - can you remove the `locals` assignment please?

(`locals` is an option that's not supported by the CUDA target).

numba_cuda/numba/cuda/dispatcher.py
gmarkall: Actually, don't worry about this comment too much - I think it will show up when we run coverage if it's not needed, so let's not think about it now.

VijayKandiah: Sounds good!

numba_cuda/numba/cuda/dispatcher.py
VijayKandiah: I removed `locals` from CUDADispatcher and _FunctionCompiler. We are now passing `locals={}` to `compiler.compile_extra` similar to how we invoke `compiler.compile_extra` in `numba/cuda/compiler.py`.

PR #342, opened 2025-07-22
[RFC,TESTING] Add filecheck test infrastructure

ashermancinelli:
Adding pattern-based checking infrastructure allows us to describe the IR and assembly expected under different conditions right next to the code that produces it. This is closer to the LLVM testing standards, and will help is iterate more quickly and add test coverage while we are in the process of vendoring in the components from numba core we need to modify in a cuda-specific way.

This PR changes a couple tests to use the filecheck patterns. For example, in numba-cuda today there tests like `self.assertIn('function_name', llvm_ir)`. Now, we can express the same check in the docstring of the kernel, and all the checks we would like to perform may live inside the kernel they come from.

This is a request-for-comment, as filecheck tests will be meaningfully different from existing numba-cuda tests.

Note also that the filecheck that has been added to the list of dependencies is _not_ the filecheck that is compiled as a part of the LLVM project, but [is an attempt to replicate the functionality in a pure python package](https://github.com/AntonLydike/filecheck/tree/main).
I ran some tests to see how well the usual FileCheck features work with the Python version, and it seems okay.
If we are uncomfortable with this development dependency, we can reevaluate.

TODO:
- [x] CUDA Test Case should have file check available by default
- [x] Review type-hint inconsistencies
- [x] Skip filecheck-based tests in python 3.9
- [x] Add dummy class for `CUDADispatcher` when running with the simulator

---

Another solution I considered would be to structure our tests much more like LLVM lit tests, not just using the filecheck pattern like I did in this patch.

In this alternative, I imagine test kernels would be written in their own files:
```python
# RUN: numba-cuda-to-llvm-ir %s | filecheck %s
# ^^ the above line converts the JIT'ed kernels in this Python module to
#    LLVM IR and checks the check-comments in the module against the IR

from numba.cuda import jit

# CHECK-LABEL: foo
# CHECK: fadd
@jit((float, float))
def test_foo(a, b):
    a = b
```


numba_cuda/numba/cuda/tests/cudapy/test_inspect.py
```diff
@@ -27,6 +27,16 @@ def test_monotyped(self):
 
         @cuda.jit(sig)
         def foo(x, y):
+            """
+            // LLVM: foo
+            // LLVM-LABEL: entry:
+            // LLVM-NEXT:         br label %"[[VAL_0:.*]]"
+            // LLVM-NEXT:       [[VAL_0]]:
```
isVoid, line 35: Is `[[VAL_0]]` a placeholder for jump label? It just needs matching labels between this line and previous line right?

numba_cuda/numba/cuda/tests/cudapy/test_inspect.py
```diff
@@ -37,28 +47,47 @@ def foo(x, y):
...
-        # Kernel in LLVM
-        self.assertIn("define void @", llvm)
-
-        asm = foo.inspect_asm(sig)
-
-        # Function name in PTX
-        self.assertIn("foo", asm)
-        # NVVM inserted comments in PTX
-        self.assertIn("Generated by NVIDIA NVVM Compiler", asm)
+        filecheck = FileCheckKernel(foo)
```
isVoid: I think `foo` is decorated with cuda.jit and is returned as a `CUDADispatcher` object. You might not access the `__doc__ ` attribute from original kernel.

numba_cuda/numba/cuda/tests/cudapy/test_inspect.py
```diff
@@ -37,28 +47,47 @@ def foo(x, y):
...
+        self.assertTrue(filecheck.check_llvm(sig))
+        self.assertTrue(filecheck.check_asm(sig))
 
     def test_polytyped(self):
         @cuda.jit
         def foo(x, y):
+            """
+            // LLVM: define void
+            // LLVM-SAME: foo
```
isVoid, line 60: From the [doc](https://llvm.org/docs/CommandGuide/FileCheck.html#the-check-same-directive) they use `CHECK:` everywhere, so I will interpret `LLVM-x` the same as `CHECK-x`. Please correct me if I'm correct about these lines:

You first use `CHECK: define void` to find the kernel signature declaration. Then use `CHECK-SAME: foo` to check that `foo` exists in the same line as `define void`. (Regardless of whether the kernel name is mangled or not, as long as `foo` exists, this check passes).

The following lines are definitions of directives that belongs to two separate FileCheck runs. One with `LLVM_INT` and one with `LLVM_FLOAT` prefix, corresponds to the two kernel launches respectively.

isVoid (commented):
I'm trying to understand the use of this tool. Also a broader question, the result of the compiler are in fact legal nvvm IRs, should we use `NVVM` as the prefix instead?

numba_cuda/numba/cuda/tests/cudapy/test_inspect.py
ashermancinelli: Yes, that is exactly right. The lines with only `LLVM` are shared between all tests that use the LLVM prefix, and the LLVM_INT and LLVM_FLOAT checks are specific to that datatype. In this way, we can interpolate the checks that should run for the IR with a particular signature into the same list of checks.

ashermancinelli: We can use CHECK as well, but I used several check prefixes in these tests to exemplify how our tests may look if we go in this direction.

numba_cuda/numba/cuda/tests/cudapy/test_inspect.py
ashermancinelli: The `__doc__` field is preserved, which I believe is being propagated by `functools.wraps` here.

numba_cuda/numba/cuda/tests/cudapy/test_inspect.py
ashermancinelli: That is right - with FileCheck we can create _captured variables_ the first time a pattern is encountered, and re-use those variables later on to ensure that values we see in one part of the text also appear in another.

ashermancinelli:
> should we use NVVM [instead of LLVM] as the prefix instead?

Sure! Whatever will most precisely differentiate the checks from one another.

ashermancinelli:
@isVoid I tried answering all your questions, but let me speak about the motivation more generally: FileCheck allows us to easily create meaningful tests (in a semi-automated way) to increase the test coverage we have so we feel more secure making larger changes. [Consider this test in LLVM Flang](https://github.com/llvm/llvm-project/blob/main/flang/test/Lower/Intrinsics/math-runtime-options.f90). There are many ways to compile the same test, some of which result in the same IR. The same source code can be reused to create 6 or so test cases, some of which share the same checks. There are scripts in the LLVM project to generate and update these tests automatically.

When we vendor in components for the purpose of cuda-specific extensions, instead of vendoring in the unit tests from numba core into numba cuda directly, we can vendor in the kernels and use `cuda.jit` instead of the regular numba jit, and ensure the IR is the same. Then, as we change the components, we can be sure we have not changed from the IR we were getting with the non-vendored version (or if we have changed the IR, it is in the way we expect).

This is also just a suggestion/RFC. We will discuss this further before submitting any changes. Thank you for the review and please ask more questions if I left anything unclear!

numba_cuda/numba/cuda/testing.py
```diff
@@ -203,3 +207,75 @@ class ForeignArray(object):
...
         self._arr = arr
         self.__cuda_array_interface__ = arr.__cuda_array_interface__
+
+
+def filecheck_ir(
+    ir_content: str,
+    check_patterns: str,
+    check_prefixes: list[str] = ["CHECK"],
+    **extra_filecheck_options: dict[str, str | int],
+) -> bool:
```
gmarkall: If it always succeeds or raises an `AssertionError`, then there's probably no point returning anything.

numba_cuda/numba/cuda/tests/cudapy/test_inspect.py
gmarkall: I think we even have a test that `__doc__` is preserved - it has to work because users might want autodoc generation for kernels, and also to be able to see the docstring of kernels in interactive sessions.

pyproject.toml
```diff
@@ -42,6 +42,7 @@ test = [
     "psutil",
     "cffi",
     "pytest",
+    "filecheck",
```
gmarkall, line 45: It's a little unfortunate there's no conda package for filecheck (as far as I can tell).

numba_cuda/numba/cuda/testing.py
```diff
@@ -203,3 +207,75 @@ class ForeignArray(object):
...
+    """Run filecheck on IR content with check patterns. Raises an AssertionError if the checks fail."""
+    opts = Options(
+        match_filename="-",
+        check_prefixes=check_prefixes,
+        **extra_filecheck_options,
+    )
+    fin = FInput(fname="-", content=ir_content)
+    parser = Parser(opts, StringIO(check_patterns), *pattern_for_opts(opts))
+    matcher = Matcher(opts, fin, parser)
+    result = matcher.run()
```
gmarkall: Can the matcher produce any output (e.g. `stderr`) that we can capture and provide to the user?

numba_cuda/numba/cuda/testing.py
```diff
@@ -203,3 +207,75 @@ class ForeignArray(object):
...
+                f"FileCheck failed with exit code {result}\n\n"
+                f"Check prefixes:\n{check_prefixes}\n\n"
+                f"Check patterns:\n{check_patterns}\n"
+                f"IR:\n{ir_content}\n\n"
+            )
+        )
+    return True
+
+
+class FileCheckKernel:
```
gmarkall: I think rather than a separate `FileCheckKernel` object, I'd prefer if we moved the `check_*` methods into `CUDATestCase`, and then could call:

```
self.check_llvm(kernel, signature)
```

instead. Instead of raising an `AssertionError`, `self.fail` should then be called with the same message that the `AssertionError` is currently constructed with.

I would prefer this because the logic seems a bit more complicated than necessary, both in the implementation and at the point of use - we create a new object, then during the check we raise an `AssertionError`, and have to return a `True` that we can assert in the test case.

numba_cuda/numba/cuda/tests/cudapy/test_inspect.py
```diff
@@ -37,28 +47,47 @@ def foo(x, y):
...
-
-        asm = foo.inspect_asm(sig)
-
-        # Function name in PTX
-        self.assertIn("foo", asm)
-        # NVVM inserted comments in PTX
-        self.assertIn("Generated by NVIDIA NVVM Compiler", asm)
+        filecheck = FileCheckKernel(foo)
+        self.assertTrue(filecheck.check_llvm(sig))
+        self.assertTrue(filecheck.check_asm(sig))
```
gmarkall: With the suggestion above (https://github.com/NVIDIA/numba-cuda/pull/342/files#r2225538599), I'd envisage this looking like:

```suggestion
        self.check_llvm(kernel, sig)
        self.check_asm(kernel, sig)
```

which is a bit closer to the idioms of the rest of the unittest suite.

gmarkall (changes requested):
I really like the idea here! This looks like a nice way to check the generated code.

I do think we could simplify the implementation and API, and make it a bit closer to the idioms of the rest of the test suite - see comments on the diff for these suggestions.

I also wonder - does FileCheck identify the failing line and provide error messages on stderr? I notice the `Matcher` class has an `stderr` member but I haven't dived deeper - I am looking to understand whether the framework can provide more precise error messages when failures occur.

numba_cuda/numba/cuda/testing.py
ashermancinelli: Good catch, I forgot to update the return value and type annotations after making a change. Updated!

numba_cuda/numba/cuda/testing.py
ashermancinelli: Yes, it does provide output. I return the patterns and the IR in the assertion error so it's easy to see what's wrong and generate new patterns, but I'm open to any suggestions since I'm not familiar with the testing conventions yet!

numba_cuda/numba/cuda/testing.py
ashermancinelli: This seems like a great idea! Let me give it a try.

ashermancinelli:
> I also wonder - does FileCheck identify the failing line and provide error messages on stderr?

It does - but please note that this is a Python re-implementation of the FileCheck executable program which is a part of the LLVM project, so I am also unfamiliar with the API of this Python module. I will investigate further and see how we can make this more ergonomic. Thank you for the review!

numba_cuda/numba/cuda/testing.py
```diff
@@ -203,3 +209,96 @@ class ForeignArray(object):
...
+    CUDADispatcher and assert that the compmilation artifacts match the
+    FileCheck checks given in the kernel's docstring.
+
+    Method assertFileCheckMatches can be used to assert that a given string
+    matches FileCheck checks, and is not specific to CUDADispatcher.
+    """
+
+    def assert_filecheck_asm(
+        self,
+        ir_producer: CUDADispatcher,
```
gmarkall: The `CUDADispatcher` type is not the same as the `CUDADispatcher` dispatcher object returned by the `cuda.jit` decorator.

ashermancinelli: Thank you!

numba_cuda/numba/cuda/tests/cudapy/test_inspect.py
ashermancinelli: What do you think of the more verbose names I chose?
```python
        self.assert_filecheck_llvm(foo, sig)
        self.assert_filecheck_asm(foo, sig)
```
Note that I stuck to snake_case naming conventions while the unittest framework uses mixedCase. I am unsure about this decision.

numba_cuda/numba/cuda/testing.py
ashermancinelli: I was able to set the stderr member to a stringio object to capture the errors. My assertions now manifest like this:

<details><summary>Click to expand</summary>

```shell
AssertionError: FileCheck failed:
-:7: error: Couldn't match "bar".
Current position at <stdin>:4:1
}
^


Check prefixes:
['LLVM']

Check patterns:

// LLVM: foo
// LLVM-LABEL: entry:
// LLVM-NEXT:         br label %"[[VAL_0:.*]]"
// LLVM-NEXT:       [[VAL_0]]:
// LLVM-NEXT:         ret void
// LLVM: bar

// ASM: Generated by NVIDIA NVVM Compiler
// ASM: foo

IR:
; ModuleID = "TestInspect.test_monotyped.<locals>.foo$3"
target triple = "nvptx64-nvidia-cuda"
target datalayout = "e-p:64:64:64-i1:8:8-i8:8:8-i16:16:16-i32:32:32-i64:64:64-i128:128:128-f32:32:32-f64:64:64-v16:16:16-v32:32:32-v64:64:64-v128:128:128-n16:32:64"

@"_ZN08NumbaEnv5numba4cuda5tests6cudapy12test_inspect11TestInspect14test_monotyped12_3clocals_3e3fooB2v3B104cw51cXTLSUwv1sDUaKthqaNgoKmjgOR3W3Cw6igA9duC0hkwnNGiHEkYkrqQBFBTDEwCyQe21eqwcFW3UoDGjzpQEsgzrtUEAA_3d_3dEfi" = common global i8* null
define void @"_ZN5numba4cuda5tests6cudapy12test_inspect11TestInspect14test_monotyped12_3clocals_3e3fooB2v3B104cw51cXTLSUwv1sDUaKthqaNgoKmjgOR3W3Cw6igA9duC0hkwnNGiHEkYkrqQBFBTDEwCyQe21eqwcFW3UoDGjzpQEsgzrtUEAA_3d_3dEfi"(float %"x", i32 %"y")
{
entry:
  br label %"B0"
B0:
  ret void
}

@"llvm.used" = appending global [1 x i8*] [i8* bitcast (void (float, i32)* @"_ZN5numba4cuda5tests6cudapy12test_inspect11TestInspect14test_monotyped12_3clocals_3e3fooB2v3B104cw51cXTLSUwv1sDUaKthqaNgoKmjgOR3W3Cw6igA9duC0hkwnNGiHEkYkrqQBFBTDEwCyQe21eqwcFW3UoDGjzpQEsgzrtUEAA_3d_3dEfi" to i8*)], section "llvm.metadata"
!nvvmir.version = !{ !0 }
!nvvm.annotations = !{ !1 }
!0 = !{ i32 2, i32 0, i32 3, i32 2 }
!1 = !{ void (float, i32)* @"_ZN5numba4cuda5tests6cudapy12test_inspect11TestInspect14test_monotyped12_3clocals_3e3fooB2v3B104cw51cXTLSUwv1sDUaKthqaNgoKmjgOR3W3Cw6igA9duC0hkwnNGiHEkYkrqQBFBTDEwCyQe21eqwcFW3UoDGjzpQEsgzrtUEAA_3d_3dEfi", !"kernel", i32 1 }
```

</details>

I think this is enough information for test authors to update their tests. One may even pipe the results to the llvm test generation scripts to get auto-generated checks from unit test error messages.

ashermancinelli:
/ok to test [c45924d](https://github.com/NVIDIA/numba-cuda/pull/342/commits/c45924dcd3e2ee8de5b02a2a04b7e60f960f9486)

pyproject.toml
ashermancinelli: I see that.. ~that's why the CI is failing. Any ideas what we should do about this?~ Is this a blocker?

pyproject.toml
```diff
@@ -42,7 +42,6 @@ test = [
     "psutil",
     "cffi",
     "pytest",
-    "filecheck",
```
gmarkall: I think we need this - was it accidentally removed?

ashermancinelli: Yes!

pyproject.toml
gmarkall: No, you just need to install it with `pip` in the CI script, like

```
pip install filecheck
```

ashermancinelli:
/ok to test [df00dc7](https://github.com/NVIDIA/numba-cuda/pull/342/commits/df00dc7e86f5cb208b4dcd388ed12cde2f5a3f35)

numba_cuda/numba/cuda/tests/cudapy/test_inspect.py
gmarkall: I think the names you chose are worded better than my suggestion. I think mixedCase should be used to be consistent with unittest, like:

```python
self.assertFilecheckAsm(foo, sig)
```

The other Numba test suite extensions to unittest do this (e.g. `assertRefCount()`, `assertNoNRTLeak()`, etc.). `unittest` is weird for Python (IMHO), I think it's because it was derived from / inspired by Java's unittest).

ashermancinelli: Updated!

ashermancinelli:
/ok to test [0552773](https://github.com/NVIDIA/numba-cuda/pull/342/commits/05527730bb1531802cc1e58371dc6eee183ff2fd)

isVoid:
@ashermancinelli thanks for the explanation. When we write tests, we are limited by the tool that we have so that we can't test more complicated pattern. This opens up a whole range of options for us. Just a simple grab through the code base with `self\.assertIn\(.*, (asm|ptx|sass|ir|llvm|llvmirs|llvmirs.*)\)` we have these tests that can potentially reuse this test infra.

<details>

```
/workspace/numbast-nvshmem/numba-cuda/numba_cuda/numba/cuda/tests/cudadrv/test_nvvm_driver.py
  107,9:         self.assertIn(".target sm_{0}{1}".format(*arch), ptx)
  108,9:         self.assertIn("simple", ptx)
  109,9:         self.assertIn("ave", ptx)

/workspace/numbast-nvshmem/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_atomics.py
  691,17:                 self.assertIn("atom.shared.cas.b64", asm)
  693,17:                 self.assertIn("atom.cas.b64", asm)

/workspace/numbast-nvshmem/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_casting.py
  138,13:             self.assertIn(f"cvt.rni.s{size}.f16", ptx)
  160,13:             self.assertIn(f"cvt.rni.u{size}.f16", ptx)
  192,13:             self.assertIn(f"cvt.rn.f16.s{size}", ptx)
  201,13:             self.assertIn(f"cvt.rn.f16.u{size}", ptx)
  229,13:             self.assertIn(f"cvt.{postfix}.f16", ptx)

/workspace/numbast-nvshmem/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_compiler.py
  35,9:         self.assertIn(".visible .entry", ptx)
  48,9:         self.assertIn("func_retval", ptx)
  50,9:         self.assertIn(".visible .func", ptx)
  78,9:         self.assertIn("fma.rn.f32", ptx)
  79,9:         self.assertIn("div.rn.f32", ptx)
  80,9:         self.assertIn("sqrt.rn.f32", ptx)
  85,9:         self.assertIn("fma.rn.ftz.f32", ptx)
  86,9:         self.assertIn("div.approx.ftz.f32", ptx)
  87,9:         self.assertIn("sqrt.approx.ftz.f32", ptx)
  270,9:         self.assertIn(target, ptx)

/workspace/numbast-nvshmem/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_device_func.py
  129,9:         self.assertIn(fname, llvm)
  146,9:         self.assertIn(fname, ptx)

/workspace/numbast-nvshmem/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_inspect.py
  36,9:         self.assertIn("foo", llvm)
  39,9:         self.assertIn("define void @", llvm)
  44,9:         self.assertIn("foo", asm)
  46,9:         self.assertIn("Generated by NVIDIA NVVM Compiler", asm)
  70,9:         self.assertIn((intp, intp), llvmirs)
  71,9:         self.assertIn((float64, float64), llvmirs)
  74,9:         self.assertIn("foo", llvmirs[intp, intp])
  75,9:         self.assertIn("foo", llvmirs[float64, float64])
  78,9:         self.assertIn("define void @", llvmirs[intp, intp])
  79,9:         self.assertIn("define void @", llvmirs[float64, float64])
  107,9:         self.assertIn("S2R", sass)  # Special register to register
  108,9:         self.assertIn("BRA", sass)  # Branch
  109,9:         self.assertIn("EXIT", sass)  # Exit program

/workspace/numbast-nvshmem/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_intrinsics.py
  645,9:         self.assertIn("add.f16", ptx)
  672,9:         self.assertIn("fma.rn.f16", ptx)
  697,9:         self.assertIn("sub.f16", ptx)
  722,9:         self.assertIn("mul.f16", ptx)
  767,9:         self.assertIn("neg.f16", ptx)
  790,9:         self.assertIn("abs.f16", ptx)

/workspace/numbast-nvshmem/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_libdevice.py
  165,9:         self.assertIn("ld.param", ptx)
  169,9:         self.assertIn("st.global", ptx)

/workspace/numbast-nvshmem/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_lineinfo.py
  216,9:         self.assertIn(target, ptx)

/workspace/numbast-nvshmem/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_minmax.py
  46,9:         self.assertIn(ptx_instruction, ptx)

/workspace/numbast-nvshmem/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_operator.py
  182,17:                 self.assertIn(instr, ptx)
  216,17:                 self.assertIn(instr, ptx)
  259,9:         self.assertIn("neg.f16", ptx)
  266,9:         self.assertIn("abs.f16", ptx)
  400,17:                 self.assertIn(s, ptx)
  435,17:                 self.assertIn(opstring[op], ptx)
  481,17:                 self.assertIn(ops, ptx)

/workspace/numbast-nvshmem/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_optimization.py
  70,17:                 self.assertIn(fragment, ptx)
  77,9:         self.assertIn("fma.rn.f64", ptx)

/workspace/numbast-nvshmem/numba-cuda/numba_cuda/numba/cuda/tests/nocuda/test_nvvm.py
  52,9:         self.assertIn(myconstant, ptx)

/workspace/numba-cuda/numba_cuda/numba/cuda/tests/cudadrv/test_nvvm_driver.py
  107,9:         self.assertIn(".target sm_{0}{1}".format(*arch), ptx)
  108,9:         self.assertIn("simple", ptx)
  109,9:         self.assertIn("ave", ptx)

/workspace/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_atomics.py
  691,17:                 self.assertIn("atom.shared.cas.b64", asm)
  693,17:                 self.assertIn("atom.cas.b64", asm)

/workspace/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_casting.py
  138,13:             self.assertIn(f"cvt.rni.s{size}.f16", ptx)
  160,13:             self.assertIn(f"cvt.rni.u{size}.f16", ptx)
  192,13:             self.assertIn(f"cvt.rn.f16.s{size}", ptx)
  201,13:             self.assertIn(f"cvt.rn.f16.u{size}", ptx)
  229,13:             self.assertIn(f"cvt.{postfix}.f16", ptx)

/workspace/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_compiler.py
  35,9:         self.assertIn(".visible .entry", ptx)
  48,9:         self.assertIn("func_retval", ptx)
  50,9:         self.assertIn(".visible .func", ptx)
  78,9:         self.assertIn("fma.rn.f32", ptx)
  79,9:         self.assertIn("div.rn.f32", ptx)
  80,9:         self.assertIn("sqrt.rn.f32", ptx)
  85,9:         self.assertIn("fma.rn.ftz.f32", ptx)
  86,9:         self.assertIn("div.approx.ftz.f32", ptx)
  87,9:         self.assertIn("sqrt.approx.ftz.f32", ptx)
  270,9:         self.assertIn(target, ptx)

/workspace/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_device_func.py
  129,9:         self.assertIn(fname, llvm)
  146,9:         self.assertIn(fname, ptx)

/workspace/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_inspect.py
  36,9:         self.assertIn("foo", llvm)
  39,9:         self.assertIn("define void @", llvm)
  44,9:         self.assertIn("foo", asm)
  46,9:         self.assertIn("Generated by NVIDIA NVVM Compiler", asm)
  70,9:         self.assertIn((intp, intp), llvmirs)
  71,9:         self.assertIn((float64, float64), llvmirs)
  74,9:         self.assertIn("foo", llvmirs[intp, intp])
  75,9:         self.assertIn("foo", llvmirs[float64, float64])
  78,9:         self.assertIn("define void @", llvmirs[intp, intp])
  79,9:         self.assertIn("define void @", llvmirs[float64, float64])
  107,9:         self.assertIn("S2R", sass)  # Special register to register
  108,9:         self.assertIn("BRA", sass)  # Branch
  109,9:         self.assertIn("EXIT", sass)  # Exit program

/workspace/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_intrinsics.py
  645,9:         self.assertIn("add.f16", ptx)
  672,9:         self.assertIn("fma.rn.f16", ptx)
  697,9:         self.assertIn("sub.f16", ptx)
  722,9:         self.assertIn("mul.f16", ptx)
  767,9:         self.assertIn("neg.f16", ptx)
  790,9:         self.assertIn("abs.f16", ptx)

/workspace/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_libdevice.py
  165,9:         self.assertIn("ld.param", ptx)
  169,9:         self.assertIn("st.global", ptx)

/workspace/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_lineinfo.py
  216,9:         self.assertIn(target, ptx)

/workspace/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_minmax.py
  46,9:         self.assertIn(ptx_instruction, ptx)

/workspace/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_operator.py
  182,17:                 self.assertIn(instr, ptx)
  216,17:                 self.assertIn(instr, ptx)
  259,9:         self.assertIn("neg.f16", ptx)
  266,9:         self.assertIn("abs.f16", ptx)
  400,17:                 self.assertIn(s, ptx)
  435,17:                 self.assertIn(opstring[op], ptx)
  481,17:                 self.assertIn(ops, ptx)

/workspace/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_optimization.py
  70,17:                 self.assertIn(fragment, ptx)
  77,9:         self.assertIn("fma.rn.f64", ptx)

/workspace/numba-cuda/numba_cuda/numba/cuda/tests/nocuda/test_nvvm.py
  52,9:         self.assertIn(myconstant, ptx)
```

</details>

ci/test_conda.sh
```diff
@@ -35,6 +35,8 @@ set +u
 conda activate test
 set -u
 
+pip install filecheck
```
gmarkall, line 38: I think you will need this in the other CI test scripts as well.

numba_cuda/numba/cuda/testing.py
```diff
@@ -5,16 +5,114 @@
...
 test_data_dir = numba_cuda_dir / "tests" / "data"
 
 
-class CUDATestCase(SerialMixin, TestCase):
+class FileCheckTestCaseMixin:
+    """
+    Mixin for tests that use FileCheck.
+
+    Methods assertFileCheckAsm and assertFileCheckLLVM will inspect a numba
+    CUDADispatcher and assert that the compmilation artifacts match the
```
gmarkall: Minor edits:

```suggestion
    Methods assertFileCheckAsm and assertFileCheckLLVM will inspect a
    CUDADispatcher and assert that the compilation artifacts match the
```

numba_cuda/numba/cuda/testing.py
```diff
@@ -5,16 +5,114 @@
...
+
+    Method assertFileCheckMatches can be used to assert that a given string
+    matches FileCheck checks, and is not specific to CUDADispatcher.
+    """
+
+    def assertFileCheckAsm(
+        self,
+        ir_producer: CUDADispatcher,
+        signature: tuple[type, ...] | None = None,
+        check_prefixes: list[str] = ["ASM"],
```
gmarkall: This use of a list as a default argument is probably safe as it's not mutated, but it always sets off alarm bells to see a mutable default in a Python function definition. I think you could either use `None` as in https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments, or have a tuple for the default instead (I assume filecheck will accept a tuple as well).

ashermancinelli:
/ok to test [8d6dcb8](https://github.com/NVIDIA/numba-cuda/pull/342/commits/8d6dcb87ec6df9a249d904e665fd5a26ebf21f52)

numba_cuda/numba/cuda/testing.py
ashermancinelli: A tuple or frozenset would be okay with me, let me try that out. Thanks!

numba_cuda/numba/cuda/testing.py
```diff
@@ -5,16 +5,114 @@
...
+            check_prefixes=check_prefixes,
+            **extra_filecheck_options,
+        )
+        input_file = FInput(fname="-", content=ir_content)
+        parser = Parser(opts, StringIO(check_patterns), *pattern_for_opts(opts))
+        matcher = Matcher(opts, input_file, parser)
+        matcher.stderr = StringIO()
+        result = matcher.run()
+        if result != 0:
+            raise AssertionError(
```
gmarkall: In line with the implementation of other `assert` methods:

```suggestion
            self.fail(
```

The outcome is the same in the context of the Numba-CUDA test suite, but using `self.fail` is compatible with someone changing the failure exception class for a test class - see https://docs.python.org/3/library/unittest.html#unittest.TestCase.failureException

numba_cuda/numba/cuda/testing.py
```diff
@@ -5,16 +5,114 @@
 from numba.tests.support import SerialMixin
 from numba.cuda.cuda_paths import get_conda_ctk
 from numba.cuda.cudadrv import driver, devices, libs
+from numba.cuda.dispatcher import CUDADispatcher
```
gmarkall, line 9: The simulator works by interacting with the import machinery to avoid importing real CUDA target classes, but instead its "dummy" classes that provide just enough of an empty implementation for imports to succeed. This import here is creating a problem for the simulator by getting the real target imported without giving the simulator a chance to prevent that.

We have a couple of options:

1. Remove this import and use the string `"CUDADispatcher"` for the type hint (I think that's an option, based on my limited understanding of Python typing)
2. Add a dummy dispatcher module to the simulator.

The dummy dispatcher module would need:

* A new file, `numba_cuda/numba/cuda/simulator/dispatcher.py` which contains only `class CUDADispatcher: ...`
* An addition to insert `numba.cuda.dispatcher` into `sys.modules`, referencing the new dummy `simulator.dispatcher`, similar to the setup for other modules at https://github.com/NVIDIA/numba-cuda/blob/392fed00cfcfc5c6e9ee14468f1c13c9edd96ce5/numba_cuda/numba/cuda/simulator/__init__.py#L33-L57

ashermancinelli: I would like to use the most precise type annotations I can, but it is ultimately not that important for this PR. Maybe we can look at doing this in a follow-up change so we can use filecheck-based tests sooner rather than later. Thank you for the suggestions!

ashermancinelli:
The filecheck module compatible with python 3.9 is too old to be used here.

ashermancinelli:
/ok to test [8fb7920](https://github.com/NVIDIA/numba-cuda/pull/342/commits/8fb792016dec69e08bc910f837532e326d64c9bb)

numba_cuda/numba/cuda/testing.py
ashermancinelli: Added the dummy class instead, 3.9 looks okay now. Will be watching CI.

ashermancinelli:
/ok to test [dd6eb32](https://github.com/NVIDIA/numba-cuda/pull/342/commits/dd6eb32d68c6b26102c170293afe30b4214a4a26)

gmarkall (approved):

gmarkall:
I think the PR is good. I am just running another check with a merge of `main` to make sure that all is well with pytest being the test framework before merging this PR.

PR #345, opened 2025-07-23
[NFC] Document development dependencies and dependency groups

ashermancinelli:
Follow-up to 1a69197, running the following should pull all the dependencies needed to develop numba-cuda for a given cuda version.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,11 +43,11 @@ These dependencies are listed under the `test-cu11` and `test-cu12` optional dep
 To install Numba-CUDA for development, run this in the root of the repository:
 
 ```shell
-pip install -e ".[cu11,test-cu11]"
+pip install -e ".[test-cu11]"
 ```
 or
 ```sh
-pip install -e ".[cu12,test-cu12]"
+pip install -e ".[test-cu12]"
 ```
 
 Numba-CUDA uses [`pre-commit`](https://pre-commit.com/) to run a number of style
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,32 +35,34 @@ into three categories:
 Remember, if you are unsure about anything, don't hesitate to comment on issues
 and ask for clarifications!
 
-## Pre-commit checks
+## Installation of Development Dependencies
+
+There are some dependencies that are required for developing Numba-CUDA, but are not required for installation or distribution.
+These dependencies are listed under the `test-cu11` and `test-cu12` optional dependency groups in our project configuration.
+
+To install Numba-CUDA for development, run this in the root of the repository:
+
+```shell
+pip install -e ".[cu11,test-cu11]"
+```
+or
+```sh
+pip install -e ".[cu12,test-cu12]"
+```
 
 Numba-CUDA uses [`pre-commit`](https://pre-commit.com/) to run a number of style
 checks in CI. To ensure your contribution will pass the checks, you can also set
 up pre-commit locally to run the checks prior to committing.
 
-To set up pre-commit with Numba-CUDA, run
-
-```
-pip install pre-commit
-```
-
-or
-
-```
-conda install pre-commit
-```
-
-depending on your package manager, then run
-
-```
+```shell
+# This will run a small set of checks every time you commit.
 pre-commit install
 ```
 
-in the root of the Numba-CUDA repository to set up the pre-commit hooks to run
-checks prior to each commit.
+To run them manually without committing, run this in the root of the repository:
+```shell
+pre-commit run --all-files
+```
 
 ## Releases
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,14 @@ cu12 = [
     "nvidia-cuda-nvrtc-cu12",
     "nvidia-nvjitlink-cu12"
 ]
+test = [
+    "pre-commit",
+    "psutil",
+    "cffi",
+    "pytest",
+    "pytest-xdist",
+    "filecheck",
+]
 test-cu11 = [
     "pre-commit",
     "psutil",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,36 +38,24 @@ cu12 = [
     "nvidia-cuda-nvrtc-cu12",
     "nvidia-nvjitlink-cu12"
 ]
-test = [
+test-cu11 = [
+    "pre-commit",
     "psutil",
     "cffi",
     "pytest",
     "pytest-xdist",
     "filecheck",
-]
-test-cu11 = [
-    "numba-cuda[test]",
     "nvidia-curand-cu11",
 ]
 test-cu12 = [
-    "numba-cuda[test]",
+    "pre-commit",
+    "psutil",
+    "cffi",
+    "pytest",
+    "pytest-xdist",
+    "filecheck",
     "nvidia-curand-cu12",
 ]
-dev = [
-    "pre-commit",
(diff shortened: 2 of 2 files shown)


ashermancinelli:
/ok to test [bd65778](https://github.com/NVIDIA/numba-cuda/pull/345/commits/bd657789be48cfb1710fcedcfa1baa0ee2cd5e86)

gmarkall:
We should probably update the readme to mention the use of `[dev-cuXX]` when installing from source... I had missed (or forgotten) the existence of these and had manually been installing dependencies myself.

ashermancinelli:
I added them in #341 to make installing the development dependencies more straightforward. I should definitely update the docs as well.

ashermancinelli:
Re-requesting review since I updated the documentation. 

Note that the docs also mention installing with conda, however this will require installling `filecheck` with pip anyways if #342, so I don't know if it will be possible to use the optional dependency groups with conda. How much of a priority is it to document how to set up Numba-CUDA with conda as the package manager?

CONTRIBUTING.md
```diff
@@ -35,32 +35,34 @@ into three categories:
...
+```shell
+pip install -e ".[dev-cu11]"
 ```
-pip install pre-commit
-```
-
 or
-
-```
-conda install pre-commit
```
kkraus14, line 53: We shouldn't get rid of the conda directions where if someone wishes to continue using the conda ecosystem we should continue supporting them

ashermancinelli: Done!

CONTRIBUTING.md
```diff
@@ -35,32 +35,39 @@ into three categories:
...
 or
-
+```sh
+pip install -e ".[dev-cu12]"
 ```
+
+Or, if you are using conda, run this in the root of the repository:
+```shell
 conda install pre-commit
 ```
```
kkraus14: Can we move this below the sentences below related to `pre-commit`. This doesn't need to be run in the root of the repo.

ashermancinelli: Thank you again for the review!

ashermancinelli:
/ok to test [380c680](https://github.com/NVIDIA/numba-cuda/pull/345/commits/380c680e253d3f3303e28f50c3d2248a3cf1a7c8)

CONTRIBUTING.md
```diff
@@ -69,6 +64,11 @@ To run them manually without committing, run this in the root of the repository:
 pre-commit run --all-files
 ```
 
+Or, if you are using conda:
+```shell
+conda install pre-commit
+```
```
kkraus14: This makes it sound like you can run the `conda install` command instead of the `pre-commit` command. In hindsight, placing this anywhere is awkward where maybe you were right to just remove it for now.

We could add a `pixi.toml` file in the future for conda users to have a development environment.

ashermancinelli: Okay, if @gmarkall doesn't have any other review comments I'll remove this comment again tomorrow. Anything else you'd like to see in the contributing doc?

kkraus14: No. I apologize for the churn here!

ashermancinelli: I appreciate your input!

ashermancinelli:
Optional dependency groups have some gotchas I was not aware of. I will need to make a couple changes first.

ashermancinelli:
/ok to test [70581bc](https://github.com/NVIDIA/numba-cuda/pull/345/commits/70581bcef6932e5e4d717771818f84f0a91ce7cc)

ashermancinelli:
/ok to test [3bb1ee3](https://github.com/NVIDIA/numba-cuda/pull/345/commits/3bb1ee3e33f715a45387534076c81501b66bcf5e)

ashermancinelli:
/ok to test [8357994](https://github.com/NVIDIA/numba-cuda/pull/345/commits/8357994df20594ada2367711bf13b7e88e61c78f)

ashermancinelli:
/ok to test [a7f6247](https://github.com/NVIDIA/numba-cuda/pull/345/commits/a7f6247bb37331fca2e1b7a1f99c9a53e4936c2c)

gmarkall (approved):
The documentation changes look fine to me. I don't think we need to also document the steps for conda separately - I think they look fine as they are.

I don't really understand the changes in pyproject.toml, but if it works / makes sense for others then it seems good to me!

gmarkall:
Having second thoughts - seeing issues like https://github.com/microsoft/syntheseus/pull/125 makes me wonder if we're going to be prone to some kind of breakage until self-referential groups are officially supported (it seems from https://github.com/pypa/pip/issues/11296 that there's no test of self-referential groups). Maybe it's not such a good idea to rely on them?

ashermancinelli:
The mysterious errors in the syntheseus issue you linked do not look fun to debug. I will remove the self-referential optional installation groups. Thanks for taking a look!

ashermancinelli:
/ok to test [24baa49](https://github.com/NVIDIA/numba-cuda/commit/24baa49d3e65a99594fc1e6010c6a5a89c20221f)

ashermancinelli:
/ok to test [da9ab24](https://github.com/NVIDIA/numba-cuda/commit/da9ab24457f2bf73f75dde72999ab3fad4ee0e54)

leofang:
FWIW Keith suggested to use the self-reference trick and it's been working fine in cuda python.

kkraus14 (approved):

pyproject.toml
```diff
@@ -39,36 +39,32 @@ cu12 = [
...
     "filecheck",
 ]
 test-cu11 = [
-    "numba-cuda[test]",
+    "pre-commit",
+    "psutil",
+    "cffi",
+    "pytest",
+    "pytest-xdist",
+    "filecheck",
```
kkraus14: Should replace all of these with `numba-cuda[test]` similar to https://github.com/NVIDIA/cuda-python/blob/main/cuda_core/pyproject.toml#L52-L54

pyproject.toml
```diff
@@ -39,36 +39,32 @@ cu12 = [
...
     "nvidia-curand-cu11",
 ]
 test-cu12 = [
```
kkraus14, line 54: I think we're missing `pre-commit` from here and is another reason to use `numba-cuda[test]` instead of duplicating all of the dependencies

ashermancinelli: Thanks for pointing out the missing dependency!

I had originally used self-referential dependency groups, and Graham suggested that we not use them. @gmarkall do you have any other thoughts about this? Are we concerned about self-referential dependency groups if other packages in our ecosystem are using them?

pyproject.toml
ashermancinelli: See my response below. Thanks for the review!

pyproject.toml
gmarkall: Sorry, I should be clear that I had not noticed that Keith suggested them. I think if/when others (Keith, Leo) are happy then this can be merged, and I will bow out  of commenting further as I've only added confusion here - many thanks!

gmarkall (approved):

ashermancinelli:
/ok to test [6837ffb](https://github.com/NVIDIA/numba-cuda/pull/345/commits/6837ffb31bbe13e336913c3102a9becc8b6901c4)

pyproject.toml
ashermancinelli: Okay, I've added them back again. Keith, does this look about right to you?

kkraus14 (approved):

PR #347, opened 2025-07-23
Migrate test infra to pytest

ZzEeKkAa:
Apparently `pytest` was already in dependencies but it was not working. There was no configuration and pytest could not recognize the layout. Now you can both run as file and as a package. Everything works:
```
pytest ./
pytest --pyargs numba.cuda.tests -v
pytest numba_cuda/numba/cuda/tests/cudapy/test_operator.py
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ If you want to manage all run-time dependencies yourself, also pass the `--no-de
 ## Running tests
 
 ```
-python -m numba.runtests numba.cuda.tests
+pytest --pyargs numba.cuda.tests -v
 ```
 
 This should discover the `numba.cuda` module from the `numba_cuda` package. You
--- a/ci/test_conda.sh
+++ b/ci/test_conda.sh
@@ -76,7 +76,7 @@ fi
 
 
 rapids-logger "Run Tests"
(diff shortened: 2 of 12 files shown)


kkraus14:
Does everywhere that we're currently using `unittest`, i.e. https://github.com/NVIDIA/numba-cuda/blob/main/numba_cuda/numba/cuda/testing.py, need to be migrated to use pytest as well?

ZzEeKkAa:
> Does everywhere that we're currently using `unittest`, i.e. https://github.com/NVIDIA/numba-cuda/blob/main/numba_cuda/numba/cuda/testing.py, need to be migrated to use pytest as well?

`pytest` supports running Python `unittest`-based tests out of the box. https://docs.pytest.org/en/stable/how-to/unittest.html . So we don't have to

ZzEeKkAa:
For some reason wheel-test workflow does not see `pytest` as an executable, maybe some terminal settings, like user mode `shell: bash -l {0}`? Logs show that pytest was installed. Conda tests pass so.

UPD: decided to go with `python -m pytest` way.

gmarkall:
>  I'll experiment with / investigate pytest parallel runs separately later.

Followed up in #350.

PR #349, opened 2025-07-23
[REVIEW][NFC] Vendor in serialize to allow for future CUDA-specific refactoring and changes

atmnp:
The serialize file contain a bunch of helper functions that we may want to specialize in the future since these are used in many of our functions. This vendors them into this repo. This is an incremental NFC change (so no extra unit tests).


gmarkall:
I think it's worth a look at Numba's `test_serialize.py` to see what tests there are directly testing this - perhaps we should be testing the vendored code a little more directly.

atmnp:
Yep silly oversight, added it now. I saw the similarly named file existing in our repo and thought it'd already been checked against.

gmarkall:
> We shouldn't bring in test cases that use the CPU target - these are all the serialize use cases. Test cases should either be modified to use the CUDA target, or not brought into Numba-CUDA, depending on how difficult it is to modify them / how much functionality they appear to cover.

I had a rethink about this - I think in the long term we should be able to get rid of `serialize`, since it appears to be related to pickling object mode code. Therefore, I think the existing tests are OK for now, and can be modified / removed if we make `serialize`-related changes.

Issue #317, opened 2025-07-03
[BUG] Occasional test suite deadlock in `test_parallel_tasks` or `test_multiple_async_done_multiple_streams`


gmarkall:
Locally the test suite has just deadlocked for me, ending in:

```
test_add_callback (numba.cuda.tests.cudadrv.test_streams.TestCudaStream.test_add_callback) ... ok
test_add_callback_with_default_arg (numba.cuda.tests.cudadrv.test_streams.TestCudaStream.test_add_callback_with_default_arg) ... ok
test_async_done (numba.cuda.tests.cudadrv.test_streams.TestCudaStream.test_async_done) ... ok
test_cancelled_future (numba.cuda.tests.cudadrv.test_streams.TestCudaStream.test_cancelled_future) ... ok
test_multiple_async_done (numba.cuda.tests.cudadrv.test_streams.TestCudaStream.test_multiple_async_done) ... ok
test_multiple_async_done_multiple_streams (numba.cuda.tests.cudadrv.test_streams.TestCudaStream.test_multiple_async_done_multiple_streams) ... ok
test_parallel_tasks (numba.cuda.tests.cudadrv.test_streams.TestCudaStream.test_parallel_tasks) ... ^C^C
```

Hitting ctrl-C to get a traceback has no effect.

A CI run for a previous test also exhibited similar behaviour (will add more details when I can retrieve them).

gmarkall:
`info threads`:

```
(gdb) info threads
  Id   Target Id                                            Frame 
* 1    Thread 0x7efaec9d8740 (LWP 146484) "python"          0x00007ef9ef64f83e in ?? () from /lib/x86_64-linux-gnu/libcuda.so
  2    Thread 0x7ef9ea1ff6c0 (LWP 146529) "cuda00001800007" 0x00007efaec71b4cd in __GI___poll (fds=0x64682dac4b80, nfds=3, timeout=-1)
    at ../sysdeps/unix/sysv/linux/poll.c:29
  3    Thread 0x7ef9dc3ff6c0 (LWP 147272) "python"          0x00007efaec698d71 in __futex_abstimed_wait_common64 (private=32505, 
    cancel=true, abstime=0x7ef9dc3fee40, op=393, expected=0, futex_word=0x64682e124db0) at ./nptl/futex-internal.c:57
[...]

PR #350, opened 2025-07-24
Support running tests in parallel

gmarkall:
- Use pytest-xdist to run tests in parallel.
- Install pytest-xdist in CI where necessary.
- Use `--dist loadscope` to enforce the required grouping of test methods in classes, so that a `ContextResettingTestCase` method can't reset the context that might still be needed by methods in a `CUDATestCase` test.
- Remove use of the `SerialMixin` class and `_numba_parallel_test_` markers. These were only respected by `numba.runtests` anyway. Some of the tests were marked as serial because they used a `multiprocessing.Pool`. This did not work with `numba.runtests` because it also used a `multiprocessing.Pool`, which created daemonic processes and would lead to the error message "daemonic processes are not allowed to have children". pytest worker processes are not daemonic, and therefore we can also run test cases that use a `multiprocessing.Pool` on them.
- Split up the ufunc tests. These are very long running, and left one worker churning through them at the end whilst the other workers sit idle.

I've been running this in a loop on my local machine and have not noticed any issues resulting from it. The test execution time on my laptop is reduced from 170 seconds (serial) to 38 seconds (parallel).

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,16 +30,9 @@ If you want to manage all run-time dependencies yourself, also pass the `--no-de
 ## Running tests
 
 ```
-pytest --pyargs numba.cuda.tests -v
+pytest -n auto --pyargs numba.cuda.tests -v
 ```
 
-Tests can be run in parallel with:
-
-```
-# -n auto: Automatically determine number of threads
-# --dist loadscope: Ensure tests are grouped as required by `CUDATestCase`
-pytest -n auto --dist loadscope --pyargs numba.cuda.tests -v
-```
 
 Testing should discover the `numba.cuda` module from the `numba_cuda` package. You
 can check where `numba.cuda` files are being located by running
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -92,6 +92,8 @@ include = ["numba_cuda*"]
 minversion = "8.0"
 testpaths = ["numba_cuda/numba/cuda/tests"]
 consider_namespace_packages = true
+# loadscope ensures the grouping required by CUDATestCase
+addopts = "--dist loadscope"
 
 [tool.ruff]
 line-length = 80


gmarkall:
It seems that loading up the CI machines with a parallel run makes it much easier to hit the async + CUDA lock + GIL deadlock (#317), so I've switched this to run in serial by default and added an explanation of how to run in parallel to the readme.

kkraus14:
> It seems that loading up the CI machines with a parallel run makes it much easier to hit the async + CUDA lock + GIL deadlock (#317), so I've switched this to run in serial by default and added an explanation of how to run in parallel to the readme.

Doesn't pytest-xdist run across multiple processes? It shouldn't really change the ability to hit that deadlock since it's dependent on behaviors in a single process. I guess in practice with multiple processes using the CUDA driver it could make it more likely to end up in a situation where you're waiting for the CUDA driver lock while holding the GIL?

ZzEeKkAa:
> The test execution time on my laptop is reduced from 170 seconds (serial) to 38 seconds (parallel)

🔥 

> Use --dist loadscope to enforce the required grouping of test methods in classes, so that a ContextResettingTestCase method can't reset the context that might still be needed by methods in a CUDATestCase test.

So does it mean that tests in one class are executed sequentially? Or tests from different classes can't be executed in parallel?

ZzEeKkAa (approved):
LGTM!

Checked locally, also works without issues. Thank you!

gmarkall:
> So does it mean that tests in one class are executed sequentially? Or tests from different classes can't be executed in parallel?

I believe that tests in one class are executed sequentially - they are all assigned to the same worker, which then executes them sequentially. Different workers are assigned different classes, and they execute the different classes in parallel.

gmarkall:
> Doesn't pytest-xdist run across multiple processes? It shouldn't really change the ability to hit that deadlock since it's dependent on behaviors in a single process. I guess in practice with multiple processes using the CUDA driver it could make it more likely to end up in a situation where you're waiting for the CUDA driver lock while holding the GIL?

I think the effect is observed because there are multiple threads involved in the deadlock (a CUDA driver thread and the Python main thread) and the system is more heavily loaded by having multiple processes, so I imagine latencies for everything are a bit stretched out due to all the factors that entails, leading to the deadlock becoming more likely. I wonder if multiple processes all interacting with a single device through the driver increases the latency for the driver to make the callback that's involved in the deadlock too - if there's only one GPU it can only do so much at once. (end of handwaving :slightly_smiling_face:  )

kkraus14:
We may want to consider just disabling that test in the case we're using RMM as the memory manager for the time being given that the reason for the deadlock is well understood at this point.

gmarkall:
The deadlock that happened here doesn't need RMM to be triggered - RMM triggers a much more "reliable" deadlock in another test case. We should probably:

- Have a test run that also involves RMM, since we don't test with it in CI at all at the moment.
- Also disable the async test cases in general to prevent deadlocks in CI, if it becomes a reguler issue.

ZzEeKkAa:
Do we want to add `--dist loadscope` to configuration like that:
```ini
[pytest]
addopts = --dist loadscope
```
https://pytest-xdist.readthedocs.io/en/stable/changelog.html#id42

kkraus14:
> The deadlock that happened here doesn't need RMM to be triggered - RMM triggers a much more "reliable" deadlock in another test case. We should probably:
> 
> * Have a test run that also involves RMM, since we don't test with it in CI at all at the moment.
> * Also disable the async test cases in general to prevent deadlocks in CI, if it becomes a reguler issue.

I think that we can still deadlock because we aren't releasing the GIL in `cuda.bindings` currently. This has been fixed in the top of tree and will be fixed in the next release, where I'm curious if we'll still see a potential deadlock once that happens.

gmarkall:
> > The deadlock that happened here doesn't need RMM to be triggered - RMM triggers a much more "reliable" deadlock in another test case. We should probably:
> > 
> > * Have a test run that also involves RMM, since we don't test with it in CI at all at the moment.
> > * Also disable the async test cases in general to prevent deadlocks in CI, if it becomes a reguler issue.
> 
> I think that we can still deadlock because we aren't releasing the GIL in `cuda.bindings` currently. This has been fixed in the top of tree and will be fixed in the next release, where I'm curious if we'll still see a potential deadlock once that happens.

Yes, we can still deadlock because of that. Even with the CUDA Python fix, RMM can still cause a deadlock with the test suite, because they're two distinct issues - RMM calls `cudaFree` without releasing the GIL independently of the CUDA Python bindings.

I've disabled these tests for now in #356.

gmarkall:
> Do we want to add `--dist loadscope` to configuration like that:
> 
> ```ini
> [pytest]
> addopts = --dist loadscope
> ```
> 
> https://pytest-xdist.readthedocs.io/en/stable/changelog.html#id42

Yes, I think so. Initially I didn't add that because it errored out when pytest-xdist was not installed, but now that I made pytest-xdist part of the requirements, that shouldn't present an issue.

isVoid (approved):
lgtm, the test also passes for me locally.

Issue #351, opened 2025-07-24
[BUG] Copying bool arrays with ndim>1 to device fails


ronpandolfi:
**Describe the bug**
When copying bool array with ndim>1 a ValueError is produced.

**Steps/Code to reproduce bug**
```python
import numpy as np
from numba import cuda

def test_bool_to_dev(shape):
    bool_data = np.random.random(shape) < 0.5
    bool_data_dev = cuda.device_array(bool_data, dtype=np.bool)

test_bool_to_dev((10,)) # this is fine
test_bool_to_dev((10, 10)) # this produces an error
```

Error result:

```python
Traceback (most recent call last):
  File "C:\Users\rp\AppData\Roaming\JetBrains\PyCharm2025.1\scratches\scratch_1.py", line 9, in <module>
    test_bool_to_dev((10, 10)) # this produces an error
    ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\rp\AppData\Roaming\JetBrains\PyCharm2025.1\scratches\scratch_1.py", line 6, in test_bool_to_dev
    bool_data_dev = cuda.device_array(bool_data, dtype=np.bool)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\rp\PycharmProjects\g2viewer\.venv\Lib\site-packages\numba\cuda\cudadrv\devices.py", line 232, in _require_cuda_context
    return fn(*args, **kws)
           ^^^^^^^^^^^^^^^^
  File "C:\Users\rp\PycharmProjects\g2viewer\.venv\Lib\site-packages\numba\cuda\api.py", line 144, in device_array
    return devicearray.DeviceNDArray(shape=shape, strides=strides, dtype=dtype,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\rp\PycharmProjects\g2viewer\.venv\Lib\site-packages\numba\cuda\cudadrv\devicearray.py", line 92, in __init__
    self._dummy = dummyarray.Array.from_desc(0, shape, strides,
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #352, opened 2025-07-24
FIX: Add type check for shape elements in DeviceNDArrayBase constructor

gnzng:
This PR addresses issue #351 by adding a type check to ensure all elements of the `shape` argument are integers during `DeviceNDArrayBase` construction.

### Description

`numba_cuda.device_array` would accept a `shape` argument containing non-integer types, such as booleans or floats. The values were implicitly cast to floats, which could lead to unexpected behavior or downstream errors.

**Behavior Before This PR**

The following code would execute without raising an error, creating an array with a shape composed of floats:

```python
import numpy as np
from numba_cuda import device_array

# No error was raised
arr = device_array(shape=np.array([True, False, 10, 10.12]))

# Resulting shape would be (1.0, 0.0, 10.0, 10.12)
print(arr.shape)
```

**Behavior After This PR**

With this change, providing non-integer elements in the `shape` argument now raises an immediate `TypeError`, ensuring that only valid shapes are used for array creation.

```python
import numpy as np
from numba_cuda import device_array

# Now raises a TypeError
arr = device_array(shape=np.array([True, False, 10, 10.12]))
```

The resulting error is:
```
TypeError: all elements of shape must be ints
```

This change improves input validation and prevents the creation of improperly configured device arrays.

### DIFF
--- a/numba_cuda/numba/cuda/api_util.py
+++ b/numba_cuda/numba/cuda/api_util.py
@@ -4,16 +4,11 @@ import numpy as np
 def prepare_shape_strides_dtype(shape, strides, dtype, order):
     dtype = np.dtype(dtype)
     if isinstance(shape, (float, np.floating)):
-        raise TypeError(
-            "shape must be an integer or tuple of integers"
-        )
-    if (
-        isinstance(shape, np.ndarray)
-        and np.issubdtype(shape.dtype, np.floating)
+        raise TypeError("shape must be an integer or tuple of integers")
+    if isinstance(shape, np.ndarray) and np.issubdtype(
+        shape.dtype, np.floating
     ):
-        raise TypeError(
-            "shape must be an integer or tuple of integers"
-        )
+        raise TypeError("shape must be an integer or tuple of integers")
     if isinstance(shape, int):
         shape = (shape,)
     if isinstance(strides, int):


gmarkall:
Thanks for the PR!

When a scalar float is passed for the shape, we get an error before the new check sees it (and I think the new check would also fail as a `float` is not iterable):

```
In [3]: cuda.device_array(1.2)

File ~/numbadev/numba-cuda/numba_cuda/numba/cuda/api.py:158, in device_array(shape, dtype, strides, order, stream)
    152 @require_context
    153 def device_array(shape, dtype=np.float64, strides=None, order="C", stream=0):
    154     """device_array(shape, dtype=np.float64, strides=None, order='C', stream=0)
    155 
    156     Allocate an empty device ndarray. Similar to :meth:`numpy.empty`.
    157     """
--> 158     shape, strides, dtype = prepare_shape_strides_dtype(
    159         shape, strides, dtype, order
    160     )
    161     return devicearray.DeviceNDArray(
    162         shape=shape, strides=strides, dtype=dtype, stream=stream
    163     )

File ~/numbadev/numba-cuda/numba_cuda/numba/cuda/api_util.py:11, in prepare_shape_strides_dtype(shape, strides, dtype, order)
      9     strides = (strides,)
     10 else:
---> 11     strides = strides or _fill_stride_by_order(shape, dtype, order)
     12 return shape, strides, dtype

File ~/numbadev/numba-cuda/numba_cuda/numba/cuda/api_util.py:16, in _fill_stride_by_order(shape, dtype, order)
     15 def _fill_stride_by_order(shape, dtype, order):
---> 16     nd = len(shape)
     17     if nd == 0:
     18         return ()


TypeError: object of type 'float' has no len()
```

What do you think about addressing this variant of the error too?

Would you also like to add a test of the fix(es)?

gnzng:
Thanks a lot!

I added the float check further upstream in `prepare_shape_strides_dtype()` ae2305b

Added tests for testing various inputs and input dtypes:
- 36ed9ef (test for right errors) 
- 6274af6 (make sure list and np.arrays are still working with valid ints)

gnzng:
Fix style via ruff format `api_util.py` since style CI `ci/check_style.sh` was failing a06a697

gmarkall (approved):
Many thanks!

PR #353, opened 2025-07-25
[WIP] Add Windows CI testing

gmarkall:
Under development for the time being.

### DIFF
--- a/ci/test_wheel.ps1
+++ b/ci/test_wheel.ps1
@@ -1,0 +1,57 @@
+# Copyright (c) 2023-2024, NVIDIA CORPORATION
+
+$ErrorActionPreference = 'Stop'
+Set-StrictMode -Version Latest
+
+function rapids-logger {
+    param (
+        [Parameter(Mandatory=$true)]
+        [string]$Text
+    )
+
+    # Determine padding and box width
+    $padding = 2
+    $boxWidth = $Text.Length + ($padding * 2)
+    $topBottom = '+' + ('-' * $boxWidth) + '+'
+    $middle = '|' + (' ' * $padding) + $Text + (' ' * $padding) + '|'
+
+    # Print the box in green
+    Write-Host $topBottom -ForegroundColor Green
+    Write-Host $middle    -ForegroundColor Green
+    Write-Host $topBottom -ForegroundColor Green
+}
+
+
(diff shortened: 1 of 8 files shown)


ci/test_wheel.ps1
```diff
@@ -0,0 +1,54 @@
...
+}
+
+
+$CUDA_VER_MAJOR_MINOR = ($env:CUDA_VER -split '\.')[0..1] -join '.'
+$CUDA_VER_MAJOR = ($env:CUDA_VER -split '\.')[0] -join '.'
+
+rapids-logger "Install wheel with test dependencies"
+$package = Resolve-Path wheel\numba_cuda*.whl | Select-Object -ExpandProperty Path
+echo "Package path: $package"
+python -m pip install "$package[test,cu12]" "cuda-python==$CUDA_VER_MAJOR.*" "cuda-core==0.3.*"
```
leofang: I think on Windows we don't use any (RAPIDS) container that has lots of CUDA things pre-installed, so we should follow `test_wheel_deps_wheels.sh` instead of `test_wheel.sh`?
https://github.com/NVIDIA/numba-cuda/blob/40cdc764f6b09ae56b12efccb0ce015d2a0b7d2e/ci/test_wheel_deps_wheels.sh#L17

gmarkall: Yeah, I started out with the most minimal thing because there seems to be a lot of cruft in our workflow and I didn't want to just propagate it without knowing which bits were actually needed.

I was keeping things a bit more literal for the time being because I find it easy to make a mistake with variable substitutions and the debug cycle is painfully long.

leofang: Yeah the turnaround time is crazy especially with driver installation...

Another nice trick to help you keep the changes minimized is if you install Git for Windows
https://github.com/NVIDIA/cuda-python/blob/dcbf5b2a2f03401180582c94c7c27f61e2d5979e/.github/workflows/test-wheel-windows.yml#L99-L111
then most of existing bash scripts can be reused with very little changes, ex:
https://github.com/NVIDIA/cuda-python/blob/dcbf5b2a2f03401180582c94c7c27f61e2d5979e/.github/workflows/test-wheel-windows.yml#L273-L278

gmarkall:
Closing as superseded by #396.

PR #354, opened 2025-07-26
[NFC] FileCheck tests check all overloads

ashermancinelli:
* Clean up type hints
* On failure, dump the failing IR to a file and provide a reproducer command
* When multiple overloaded IR is available for a kernel and no signature has been passed, check all of the overloads

TODO:
- [ ] Need to decide: is an environment variable how we want to toggle the dumping? I don't think it should happen by default in case multiple filecheck tests run and fail in parallel. When updating tests, this is extremely helpful in my opinion.

### DIFF
--- a/numba_cuda/numba/cuda/testing.py
+++ b/numba_cuda/numba/cuda/testing.py
@@ -1,7 +1,6 @@
 import os
 import platform
 import shutil
-
 from numba.core.utils import PYVERSION
 from numba.tests.support import SerialMixin
 from numba.cuda.cuda_paths import get_conda_ctk
@@ -10,12 +9,14 @@ from numba.cuda.dispatcher import CUDADispatcher
 from numba.core import config
 from numba.tests.support import TestCase
 from pathlib import Path
-from typing import Union
+
+from typing import Iterable, Union
 from io import StringIO
 import unittest
 
 if PYVERSION >= (3, 10):
-    from filecheck.matcher import Matcher, Options
+    from filecheck.matcher import Matcher
+    from filecheck.options import Options
     from filecheck.parser import Parser, pattern_for_opts
     from filecheck.finput import FInput
 
@@ -23,7 +24,7 @@ numba_cuda_dir = Path(__file__).parent
 test_data_dir = numba_cuda_dir / "tests" / "data"
 
 
-class FileCheckTestCaseMixin:
+class FileCheckTestCaseMixin(unittest.TestCase):
     """
     Mixin for tests that use FileCheck.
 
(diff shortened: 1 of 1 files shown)


ashermancinelli:
/ok to test [66f2f8c](https://github.com/NVIDIA/numba-cuda/pull/354/commits/66f2f8cd7f681bd0263260ade6fdc616122db748)

numba_cuda/numba/cuda/testing.py
```diff
@@ -10,20 +9,22 @@
...
+    from filecheck.options import Options
     from filecheck.parser import Parser, pattern_for_opts
     from filecheck.finput import FInput
 
 numba_cuda_dir = Path(__file__).parent
 test_data_dir = numba_cuda_dir / "tests" / "data"
 
 
-class FileCheckTestCaseMixin:
+class FileCheckTestCaseMixin(unittest.TestCase):
```
gmarkall: `CUDATestCase` inheriting `TestCase` twice via `FileCheckTestCaseMixin` does not seem good. It's not clear to me that we benefit from a separate mixin here anyway - can we move the methods of `FileCheckTestCaseMixin` into the `CUDATestCase`, and get rid of the mixin?

numba_cuda/numba/cuda/testing.py
```diff
@@ -35,56 +36,81 @@ class FileCheckTestCaseMixin:
...
-        if signature:
-            ir_content = ir_content[signature]
-        check_patterns = ir_producer.__doc__
-        self.assertFileCheckMatches(
-            ir_content,
-            check_patterns=check_patterns,
-            check_prefixes=check_prefixes,
-            **extra_filecheck_options,
+        ir_contents = self._getIRContent(ir_producer.inspect_asm(), signature)
+        assert ir_contents
```
gmarkall: Hitting a bare assert is usually not the nicest debug experience. What is this guarding against?

numba_cuda/numba/cuda/testing.py
```diff
@@ -35,56 +36,81 @@ class FileCheckTestCaseMixin:
...
-        ir_content = ir_producer.inspect_asm()
-        if signature:
-            ir_content = ir_content[signature]
-        check_patterns = ir_producer.__doc__
-        self.assertFileCheckMatches(
-            ir_content,
-            check_patterns=check_patterns,
-            check_prefixes=check_prefixes,
-            **extra_filecheck_options,
+        ir_contents = self._getIRContent(ir_producer.inspect_asm(), signature)
```
gmarkall: If `_getIRContent` is returning `ir_contents`, should it be called `_getIRContents`?

gmarkall:
> Need to decide: is an environment variable how we want to toggle the dumping? I don't think it should happen by default in case multiple filecheck tests run and fail in parallel.

Now we've switched over to pytest, it might be nice to add an option with https://docs.pytest.org/en/stable/reference/reference.html#pytest.hookspec.pytest_addoption so that user can run e.g.:

```
pytest --dump-failed-filechecks --pyargs numba.cuda.tests
```

I don't think the name of the file should be hardcoded - perhaps file names could be derived from the name of the test case (and potentially suffixed with a numeric index if there is a name clash due to things like subtests?

numba_cuda/numba/cuda/testing.py
ashermancinelli: That the IR given to us by the user was not empty - Are assertions in the assert methods of a test case still used to indicate a violated precondition, and `self.fail/skip` is used to indicate the result of the assertion represented by the method? Assuming so, I just added messages to these assertions. Please let me know if you had something else in mind.

ashermancinelli:
> > Need to decide: is an environment variable how we want to toggle the dumping? I don't think it should happen by default in case multiple filecheck tests run and fail in parallel.
> 
> Now we've switched over to pytest, it might be nice to add an option with https://docs.pytest.org/en/stable/reference/reference.html#pytest.hookspec.pytest_addoption so that user can run e.g.:
> 
> ```
> pytest --dump-failed-filechecks --pyargs numba.cuda.tests
> ```
> 
> I don't think the name of the file should be hardcoded - perhaps file names could be derived from the name of the test case (and potentially suffixed with a numeric index if there is a name clash due to things like subtests?

I love this suggestion. I added it in the latest revision, so now the failing IR and checks are dumped to files like this: `numba_cuda_tests_cudapy_test_inspect_TestInspect_test_monotyped.ll`. Is that what you had in mind? Do you think we should try to rename it with a numeric suffix if it already exists, or is the name so specific that we don't mind overwriting a dump from a previous failing run?

ashermancinelli:
/ok to test [67642ed](https://github.com/NVIDIA/numba-cuda/commit/67642ed242f6bc02ed49376c938e92e7eef661c2)

ashermancinelli:
/ok to test [67642ed](https://github.com/NVIDIA/numba-cuda/commit/67642ed242f6bc02ed49376c938e92e7eef661c2)

ashermancinelli:
/ok to test [49a007a](https://github.com/NVIDIA/numba-cuda/pull/354/commits/49a007abfa08e2f8e4077b01b1bcbe9821c08f50)

isVoid:
> numba_cuda_tests_cudapy_test_inspect_TestInspect_test_monotyped.ll. Is that what you had in mind? Do you think we should try to rename it with a numeric suffix if it already exists, or is the name so specific that we don't mind overwriting a dump from a previous failing run?

One idea of separating them from one test run from another is placing the dump files in a folder named `numba-cuda-tests_YYYY:MM:DD:HH:MM:SS` and place the dump there. But second level internal may not be sufficient to isolate repeated runs.

I think isolation is necessary when `pytest --count` is used to test bugs which is not deterministic.

ashermancinelli:
> > numba_cuda_tests_cudapy_test_inspect_TestInspect_test_monotyped.ll. Is that what you had in mind? Do you think we should try to rename it with a numeric suffix if it already exists, or is the name so specific that we don't mind overwriting a dump from a previous failing run?
> 
> One idea of separating them from one test run from another is placing the dump files in a folder named `numba-cuda-tests_YYYY:MM:DD:HH:MM:SS` and place the dump there. But second level internal may not be sufficient to isolate repeated runs.
> 
> I think isolation is necessary when `pytest --count` is used to test bugs which is not deterministic.

I like that idea! Feel free to let me know what's most intuitive and I'll do it. If I had to guess, I think the workflow will look like this:

1. Make a commit that changes the IR we produce somehow
2. Run the tests and dump the failing IR
3. Make sure the differences are what we expected
4. Regenerate and update the checks in one test
5. Rerun, and if any tests are still failing, goto step 2

I don't envision myself needing to keep multiple directories with IR dumps from failing tests. I think a directory with a timestamp in the name would be a great way to accomplish that though.

isVoid:
@ashermancinelli The workflow sounds right. I think my original language is misleading as "second level interval" is not referring to multiple directories. I meant using second as the smallest time interval may not be sufficient as `pytest --count` can rerun tests in sub-second frequency. What do you think of adding millisecond interval in the timestamp?

ashermancinelli:
Okay, the IR is now dumped in this format on error:

```

> filecheck --check-prefixes=LLVM numba-ir-2025_07_28_15_49_27/numba_cuda_tests_cudapy_test_inspect_TestInspect_test_monotyped.checks --input-file=numba-ir-2025_07_28_15_49_27/numba_cuda_tests_cudapy_test_inspect_TestInspect_test_monotyped.ll
numba-ir-2025_07_28_15_49_27/numba_cuda_tests_cudapy_test_inspect_TestInspect_test_monotyped.checks:9: error: Couldn't match "bar".
Current position at numba-ir-2025_07_28_15_49_27/numba_cuda_tests_cudapy_test_inspect_TestInspect_test_monotyped.ll:9:1
}
^
```

It is very verbose, but we will be unlikely to see collisions. Thanks for the review @isVoid!

ashermancinelli:
/ok to test [10e93cc](https://github.com/NVIDIA/numba-cuda/pull/354/commits/10e93ccc1aaf4933d021d87177007e95f8fa63f9)

ashermancinelli:
/ok to test [d41fb5e](https://github.com/NVIDIA/numba-cuda/pull/354/commits/d41fb5e64bde21307965240f54c2fd7ecb0b2c58)

ashermancinelli:
/ok to test [b0bed24](https://github.com/NVIDIA/numba-cuda/pull/354/commits/b0bed24e94aa7ac96aa337c2cc82ce7a4b7bf512)

numba_cuda/numba/cuda/testing.py
```diff
@@ -107,39 +154,35 @@ def assertFileCheckMatches(
...
+                ir_dump = dump_directory / Path(base_path).with_suffix(".ll")
+                checks_dump = dump_directory / Path(base_path).with_suffix(
+                    ".checks"
+                )
+                with (
+                    open(ir_dump, "w") as ir_file,
+                    open(checks_dump, "w") as checks_file,
+                ):
+                    _ = ir_file.write(ir_content + "\n")
+                    _ = checks_file.write(check_patterns)
```
gmarkall, line 174: I'm a little confused here: if you don't want the return value, why assign it?

gmarkall (approved):

gmarkall:
I forgot to mention - I think the dumps for subtests will still overwrite each other if they don't form part of the test's `id()` (I'm not sure if they do or not) but I don't think this is a massive problem - one case is sufficient to identify an issue, and after fixing it if a different subtest still has a problem that will show up on a re-run.

PR #355, opened 2025-07-26
Vendor in Bytecode Interpreter

ashermancinelli:
I only vendored-in the interpreter with Python > 3.9 because:
* The main branches of Numba core and llvmlite do not support Python <= 3.9, and there are lots of checks for Python versions in this source file.
* I would like the first check-in of the interpreter to be unchanged, so that the original version exists in our git history in case we need to track down bugs introduced in our interpreter with CUDA-specific changes.

This is why the original Numba interpreter is still imported in 3.9.
If we like, I will go through the version-specific parts of the interpreter as a follow-up, and try to edit the interpreter such that we can still use it with older pythons.
I would like to do that in a second step to be sure I do not introduce any bugs in the vendoring-in process.

Code coverage reported 46% from this source file after running our test suite. There are a few reasons.
* Much of the interpreter covers features that I don't believe we support, namely dictionaries and exceptions (please correct me if I'm wrong)
* Some of it is gated behind Python versions (eg a different method for 3.10 vs 3.11)
* I do not see any tests in Numba checking the behavior of specific opcodes - they appear to be integrated tests where the kernel's behavior and IR are inspected, but the opcodes used by the interpreter are not.

The interpreter was pulled in from numba commit 262b49a3e

### DIFF
--- a/numba_cuda/numba/cuda/core/interpreter.py
+++ b/numba_cuda/numba/cuda/core/interpreter.py
@@ -1,0 +1,3592 @@
+import builtins
+import collections
+import dis
+import operator
+import logging
+import textwrap
+
+from numba.core import errors, ir, config
+from numba.core.errors import (
+    NotDefinedError,
+    UnsupportedBytecodeError,
+    error_extras,
+)
+from numba.core.ir_utils import get_definition, guard
+from numba.core.utils import (
+    PYVERSION,
+    BINOPS_TO_OPERATORS,
+    INPLACE_BINOPS_TO_OPERATORS,
+    _lazy_pformat,
+)
+from numba.core.byteflow import Flow, AdaptDFA, AdaptCFA, BlockKind
+from numba.core.unsafe import eh
+from numba.cpython.unsafe.tuple import unpack_single_tuple
+
+
+if PYVERSION in ((3, 12), (3, 13)):
+    # Operands for CALL_INTRINSIC_1
+    from numba.core.byteflow import CALL_INTRINSIC_1_Operand as ci1op
+elif PYVERSION in ((3, 10), (3, 11)):
+    pass
+else:
+    raise NotImplementedError(PYVERSION)
+
+
+class _UNKNOWN_VALUE(object):
+    """Represents an unknown value, this is for ease of debugging purposes only."""
+
+    def __init__(self, varname):
+        self._varname = varname
+
+    def __repr__(self):
+        return "_UNKNOWN_VALUE({})".format(self._varname)
+
+
+_logger = logging.getLogger(__name__)
+
+
+class Assigner(object):
+    """
+    This object keeps track of potential assignment simplifications
+    inside a code block.
+    For example `$O.1 = x` followed by `y = $0.1` can be simplified
+    into `y = x`, but it's not possible anymore if we have `x = z`
+    in-between those two instructions.
+
+    NOTE: this is not only an optimization, but is actually necessary
+    due to certain limitations of Numba - such as only accepting the
+    returning of an array passed as function argument.
+    """
+
+    def __init__(self):
+        # { destination variable name -> source Var object }
+        self.dest_to_src = {}
+        # Basically a reverse mapping of dest_to_src:
+        # { source variable name -> all destination names in dest_to_src }
+        self.src_invalidate = collections.defaultdict(list)
+        self.unused_dests = set()
+
+    def assign(self, srcvar, destvar):
+        """
+        Assign *srcvar* to *destvar*. Return either *srcvar* or a possible
+        simplified assignment source (earlier assigned to *srcvar*).
+        """
+        srcname = srcvar.name
+        destname = destvar.name
+        if destname in self.src_invalidate:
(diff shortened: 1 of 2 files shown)


ashermancinelli:
/ok to test [01c0f70](https://github.com/NVIDIA/numba-cuda/pull/355/commits/01c0f70e5de01966b6aa85973b6c79b89351122a)

ashermancinelli:
/ok to test [01c0f70](https://github.com/NVIDIA/numba-cuda/pull/355/commits/01c0f70e5de01966b6aa85973b6c79b89351122a)

numba_cuda/numba/cuda/core/interpreter.py
```diff
@@ -0,0 +1,3588 @@
+import builtins
+import collections
+import dis
+import operator
+import logging
+import textwrap
+
+from numba.core import errors, ir, config
```
isVoid, line 8: This file has a lot of isinstance check against upstream Numba IR. I think @gmarkall mentioned that when we vend in `Numba.IR`, we want to modify this import to the vended IR module. Just putting this out there to validate.

ashermancinelli: That is my understanding!

numba_cuda/numba/cuda/core/interpreter.py
```diff
@@ -0,0 +1,3588 @@
...
+
+    def _op_JUMP_IF(self, inst, pred, iftrue):
+        brs = {
+            True: inst.get_jump_target(),
+            False: inst.next,
+        }
+        truebr = brs[iftrue]
+        falsebr = brs[not iftrue]
+
+        name = "bool%s" % (inst.offset)
```
gmarkall: Are you sure you've vendored the interpreter from a recent commit? In the commit you linked to, it seems that this line is 

```suggestion
        name = "$bool%s" % (inst.offset)
```

which is a fix for a debuginfo issue that we upstreamed. (observed from https://github.com/numba/numba/blob/1358274054f88c19838ba11543b8f87b05cf0632/numba/core/interpreter.py#L3059)

Maybe I'm getting something backwards / wrong here - if so, please point me in the correct direction! :slightly_smiling_face:

ashermancinelli: Hm, I checked the commit hash and copied it over directly🤔 I will pull the latest version and copy it in again. How does that sound?

ashermancinelli: That was about the only difference between my last commit and the most recent interpreter.py:
```diff
diff --git a/numba_cuda/numba/cuda/core/interpreter.py b/numba_cuda/numba/cuda/core/interpreter.py
index 6c6b5c5..abf8216 100644
--- a/numba_cuda/numba/cuda/core/interpreter.py
+++ b/numba_cuda/numba/cuda/core/interpreter.py
@@ -16,11 +16,13 @@ from numba.core.utils import (
     PYVERSION,
     BINOPS_TO_OPERATORS,
     INPLACE_BINOPS_TO_OPERATORS,
+    _lazy_pformat,
 )
 from numba.core.byteflow import Flow, AdaptDFA, AdaptCFA, BlockKind
 from numba.core.unsafe import eh
 from numba.cpython.unsafe.tuple import unpack_single_tuple
 
+
 if PYVERSION in ((3, 12), (3, 13)):
     # Operands for CALL_INTRINSIC_1
     from numba.core.byteflow import CALL_INTRINSIC_1_Operand as ci1op
@@ -1420,7 +1422,9 @@ class Interpreter(object):
             self.arg_count,
             self.arg_names,
         )
-        _logger.debug(func_ir.dump_to_string())
+        _logger.debug(
+            _lazy_pformat(func_ir, lazy_func=lambda x: x.dump_to_string())
+        )
 
         # post process the IR to rewrite opcodes/byte sequences that are too
         # involved to risk handling as part of direct interpretation
@@ -3183,7 +3187,7 @@ class Interpreter(object):
         truebr = brs[iftrue]
         falsebr = brs[not iftrue]
 
-        name = "bool%s" % (inst.offset)
+        name = "$bool%s" % (inst.offset)
         gv_fn = ir.Global("bool", bool, loc=self.loc)
         self.store(value=gv_fn, name=name)
```

ashermancinelli:
/ok to test [c367c0f](https://github.com/NVIDIA/numba-cuda/commit/c367c0fae73e78161c956ba3eeb1c463faee6c9c)

ashermancinelli:
/ok to test [c367c0f](https://github.com/NVIDIA/numba-cuda/commit/c367c0fae73e78161c956ba3eeb1c463faee6c9c)

ashermancinelli:
/ok to test [2a89771](https://github.com/NVIDIA/numba-cuda/commit/2a8977119e7c7f9319f062b4613922f9981f0559)

numba_cuda/numba/cuda/core/interpreter.py
gmarkall: These are the same differences that I saw - thanks for confirming!

gmarkall (approved):

PR #357, opened 2025-07-28
Remove remaining references/dependencies on `pynvjitlink`

brandon-b-miller:
- Removes any straggling references to `pynvjitlink` from the codebase and tests
- Clean up a place where we still depend on it for our wheel
- Fix logic that shouldn't warn if it happens to be present


gmarkall (approved):
This looks good to me. @ashermancinelli Does this solve the issue you were seeing with the warnings being generated?

ashermancinelli:
It does, thank you!

ashermancinelli (approved):

PR #358, opened 2025-07-28
Add more debuginfo tests

ashermancinelli:
Additional debug info tests were adapted from Numba core debuginfo tests, but using filecheck. For example, the original tests searched the IR with regular expressions and built up a dictionary of debuginfo values and the types they correspond to - now we use filecheck capture variables the first time they are encountered, and we refer back to them later in the checks.

### DIFF
--- a/numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
+++ b/numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
@@ -403,6 +403,387 @@ class TestCudaDebugInfo(CUDATestCase):
         match = re.compile(pat6).search(llvm_ir)
         self.assertIsNotNone(match, msg=llvm_ir)
 
+    def test_DW_LANG(self):
+        @cuda.jit(debug=True)
+        def foo():
+            """
+            CHECK: distinct !DICompileUnit
+            CHECK-SAME: emissionKind: FullDebug
+            CHECK-SAME: isOptimized: true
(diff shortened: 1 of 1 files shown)


ashermancinelli:
/ok to test [276ad34](https://github.com/NVIDIA/numba-cuda/pull/358/commits/276ad34456fd486dd1f197a2eb0bb93bafbef16d)

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
```diff
@@ -403,6 +403,387 @@ def f(x, y):
         match = re.compile(pat6).search(llvm_ir)
         self.assertIsNotNone(match, msg=llvm_ir)
 
+    def test_DW_LANG(self):
+        @cuda.jit(debug=True)
```
gmarkall, line 407: No need to change things here (unless you want to, possibly in a follow-up), but if you specify a signature then the kernel will be compiled when it is defined for the given signature, which saves launching it as part of the test. For this kernel it would be

```suggestion
        @cuda.jit("void()", debug=True)
```

Then `foo[1, 1]()` below can be omitted and the LLVM IR will still be present for FileCheck to check.

gmarkall: Ah, I see now that this pattern is used in `test_DITypes` below.

gmarkall (approved):

PR #359, opened 2025-07-29
Vendor in usecases used in testing

atmnp:
Small change to vendor in some enums/functions used in testing. This cuts down our number of unique used imports by quite a bit.

### DIFF
--- a/numba_cuda/numba/cuda/tests/complex_usecases.py
+++ b/numba_cuda/numba/cuda/tests/complex_usecases.py
@@ -1,0 +1,113 @@
+import cmath
+
+
+def div_usecase(x, y):
+    return x / y
+
+
+def real_usecase(x):
+    return x.real
+
+
+def imag_usecase(x):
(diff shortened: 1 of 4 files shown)


numba_cuda/numba/cuda/tests/complex_usecases.py
```diff
@@ -0,0 +1,119 @@
...
+def polar_usecase(x):
+    return cmath.polar(x)
+
+
+_two = 2.0
+
+
+def polar_as_complex_usecase(x):
+    # HACK: clear errno by invoking float.__pow__
+    # (workaround for http://bugs.python.org/issue24489)
```
brandon-b-miller: This looks like it was fixed a while back, is it still relevant to the versions of python we support?

atmnp: Nope, fixed it in the rebase. It was for older python versions than we intend to support.

PR #360, opened 2025-07-29
[Refactor] Vendor in Typing Templates

atmnp:
The Signature class is also modified to inherit from the upstream `numba.core.typing.templates.Signature` to pass subclass checks in the upstream `AbstractTemplate` that is still used by parts of upstream that haven't been vendored in yet. Likewise, there are calls to the vendored-in `apply` for `AbstractTemplate` that pass in the upstream `Signature` so the subclass check in the vendored-in `AbstractTemplate` is extended to check for either.

These are hacks marked as such, so we can undo them as soon as we can. Afaict, we can't use subclass checks/hooks since this would require changing the upstream Signature class.

### DIFF
--- a/numba_cuda/numba/cuda/_internal/cuda_bf16.py
+++ b/numba_cuda/numba/cuda/_internal/cuda_bf16.py
@@ -23,8 +23,8 @@ from numba.core.extending import (
     make_attribute_wrapper,
     register_model,
 )
-from numba.core.typing import signature
-from numba.core.typing.templates import AttributeTemplate, ConcreteTemplate
+from numba.cuda.typing import signature
+from numba.cuda.typing.templates import AttributeTemplate, ConcreteTemplate
 from numba.cuda import CUSource, declare_device
 from numba.cuda.cudadecl import register, register_attr, register_global
 from numba.cuda.cudaimpl import lower
--- a/numba_cuda/numba/cuda/typing/templates.py
+++ b/numba_cuda/numba/cuda/typing/templates.py
@@ -1,0 +1,1427 @@
+"""
+Define typing templates
+"""
+
+from abc import ABC, abstractmethod
+import functools
+import sys
+import inspect
+import os.path
+from collections import namedtuple
+from collections.abc import Sequence
+from types import MethodType, FunctionType, MappingProxyType
+
+import numba
+from numba.core import types, utils, targetconfig
+from numba.core.errors import (
+    TypingError,
+    InternalError,
+)
+from numba.core.cpu_options import InlineOptions
+from numba.core.typing.templates import Signature as CoreSignature
+
+# info store for inliner callback functions e.g. cost model
+_inline_info = namedtuple("inline_info", "func_ir typemap calltypes signature")
+
+
+# HACK: Remove this inheritance once all references to CoreSignature are removed
+class Signature(CoreSignature):
+    """
+    The signature of a function call or operation, i.e. its argument types
+    and return type.
+    """
+
+    # XXX Perhaps the signature should be a BoundArguments, instead
+    # of separate args and pysig...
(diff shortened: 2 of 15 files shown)


numba_cuda/numba/cuda/_internal/cuda_bf16.py
```diff
@@ -23,8 +23,8 @@
     make_attribute_wrapper,
     register_model,
 )
-from numba.core.typing import signature
-from numba.core.typing.templates import AttributeTemplate, ConcreteTemplate
+from numba.cuda.typing import signature
```
gmarkall, line 26: This is an autogenerated file, so we need to update Numbast to reflect these changes. cc @isVoid

gmarkall: (Not an action for this PR, but noting the requirement)

gmarkall: https://github.com/NVIDIA/numbast/issues/163 created to record this.

numba_cuda/numba/cuda/typing/templates.py
```diff
@@ -0,0 +1,1427 @@
...
+        #
+        # NOTE: This will need subsequently fixing again when targets use solely
+        # the extension APIs to describe their implementation. The issue will be
+        # that the builtin_registry should contain _just_ the stack allocated
+        # implementations and low level target invariant things and should not
+        # be modified further. It should be acceptable to remove the `then`
+        # branch and just keep the `else`.
+
+        # In case the target has swapped, e.g. cuda borrowing cpu, refresh to
+        # populate.
```
gmarkall: This comment needs some thinking about before this PR seems safe to merge, because we're fiddling with the structure of something already quite fragile.

numba_cuda/numba/cuda/typing/templates.py
```diff
@@ -0,0 +1,1427 @@
...
+    An incremental loader for a typing registry.
+    """
+
+    registry_items = ("functions", "attributes", "globals")
+
+
+builtin_registry = Registry()
+infer = builtin_registry.register
+infer_getattr = builtin_registry.register_attr
+infer_global = builtin_registry.register_global
```
gmarkall: My initial impression was that these should be deleted, because we don't want to instantiate another registry in a new place. However, I see that the code I noticed at https://github.com/NVIDIA/numba-cuda/pull/360/files#r2245580196 refers to it, so there's some risk of confusion between registries with this PR as-is. The test suite seems to be OK, but I foresee it breaking users of the extension API.

gmarkall (changes requested):
The PR seems good from the standpoint of "does what it says it will do" but I'm a bit worried about the fragility of some of this implementation, as noted on the diff. I think I need to think about how we can remove the instantiation of new registries before it can proceed.

(Marking as changes required for this reason)

numba_cuda/numba/cuda/typing/templates.py
gmarkall: Relevant commit / PR: https://github.com/numba/numba/pull/6898/commits/900532d64222d37fe8ac0beaf2969b10758d159a

numba_cuda/numba/cuda/typing/templates.py
gmarkall: I removed these and added appropriate comments.

numba_cuda/numba/cuda/typing/templates.py
gmarkall: I removed the reference to `builtin_registry` from below here. This works for now because this code isn't hooked up yet, so I added some extra comments indicating what was done, to help figure out what's going on if we get any strange behaviour if this gets hooked up later.

gmarkall (approved):

PR #364, opened 2025-07-29
Vendor in additional debug info tests

ashermancinelli:
Adapted more tests in numba core to work with numba-cuda's `cuda.jit`.


ashermancinelli:
/ok to test [0d9f022](https://github.com/NVIDIA/numba-cuda/pull/364/commits/0d9f022201f6c76c04db409e1d25272942ebc422)

ashermancinelli:
/ok to test [bb3b518](https://github.com/NVIDIA/numba-cuda/pull/364/commits/bb3b518c31ee138b63df799d869b0885d212d6ac)

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
```diff
@@ -784,6 +790,236 @@ def foo(a):
...
+        msg = str(found.message)
+        # make sure the warning contains the right message
+        self.assertIn("Could not find source for function", msg)
+        # and refers to the offending function
+        self.assertIn(str(foo.py_func), msg)
+
+    def test_irregularly_indented_source(self):
+        @cuda.jit(tuple(), debug=True)
+        def foo():
+            # NOTE: THIS COMMENT MUST START AT COLUMN 0 FOR THIS SAMPLE CODE TO BE VALID # noqa: E115, E501
```
gmarkall: This doesn't start at column 0 - I suspect the filecheck tests below might be checking for the wrong thing?

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
```diff
@@ -784,6 +790,236 @@ def foo(a):
...
+            foo.inspect_llvm()[tuple()],
+            """
+            CHECK: define void @{{.+}}foo
+            CHECK: !DILocation(
+            CHECK-SAME: column: 1
+            CHECK-NOT: DILocation
+            """,
+        )
+
+    def test_no_if_op_bools_declared(self):
```
gmarkall, line 535: This overlaps with `test_issue_9888` (`test_no_if_op_bools_declared` was the upstreamed version of that test), but I think having both is fine, particularly as this variant uses filecheck.

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
```diff
@@ -784,6 +790,236 @@ def foo(a):
...
+
+        @cuda.jit(opt=True)
+        def foo(x, y):
+            """
+            CHECK-LABEL: define void @{{.+}}foo
+            CHECK: call i32 @"[[BAR:.+]]"(
+            CHECK: call i32 @"[[BAZ:.+]]"(
+
+            CHECK-DAG: declare i32 @"[[BAR]]"({{.+}}alwaysinline
+            CHECK-DAG: declare i32 @"[[BAZ]]"(
```
gmarkall, line 582: Should we also be checking for `noinline` on `BAZ`? And for the `llvm.sin.f64` intrinsic?

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
```diff
@@ -784,6 +790,236 @@ def foo(a):
...
+            CHECK: store i64 5, i64* %"c{{.+}} !dbg ![[STORE5:.+]]
+            CHECK: store i64 1, i64* %"c{{.+}} !dbg ![[STORE1:.+]]
+            CHECK: [[STORE5]] = !DILocation(
+            CHECK: [[STORE1]] = !DILocation(
+            """
+            if n:
+                c = 5
+            else:
+                c = 1
+            # prevents inline of return on py310
```
gmarkall: These assignments may not be necessary as we're not returning `c` in the adapted test case (I am not sure, but suspect).

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
```diff
@@ -784,6 +790,236 @@ def foo(a):
...
+        """Tests that DILocation information for versions of variables matches
+        up to their definition site."""
+
+        @cuda.jit(debug=True)
+        def foo(dest, n):
+            """
+            CHECK: define void @{{.+}}foo
+            CHECK: store i64 5, i64* %"c{{.+}} !dbg ![[STORE5:.+]]
+            CHECK: store i64 1, i64* %"c{{.+}} !dbg ![[STORE1:.+]]
+            CHECK: [[STORE5]] = !DILocation(
```
gmarkall, line 646: Should we be checking that the locations match the source locations here? The original test seems to do this.

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
```diff
@@ -784,6 +790,236 @@ def foo(a):
...
+            )
+
+            ir_content = g.inspect_llvm()[g.signatures[0]]
+            self.assertFileCheckMatches(ir_content, checks)
+
+    def test_nrt_debuginfo_refcount(self):
+        with override_config("CUDA_ENABLE_NRT", True):
+
+            @cuda.jit(debug=True)
+            def sink(*x):
```
gmarkall: I'm not sure `*args` is properly supported in the CUDA target, so it'd probably be safer to write this as

```suggestion
            def sink(x):
```

With `*x`, some innocuous change might cause this to fail otherwise.

gmarkall (commented):
Looks good in general, just a few notes on the diff.

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
ashermancinelli: You're right, looks like my formatter indented this for me and it still passed so I didn't notice. The DILocation in the IR still has `column: 1`, and I'm not sure why.

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
ashermancinelli: `noinline` is not set in these test cases, which I assumed was from this:
```python
# nvvm.py
# The noinline attribute is not permitted, because this causes NVVM to emit
#       a warning, which counts as failing IR verification.
```
Do you expect to see noinline in this test case just like upstream?

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
ashermancinelli: It looks like the original test was not checking for column zero, just that there was one debug location? I'm not sure how useful of a test it is, but to match the upstream test, we can keep the check for one DIlocation and no others, I think.

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
ashermancinelli: You're right, and there's something off/I don't understand about the line info for versioned variables.

ashermancinelli:
/ok to test [30c0832](https://github.com/NVIDIA/numba-cuda/pull/364/commits/30c083265ddd58c28adca010c9b834f22ac6a2a0)

ashermancinelli:
/ok to test [ca096e9](https://github.com/NVIDIA/numba-cuda/pull/364/commits/ca096e9211d71bd2c03506265de36315847a0399)

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
ashermancinelli: The checks are at least checking what the original test was checking - except I cannot get ruff to stop re-indenting the comment, so I am not sure how to check this part in.

ashermancinelli:
Thank you for the review!

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
```diff
@@ -784,6 +791,282 @@ def foo(a):
...
+        self.assertEqual(len(w), 1)
+        found = w[0]
+        self.assertEqual(found.category, NumbaDebugInfoWarning)
+        msg = str(found.message)
+        # make sure the warning contains the right message
+        self.assertIn("Could not find source for function", msg)
+        # and refers to the offending function
+        self.assertIn(str(foo.py_func), msg)
+
+    def test_irregularly_indented_source(self):
```
jiel-nv: I noticed in the effort of the recent migrating debug info tests, most of the decorator lines have option `debug=True` but missing option `opt=False`. That results in generating the following warning message during the test run,

`NumbaInvalidConfigWarning: debug=True with opt=True is not supported by CUDA. This may result in a crash - set debug=False or opt=False.
    warn(NumbaInvalidConfigWarning(msg))
`

Can we add `opt=False` to eliminate the warning? Or otherwise is it intended to produce the warning from running tests?

ashermancinelli: Yes! This became apparent to me too, once the nvjitlink warnings were suppressed. I will fix this.

ashermancinelli:
/ok to test [bc20865](https://github.com/NVIDIA/numba-cuda/pull/364/commits/bc208655c98d8c160b5753ec341e4d55a58ace79)

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
ashermancinelli: Resolved in the latest commit.

numba_cuda/numba/cuda/testing.py
```diff
@@ -177,6 +177,8 @@ def assertFileCheckMatches(
 
             self.fail(
                 f"FileCheck failed:\n{matcher.stderr.getvalue()}\n\n"
+                + f"Check prefixes:\n{check_prefixes}\n\n"
```
gmarkall: I think this was removed in #366 - is this added back here by mistake, or because removing it was a mistake?

When I saw it removed in #366, I assumed that was fine because it makes the test output a little more concise, and the prefixes and patterns could always be observed in the dumps if one is interested. Was that a reasonable evaluation?

gmarkall:
What happened to this PR? It seems to have lost the additional tests it added - unless the Github UI is somehow glitching for me.

ashermancinelli:
Yes, something in my rebase seems to have gone wrong. Please let me revisit this.

numba_cuda/numba/cuda/testing.py
ashermancinelli: Yes, this change was mistakenly taken from the rebase, and it has been removed. Thank you!

ashermancinelli:
/ok to test [7ce3c56](https://github.com/NVIDIA/numba-cuda/pull/364/commits/7ce3c56e94c5425db5cfce554f4d9f7104490b01)

ashermancinelli:
/ok to test [b41097c](https://github.com/NVIDIA/numba-cuda/pull/364/commits/b41097c15090dc8091044fceec29c9461096783e)

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
ashermancinelli: CI on commit [b41097c](https://github.com/NVIDIA/numba-cuda/pull/364/commits/b41097c15090dc8091044fceec29c9461096783e) is failing because ruff still refuses this line. I bypassed the pre-commit checks and added `fmt: off/on` before and after the irregularly indented line, and `fmt: skip` on the line itself, but none seems to help.

ashermancinelli:
/ok to test [4888fd0](https://github.com/NVIDIA/numba-cuda/pull/364/commits/4888fd0f4be8ff2e354d35cbd173f32fa9ac3a9a)

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
ashermancinelli: Test removed from this PR because of issues described above.

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
ashermancinelli: I added more rigorous tests for the device functions calling math functions. I wonder if the checks for inline attributes (and lack of checks for `noinline`) should be added either way because they test our behavior today - and if we want to change that behavior, maybe it should be a separate issue. What do you think?

ashermancinelli:
/ok to test [1cfebf3](https://github.com/NVIDIA/numba-cuda/pull/364/commits/1cfebf31beae6fb6866da8a56e5a415f7a2fd176)

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
gmarkall: The comment
```
# nvvm.py
# The noinline attribute is not permitted, because this causes NVVM to emit
#       a warning, which counts as failing IR verification.
```

refers to kernels only, whereas `bar()` and `baz()` are device functions.

As far as I can tell this test as-is makes sense, I think the upstream one might have evolved over time a bit and the comments got out of sync with the generated code.

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
gmarkall: Did you have a check for the locations to add still?

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
gmarkall: Did these turn out to be needed?

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
ashermancinelli: Yes, I can add this. I will review this test case again.

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
ashermancinelli: No, they were not.

numba_cuda/numba/cuda/tests/cudapy/test_debuginfo.py
ashermancinelli: They have been added in the most recent revision. Thanks!

ashermancinelli:
/ok to test [276ce2d](https://github.com/NVIDIA/numba-cuda/pull/364/commits/276ce2de7a4fed9e9eaa9362a8bf7955ac29a0e7)

gmarkall (approved):
Will re-run CI after merging the CUDA 13 PR, but otherwise I think this is good.

PR #366, opened 2025-07-31
Split out debuginfo tests for types

ashermancinelli:
The basic scalar and array types changed slightly in llvmlite 45; Jie submitted some fixes to eliminate redundant types, which causes some differences in the IR we generate.

This PR addresses the issue in two ways:
- For the scalar types, each type is broken out into its own test case. This removes most of the differences we see in llvmlite45, because there are no common debuginfo metadata that can be eliminated in the first place.
- For array types (float32 is the only one tested this way in both upstream and numba-cuda) and aggregate types (complex numbers) I added a check prefix for llvmlite version >=45 and <=44. The IR is just different enough to make it difficult for the same checks to work with both versions. After Jie's fix, llvmlite no longer creates redundant debuginfo metadata nodes for basic scalar types.

Although it takes up quite a bit of space, I find it easier to navigate the checks with IR snippets in the test code than reading regular expressions. This is a matter of taste though, so I'm open to suggestions.

Note the use of subtests; this should be removed in favor of parameterized tests with pytests once we are fully moved over.

This might also be a good prompt for us to start testing with newer/development versions of llvmlite.

### DIFF
--- a/numba_cuda/numba/cuda/tests/cudapy/test_debuginfo_types.py
+++ b/numba_cuda/numba/cuda/tests/cudapy/test_debuginfo_types.py
@@ -1,5 +1,5 @@
 import numba.cuda as cuda
-from numba.cuda.testing import CUDATestCase
+from numba.cuda.testing import CUDATestCase, skip_on_cudasim
 import llvmlite
 from numba import types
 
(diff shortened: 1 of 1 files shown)


ashermancinelli:
/ok to test [2aec167](https://github.com/NVIDIA/numba-cuda/pull/366/commits/2aec167e7b3a8e041ec554fd51efc76d2613da23)

ashermancinelli:
/ok to test [f272792](https://github.com/NVIDIA/numba-cuda/pull/366/commits/f272792ee5332f1572eabd6e8c3d7cd2ca920c6a)

jiel-nv:
Overall looks good to me. Thank you Asher!

PR #370, opened 2025-08-04
Support CUDA 13, drop support for CUDA 11

gmarkall:
Required changes:

- Test third party libraries with CUDA 12 only
- Update matrix:
  - Drop CUDA 11
  - Change 12.8.0 to 12.9.1
  - Add CUDA 13
- Only run apt commands on Ubuntu - this was coincidentally handled by CTK 11 vs
  12, where 11 ran on Rocky Linux and 12 on Ubuntu
- Remove `CTK_CURAND_VMAP` - this is no longer used, and the cuRAND installation
  is handled as a dependency of the numba-cuda package.
- Updates documentation to drop references to CTK 11, and add references to CTK
  13.
  - MVC is handled automatically, so most of the documentation for it is
    dropped.
- Remove code relating to CTK 11 only.
- Add handling for CUDA 13 paths when using the ctypes binding
- Prevent `NVRTC()` objects being creating when using the NVIDIA cuda-python
  bindings, and replace uses of it with . The `NVRTC()` class directly opens the
  NVRTC DLL / SO, so it should not be used in conjunction with the cuda-python
  bindings.
- In test binaries, the lowest CC we should generate code for is 7.5, because
  this is the minimum supported by CTK 13.
- The CCCL include path has changed in CUDA 13, so we add that to the include
  path for test binary generation.


brandon-b-miller:
Can't comment on the exact line but we should remove `NUMBA_CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY` `docs/source/reference/envvars.rst`

pyproject.toml
```diff
@@ -38,37 +30,46 @@ cu12 = [
...
     "nvidia-nvjitlink-cu12"
 ]
+cu13 = [
+    "cuda-bindings==13.0.*",
+    "cuda-core==0.3.*",
+    "cuda-python==13.0.*",  # supports all CTK 12.x
+    "nvidia-cuda-nvcc-cu13",  # for libNVVM
+    "nvidia-cuda-runtime-cu13",
+    "nvidia-cuda-nvrtc-cu13",
+    "nvidia-nvjitlink-cu13"
```
kkraus14: The scheme has changed for CUDA 13 where this is now `nvidia-cuda-nvcc=13.*` for example

pyproject.toml
```diff
@@ -53,7 +53,7 @@ test-cu12 = [
 ]
 test-cu13 = [
     "numba-cuda[test]",
-    "nvidia-curand-cu13",
+    "nvidia-curand-10.4.*",
```
kkraus14:
```suggestion
    "nvidia-curand=10.4.*",
```

docs/source/user/installation.rst
```diff
@@ -30,23 +28,21 @@ versions, refer to :ref:`minor-version-compatibility`.
 Installation with a Python package manager
 ==========================================
 
-Conda users can install the CUDA Toolkit into a conda environment.
-
-For CUDA 12::
+Conda users can install the CUDA Toolkit into a conda environment::
 
     $ conda install -c conda-forge numba-cuda "cuda-version>=12.0"
```
bdice: This will allow `cuda-version 13.0` as written.
```suggestion
    $ conda install -c conda-forge numba-cuda "cuda-version=12"
```

docs/source/user/installation.rst
```diff
@@ -30,23 +28,21 @@ versions, refer to :ref:`minor-version-compatibility`.
...
 
-Alternatively, you can install all CUDA 12 dependencies from PyPI via ``pip``::
+Or for CUDA 13::
 
-    $ pip install numba-cuda[cu12]
+    $ conda install -c conda-forge numba-cuda "cuda-version>=13.0"
```
bdice:
```suggestion
    $ conda install -c conda-forge numba-cuda "cuda-version=13"
```

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -348,16 +336,10 @@ def compile(src, name, cc, ltoir=False):
...
-    nvrtc_ver_major = nvrtc_version[0]
-
     cudadrv_path = os.path.dirname(os.path.abspath(__file__))
     numba_cuda_path = os.path.dirname(cudadrv_path)
 
-    if nvrtc_ver_major == 11:
-        numba_include = f"{os.path.join(numba_cuda_path, 'include', '11')}"
-    else:
-        numba_include = f"{os.path.join(numba_cuda_path, 'include', '12')}"
+    numba_include = f"{os.path.join(numba_cuda_path, 'include', '12')}"
```
bdice, line 355: Might need additional logic for CUDA 13.

pyproject.toml
```diff
@@ -38,37 +30,46 @@ cu12 = [
     "nvidia-cuda-nvrtc-cu12",
     "nvidia-nvjitlink-cu12"
 ]
+cu13 = [
+    "cuda-bindings==13.0.*",
+    "cuda-core==0.3.*",
+    "cuda-python==13.0.*",  # supports all CTK 12.x
```
bdice:
```suggestion
    "cuda-python==13.0.*",  # supports all CTK 13.x
```

pyproject.toml
```diff
@@ -38,37 +30,46 @@ cu12 = [
...
+    "nvidia-cuda-nvcc=13.*",  # for libNVVM
+    "nvidia-cuda-runtime=13.*",
+    "nvidia-cuda-nvrtc=13.*",
+    "nvidia-nvjitlink=13.*"
```
bdice: I think CI is failing because of incompatible dependency specifiers.
```
 configuration error: `project.optional-dependencies.cu13[3]` must be pep508
```
```suggestion
    "nvidia-cuda-nvcc==13.*",  # for libNVVM
    "nvidia-cuda-runtime==13.*",
    "nvidia-cuda-nvrtc==13.*",
    "nvidia-nvjitlink==13.*"
```

pyproject.toml
```diff
@@ -38,37 +30,46 @@ cu12 = [
...
-    "numba-cuda[test]",
-    "nvidia-curand-cu11",
-]
 test-cu12 = [
     "numba-cuda[test]",
     "nvidia-curand-cu12",
 ]
+test-cu13 = [
+    "numba-cuda[test]",
+    "nvidia-curand=10.4.*",
```
bdice:
```suggestion
    "nvidia-curand==10.4.*",
```

kkraus14:
It looks like most of the failures are from failing to find `<cuda/atomic>` which is part of CCCL.

I think this is because we aren't handling `SYSTEM_INCLUDES` here: https://github.com/NVIDIA/numba-cuda/blob/be32e90da3424b9a0475f3239cce8f97c3fb1765/numba_cuda/numba/cuda/tests/test_binary_generation/generate_raw_ltoir.py#L43 and CUDA 13.0 moved CCCL to a new location which made it a system include instead of a user include.

kkraus14:
We also need to add 13.0 here: https://github.com/NVIDIA/numba-cuda/blob/be32e90da3424b9a0475f3239cce8f97c3fb1765/ci/test_wheel_deps_wheels.sh#L8

docs/source/user/installation.rst
```diff
@@ -11,17 +11,15 @@ Supported GPUs
 --------------
 
 Numba supports all NVIDIA GPUs that are supported by the CUDA Toolkit it uses.
-Presently for CUDA 11 this ranges from Compute Capabilities 3.5 to 9.0, and for
-CUDA 12 this ranges from 5.0 to 12.1, depending on the exact installed version.
+Presently for CUDA 12 this ranges from Compute Capabilities 5.0 to 12.1, and for
+CUDA 13 this ranges from 5.0 to XXX, depending on the exact installed version.
```
kkraus14: Do we want to leave this as `XXX` or update it as we go?

pyproject.toml
```diff
@@ -38,6 +30,16 @@ cu12 = [
     "nvidia-cuda-nvrtc-cu12",
     "nvidia-nvjitlink-cu12"
 ]
+cu13 = [
+    "cuda-bindings==13.0.*",
+    "cuda-core==0.3.*",
+    "cuda-python==13.0.*",  # supports all CTK 13.x
+    "nvidia-cuda-nvcc==13.*",  # for libNVVM
```
kkraus14: In CUDA 13 there is now an `nvidia-nvvm` package

```suggestion
    "nvidia-nvvm==13.*",
```

pyproject.toml
```diff
@@ -38,6 +30,16 @@ cu12 = [
     "nvidia-cuda-nvrtc-cu12",
     "nvidia-nvjitlink-cu12"
 ]
+cu13 = [
+    "cuda-bindings==13.0.*",
+    "cuda-core==0.3.*",
+    "cuda-python==13.0.*",  # supports all CTK 13.x
```
kkraus14: I think we can pin this to just `13.*` since we're guaranteeing not to have API / ABI breaking changes?

```suggestion
    "cuda-python==13.*",
```

pyproject.toml
```diff
@@ -38,6 +30,16 @@ cu12 = [
     "nvidia-cuda-nvrtc-cu12",
     "nvidia-nvjitlink-cu12"
 ]
+cu13 = [
+    "cuda-bindings==13.0.*",
```
kkraus14:
```suggestion
    "cuda-bindings==13.*",
```

pyproject.toml
```diff
@@ -46,16 +48,16 @@ test = [
...
-]
 test-cu12 = [
     "numba-cuda[cu12]",
     "numba-cuda[test]",
     "nvidia-curand-cu12",
 ]
+test-cu13 = [
+    "numba-cuda[cu13]",
+    "numba-cuda[test]",
+    "nvidia-curand==10.4.*",
```
kkraus14, line 60: This doesn't allow floating forward in minor versions for cuRAND. Should we make this `10.*` instead?

ci/matrix.yml
```diff
@@ -3,13 +3,13 @@ build-matrix:
...
-  - { CUDA_VER: '11.8.0', ARCH: 'amd64', PY_VER: '3.10', LINUX_VER: 'ubuntu20.04', GPU: 'l4',   DRIVER: 'latest' }
-  - { CUDA_VER: '12.0.1', ARCH: 'amd64', PY_VER: '3.11', LINUX_VER: 'ubuntu22.04', GPU: 'l4',   DRIVER: 'latest' }
-  - { CUDA_VER: '12.2.2', ARCH: 'amd64', PY_VER: '3.12', LINUX_VER: 'ubuntu22.04', GPU: 'l4',   DRIVER: 'latest' }
-  - { CUDA_VER: '12.8.0', ARCH: 'amd64', PY_VER: '3.13', LINUX_VER: 'ubuntu24.04', GPU: 'l4',   DRIVER: 'latest' }
-  - { CUDA_VER: '11.8.0', ARCH: 'arm64', PY_VER: '3.9',  LINUX_VER: 'rockylinux8', GPU: 'a100', DRIVER: 'earliest' }
-  - { CUDA_VER: '11.8.0', ARCH: 'arm64', PY_VER: '3.10', LINUX_VER: 'ubuntu20.04', GPU: 'a100', DRIVER: 'latest' }
-  - { CUDA_VER: '12.2.2', ARCH: 'arm64', PY_VER: '3.11', LINUX_VER: 'ubuntu22.04', GPU: 'a100', DRIVER: 'latest' }
-  - { CUDA_VER: '12.2.2', ARCH: 'arm64', PY_VER: '3.12', LINUX_VER: 'ubuntu22.04', GPU: 'a100', DRIVER: 'latest' }
-  - { CUDA_VER: '12.8.0', ARCH: 'arm64', PY_VER: '3.13', LINUX_VER: 'ubuntu24.04', GPU: 'a100', DRIVER: 'latest' }
+  - { CONTAINER_CUDA_VER: '12.0.1', CUDA_VER: '12.0.1', ARCH: 'amd64', PY_VER: '3.9',  LINUX_VER: 'rockylinux8', GPU: 'l4',   DRIVER: 'earliest' }
```
bdice: I think 57d504766170604a530434df4f58957126ff14af can be reverted. We are shipping CUDA 13.0 ci-conda images now.

docs/source/user/installation.rst
bdice: We definitely need the lower bound to be 7.5. Maxwell, Pascal, and Volta were dropped in CUDA 13. I'd leave the upper bound at 12.1 but it can be qualified with "as of now" in case this is missed during future updates.
```suggestion
CUDA 13 this ranges from 7.5 to 12.1 (the latest as of CUDA 13.0), depending on the exact installed version.
```

numba_cuda/numba/cuda/cudadrv/libs.py
```diff
@@ -2,8 +2,7 @@
 
 CUDA Toolkit libraries can be available via either:
 
-- the `cuda-nvcc` and `cuda-nvrtc` conda packages for CUDA 12,
-- the `cudatoolkit` conda package for CUDA 11,
+- the `cuda-nvcc` and `cuda-nvrtc` conda packages,
```
bdice, line 5: I believe that with some effort, we should be able to support CUDA 13 from wheels too. CUDA 13 ships a full compiler in the `nvidia-cuda-nvcc` package. That can be out of scope for the first pass of CUDA 13 support, but will be important for RAPIDS over time so we're no longer tied to system CTK packages when using wheels.

pyproject.toml
bdice: We could also use the [brand-new `cuda-toolkit` metapackage](https://pypi.org/project/cuda-toolkit/#description) to get a curand that ships with any CUDA 13 toolkit, like this:
```suggestion
    "cuda-toolkit[curand]==13.*",
```

bdice (commented):
A few more bits of feedback to help this along.

There might be a couple remaining references to `CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY` in test code that could be removed. These could be cleaned up in a later PR, or left in place for test coverage if we think that there could be code paths that trigger something somehow.

It should be safe to delete these test utilities and all references to them.
https://github.com/gmarkall/numba-cuda/blob/1bd3b2290268b51833ded617bc28f11fd1dee92f/numba_cuda/numba/cuda/testing.py#L265-L284

Another test using `CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY`:
https://github.com/gmarkall/numba-cuda/blob/1bd3b2290268b51833ded617bc28f11fd1dee92f/numba_cuda/numba/cuda/tests/cudadrv/test_mvc.py#L12-L22

numba_cuda/numba/cuda/cudadrv/libs.py
gmarkall: We have the test-wheel-deps-wheels test that tests this, so I think it is supported, but through the cuda-python bindings instead of the ctypes bindings that `libs.py` is in support of. So I'll leave this as-is for now, on the basis that it should be removed once the ctypes binding is removed.

numba_cuda/numba/cuda/cudadrv/nvrtc.py
gmarkall: I agree - I think it depends if we vendor in headers from the CUDA 13 toolkit, or if the 12 ones contain the functionality we need adequately.

pyproject.toml
gmarkall: I got the impression that 10.3 was for CUDA 12 and 10.4 for CUDA 13 so I wasn't sure about making it 10.*. Is that a misunderstanding?

Does `cuda-toolkit[curand]==13.*` install only curand?

gmarkall: Ah, it seems like that is the case. Maybe it can be used for other toolkit packages too.

bdice: Yes, see https://pypi.org/project/cuda-toolkit/ for the description. Various packages can be selected via extras, like `cuda-toolkit[curand,cublas,nvcc]`.

kkraus14:
The failing jobs are from failing to dlopen `libnvrtc.so` from wheels.

.github/workflows/pr.yaml
```diff
@@ -146,7 +146,7 @@ jobs:
       build_type: pull-request
       script: "ci/test_thirdparty.sh"
       matrix: ${{ needs.compute-matrix.outputs.TEST_MATRIX }}
-      matrix_filter: map(select(.ARCH == "amd64" and (.CUDA_VER | split(".") | .[0] | tonumber >= 12))) | group_by(.CUDA_VER|split(".")|map(tonumber)|.[0]) | map(max_by([(.PY_VER|split(".")|map(tonumber)), (.CUDA_VER|split(".")|map(tonumber))]))
+      matrix_filter: map(select(.ARCH == "amd64" and (.CUDA_VER | split(".") | .[0] | tonumber == 12))) | group_by(.CUDA_VER|split(".")|map(tonumber)|.[0]) | map(max_by([(.PY_VER|split(".")|map(tonumber)), (.CUDA_VER|split(".")|map(tonumber))]))
```
bdice, line 150: Do we need a TODO to re-enable tests with CUDA 13? What requirements are missing? (e.g. cupy, …?)

.github/workflows/pr.yaml
```diff
@@ -162,4 +162,4 @@ jobs:
       build_type: pull-request
       script: "ci/coverage_report.sh"
       matrix: ${{ needs.compute-matrix.outputs.TEST_MATRIX }}
-      matrix_filter: 'map(select(.ARCH == "amd64" and .CUDA_VER == "12.8.0" and .PY_VER == "3.13")) | .[0:1]'
+      matrix_filter: 'map(select(.ARCH == "amd64" and .CUDA_VER == "12.9.1" and .PY_VER == "3.11")) | .[0:1]'
```
bdice, line 166: Why was Python downgraded?

ci/matrix.yml
```diff
@@ -3,13 +3,13 @@ build-matrix:
...
-  - { CUDA_VER: '12.0.1', ARCH: 'amd64', PY_VER: '3.11', LINUX_VER: 'ubuntu22.04', GPU: 'l4',   DRIVER: 'latest' }
-  - { CUDA_VER: '12.2.2', ARCH: 'amd64', PY_VER: '3.12', LINUX_VER: 'ubuntu22.04', GPU: 'l4',   DRIVER: 'latest' }
-  - { CUDA_VER: '12.8.0', ARCH: 'amd64', PY_VER: '3.13', LINUX_VER: 'ubuntu24.04', GPU: 'l4',   DRIVER: 'latest' }
-  - { CUDA_VER: '11.8.0', ARCH: 'arm64', PY_VER: '3.9',  LINUX_VER: 'rockylinux8', GPU: 'a100', DRIVER: 'earliest' }
-  - { CUDA_VER: '11.8.0', ARCH: 'arm64', PY_VER: '3.10', LINUX_VER: 'ubuntu20.04', GPU: 'a100', DRIVER: 'latest' }
-  - { CUDA_VER: '12.2.2', ARCH: 'arm64', PY_VER: '3.11', LINUX_VER: 'ubuntu22.04', GPU: 'a100', DRIVER: 'latest' }
-  - { CUDA_VER: '12.2.2', ARCH: 'arm64', PY_VER: '3.12', LINUX_VER: 'ubuntu22.04', GPU: 'a100', DRIVER: 'latest' }
-  - { CUDA_VER: '12.8.0', ARCH: 'arm64', PY_VER: '3.13', LINUX_VER: 'ubuntu24.04', GPU: 'a100', DRIVER: 'latest' }
+  - { CUDA_VER: '12.0.1', ARCH: 'amd64', PY_VER: '3.9',  LINUX_VER: 'rockylinux8', GPU: 'l4',   DRIVER: 'earliest' }
+  - { CUDA_VER: '12.2.2', ARCH: 'amd64', PY_VER: '3.10', LINUX_VER: 'ubuntu20.04', GPU: 'l4',   DRIVER: 'latest' }
```
bdice: We are no longer producing ubuntu20.04 images (EOL) so these may be out of date.

numba_cuda/numba/cuda/cudadrv/nvrtc.py
bdice: It looks like all of these headers are for bfloat16 support. Is this something that we still need to vendor or can we get it from the Toolkit? I saw “ bfloat16 requires compute capability 8.0+ and CUDA version>= 12.0” which makes me think we should not need to vendor?

Can we delete the directory of CUDA 11 includes? https://github.com/gmarkall/numba-cuda/tree/cuda-13/numba_cuda/numba/cuda/include/11

pyproject.toml
```diff
@@ -38,6 +30,16 @@ cu12 = [
     "nvidia-cuda-nvrtc-cu12",
     "nvidia-nvjitlink-cu12"
 ]
+cu13 = [
+    "cuda-bindings==13.*",
+    "cuda-core==0.3.*",
```
bdice: Might need a minimum patch version. Something like this, but check my values for the first patch release to support CUDA 13:
```suggestion
    "cuda-core>=0.3.2,<0.4.0dev0",
```

.github/workflows/pr.yaml
gmarkall: The thirdparty test is presently cuDF, so having a version of cuDF that officially supports CUDA 13 is probably a good criteria for this - I think there's not a CUDA 13-supporting version yet?

.github/workflows/pr.yaml
gmarkall: Because the matrix changed so that the newer Pythons were tested with CUDA 13, and I wanted to keep the coverage report on a CUDA 12 version.

ci/matrix.yml
gmarkall: Ah, thanks for the heads-up - I'll put this as a to-do item to upgrade to 22.04, but will leave as-is for now as it is working.

numba_cuda/numba/cuda/cudadrv/nvrtc.py
gmarkall: We vendor them to avoid having to discover the correct version of the headers (or assume their existence on the system).

I don't mind considering getting rid of the vendoring in the long term, but I think this causes no issue for now, and trying to rely on finding them on the users' system and getting it right in all cases will likely require a quantity of effort that I just don't think makes sense to prioritise with the massive queue of outstanding other items right now.

We can delete the CUDA 11 includes.

ci/matrix.yml
gmarkall: Actually nevermind, I'm going to rework this right now.

pyproject.toml
gmarkall: I'll add a TODO for this.

kkraus14 (approved):

numba_cuda/numba/cuda/cudadrv/nvrtc.py
leofang: FWIW the pathfinder will gain header search capability soon-ish: https://github.com/NVIDIA/cuda-python/issues/832

pyproject.toml
```diff
@@ -38,6 +30,17 @@ cu12 = [
     "nvidia-cuda-nvrtc-cu12",
     "nvidia-nvjitlink-cu12"
 ]
+# TODO: Use cuda-toolkit package dependencies - e.g. cuda-toolkit[curand,nvvm,nvrtc]=13.*
+cu13 = [
+    "cuda-bindings==13.*",
+    "cuda-core==0.3.2,<0.4.0dev0",
```
leofang, line 36: In the near future we should aim to relax this pinning, I think it's too tight. The cuda-core features used by numba-cuda has been stabilized since 0.3.0.

pyproject.toml
```diff
@@ -38,6 +30,17 @@ cu12 = [
...
+    "cuda-python==13.*",
```
leofang, line 37: We actually don't need to depend on `cuda-python` anymore, since numba-cuda already depends on the needed components.

.github/workflows/pr.yaml
bdice: Correct. We are waiting on numba-cuda. 🙃 We'll get this sorted out soon enough.

pyproject.toml
gmarkall: Do you think
```
cuda-core>=0.3.2
```
is now appropriate?

pyproject.toml
gmarkall: Does it depend on them by depending on cuda-core?

PR #371, opened 2025-08-05
Relax subclass check in vendored in typing templates

atmnp:
Relaxes subclass check for Function and AbstractTemplate as an interim solution for nvmath-python until next release of Numba CUDA.


ZzEeKkAa (approved):
I can confirm that this fixed the problem for me, thank you! But would wait @gmarkall approval before merging

gmarkall (approved):

gmarkall:
This makes sense and seems good, but I do want to wait until CI is back before merging it.

PR #372, opened 2025-08-05
remove uses of rapids-configure-conda-channels

jameslamb:
Contributes to https://github.com/rapidsai/gha-tools/issues/145

In RAPIDS, we're planning to remote the `rapids-configure-conda-channels` tool. This removes the one use of that here.

I don't think it's necessary here. That tool just exists to choose between the `rapidsai` and `rapidsai-nightly` conda channels based on whether or not a given CI run is a release (i.e. triggered by a tag/release).

source code: https://github.com/rapidsai/gha-tools/blob/3996b1773127b5b5b3969accf8a7660067571ab7/tools/rapids-configure-conda-channels#L28-L36

`numba-cuda` doesn't rely on either of those channels in its CI. As far as I can tell, the only RAPIDS dependencies this project has are the *optional* dependencies on `cubinlinker` and `ptxcompiler`, which aren't ever installed in CI and just documented as a CUDA-11-only opt-in:

https://github.com/NVIDIA/numba-cuda/blob/b13d062bc254bf5fec8ee355f0f584e8dc033cad/docs/source/user/minor_version_compatibility.rst?plain=1#L28-L36


gmarkall (approved):

gmarkall:
Thansk @jameslamb !

PR #373, opened 2025-08-05
[Refactor] Vendor in _dispatcher, _devicearray, mviewbuf C extension for CUDA-specific customization

VijayKandiah:
This PR vendors in the _dispatcher, _devicearray, mviewbuf C extensions for CUDA-specific changes.


VijayKandiah:
Currently, `pip install -e .` does not seem to build the extension, so we have to use `python setup.py build_ext --inplace` to build the extension. This will probably fail CI testing until we either update CI or make `pip install -e .` automatically build the extension.

gmarkall:
> Currently, `pip install -e .` does not seem to build the extension, so we have to use `python setup.py build_ext --inplace` to build the extension. This will probably fail CI testing until we either update CI or make `pip install -e .` automatically build the extension.

I just had a vague recollection about this - do you have setuptools > 78 installed? If so, does going back to setuptools 78 or earlier make `pip install -e .` build the extension?

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -1540,3 +2135,16 @@ def _reduce_states(self):
         Compiled definitions are discarded.
         """
         return dict(py_func=self.py_func, targetoptions=self.targetoptions)
+
+
+if config.USE_LEGACY_TYPE_SYSTEM:  # Old type system
```
gmarkall: There is only the "legacy" type system, so we don't need to check this - we can just use the branch for "legacy". (The new type system was started, then abandoned)

VijayKandiah: Sounds good, I changed it.

VijayKandiah:
> > Currently, `pip install -e .` does not seem to build the extension, so we have to use `python setup.py build_ext --inplace` to build the extension. This will probably fail CI testing until we either update CI or make `pip install -e .` automatically build the extension.
> 
> I just had a vague recollection about this - do you have setuptools > 78 installed? If so, does going back to setuptools 78 or earlier make `pip install -e .` build the extension?

The issue was not a newer setuptools version. It was simply that `is_building` method in `setup.py` was not looking for `editable_wheel` as one of the arguments to enable building the C extension for. Fixed it with my recent commit.

VijayKandiah:
I had to include `numpy` as a build requirement for `setuptools.build_meta` in `pyproject.toml`. However, I am seeing build errors in our CI with numpy not found.
```
ModuleNotFoundError: No module named 'numpy'
  error: subprocess-exited-with-error
  
  × Building wheel for numba-cuda (pyproject.toml) did not run successfully.
  │ exit code: 1
 ```

 Is there a way to update CI to add this build requirement?

**Comment by VijayKandiah** [comment]
My latest commit also vendors in `_devicearray` and `mviewbuf` C extension. This should encompass all the C extensions we use in numba-cuda. 

I noticed a circular import issue if `_devicearray` cext was placed at `numba_cuda.numba.cuda`, and moving it to the top level `numba_cuda` fixed this. For uniformity, I ended up placing all cexts at `numba_cuda.cext`. @gmarkall please let me know your thoughts on the updated folder structure here.

**Comment by ashermancinelli** [comment]
The failures in CI fail on the `python -m numba --sysinfo` command, which I can reproduce locally with your branch if I perform a non-editable install. FWIW, If I put an `__init__.py` in `numba_cuda/cext/` and reinstall, then I can run the sysinfo command successfully. I'm not yet sure if there are more issues though.

**Comment by gmarkall** [comment]
Just to record where we are after discussion - I think this requires the extension modules to be moved under `numba_cuda.numba.cuda.cext` due to a potential bad interaction with the redirector in their current location, before we can get it to work. It will also need the build matrix modifying to build separately for `amd64` and `aarch64`.

File: ci/test_simulator.sh [hunk 2]
```diff
@@ -21,7 +21,18 @@ set -u
 
 pip install filecheck
 
-rapids-mamba-retry install -c `pwd`/conda-repo numba-cuda
+# Detect system architecture to set conda repo path
+ARCH=$(uname -m)
```
rparolin: Can we hoist this out into a common place so we don't repeat it in all our .sh files?

VijayKandiah: Yes we should. The latest version of this PR is at https://github.com/NVIDIA/numba-cuda/pull/451 as Asher and Graham took a crack as fixing CI there. We'll add a common utility function to calculate the ARCH and folder path once we get CI working there.

VijayKandiah:
Closing this one as the follow up is at https://github.com/NVIDIA/numba-cuda/pull/451.

VijayKandiah:
This PR currently contains some import hacks in `cext/__init__.py` to help with our CI testing. The attempt to remove import hacks is ongoing at https://github.com/NVIDIA/numba-cuda/pull/451. As long as CI passes here, and everything else looks good to reviewers, I am in favor of getting this PR in and create a follow up PR for removing import hacks.

PR #374, opened 2025-08-05
[WIP][Refactor] Vendor in npyfuncs for CUDA-specific customization

atmnp:
This moves the func impls used in our ufunc_db into this repo. I no longer think that this needs additional comprehensive testing because the vast majority if not all of the impls are already tested within `test_ufuncs.py`


numba_cuda/numba/cuda/np/npyfuncs.py
```diff
@@ -0,0 +1,1785 @@
...
+
+    res = context.compile_internal(builder, lcm, sig, args)
+    return impl_ret_untracked(context, builder, sig.return_type, res)
+
+
+########################################################################
+# Numpy style complex sign
+
+
+def np_complex_sign_impl(context, builder, sig, args):
```
gmarkall, line 630: Why does this only include the NumPy 1.x-equivalent implementation? Tests are not failing because we haven't implemented `np.sign` yet, but that's only because we never got around to adding all the ufuncs. Once we add it I would not expect this to behave correctly with NumPy 2.x. If there's nothing blocking it, can you add the NumPy 2.x version back please?

gmarkall (commented):
Looks good (assuming CI still passes, I'm just waiting for it) - with one question on the diff.

numba_cuda/numba/cuda/np/npyfuncs.py
gmarkall: To clarify a bit - in upstream Numba there is a branch that does things differently with NumPy 2.x:

https://github.com/numba/numba/blob/b8db41c873640fabdcebeb241b98556b693ef02b/numba/np/npyfuncs.py#L584-L599

This implementation only handles the NumPy 1.x side of that branch - I think that the branch and both implementations should be present here too.

numba_cuda/numba/cuda/np/npyfuncs.py
```diff
@@ -51,7 +51,7 @@ def _check_arity_and_homogeneity(sig, args, arity, return_type=None):
         assert False, msg
 
 
-if config.USE_LEGACY_TYPE_SYSTEM:
+if getattr(config, "USE_LEGACY_TYPE_SYSTEM", True):
```
gmarkall, line 54: For future reference, you can always assume `USE_LEGACY_TYPE_SYSTEM` is `True`, and not include the code for the case when it is `False`.

gmarkall (approved):

PR #376, opened 2025-08-05
Improve Bfloat16 Support in Numba-CUDA

isVoid:
This is an overhaul of the bfloat16 support. The PR improves bfloat16 by implementing a unified type object to the constructor handle, defining proper cast and unify rules of bfloat16 to other numeric types, adding *initial* host side support of this type (leveraging ml_dtypes package), printing support in kernel, as well as plethora of intrinsics.

Details:
1. User can now write `from numba.cuda.types import bfloat16` and directly construct one `x = bfloat16(3.14)` on both host and device. (Host side requires `ml_dtypes` package).
2. You may now `print(x)` on device.
3. You may now use comparison intrinsics, such as heq, hmax, hmax_nan etc. on device.
4. You may also use conversion intrinsics, such as `int64_to_bfloat16_rz` etc. on device; bit cast intrinsics, such as `int16_as_bfloat16` on device.


numba_cuda/numba/cuda/models.py
```diff
@@ -42,3 +43,35 @@ def __init__(self, dmm, fe_type):
...
+
+def _as_bfloat(value):
+    # Step 1: Convert to float
+    f = ir.types._as_float(value)
+    # Step 2: Truncate (or round, we choose truncate) last 16 bits
+    bf = f >> 16
+    return bf
+
+
+class BfloatType(ir.types._BaseFloatType):
```
gmarkall: Should this be:

```suggestion
class Bfloat16Type(ir.types._BaseFloatType):
```

I don't believe there are other bitwidths, but would this be more consistent naming? Or is there some other subtlety here?

numba_cuda/numba/cuda/models.py
```diff
@@ -42,3 +43,35 @@ def __init__(self, dmm, fe_type):
...
+    """Brain-float type"""
+
+    null = "0.0"
+    intrinsic_name = "bfloat"
+
+    def __str__(self):
+        return "bfloat"
+
+    def format_constant(self, value):
```
gmarkall: How is this method intended to be used? For a `FloatType`, I can do:

```python
from llvmlite import ir

float_type = ir.FloatType()
float_type.format_constant(1)
# '0x3ff0000000000000'
```

but with the `BfloatType` the same yields an error:

```python
from numba.cuda.models import BfloatType

bfloat_type = BfloatType()
bfloat_type.format_constant(1)
# TypeError: unsupported operand type(s) for >>: 'float' and 'int'
```

numba_cuda/numba/cuda/models.py
isVoid: I was thinking largely due to the native type name in llvm for bf16 is `bfloat`. Trying to be consistent with llvm.

gmarkall: Ah, makes sense!

numba_cuda/numba/cuda/models.py
isVoid: Updated. This method is defined for all other float types in llvmlite. So I also implemented it following the pattern. This function formats the constant in the IEEE754 double representation of the corresponding float value.

numba_cuda/numba/cuda/types.py
```diff
@@ -38,3 +39,51 @@ class CUDADispatcher(types.Dispatcher):
...
+        if isinstance(other, (types.Float, types.Integer)):
+            return typingctx.unify_pairs(self, other)
+
+    def cast_python_value(self, value):
+        try:
+            import ml_dtypes  # noqa: F401
+
+            return ml_dtypes.bfloat16(value)
+        except ImportError:
+            raise NotImplementedError
```
gmarkall: I forgot to install `ml_dtypes` and all I got was a `NotImplementedError`. Maybe 
```suggestion
            raise NotImplementedError("Please install ml_dtypes to use bfloat16 on the host")
```

or similar would be a bit more friendly?

numba_cuda/numba/cuda/types.py
```diff
@@ -38,3 +39,51 @@ class CUDADispatcher(types.Dispatcher):
...
+
+    def can_convert_from(self, typingctx, other):
+        if isinstance(other, types.Float):
+            return Conversion.unsafe
+
+        elif isinstance(other, types.Integer):
+            if other.bitwidth == 8:
+                return Conversion.safe
+
+        return Conversion.unsafe
```
gmarkall: If the other type is not a float or an integer, then I think it can't be converted - is that the case? If so, we should return `None` for conversions that aren't possible. Should this be:

```suggestion
            if other.bitwidth == 8:
                return Conversion.safe
            else:
                return Conversion.unsafe
```

gmarkall: E.g. I think we probably don't want to be able to convert from `complex64`

numba_cuda/numba/cuda/types.py
```diff
@@ -38,3 +39,51 @@ class CUDADispatcher(types.Dispatcher):
...
+    def can_convert_to(self, typingctx, other):
+        if isinstance(other, types.Float):
+            if other.bitwidth >= 32:
+                return Conversion.safe
+            else:
+                return Conversion.unsafe
+        elif isinstance(other, types.Integer):
+            return Conversion.unsafe
+
+        return Conversion.unsafe
```
gmarkall: A similar question here about conversions from non-integer/floating point types.

numba_cuda/numba/cuda/printimpl.py
```diff
@@ -48,6 +48,17 @@ def real_print_impl(ty, context, builder, val):
     return "%f", [lld]
 
 
+@print_item.register(Bfloat16)
+def bfloat16_print_impl(ty, context, builder, val):
```
gmarkall, line 55: Is this tested? I can't see a print being checked in the tests.

numba_cuda/numba/cuda/models.py
```diff
@@ -42,3 +44,48 @@ def __init__(self, dmm, fe_type):
...
+class _model___nv_bfloat16(PrimitiveModel):
+    def __init__(self, dmm, fe_type):
+        from numba.cuda.api import get_current_device
+
+        major, minor = get_current_device().compute_capability
+
+        # Blackwell device leverage latest nvvm (llvm 20+ dialect) which has
+        # bfloat type
+        if major >= 10:
+            be_type = BfloatType()
```
gmarkall: Did you find out if this mirrors what NVCC / NVRTC do on Blackwell?

numba_cuda/numba/cuda/printimpl.py
gmarkall: ... The reason I ask is, I thought a bitcast might be necessary if `val` is not an integer (e.g. with the data model as it is for CC 10 and above), and wondered why that didn't show up as a potential issue.

numba_cuda/numba/cuda/tests/cudapy/test_bfloat16.py
```diff
@@ -1,5 +1,112 @@
-from numba import cuda, float32
-from numba.cuda.bf16 import bfloat16
+import unittest
+from importlib.util import find_spec
+import numpy as np
+from ml_dtypes import bfloat16 as mldtypes_bf16
```
gmarkall, line 5: This makes ml_dtypes a hard requirement for testing. Even if we're not necessarily going to add it as a hard dependency of numba-cuda without some further deliberation / enquiry, maybe it should be added to the test dependencies at least?

gmarkall (changes requested):
I think this is largely looking good - just a couple of questions / comments on the diff.

numba_cuda/numba/cuda/tests/cudapy/test_bfloat16.py
isVoid: Yes - we should add this to the test dependencies.

numba_cuda/numba/cuda/printimpl.py
isVoid: Good catch, it isn't tested. Since we dropped adopting native type for cc10, it wouldn't be an issue for newer archs.

numba_cuda/numba/cuda/models.py
isVoid: [edea3c3](https://github.com/NVIDIA/numba-cuda/pull/376/commits/edea3c3e7022e8f6b32a748b3638801b8feb3380)

numba_cuda/numba/cuda/printimpl.py
isVoid: [edea3c3](https://github.com/NVIDIA/numba-cuda/pull/376/commits/edea3c3e7022e8f6b32a748b3638801b8feb3380)

numba_cuda/numba/cuda/types.py
isVoid: Ah that's right. When I wrote this, I was only thinking about scalar numeric types.

numba_cuda/numba/cuda/models.py
isVoid: I found out. And this does not mirror nvcc/nvrtc. They use IntType16 as underlying type.

isVoid: [bc3b27d](https://github.com/NVIDIA/numba-cuda/pull/376/commits/bc3b27de97e60cab7a0abf4df793ff0875c3caa9)

numba_cuda/numba/cuda/types.py
isVoid: [0fa0174](https://github.com/NVIDIA/numba-cuda/pull/376/commits/0fa0174d42320b011f3fd7b3c563042d48f80c5d)

numba_cuda/numba/cuda/types.py
isVoid: [0fa0174](https://github.com/NVIDIA/numba-cuda/pull/376/commits/0fa0174d42320b011f3fd7b3c563042d48f80c5d)

numba_cuda/numba/cuda/tests/cudapy/test_bfloat16.py
isVoid: [e09ffc6](https://github.com/NVIDIA/numba-cuda/pull/376/commits/e09ffc6dff6a935c4cd9325f1d4eab4120921787)

gmarkall (approved):
Looks good - I think once CI is passing / any conflicts are resolved, this will be good to merge.

numba_cuda/numba/cuda/types.py
```diff
@@ -38,3 +39,68 @@ class CUDADispatcher(types.Dispatcher):
...
+
+    def unify(self, typingctx, other):
+        if isinstance(other, (types.Float, types.Integer)):
+            return typingctx.unify_pairs(self, other)
+
+    def cast_python_value(self, value):
+        try:
+            import ml_dtypes  # noqa: F401
+
+            return ml_dtypes.bfloat16(value)
```
kkraus14: Is there any way we could do this without the `ml_dtypes` dependency? It depends on NumPy and is relatively aggressive about depending on newer versions of NumPy that our numba-cuda users have expressed causing difficulties for them.

isVoid: `ml_dtypes.bfloat16` is mainly used on host to support storage and casting functions. @ZzEeKkAa also has requested a native representation of bfloat16 in Numba, one that's also compatible with Numpy. I can look into implementing that as a follow up.

For this PR, can I only remove the ml_dtypes dependency in core code base, but rely on it at test time to compare it with the cast result on host?

isVoid: Discussed offline - it's ok to use `ml_dtypes` as optional dependency for core lib. And we can use `ml_dtypes` as hard dependency in test. I plan to add host side bfloat support in a follow up PR.

PR #377, opened 2025-08-06
Vendor-in IR utilities for CUDA-specific refactoring

ashermancinelli:
Vendor in the numba core IR utilities. There were several unused variables in this component; I removed them as I found them, and when they were the result of a function call, the function call was preserved so I believe the original behavior has been preserved.

NOTE: A significant portion of this module is covered by neither our numba-cuda tests nor the upstream `test_ir_utils.py` test. We may be able to trim this module significantly.
NOTE: the original copyright in ir_utils.py has been preserved.

Numba commit 262b49a3e was used.

### DIFF
--- a/numba_cuda/numba/cuda/core/interpreter.py
+++ b/numba_cuda/numba/cuda/core/interpreter.py
@@ -11,7 +11,7 @@ from numba.core.errors import (
     UnsupportedBytecodeError,
     error_extras,
 )
-from numba.cuda.core.ir_utils import get_definition, guard
+from numba.cuda.core import ir_utils
 from numba.core.utils import (
     PYVERSION,
     BINOPS_TO_OPERATORS,
@@ -1296,10 +1296,10 @@ def _build_new_build_map(func_ir, name, old_body, old_lineno, new_items):
     values = []
     for pair in new_items:
         k, v = pair
-        key_def = guard(get_definition, func_ir, k)
+        key_def = ir_utils.guard(ir_utils.get_definition, func_ir, k)
         if isinstance(key_def, (ir.Const, ir.Global, ir.FreeVar)):
             literal_keys.append(key_def.value)
-        value_def = guard(get_definition, func_ir, v)
+        value_def = ir_utils.guard(ir_utils.get_definition, func_ir, v)
         if isinstance(value_def, (ir.Const, ir.Global, ir.FreeVar)):
             values.append(value_def.value)
         else:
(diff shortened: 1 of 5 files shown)


ashermancinelli:
/ok to test [618cf36](https://github.com/NVIDIA/numba-cuda/pull/377/commits/618cf366f176eda0d864f53fa1d253fa637d607c)

ashermancinelli:
The CI failures appear to be unrelated.

numba_cuda/numba/cuda/core/interpreter.py
```diff
@@ -1296,10 +1296,10 @@ def _build_new_build_map(func_ir, name, old_body, old_lineno, new_items):
     values = []
     for pair in new_items:
         k, v = pair
-        key_def = guard(get_definition, func_ir, k)
+        key_def = ir_utils.guard(ir_utils.get_definition, func_ir, k)
```
gmarkall, line 1299: I'm curious - do our tests all pass if we use 

```suggestion
        key_def = func_ir.get_definition(k)
```

here (and similar for `value_def` below) instead?

ashermancinelli: I will check!

numba_cuda/numba/cuda/typing/templates.py
```diff
@@ -706,7 +707,7 @@ def generic(self, args, kws):
                 )
             )
             ir = PreLowerStripPhis()._strip_phi_nodes(ir)
-            ir._definitions = numba.core.ir_utils.build_definitions(ir.blocks)
+            ir._definitions = ir_utils.build_definitions(ir.blocks)
```
gmarkall, line 710: I wonder if we can just vendor the `build_definitions` function and `build_defs_extensions` dict from `ir_utils` into here?

numba_cuda/numba/cuda/core/interpreter.py
gmarkall: (If not, perhaps we can still do the above suggestion and catch a `KeyError` rather than keeping get_definition and guard)

gmarkall (commented):
Following conversation outside of this PR, there may be items that depend on ir_utils that we add later, so we should tidy up later rather than right now - ignore comments on the diff!

numba_cuda/numba/cuda/typing/templates.py
ashermancinelli: This is a good point - but other components in core numba use these utilities (and others) so I wonder if we ought to clean out dead code once we have already vendored-in what we need, and we can look at the total coverage report. This way we do not need to revisit components.

gmarkall (approved):

PR #378, opened 2025-08-06
[Refactor] Vendor in cpython math impls for CUDA-specific customization

atmnp:
This moves the cpython func impls (cmathimpl, mathimpl, and numbers) used in our ufunc_db into this repo. I no longer think that this needs additional comprehensive testing because the vast majority if not all of the impls are already tested within test_ufuncs.py, similar to https://github.com/NVIDIA/numba-cuda/pull/374. 

I've removed the registry re-definitions like we did for typing templates, and left identical comments (with the correct registry names). For now, these impls are registering into the `cudaimpl` registry.


numba_cuda/numba/cuda/target.py
```diff
@@ -140,7 +140,7 @@ def init(self):
         self._target_data = None
 
     def load_additional_registries(self):
-        # side effect of import needed for numba.cpython.*, the builtins
+        # side effect of import needed for numba.cpython.*, numba.cuda.cpython.*, the builtins
         # registry is updated at import time.
         from numba.cpython import numbers, tupleobj, slicing  # noqa: F401
```
gmarkall: Should any of these imports change?

numba_cuda/numba/cuda/cpython/cmathimpl.py
```diff
@@ -0,0 +1,566 @@
...
+"""
+
+import cmath
+import math
+
+from numba.core.imputils import impl_ret_untracked
+from numba.core import types
+from numba.core.typing import signature
+from numba.cuda.cpython import mathimpl
+from numba.core.extending import overload
```
gmarkall, line 15: I think the uses of overload should be modified to be

```python
@overload(func, target="cuda")
```

(i.e. make them CUDA target-specific)

so that we're not adding duplicate overloads for the generic target in Numba.

numba_cuda/numba/cuda/cpython/cmathimpl.py
```diff
@@ -0,0 +1,566 @@
...
+# ---------------------------------------------------------------------
+# XXX: In upstream Numba, this file would create a cmathimpl registry
+# if it was installed in the target (as it is for the CUDA target).
+# The cmathimpl registry has been removed from this file (it was
+# initialized as `registry = Registry('cmathimpl')`) as it would duplicate
+# the cmathimpl registry in upstream Numba, which would be likely to lead
+# to confusion / mixing things up between two cmathimpl registries. The
+# comment that accompanied this behaviour is left here, even though the
+# code that would pick the cmathimpl registry has been removed, for the
+# benefit of future understanding.
```
gmarkall: I'm not sure if this comment is incorrect, or if I'm just not understanding it correctly. I don't see why creating a new registry here, and then installing it into the CUDA target would cause problems. Did you try this? If so, what issue do you run into?

numba_cuda/numba/cuda/cpython/numbers.py
```diff
@@ -0,0 +1,1471 @@
+import math
+import numbers
+
+import numpy as np
+import operator
+
+from llvmlite import ir
+from llvmlite.ir import Constant
+
+from numba.core.imputils import (
```
gmarkall: By importing and using `lower_*` from Numba core, we are re-registering duplicate implementations in the core of Numba. I think we should be using our own registry (that we can create in this module) here.

gmarkall (changes requested):
Having looked over the vendored code, the implementations are all fine to be vendored over. However, I think we should have a registry for each module, and then install each registry in the CUDA target instead of using upstream registries, or adding them directly to the `cudaimpl` registry.

gmarkall:
@atmnp Just to check, are you waiting on me to look at this? Or still planning to make the changes to add back the registries?

numba_cuda/numba/cuda/cpython/cmathimpl.py
```diff
@@ -8,24 +8,16 @@
 import cmath
 import math
 
-from numba.core.imputils import impl_ret_untracked
+from numba.core.imputils import Registry
+from numba.cuda.cudaimpl import impl_ret_untracked
```
gmarkall: Putting `impl_ret_untracked` in `cudaimpl` is a bit confusing for later vendoring of `imputils`. It would be better to keep `impl_ret_untracked` being imported from `numba.core.imputils`, then when someone vendors `imputils` later, it will be more straightforward for them.

PR #380, opened 2025-08-06
[Refactor][NFC] Vendor in CloudPickler

atmnp:
This has been modified to be merged with our NumbaPickler to further simplify the implementation.

### DIFF
--- a/numba_cuda/numba/cuda/deviceufunc.py
+++ b/numba_cuda/numba/cuda/deviceufunc.py
@@ -10,14 +10,78 @@ from collections import OrderedDict
 import operator
 import warnings
 from functools import reduce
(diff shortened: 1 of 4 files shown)


gmarkall (changes requested):
It's quite hard to figure out the level of risk in the changes because of everything being squashed into one commit. I'm also not sure of the value of trying to make this simplification, and I think it's going to be hard to think about / understand.

I think it'd be better to get this in quicker rather than try to change the way it works, as I believe this is only really essential for supporting pickling of object mode code, which is not supported in CUDA. If just a simple vendoring without other code changes works, can you please change this to just vendor the code without the refactoring?

PR #381, opened 2025-08-06
Add locals attribute to _FunctionCompiler

VijayKandiah:
We had initially removed the `locals` attribute from `_FunctionCompiler` while vendoring it in for CUDA-Specific changes in https://github.com/NVIDIA/numba-cuda/pull/338. This seemed to have caused `AttributeError` for a nvmath-python test https://github.com/NVIDIA/nvmath-python/blob/main/tests/nvmath_tests/device/test_cublasdx_numba.py. This PR restores this `locals` attribute and initializes it to `{}`


ZzEeKkAa:
Thank you! This is the root problem that I've faced:
```python
import numpy as np
from numba import cuda

cond = None

@cuda.jit
def f(b_global):
    b_smem = cuda.shared.array(shape=(1,), dtype=np.float32)

    if cond:
        b_smem[0] = b_global[0]

b = np.zeros(1, np.float32)

b_d = cuda.to_device(b)

f[1, 1](b_d)
```
I guess, what triggered it is a cast from `None` to `bool` type while taking it from freevar...
May be we can add it as a test?

Issue #375, opened 2025-08-05
[BUG] `Device.name` is a `bytes` with the `ctypes` binding, but a `str` with the CUDA-Python bindings


gmarkall:
**Describe the bug**

Noted in a comment on #235 - `Device.name` is a `bytes` with the ctypes binding, but a string with the CUDA-Python binding.

**Steps/Code to reproduce bug**

```
$ python -c "from numba import cuda; print(cuda.get_current_device().name)"
NVIDIA RTX 3500 Ada Generation Laptop GPU                                                                

$ NUMBA_CUDA_USE_NVIDIA_BINDING=0 python -c "from numba import cuda; print(cuda.get_current_device().name)"
b'NVIDIA RTX 3500 Ada Generation Laptop GPU'
```

**Expected behavior**

The type of `name` should be consistent across both bindings. I'm not sure whether we should try to make `name` a `bytes` to follow the old behaviour, or whether it's better as a string though.

**Additional context**

cc @marioroy

brandon-b-miller:
Happy to come up with a fix for this 👍

PR #382, opened 2025-08-06
Make device name consistent between bindings

brandon-b-miller:
Closes https://github.com/NVIDIA/numba-cuda/issues/375


brandon-b-miller:
> This fix is confusing -- shouldn't we move `decode` to the other (ctypes) branch, and always return a str?

I went back and forth on this and decided the safest thing to do is to change it so that both branches produce the old (ctypes) behavior which is to return the result as a bytes object up to the first null byte. I think we maybe ought to change it to be a string in the future regardless of binding but at least this PR ameliorates the original bug report.

kkraus14 (approved):

PR #383, opened 2025-08-06
Remove _helperlib dep., implement attempt_nocopy_reshape in python for CUDA-specific changes

VijayKandiah:
We are only using the method `attempt_nocopy_reshape` from numba's `_helperlib` C extension. This PR introduces a python implementation of attempt_nocopy_reshape for CUDA-specific customizations and to remove dependency on `_helperlib`.


gmarkall (commented):
A quick note (not got really round to reviewing this yet) - `numba_attempt_nocopy_reshape` is already implemented in `reshape_funcs.cu`. There may be some merit to having the implementation in Python, but I think there is also the risk of it being less efficient in terms of register usage as I would guess it would use 64-bit ints in places where the `.cu` version uses `int` types.

PR #385, opened 2025-08-07
Fix CI for CUDA 13 drivers

gmarkall:
Once this is passing it should be no longer WIP. Not sure if more changes need adding yet,


gmarkall:
I'm going to merge this, though it does contain a few hacks to get things moving so I'm tracking these in #386.

PR #387, opened 2025-08-07
Require `cuda-bindings` `12.9.1`

brandon-b-miller:
`cuda-bindings` `12.9.1` is built without monitoring, which avoids https://github.com/cython/cython/issues/7050. This makes it the only safe version since the nvidia binding is used now by default. This PR requires the patched packages.


bdice (approved):

leofang:
~~I am not sure this makes sense. We don't want to pin foundational packages over tightly. cuda-bindings has been built with the profiling support since day 1 and it's been working fine, and we aren't able to reproduce the issue outside of a very specific setup. Can we move this pin to RAPIDS stack instead?~~

NVM I misread the diff, sorry

conda/recipes/numba-cuda/meta.yaml
```diff
@@ -28,7 +28,7 @@ requirements:
   run:
     - python
     - numba >=0.59.1
-    - cuda-bindings
+    - cuda-bindings >= 12.9.1
```
leofang:
```suggestion
    - cuda-bindings >=12.9.1
```

leofang:
/ok to test [b9fa346](https://github.com/NVIDIA/numba-cuda/pull/387/commits/b9fa346508671e0b82cc534f8e759364f8850134)

leofang:
I am not sure why all conda tests failed with `ModuleNotFoundError: No module named 'numba.cuda.typing'` 🤔

brandon-b-miller:
> I am not sure why all conda tests failed with `ModuleNotFoundError: No module named 'numba.cuda.typing'` 🤔

This is weird. Trying to repro locally

brandon-b-miller:
I'm not exactly sure what's wrong here but an observation I made is that the conda test script doesn't seem to be installing the locally built version of numba-cuda, instead testing a version it's pulling from `conda-forge`. 

Bad run:
```
numba-cuda     0.18.1  pyh7ef8ed5_0          conda-forge     470kB
```
Good run:
```
numba-cuda     0.18.0  py_0                  /__w/numba-cuda/numba-cuda/conda-repo     509kB
```

conda/recipes/numba-cuda/meta.yaml
```diff
@@ -28,7 +28,7 @@ requirements:
   run:
     - python
     - numba >=0.59.1
-    - cuda-bindings
+    - cuda-bindings >=12.9.1
```
bdice, line 31: At some point soon, we will probably need to produce two builds, one with `cuda-version>=12.0,<13.0a0` and another with `cuda-version>=13.0,<14.0a0`. That will be needed in order to have separate bounds on `cuda-bindings` for each CUDA major. This may intersect with #370.

bdice:
@brandon-b-miller I think this PR is fine but you'll need the changes from #370 that drop CUDA 11. The failed jobs all use CUDA 11, and the requirement for `cuda-bindings>=12.9.1` won't be satisfiable on CUDA 11. Maybe just an oversight?

Should we merge this into #370?

brandon-b-miller:
@bdice at the very least I am inclined to wait until CUDA 11 is dropped before spending much more time chasing the cause here.

ZzEeKkAa:
Does `cuda-bindings` 12.9.1 suppose to work with all and only all ctk 12.* versions?

brandon-b-miller:
> Does `cuda-bindings` 12.9.1 suppose to work with all and only all ctk 12.* versions?

I believe so, from https://github.com/NVIDIA/cuda-python/blob/main/cuda_bindings/docs/source/support.rst, 
```
cuda.bindings 12.x supports any Toolkit 12.y
```

conda/recipes/numba-cuda/meta.yaml
kkraus14: In theory we could do something like adding `cuda-bindings != 13.0` and `cuda-bindings !=13.1` to require `cuda-bindings 13.2+` in the future.

kkraus14 (approved):

conda/recipes/numba-cuda/meta.yaml
bdice: That is possible but the mechanics of such a selector become unwieldy for both human readers/writers and environment solvers pretty fast...

PR #391, opened 2025-08-08
Vendor in bytecode.py for CUDA-specific refactoring

ashermancinelli:
Vendors in the bytecode module for CUDA-specific refactoring.  I am working towards vendoring in ir.py, but this has many dependencies and many modules use isinstance checks on classes they import from numba.core.ir directly, meaning these _users_ of numba.core.ir will all need to be vendored in before numba.core.ir itself can be.

NOTE: This module was adapted to work with Python 3.9 again. I looked at the difference between the current numba development branch and [numba commit 2c9f1caa3](https://github.com/numba/numba/commit/2c9f1caa3df941bf7066f7e977a18e122d363d42). For the most part, Python 3.9 took the same branches as 3.10, so that is what I updated. The primary difference was in `get_jump_target`, because Python 3.10 changed the addressing of jump instructions.

```
commit 2c9f1caa3df941bf7066f7e977a18e122d363d42
Commit:     kc611 <ckaustubhm06@gmail.com>
CommitDate: Wed Sep 18 17:43:06 2024 +0530

    Removed references to Python 3.9 within core
```

I intend to do the same thing with interpreter.py. bytecode.py is imported from a few places, so it was easier to make this change in the initial commit than guard the imports at the import site.

bytecode.py has few dependencies and is not directly tested.

### DIFF
--- a/numba_cuda/numba/cuda/core/bytecode.py
+++ b/numba_cuda/numba/cuda/core/bytecode.py
@@ -1,0 +1,711 @@
+from collections import namedtuple, OrderedDict
+import dis
+import inspect
+import itertools
+
+from types import CodeType, ModuleType
+
+from numba.core import errors, utils, serialize
+from numba.core.utils import PYVERSION
+
+
+if PYVERSION in ((3, 12), (3, 13)):
+    from opcode import _inline_cache_entries
+
+    # Instruction/opcode length in bytes
+    INSTR_LEN = 2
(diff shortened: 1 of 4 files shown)


ashermancinelli:
/ok to test [28bfb7e](https://github.com/NVIDIA/numba-cuda/pull/391/commits/28bfb7e614c16004c867ced4dd015a57b64b5f74)

ashermancinelli:
This is blocking the vendoring of ir.

PR #392, opened 2025-08-08
Backport interpreter.py for Python 3.9

ashermancinelli:
Adds Python 3.9 to the supported versions in interpreter.py.

Similar to #391, I looked at the diff between the current numba development branch and this commit to find the biggest differences:
```
commit 2c9f1caa3df941bf7066f7e977a18e122d363d42
Commit:     kc611 <ckaustubhm06@gmail.com>
CommitDate: Wed Sep 18 17:43:06 2024 +0530

    Removed references to Python 3.9 within core
```

There were not many; for the most part, Python 3.9 is handled the same as 3.10.

NOTE: I am not concerned about subclassing issues from vendoring in `UnsupportedBytecodeError` based on what I see in numba-cuda because I do not see any uses of `self.assertRaises(UnsupportedBytecodeError)` or explicit catch blocks, which would need to be updated.


ashermancinelli:
/ok to test [b897f63](https://github.com/NVIDIA/numba-cuda/pull/392/commits/b897f63c2cd6ace935bc4d1c6835c3cacca30447)

numba_cuda/numba/cuda/core/interpreter.py
```diff
@@ -1437,7 +1440,7 @@ def interpret(self, bytecode):
         if PYVERSION in ((3, 10), (3, 11), (3, 12), (3, 13)):
             peepholes.append(peep_hole_list_to_tuple)
         peepholes.append(peep_hole_delete_with_exit)
-        if PYVERSION in ((3, 10), (3, 11), (3, 12), (3, 13)):
+        if PYVERSION in ((3, 9), (3, 10), (3, 11), (3, 12), (3, 13)):
```
gmarkall: I think maybe this should have not been added here, but the check above - it seems that `peep_hole_list_to_tuple` is the peephole that pertains also to 3.9.

For reference, I am looking at the interpreter.py implementation from Numba 0.60.0:

https://github.com/numba/numba/blob/53e976f1b0c6683933fa0a93738362914bffc1cd/numba/core/interpreter.py#L1421-L1432

ashermancinelli: Great catch, thank you for finding this!

ashermancinelli:
/ok to test [b16f919](https://github.com/NVIDIA/numba-cuda/pull/392/commits/b16f919cab58d053a4ff188f2b165527a9536948)

gmarkall (approved):

PR #393, opened 2025-08-08
[Refactor][NFC] Vendor in ufunc utility functions, classes for CUDA-specific optimizations

atmnp:


numba_cuda/numba/cuda/np/ufunc/ufuncbuilder.py
kkraus14, line 1: Could you add the spdx license and copyright identifiers as have been added in other files recently?

gmarkall (approved):

PR #395, opened 2025-08-11
Fix package clobbering in thirdparty CI tests

brandon-b-miller:
In our thirdparty CI tests, installation of the cuDF package is clobbering the branch version of numba-cuda we actually want to test. Ran into this when cuDF failed to import locally using `main` and trying to track down why it didn't show up in CI.

This PR will include fixes for things that are broken, currently cuDF fails to import due to an instance check against upstream numba's `AttributeTemplate` class.


brandon-b-miller:
The tests now pass here.

kkraus14 (approved):

PR #399, opened 2025-08-13
[Refactor] Vendor in typing.npydecl, typing.BaseContext, typing.Context for CUDA-specific changes

VijayKandiah:
This PR vendors in few more classes (`typing.npydecl, typing.BaseContext, typing.Context`) from `numba.core.typing` for CUDA-specific customization.

### DIFF
--- a/numba_cuda/numba/cuda/core/ir_utils.py
+++ b/numba_cuda/numba/cuda/core/ir_utils.py
@@ -12,7 +12,8 @@ import warnings
 
 import numba
 from numba.core.extending import _Intrinsic
-from numba.core import types, typing, ir, analysis, postproc, rewrites, config
+from numba.core import types, ir, analysis, postproc, rewrites, config
+from numba.cuda import typing
 from numba.core.typing.templates import signature
 from numba.core.analysis import (
     compute_live_map,
--- a/numba_cuda/numba/cuda/typing/npydecl.py
+++ b/numba_cuda/numba/cuda/typing/npydecl.py
@@ -1,0 +1,657 @@
+import numpy as np
+import operator
+
+from numba.cuda.typing.templates import AbstractTemplate, Registry, signature
+from numba.core import types
+from numba.cuda import utils
+from numba.core.errors import TypingError, NumbaTypeError
+from numba import pndindex
+from numba.np.numpy_support import (
+    ufunc_find_matching_loop,
+    supported_ufunc_loop,
+    from_dtype,
+    as_dtype,
+    resolve_output_type,
+    _ufunc_loop_sig,
+)
+
+registry = Registry()
+infer = registry.register
+infer_global = registry.register_global
+infer_getattr = registry.register_attr
+
+
+class Numpy_rules_ufunc(AbstractTemplate):
+    @classmethod
+    def _handle_inputs(cls, ufunc, args, kws):
+        """
+        Process argument types to a given *ufunc*.
+        Returns a (base types, explicit outputs, ndims, layout) tuple where:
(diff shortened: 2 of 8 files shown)


numba_cuda/numba/cuda/core/ir_utils.py
```diff
@@ -12,7 +12,8 @@
 
 import numba
 from numba.core.extending import _Intrinsic
-from numba.core import types, typing, ir, analysis, postproc, rewrites, config
+from numba.core import types, ir, analysis, postproc, rewrites, config
+from numba.cuda import typing
```
brandon-b-miller, line 15: Are we in a position where this change could be made across the codebase? Or does this PR change only the instances that are now supported alongside the changes in this PR?

brandon-b-miller:
@gmarkall it looks like this PR may vendor some of the machinery around the numpy API that historically we haven't supported within kernels, but maybe now could with memory management. Should we consider looking into that again?

brandon-b-miller (approved):

numba_cuda/numba/cuda/core/ir_utils.py
VijayKandiah: We are not in a position to make the change `numba.core.typing` --> `numba.cuda.typing` across the entire codebase yet. There are still some remaining components to be vendored in `numba.cuda.typing` before we can make that change.

numba_cuda/numba/cuda/typing/npydecl.py
```diff
@@ -0,0 +1,657 @@
+import numpy as np
+import operator
+
+from numba.cuda.typing.templates import AbstractTemplate, Registry, signature
+from numba.core import types
+from numba.cuda import utils
+from numba.core.errors import TypingError, NumbaTypeError
+from numba import pndindex
+from numba.np.numpy_support import (
```
gmarkall: We have `numba.cuda.np.numpy_support` now, so I think this can be switched.

gmarkall: (If there are some missing functions in it, then please do add them in if there's not too much to add)

numba_cuda/numba/cuda/typing/npydecl.py
```diff
@@ -0,0 +1,657 @@
+import numpy as np
+import operator
+
+from numba.cuda.typing.templates import AbstractTemplate, Registry, signature
+from numba.core import types
+from numba.cuda import utils
+from numba.core.errors import TypingError, NumbaTypeError
+from numba import pndindex
```
gmarkall: We don't need this - it was specific to the parfors target.
```suggestion
```

numba_cuda/numba/cuda/typing/npydecl.py
```diff
@@ -0,0 +1,657 @@
...
+            if not arrays:
+                return
+            arrays = list(arrays)
+        else:
+            arrays = [arrays]
+        nditerty = types.NumpyNdIterType(arrays)
+        return signature(nditerty, *args)
+
+
+@infer_global(pndindex)
```
gmarkall:
```suggestion
```

numba_cuda/numba/cuda/typing/npydecl.py
VijayKandiah: Added the needed functions to `numba.cuda.np.numpy_support`. Did not vendor in `numba.np.ufunc_db` that is used by `numba.cuda.np.numpy_support.supported_ufunc_loop` though. I feel that should be part of a separate PR vendoring in `numba.np` ufuncs.

Issue #151, opened 2025-03-10
[FEA] Make `cuda.core.Stream` recognized by numba-cuda by supporting the `__cuda_stream__` protocol


leofang:
**Is your feature request related to a problem? Please describe.**

I want to use `cuda.core.Stream` for
- kernel launch
- data copy
- ...

xref: https://nvidia.github.io/cuda-python/cuda-core/latest/interoperability.html#cuda-stream-protocol

**Describe the solution you'd like**

**Describe alternatives you've considered**

N/A

**Additional context**

N/A

brandon-b-miller:
If all we want is what's described in the OP, the following code

```python

from numba import cuda
from cuda.core.experimental import Stream
from cuda.core.experimental import Device

dev = Device()
dev.set_current()
stream = dev.create_stream()
stream2 = dev.create_stream()

@cuda.jit
def kernel():
    pass


kernel[1,1, stream]()
kernel[1,1, stream2]()

devary = cuda.to_device([1, 2, 3, 4, 5], stream=stream2)
print(devary.copy_to_host(stream=stream2))

```
```
[1 2 3 4 5] 
```

Works with just this patch:

<details>

```
diff --git a/numba_cuda/numba/cuda/cudadrv/driver.py b/numba_cuda/numba/cuda/cudadrv/driver.py
index 6d3f307..b5a7c34 100644
--- a/numba_cuda/numba/cuda/cudadrv/driver.py
+++ b/numba_cuda/numba/cuda/cudadrv/driver.py
@@ -71,6 +71,7 @@ USE_NV_BINDING = config.CUDA_USE_NVIDIA_BINDING
 if USE_NV_BINDING:
     from cuda.bindings import driver as binding
 
+
     # There is no definition of the default stream in the Nvidia bindings (nor
     # is there at the C/C++ level), so we define it here so we don't need to
     # use a magic number 0 in places where we want the default stream.
@@ -3504,7 +3505,8 @@ def host_to_device(dst, src, size, stream=0):
     varargs = []
 
     if stream:
-        assert isinstance(stream, Stream)
+        if not USE_NV_BINDING:
+            assert isinstance(stream, Stream)
         fn = driver.cuMemcpyHtoDAsync
         varargs.append(stream.handle)
     else:
@@ -3522,7 +3524,8 @@ def device_to_host(dst, src, size, stream=0):
     varargs = []
 
     if stream:
-        assert isinstance(stream, Stream)
[...]

PR #401, opened 2025-08-18
Handle `cuda.core.Stream` in driver operations

brandon-b-miller:
Closes https://github.com/NVIDIA/numba-cuda/issues/151

### DIFF
--- a/numba_cuda/numba/cuda/tests/cudadrv/test_events.py
+++ b/numba_cuda/numba/cuda/tests/cudadrv/test_events.py
@@ -5,6 +5,7 @@ import numpy as np
 from numba import cuda
 from numba.cuda.testing import unittest, CUDATestCase
 from cuda.core.experimental import Device
+from numba.cuda.testing import skip_on_cudasim
 
 
 class TestCudaEvent(CUDATestCase):
@@ -26,6 +27,7 @@ class TestCudaEvent(CUDATestCase):
         stream = cuda.stream()
         self.event_elapsed_inner(stream)
 
+    @skip_on_cudasim("Testing cuda.core events requires driver")
     def test_event_elapsed_cuda_core_stream(self):
         dev = Device()
         dev.set_current()


numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -3648,3 +3636,18 @@ def inspect_obj_content(objpath: str):
...
     return code_types
+
+
+def _stream_handle(stream):
+    if stream == 0:
+        return stream
+    elif hasattr(stream, "__cuda_stream__"):
+        _, ptr = stream.__cuda_stream__()
+        if isinstance(stream, experimental.Stream):
+            return int(ptr)
```
brandon-b-miller: https://nvidia.github.io/cuda-python/cuda-core/latest/interoperability.html#cuda-stream-protocol says 
```
both as Python int
```
but I observe the protocol to return a `<class 'cuda.bindings.driver.CUstream'>` object for the second element here.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -58,11 +58,7 @@
 
 if USE_NV_BINDING:
     from cuda.bindings import driver as binding
-    from cuda.core.experimental import (
-        Linker,
-        LinkerOptions,
-        ObjectCode,
-    )
+    from cuda.core import experimental
```
kkraus14: nitpick: but importing this as experimental makes reading downstream code a bit confusing

numba_cuda/numba/cuda/cudadrv/driver.py
isVoid: Since we have `cuda.bindings.utils.get_cuda_native_handle`, shouldn't we use that helper instead of __int__?

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2337,6 +2333,11 @@ def __int__(self):
         # The default stream's handle.value is 0, which gives `None`
         return self.handle.value or drvapi.CU_STREAM_DEFAULT
 
+    def __cuda_stream__(self):
+        if not self.handle.value:
+            return (0, drvapi.CU_STREAM_DEFAULT)
+        return 0, self.handle.value if USE_NV_BINDING else self.handle
```
kkraus14: nitpick:
```suggestion
        return (0, self.handle.value if USE_NV_BINDING else self.handle)
```

numba_cuda/numba/cuda/cudadrv/driver.py
isVoid: +1. I think we want to promote canonical `from cuda.core.experimental import X` pattern. Also, when we eventually move things to non-experimental namespace, we only need 1 line change, not everywhere change.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -3639,3 +3624,20 @@ def inspect_obj_content(objpath: str):
             code_types.add(match.group(1))
 
     return code_types
+
+
+def _stream_handle(stream):
+    if stream == 0:
```
isVoid, line 3237: Not specific to this PR change - but what does "0" mean for a stream?

isVoid: A small docstring in this helper is helpful in explaining the supported stream objects in Numba.

isVoid (commented):
I see this PR closes #151, per issue suggests that we can pass a cuda core stream object via kernel launch interface, but this PR is missing a test for this use case.

numba_cuda/numba/cuda/cudadrv/driver.py
brandon-b-miller: Ah these are both good points. This was an attempt to disambiguate the numba-cuda and cuda-core `Stream` classes, but it seems like it makes things worse. I'll come up with something else for this.

numba_cuda/numba/cuda/cudadrv/driver.py
brandon-b-miller: This should just map to the default stream, I'll add docs

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -3541,3 +3530,26 @@ def inspect_obj_content(objpath: str):
...
+    """
+    Obtain the appropriate handle for various types of
+    acceptable stream objects. Acceptable types are
+    int (0 for default stream), Stream, ExperimentalStream
+    """
+
+    if stream == 0:
+        return stream
+    elif hasattr(stream, "__cuda_stream__"):
+        _, ptr = stream.__cuda_stream__()
```
leofang: always check the version, in case we go crazy and break it later
```suggestion
        ver, ptr = stream.__cuda_stream__()
        assert ver == 0
```

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -3541,3 +3530,26 @@ def inspect_obj_content(objpath: str):
...
+        if isinstance(ptr, binding.CUstream):
+            return get_cuda_native_handle(ptr)
```
leofang, line 3248: this seems like a bug in cuda.core, I vaguely remember we discussed this somewhere, is there an issue created already?

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -3541,3 +3530,26 @@ def inspect_obj_content(objpath: str):
             code_types.add(match.group(1))
 
     return code_types
+
+
+def _stream_handle(stream):
+    """
+    Obtain the appropriate handle for various types of
+    acceptable stream objects. Acceptable types are
+    int (0 for default stream), Stream, ExperimentalStream
```
leofang, line 3234: Is the docstring outdated? int is currently not allowed

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -3420,67 +3422,54 @@ def device_to_host(dst, src, size, stream=0):
...
+    fn(*args)
 
 
 def device_memset(dst, val, size, stream=0):
-    """Memset on the device.
-    If stream is not zero, asynchronous mode is used.
+    """
+    Memset on the device.
+    If stream is 0, the call is synchronous.
+    If stream is a Stream object, asynchronous mode is used.
```
leofang, line 3140: There is a bug (or change or behavior) here and elsewhere. `stream` can be a `Stream` object from either numba-cuda or cuda.core, but still holds 0 (the default stream) under the hood. However, the call now becomes asynchronous (with respect to the host) instead of synchronous. Just wanted to call it out in case it was not the intention.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -3398,20 +3405,15 @@ def host_to_device(dst, src, size, stream=0):
...
-        varargs.append(handle)
-    else:
-        fn = driver.cuMemcpyHtoD
+        args += (_stream_handle(stream),)
 
-    fn(device_pointer(dst), host_pointer(src, readonly=True), size, *varargs)
+    fn(*args)
 
 
 def device_to_host(dst, src, size, stream=0):
```
leofang, line 3104: As mentioned below (or above), `stream` semantics is changed which probably has a bigger impact to this method, because the copy is now asynchronous and to access `src` on host a stream synchronization is needed.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2334,6 +2336,11 @@ def __int__(self):
         # The default stream's handle.value is 0, which gives `None`
         return self.handle.value or drvapi.CU_STREAM_DEFAULT
 
+    def __cuda_stream__(self):
+        if not self.handle.value:
+            return (0, drvapi.CU_STREAM_DEFAULT)
+        return (0, self.handle.value if USE_NV_BINDING else self.handle)
```
leofang: This PR now full relies on having this protocol, even internally. But my concern is removing the internal fast paths we used to have could hurt the perf. I haven't caught up with the earlier discussions, not sure how much perf change this would do & if this is acceptable.

numba_cuda/numba/cuda/tests/cudadrv/test_cuda_driver.py
```diff
@@ -154,6 +158,65 @@ def test_cuda_driver_stream_operations(self):
...
+
+        ary = cuda.to_device([0] * 100)
+
+        dev = Device()
+        dev.set_current()
+        stream = dev.create_stream()
+
+        kernel[1, 100, stream](ary)
+
+        result = ary.copy_to_host()
```
leofang: Shouldn't this copy be launched on `stream`?

numba_cuda/numba/cuda/cudadrv/driver.py
leofang: We've fixed this in v0.3.1 (as part of https://github.com/NVIDIA/cuda-python/pull/709).

numba_cuda/numba/cuda/cudadrv/driver.py
brandon-b-miller: It's slower yeah:
```python
from numba.cuda.cudadrv.driver import Device, _stream_handle

dev = Device(0)
ctx = dev.get_primary_context()
stream = ctx.get_default_stream()

def old(stream):
    return stream.handle.value

def new(stream):
    return _stream_handle(stream)
```

```
%timeit old(stream)
61.2 ns ± 0.662 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

```
%timeit new(stream)
238 ns ± 2.59 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

It's tough for me to say if this makes a material difference in the wild, but there should be a way of moving things around so that the new approach is just as fast as the old one for the internal objects.

brandon-b-miller: Adding the fastpath into the function directly gets us down to 
```
134 ns ± 1.11 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```
This might be close to function call overhead though

numba_cuda/numba/cuda/cudadrv/driver.py
brandon-b-miller: This is a really good catch. As a follow up to this, is the output here as expected, where `dev` is a `cuda.core.experimental.Device` for whom `set_current()` has been called? Should it not be `(0, 0)`?

```
>>> dev.default_stream.__cuda_stream__()
(0, 1)
```

I ask hoping there's a reliable way of detecting this situation based on the passed object.

numba_cuda/numba/cuda/cudadrv/driver.py
brandon-b-miller: Only for the special value 0 I believe.

numba_cuda/numba/cuda/tests/cudadrv/test_cuda_driver.py
```diff
@@ -154,6 +158,65 @@ def test_cuda_driver_stream_operations(self):
...
+
+    @unittest.skipIf(not _driver.USE_NV_BINDING, "NV binding not enabled")
+    def test_cuda_core_stream_launch_user_facing(self):
+        @cuda.jit
+        def kernel(a):
+            idx = cuda.grid(1)
+            if idx < len(a):
+                a[idx] = idx
+
+        ary = cuda.to_device([0] * 100)
```
isVoid: Same to @leofang [comment](https://github.com/NVIDIA/numba-cuda/pull/401/files/f3b07c0174e004cf9d1bdc820c04a9fa8b97c25e#r2302001671), h2d should also use the same stream.

numba_cuda/numba/cuda/tests/cudadrv/test_cuda_driver.py
isVoid: We'll also need `stream.sync()` afterwards because with stream kwarg the copy is async.

numba_cuda/numba/cuda/cudadrv/driver.py
brandon-b-miller: After a while searching around the codebase I concluded this was at least the original [intention](https://github.com/NVIDIA/numba-cuda/blob/391478e6aa6b814521bffe3699b1a450a48fed8b/numba_cuda/numba/cuda/cudadrv/devicearray.py#L250-L252), though these are really only used for the deprecated device array API: 

```
        If a CUDA ``stream`` is given, then the transfer will be made
        asynchronously as part as the given stream.  Otherwise, the transfer is
        synchronous: the function returns after the copy is finished.
```

So AFAICT this PR maintains the above behavior just with a new stream object. Ultimately though I'm not sure we should spend too much time thinking about it as these will be removed and users performing these types of memory transfers should use either cupy for a nice array API or `cuda.bindings` for full control of things like synchronization behavior.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -63,6 +63,15 @@
...
 )
 
+from cuda.bindings.utils import get_cuda_native_handle
+from cuda.core.experimental import (
+    Stream as ExperimentalStream,
+)
+
+# For backwards compatibility: indicate that the NVIDIA CUDA Python bindings are
+# in use. Older code checks this flag to branch on binding-specific behavior.
+USE_NV_BINDING = True
```
kkraus14: We have since removed this from the codebase so I think we need to remove this now.

brandon-b-miller: Thanks! removed.

numba_cuda/numba/cuda/cudadrv/driver.py
kkraus14: Should we consider deprecating allowing passing `0` as a Stream? The "default stream" is ambiguous in Python since PTDS is normally a host compile-time concept. We have an environment variable for controlling it in `cuda.bindings` / `cuda.core`: `CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM` which I think should be generally used.

It would be great if we could introduce a deprecation warning in some form to passing `0` as a Stream in user facing APIs.

brandon-b-miller: From the user perspective we're deprecating the apis fully in https://github.com/NVIDIA/numba-cuda/pull/546, so those should be gone entirely. But we should do a sweep and make sure we're being explicit with all our usages of streams internally.

kkraus14: Outside of the `DeviceNDArray` class, I think streams are accepted when launching kernels and using the Event APIs as well where we should properly handle there as well?

brandon-b-miller: launching is tested as part of this PR, events added in https://github.com/NVIDIA/numba-cuda/pull/401/commits/7df62cecf5a1a7e517c10e777bda2dfa52d5ffe9 though.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -64,6 +64,12 @@
     ObjectCode,
 )
 
+from cuda.bindings.utils import get_cuda_native_handle
```
leofang, line 67: @brandon-b-miller `get_cuda_native_handle` was not added until very recently (12.9.1), so this will break users who are using older `cuda-bindings` 12.x. I suggest this:
```python
try:
    from cuda.bindings.utils import get_cuda_native_handle
except ImportError:
    def get_cuda_native_handle(obj):
        return int(obj)
```

leofang: Tracked in https://github.com/NVIDIA/numba-cuda/issues/563

leofang:
Another question: @brandon-b-miller  IIRC we hit some test failures that blocked this PR from making progress. What was the fix that unblocked this?

brandon-b-miller:
> Another question: @brandon-b-miller IIRC we hit some test failures that blocked this PR from making progress. What was the fix that unblocked this?

It was the fact that elsewhere in the test suite we were blowing up all contexts out from under cuda-core, fixed by https://github.com/NVIDIA/numba-cuda/pull/507.

leofang:
Ahh yes, the context reset issue, thanks for reminder 🙏

PR #403, opened 2025-08-18
Vendor in UnsupportedBytecodeError

ashermancinelli:
Vendors in UnsuportedBytecodeError and changes imports to use other utilities that have already been vendored in. _lazy_pformat was also needed when building with an older numba, which was already vendored in, so the imports have been changed.

### DIFF
--- a/numba_cuda/numba/cuda/errors.py
+++ b/numba_cuda/numba/cuda/errors.py
@@ -12,6 +12,13 @@ class KernelRuntimeError(RuntimeError):
         super(KernelRuntimeError, self).__init__(msg)
 
 
+class UnsupportedBytecodeError(Exception):
+    """Unsupported bytecode is non-recoverable"""
+
+    def __init__(self, msg, loc=None):
+        super().__init__(f"{msg}. Raised from {loc}")
+
+
 class CudaLoweringError(LoweringError):
     pass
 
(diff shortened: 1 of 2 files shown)


ashermancinelli:
/ok to test [0021c0f](https://github.com/NVIDIA/numba-cuda/pull/403/commits/0021c0f4e857e4a83041ebd46bdb5ab167ff968f)

numba_cuda/numba/cuda/errors.py
```diff
@@ -12,6 +12,13 @@ def __init__(self, msg, tid=None, ctaid=None):
         super(KernelRuntimeError, self).__init__(msg)
 
 
+class UnsupportedBytecodeError(Exception):
```
bdice, line 15: Do we want to add a comment indicating this is vendored from `numba>=0.61.0` in order to support `numba==0.60.0`?

ashermancinelli: We can - we've vendored in lots of components, but I've only been indicating which Numba commit they come from when it's relevant for maintenance. For example, the current numba bytecode interpreter does not support Python 3.9, so I mentioned the commit I pulled it in from before making another commit to back-port it to 3.9. Because this class is just a thin wrapper and we would have vendored it in eventually anyways, I didn't see the need.

PR #406, opened 2025-08-19
Fix NRT stats tests when MVC is required and add them to thirdparty

brandon-b-miller:
This PR fixes an oversight where our memsys module isn't LTO'ed by default. This shakes out in cuDF's testing under certain MVC scenarios, and adds those tests here. 

We probably want to incorporate changes from https://github.com/rapidsai/cudf/pull/19730#pullrequestreview-3132253788 here too, so I'll wait to merge this.


brandon-b-miller:
Right this will only work if `nvjitlink` exists and is sufficiently recent. Will push some fixes.

kkraus14 (approved):

PR #407, opened 2025-08-19
Vendor-In find_lib for CUDA-specific changes

atmnp:
This vendors in the small utility class we use to find shared libraries. Also fixes some dangling imports caused by the recent vendoring PRs.


gmarkall:
Looks good, I guess it just needs the SPDX header adding.

PR #412, opened 2025-08-20
Add SPDX identifiers to all files

kkraus14:
Adds SPDX identifiers to all files. Converted existing Copyright statements to SPDX identifiers and added them where there was no existing Copyright statement.

A few things we should check before merging here:

- [x] SPDX identifiers were added to `.rst` files via block comments. Sphinx also seemingly uses comments for functionality, so we should make sure it doesn't break doc rendering and remove the SPDX identifiers if it does.
- [x] Added SPDX identifiers to the `.pth` redirector file. The Python documentation (https://docs.python.org/3/library/site.html) claims that "Blank lines and lines beginning with `#` are skipped" so this should be okay. I believe our CI suite should exercise this already, but if this is not the case we should explicitly check it.
- [x] There were a couple of places that files had Apache 2.0 licenses, presumably from being vendored from RAPIDS. I relicensed these files to BSD-2 to match the rest of the repo. There's no intellectual property concerns or patentability in these files.


kkraus14:
I downloaded the artifact from the docs build and confirmed locally that the docs all rendered and looked correct.

isVoid (approved):

PR #417, opened 2025-08-21
Vendor in SSA for CUDA-specific refactoring

ashermancinelli:
The tests had to undergo some significant refactoring - I tried to preserve as much of the original test coverage as I could, but here are notes on some tests that I had to remove entirely:
* Test test_unhandled_undefined was skipped because it relies on potentially undefined return values and we do not support return values at all.
* Test test_issue5482_objmode_expr_null_lowering was skipped because the object mode is not supported for the numba-cuda target.
* Several tests that used custom compilation pipelines were omitted because custom pipelines do not work with numba.cuda.jit. We may want to re-add these tests if we add this functionality, but I did not include custom pipelines in this PR to keep the focus on vendoring in this component and not feature work.

Otherwise, it was a 1-1 replacement of numba's jit and njit with numba.cuda.jit, and return values were replaced with assignments to output arrays.


ashermancinelli:
/ok to test [8331e9a](https://github.com/NVIDIA/numba-cuda/pull/417/commits/8331e9aefcc6c34e8fa52c4161b4fae345792434)

ashermancinelli:
/ok to test [09adca3](https://github.com/NVIDIA/numba-cuda/pull/417/commits/09adca317efd88d26688fac975aa38a5da783dde)

ashermancinelli:
/ok to test [75ef494](https://github.com/NVIDIA/numba-cuda/pull/417/commits/75ef4942814b7915d3d9532d18cca3012e6b4610)

numba_cuda/numba/cuda/tests/cudapy/test_ssa.py
```diff
@@ -0,0 +1,387 @@
...
+
+        lin = np.array([0.1, 0.6, 0.3])
+        self.check_func(foo, np.array([0.1]), np.zeros((2, 2)), lin)
+
+    def test_issue5482_missing_variable_init(self):
+        # Test error that lowering fails because variable is missing
+        # a definition before use.
+        @jit
+        def foo(result, x, v, n):
+            problematic = 0  # Initialize to avoid unbound variable
```
gmarkall: Is your type checker / linter telling you to do this? I think the point of this test is that it's not initialized on all branches.

```suggestion
```

ashermancinelli: You're right, I was mistaken to add this.

ashermancinelli:
/ok to test [348875e](https://github.com/NVIDIA/numba-cuda/pull/417/commits/348875e714b3c9825df4253080b091520532282f)

PR #418, opened 2025-08-22
Vendor in postproc module for CUDA-specific refactoring

ashermancinelli:
This module has few dependencies and no standalone tests, so it has been vendored in as-is (with the copyright modifications).

A few unrelated imports were importing modules from upstream numba that we have already vendored in, so I corrected them to use our copy. If we would like this PR to be strictly related to the postproc module, I can revert these changes.

### DIFF
--- a/numba_cuda/numba/cuda/compiler.py
+++ b/numba_cuda/numba/cuda/compiler.py
@@ -12,7 +12,6 @@ from numba.core import (
     funcdesc,
     config,
     bytecode,
-    postproc,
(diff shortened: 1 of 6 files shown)


ashermancinelli:
/ok to test [2ccc551](https://github.com/NVIDIA/numba-cuda/pull/418/commits/2ccc551b07f710a7a6b576288fcea32534ba00b9)

gmarkall (approved):

PR #419, opened 2025-08-22
Vendor in the consts module for CUDA-specific refactoring

ashermancinelli:
consts.py has few dependencies and no standalone tests that I could find, so it has simply been vendored in and the existing uses in numba-cuda have been updated.

I noticed untyped_passes was still conditionally importing the vendored-in interpreter.py based on the python version, so I updated that as well.

### DIFF
--- a/numba_cuda/numba/cuda/core/analysis.py
+++ b/numba_cuda/numba/cuda/core/analysis.py
@@ -3,7 +3,8 @@
 
 from collections import namedtuple
 from numba import types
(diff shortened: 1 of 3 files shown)


ashermancinelli:
/ok to test [e8f68f4](https://github.com/NVIDIA/numba-cuda/pull/419/commits/e8f68f436bb9567b5f3d98e6255346895be9b7a9)

PR #427, opened 2025-08-26
Compatibility fixes for Numba 0.62

gmarkall:
- `newpassmanagers` need to be used with llvmlite >= 0.45 (the legacy pass manager is removed with the move to LLVM 20)
- The `Array._allocate()` overload was made CPU-only with https://github.com/numba/numba/pull/10185, so we need one for the CUDA target.

### DIFF
--- a/numba_cuda/numba/cuda/core/typed_passes.py
+++ b/numba_cuda/numba/cuda/core/typed_passes.py
@@ -50,8 +50,13 @@ from numba.cuda.core.ir_utils import (
     replace_vars,
 )
 from numba.cuda.core import postproc
-from llvmlite import binding as llvm
 
+try:
+    # llvmlite < 0.45
+    from llvmlite.binding import passmanagers
+except ImportError:
+    # llvmlite >= 0.45
+    from llvmlite.binding import newpassmanagers as passmanagers
 
 # Outputs of type inference pass
 _TypingResults = namedtuple(
@@ -513,7 +518,7 @@ class BaseNativeLowering(abc.ABC, LoweringPass):
         calltypes = state.calltypes
         flags = state.flags
         metadata = state.metadata
-        pre_stats = llvm.passmanagers.dump_refprune_stats()
+        pre_stats = passmanagers.dump_refprune_stats()
 
         msg = "Function %s failed at nopython mode lowering" % (
             state.func_id.func_name,
@@ -576,7 +581,7 @@ class BaseNativeLowering(abc.ABC, LoweringPass):
                 )
 
             # capture pruning stats
-            post_stats = llvm.passmanagers.dump_refprune_stats()
+            post_stats = passmanagers.dump_refprune_stats()
             metadata["prune_stats"] = post_stats - pre_stats
 
             # Save the LLVM pass timings
--- a/numba_cuda/numba/cuda/memory_management/__init__.py
+++ b/numba_cuda/numba/cuda/memory_management/__init__.py
@@ -2,3 +2,7 @@
 # SPDX-License-Identifier: BSD-2-Clause
 
 from numba.cuda.memory_management.nrt import rtsys  # noqa: F401
+from numba.cuda.memory_management import arrayobj_extras  # noqa: F401
+
+
+__all__ = ["rtsys"]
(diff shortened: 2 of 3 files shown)


gmarkall:
@yashssh I think it's correct to just substitute the old for the new pass managers in this way - is that correct?

yashssh:
Yes, it looks correct. 

On a side note I can see numba-cuda has its own set of passes in `numba_cuda/numba/cuda/core/typed_passes.py` I am assuming they are independent of llvmlite's passmanager infra and managed separately?

gmarkall:
> On a side note I can see numba-cuda has its own set of passes in `numba_cuda/numba/cuda/core/typed_passes.py` I am assuming they are independent of llvmlite's passmanager infra and managed separately?

That's correct, these are passes operating on Numba IR. It looks similar to LLVM because LLVM's design inspired Numba's design in this area.

numba_cuda/numba/cuda/core/typed_passes.py
```diff
@@ -38,8 +38,13 @@
     is_operator_or_getitem,
 )
 from numba.cuda.core import postproc
-from llvmlite import binding as llvm
 
+try:
+    # llvmlite < 0.45
+    from llvmlite.binding import passmanagers
```
rparolin, line 43: The comments reflect behaviour based on `llvmlite` version but we attempt to load the legacy `passmanager` and if that fails we attempt to load the new passmanager.  Shouldn't that be the other way around?  We attempt to load the `newpassmanager` type and if it doesn't exist we fallback to the legacy `passmanger` type?

numba_cuda/numba/cuda/memory_management/nrt.py
```diff
@@ -42,6 +44,34 @@ def get_include():
...
+# Provide an implementation of Array._allocate() for the CUDA target (used
+# internally by Numba when generating the allocation of an array)
+
+
+@intrinsic
+def intrin_alloc(typingctx, allocsize, align):
+    """Intrinsic to call into the allocator for Array"""
+
+    def codegen(context, builder, signature, args):
+        [allocsize, align] = args
```
rparolin: `[allocsize, align] = args` I was concerned that this was create an unnecessary python list temporary but according to the documentation I've read this is equivalent to writing `allocsize, align = args`.  I wanted to highlight to confirm understanding.

rparolin:
I'm not sure if there are other tests covering the new allocation API we implemented in this PR.  Definitely something we should consider adding in a future PR is that isn't already covered.

numba_cuda/numba/cuda/core/typed_passes.py
gmarkall: Although it would seem cleaner / more intuitive to do it that way, we should not do that, because `newpassmanager` exists in older versions, but without the specific API we need - so it will fail if done that way round.

numba_cuda/numba/cuda/memory_management/nrt.py
gmarkall: That is correct - we can confirm something like this by inspecting the bytecode. For example:

```python
from dis import dis

def list_assignment(args):
    [a, b] = args
    return a + b

dis(list_assignment)
```

gives

```
  1           0 RESUME                   0

  2           2 LOAD_FAST                0 (args)
              4 UNPACK_SEQUENCE          2
              8 STORE_FAST               1 (a)
             10 STORE_FAST               2 (b)

  3          12 LOAD_FAST                1 (a)
             14 LOAD_FAST                2 (b)
             16 BINARY_OP                0 (+)
             20 RETURN_VALUE
```

So we see there was no list creation, it just unpacked `args` and stored them into `a` and `b`.

That said, IMHO this is not very good form and would be better without the square brackets - I'd just copy-pasted the code from upstream Numba without modification.

gmarkall:
> I'm not sure if there are other tests covering the new allocation API we implemented in this PR. Definitely something we should consider adding in a future PR is that isn't already covered.

These are covered by the NRT tests - anything that allocates an array fails without this PR with Numba 0.62.

numba_cuda/numba/cuda/core/typed_passes.py
rparolin: Makes sense. Do you think that warrants a comment explaining it?  Or would that just be noise?

PR #428, opened 2025-08-26
Bfloat Test Expects Different PTX Based on CC

isVoid:
This PR amends the bfloat16 PTX assertions to consider different PTX for different compute capabilities.


gmarkall (approved):

gmarkall:
If this resolves the issue that @ashermancinelli saw on H100, then I think this will be good to merge.

ashermancinelli (approved):
Works for me on H100!

PR #429, opened 2025-08-26
Avoid Using Device Compute Capability for Linker Class

isVoid:
When debugging with @gmarkall, we discovered that the linker class is using the current device's compute capability for cuda source compilation. Because `compile_ptx` can be used
to compile target code for a different architecture from current device. This PR pass through the CC used to construct the linker class to the cuda source compilation. This avoids
CC inconsistencies between compile and linking.


brandon-b-miller (approved):
Ah, nice catch @isVoid and @gmarkall . Thanks for the patch.

kkraus14 (approved):

kkraus14:
Can a test be added for this that invokes this codepath in a way that it previously would have generated the current CC instead of the passed in CC?

isVoid:
> Can a test be added for this that invokes this codepath in a way that it previously would have generated the current CC instead of the passed in CC?

Just want to revise a section where I mentioned that `compile_ptx` can be used to generate code for a different CC. We actually don't have a public API that does compile+linking for a different CC in Numba-CUDA. When https://github.com/NVIDIA/numba-cuda/pull/428 genereted code for a different CC, we used `NUMBA_CUDA_FORCE_CC` and `NUMBA_DUMP_ASSEMBLY` to dump linked assembly that includes external functions. It is still a useful developer feature IMO, I can implement a test of above as-is and provide a documentation in FAQs, WDYT?

gmarkall:
> Just want to revise a section where I mentioned that `compile_ptx` can be used to generate code for a different CC. We actually don't have a public API that does compile+linking for a different CC in Numba-CUDA.

I had thought in the past about augmenting the compilation APIs we have like `compile()`, `compile_ptx()`, etc. with functions to link as well. At the time I decided not to add linking APIs because I felt it would just be duplicating the efforts of CUDA Python but with a disparate interface. Now that we have a lot of Numba extensions that rely on linking of external code in a way that is invisible to the user (e.g. using bfloat16 bindings via Numbast), it does make sense to have an API that returns linked code (either PTX generated using nvjitlink, or a cubin) because the user would have no way to reliably know what to link with code generated using `compile()`.

However, that is a big scope increase for this PR that essentially fixes a small error, so I wouldn't suggest we do that here.

> When #428 genereted code for a different CC, we used `NUMBA_CUDA_FORCE_CC` and `NUMBA_DUMP_ASSEMBLY` to dump linked assembly that includes external functions. It is still a useful developer feature IMO, I can implement a test of above as-is and provide a documentation in FAQs, WDYT?

I figured out the reason that `NUMBA_FORCE_CUDA_CC` exists - it's because Numba used to have a hardcoded list of supported compute capabilities. Newer CUDA toolkits might support compute capabilities that the current version of Numba did not know about, so it could be used to force the CC to be a later version (e.g. Numba only supporting CC 3.5, but a CUDA toolkit supporting 5.0). The `NUMBA_FORCE_CUDA_CC` option does not work if you force a CC later than the CC of the device you actually have (since you can't run code for a newer SM on older hardware). I don't think this is a good developer tool because it doesn't really work, and IIRC we didn't actually get the PTX for Hopper by using it, but instead by calling `kernel.overloads[kernel.signatures[0]]._codelibrary.get_cubin(cc=(9, 0))` with `NUMBA_DUMP_ASSEMBLY` set.

In conclusion - I'd rather get rid of `NUMBA_FORCE_CUDA_CC` (because it is broken) rather than starting to lean on it as test infrastructure. For adding a test for this fix - I think it would be sufficient to write a test case that creates a new `_Linker` object (using `Linker.new()`) with a CC that is different to the current one, then adding a `CUSource` to the link, and completing the link - this should trigger a mismatch of CC between the LTOIR from NVRTC and the CC the link is for, in the absence of the fix in this PR.

isVoid:
> I think it would be sufficient to write a test case that creates a new _Linker object (using Linker.new()) with a CC that is different to the current one, then adding a CUSource to the link, and completing the link - this should trigger a mismatch of CC between the LTOIR from NVRTC and the CC the link is for, in the absence of the fix in this PR.

Of course! Testing _Linker class is most straightforward. Implemented in [e23d9f1](https://github.com/NVIDIA/numba-cuda/pull/429/commits/e23d9f1cd3481c00ff53a588e28ce0f141c6cce6)

numba_cuda/numba/cuda/tests/cudadrv/test_linker.py
```diff
@@ -329,6 +329,20 @@ def test_get_local_mem_per_specialized(self):
         calc_size = np.dtype(np.float64).itemsize * LMEM_SIZE
         self.assertGreaterEqual(local_mem_size, calc_size)
 
+    def test_link_for_different_cc(self):
+        if not config.CUDA_USE_NVIDIA_BINDING:
+            self.skipTest("Ctypes Linker does not support outputting PTX.")
```
isVoid: I tried using cuobjdump to get PTX. But it requires the output to have proper fatbin binary headers. The output of cuLink is cubin binary and cannot result in PTX.

kkraus14: You can probably use https://github.com/NVIDIA/cuda-python/blob/main/cuda_bindings/cuda/bindings/utils/_ptx_utils.py#L94 against `ptx.code`?

numba_cuda/numba/cuda/tests/cudadrv/test_linker.py
```diff
@@ -329,6 +329,20 @@ def test_get_local_mem_per_specialized(self):
...
+
+        linker = _Linker.new(cc=(12, 0))
```
gmarkall: The error only occurs when the LTOIR CC is greater than the Linker's given CC, so this will never fail. We also need to turn LTO on to trigger the issue. On my system (CC 8.6), I can use

```suggestion
        linker = _Linker.new(cc=(7, 5), lto=True)
```

to trigger it.

numba_cuda/numba/cuda/tests/cudadrv/test_linker.py
```diff
@@ -329,6 +329,20 @@ def test_get_local_mem_per_specialized(self):
...
+        code = """
+__device__ int foo(int x) {
+    return x + 1;
+}
+"""
+        linker.add_cu(code, "foo")
+        ptx = linker.complete()
```
gmarkall: We can get the PTX from the linker like:

```suggestion
        ptx = linker.get_linked_ptx().decode()
```

numba_cuda/numba/cuda/tests/cudadrv/test_linker.py
```diff
@@ -329,6 +329,20 @@ def test_get_local_mem_per_specialized(self):
...
+        assert "target sm_120" in str(ptx.code)
```
gmarkall: This would need to match the SM from above (it does match the `(12, 0)` above at present but would need to change when you use an earlier CC)

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -3003,10 +3003,7 @@ def add_ptx(self, ptx, name="<cudapy-ptx>"):
         self._object_codes.append(obj)
 
     def add_cu(self, cu, name="<cudapy-cu>"):
-        with driver.get_active_context() as ac:
-            dev = driver.get_device(ac.devnum)
-            cc = dev.compute_capability
-        obj, log = nvrtc.compile(cu, name, cc, ltoir=self.lto)
+        obj, log = nvrtc.compile(cu, name, self.cc, ltoir=self.lto)
```
gmarkall, line 3003: The same mistake is present in the `add_cu` method of `_LinkerBase` as well.

numba_cuda/numba/cuda/tests/cudadrv/test_linker.py
gmarkall: It's not possible to run this test with the ctypes linker because it doesn't have a way to do LTO, and I don't think there's a way we could reasonably implement that and have it produce LTO-d PTX - so there's not much we can check here for the ctypes linker.

numba_cuda/numba/cuda/tests/cudadrv/test_linker.py
isVoid: Right - the cc argument is passed down to the linker, and this PR is to make sure the nvrtc compile is of the same version as the linker argument.

numba_cuda/numba/cuda/tests/cudadrv/test_linker.py
```diff
@@ -329,6 +329,20 @@ def test_get_local_mem_per_specialized(self):
...
+
+        linker = _Linker.new(cc=(7, 5), lto=True)
+        code = """
+__device__ int foo(int x) {
+    return x + 1;
+}
+"""
+        linker.add_cu(code, "foo")
+        ptx = linker.get_linked_ptx().decode()
+        assert "target sm_75" in str(ptx.code)
```
gmarkall: Sorry, forgot to suggest this should have been
```suggestion
        assert "target sm_75" in ptx
```

numba_cuda/numba/cuda/tests/cudadrv/test_linker.py
gmarkall: Seems like we also need to skip the test if the toolkit version is less than 12.3, because in that case the `cuda.core` linker does not use nvjitlink.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2836,10 +2836,7 @@ def add_ptx(self, ptx, name):
     def add_cu(self, cu, name):
         """Add CUDA source in a string to the link. The name of the source
         file should be specified in `name`."""
-        with driver.get_active_context() as ac:
-            dev = driver.get_device(ac.devnum)
-            cc = dev.compute_capability
-        ptx, log = nvrtc.compile(cu, name, cc)
+        ptx, log = nvrtc.compile(cu, name, self.cc)
```
gmarkall, line 2839: I think the CC is always specified when creating a linker now, so to fix the CI issue that resulted from this addition, I think adding `self.cc = cc` to `CtypesLinker.__init__()` should be safe.

isVoid:
As it turns out this becomes a precondition to the `compile_ptx` augmentation request because we need to make linker context independent for that PR.

gmarkall:
> As it turns out this becomes a precondition to the `compile_ptx` augmentation request because we need to make linker context independent for that PR.

I think that's not the case if we don't return something that's linked - as I understood it, we would return all the linker inputs and let the user link them themselves - one motivation for doing that instead is that we think it's not possible to provide a set of LTOIR inputs to nvjitlink and then get a single LTOIR output back out.

PR #432, opened 2025-08-27
[Refactor][NFC] Vendor in testing support utils for CUDA-specific changes

atmnp:
This change vendors in numba.testing.support, specifically the Mixins that we use for testing for memory leaks and checking warnings.


numba_cuda/numba/cuda/simulator/memory_management/nrt.py
```diff
@@ -3,7 +3,19 @@
 
 from numba import config
 
-rtsys = None
+
+class RTSys:
+    def __init__(self, *args, **kwargs):
+        pass
+
+    def memsys_enable_stats(self):
```
gmarkall, line 11: For future reference - It's a little non-obvious why this was needed but the reason is that the `BaseUFuncTest` switched to the Numba-CUDA `MemoryLeakMixin` which initializes the Numba-CUDA `rtsys`. Prior to this PR, it was erroneously using the CPU-based `MemoryLeakMixin`.

gmarkall (approved):

PR #433, opened 2025-08-27
[Refactor][NFC] Vendor-in core Numba analysis utils for CUDA-specific changes

atmnp:
There are a few core utilities from Numba that we use during analysis to compute use-def chains, live maps, dead maps, etc. This change vendors those in for CUDA-specific changes.

### DIFF
--- a/numba_cuda/numba/cuda/tests/test_analysis.py
+++ b/numba_cuda/numba/cuda/tests/test_analysis.py
@@ -19,7 +19,6 @@ from numba.cuda.tests.support import (
     TestCase,
     MemoryLeakMixin,
 )
-from numba.cuda.testing import skip_on_cudasim
 from numba.cuda.core.analysis import (
     dead_branch_prune,
     rewrite_semantic_constants,
@@ -27,11 +26,16 @@ from numba.cuda.core.analysis import (
 from numba.cuda.core.untyped_passes import (
     ReconstructSSA,
 )
+import unittest
+from numba.cuda.core import config
 
 _GLOBAL = 123
 
 enable_pyobj_flags = Flags()
 enable_pyobj_flags.enable_pyobject = True
+
+if config.ENABLE_CUDASIM:
+    raise unittest.SkipTest("Analysis passes not done in simulator")
 
 
 def compile_to_ir(func):
@@ -159,7 +163,6 @@ class TestBranchPruneBase(MemoryLeakMixin, TestCase):
 
 
 class TestBranchPrune(TestBranchPruneBase):
-    @skip_on_cudasim("Analysis passes not done in simulator")
     def test_single_if(self):
         def impl(x, res):
             if 1 == 0:
@@ -253,7 +256,6 @@ class TestBranchPrune(TestBranchPruneBase):
             10,
         )
 
-    @skip_on_cudasim("Analysis passes not done in simulator")
     def test_single_if_else(self):
         def impl(x, res):
             if x is None:
@@ -274,7 +276,6 @@ class TestBranchPrune(TestBranchPruneBase):
             10,
         )
 
-    @skip_on_cudasim("Analysis passes not done in simulator")
     def test_single_if_const_val(self):
         def impl(x, res):
             if x == 100:
@@ -311,7 +312,6 @@ class TestBranchPrune(TestBranchPruneBase):
             100,
         )
 
-    @skip_on_cudasim("Analysis passes not done in simulator")
     def test_single_if_else_two_const_val(self):
         def impl(x, y, res):
             if x == y:
@@ -358,7 +358,6 @@ class TestBranchPrune(TestBranchPruneBase):
             1000,
         )
 
-    @skip_on_cudasim("Analysis passes not done in simulator")
     def test_single_if_else_w_following_undetermined(self):
         def impl(x, res):
             x_is_none_work = False
@@ -425,7 +424,6 @@ class TestBranchPrune(TestBranchPruneBase):
             10,
         )
 
-    @skip_on_cudasim("Analysis passes not done in simulator")
     def test_double_if_else_rt_const(self):
         def impl(x, res):
             one_hundred = 100
@@ -445,7 +443,6 @@ class TestBranchPrune(TestBranchPruneBase):
         self.assert_prune(impl, (types.NoneType("none"),), [False, None], None)
         self.assert_prune(impl, (types.IntegerLiteral(10),), [True, None], 10)
 
-    @skip_on_cudasim("Analysis passes not done in simulator")
     def test_double_if_else_non_literal_const(self):
         def impl(x, res):
             one_hundred = 100
@@ -458,7 +455,6 @@ class TestBranchPrune(TestBranchPruneBase):
         self.assert_prune(impl, (types.IntegerLiteral(10),), [None], 10)
         self.assert_prune(impl, (types.IntegerLiteral(100),), [None], 100)
 
-    @skip_on_cudasim("Analysis passes not done in simulator")
     def test_single_two_branches_same_cond(self):
         def impl(x, res):
             if x is None:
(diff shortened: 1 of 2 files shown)


numba_cuda/numba/cuda/tests/support.py
```diff
@@ -817,3 +821,28 @@ def check_warnings(self, messages, category=RuntimeWarning):
...
                     found += 1
         self.assertEqual(found, len(messages))
+
+
+class SerialMixin(object):
+    """Mixin to mark test for serial execution."""
+
+    _numba_parallel_test_ = False
+
+
```
gmarkall: The `SerialMixin` carries no meaning in the Numba-CUDA testsuite (it only has an effect with the Numba test runner):
```suggestion
```

numba_cuda/numba/cuda/tests/support.py
```diff
@@ -817,3 +821,28 @@ def check_warnings(self, messages, category=RuntimeWarning):
...
+        )
+        # mangle the default pipeline and inject DCE and IR preservation ahead
+        # of legalisation
+
+        # TODO: add a way to not do this! un-finalizing is not a good idea
+        pipeline._finalized = False
+        pipeline.add_pass_after(PreserveIR, IRLegalization)
+
+        pipeline.finalize()
+        return [pipeline]
```
gmarkall: I think this is no longer used in this PR (and we would not be able to use it in Numba-CUDA tests anyway, because we don't support the `pipeline_class` kwarg):
```suggestion
```

numba_cuda/numba/cuda/tests/support.py
```diff
@@ -32,6 +32,10 @@
 )
 from numba.core.datamodel.models import OpaqueModel
 from numba.cuda.np import numpy_support
+from numba.cuda.core.compiler import CompilerBase
+from numba.cuda.compiler import DefaultPassBuilder
+from numba.cuda.core.untyped_passes import PreserveIR
+from numba.cuda.core.typed_passes import IRLegalization
```
gmarkall: These are no longer necessary with the changes suggested below:
```suggestion
```

numba_cuda/numba/cuda/tests/test_analysis.py
```diff
@@ -0,0 +1,1109 @@
...
+from numba import jit, njit
+from numba.core import types, errors, ir
+from numba.cuda.utils import PYVERSION
+from numba.cuda.core import postproc, rewrites, ir_utils
+from numba.cuda.core.options import ParallelOptions
+from numba.cuda.core.inline_closurecall import InlineClosureCallPass
+from numba.cuda.tests.support import (
+    TestCase,
+    MemoryLeakMixin,
+    SerialMixin,
```
gmarkall:
```suggestion
```

numba_cuda/numba/cuda/tests/test_analysis.py
```diff
@@ -0,0 +1,1109 @@
...
+        if args is None:
+            res = cres.entry_point()
+            expected = func()
+        else:
+            res = cres.entry_point(*args)
+            expected = func(*args)
+        self.assertEqual(res, expected)
+
+
+class TestBranchPrune(TestBranchPruneBase, SerialMixin):
```
gmarkall:
```suggestion
class TestBranchPrune(TestBranchPruneBase):
```

numba_cuda/numba/cuda/tests/test_analysis.py
```diff
@@ -0,0 +1,1109 @@
...
+            impl,
+            (types.int64,),
+            [
+                None,
+            ],
+            1,
+        )
+
+
+class TestBranchPrunePredicates(TestBranchPruneBase, SerialMixin):
```
gmarkall:
```suggestion
class TestBranchPrunePredicates(TestBranchPruneBase):
```

numba_cuda/numba/cuda/tests/test_analysis.py
```diff
@@ -0,0 +1,1109 @@
...
+            pm.add_pass(PreserveIR, "preserves the IR as metadata")
+            dpb = DefaultPassBuilder
+            typed_passes = dpb.define_typed_pipeline(self.state)
+            pm.passes.extend(typed_passes.passes)
+            lowering_passes = dpb.define_nopython_lowering_pipeline(self.state)
+            pm.passes.extend(lowering_passes.passes)
+            pm.finalize()
+            return [pm]
+
+
```
gmarkall: This class doesn't seem to be doing anything - were there some tests in it that are now removed?

gmarkall (commented):
I started reviewing this and ended up with a few comments on the diff. I wanted to post the review now so that they become visible, and also to enquire about the tests in `test_analysis` - these are using Numba's CPU target for the tests. Did you plan to modify them to use the CUDA target?

atmnp:
The only tests failing right now in CI are with Python 3.9, and we intend on removing support for that Python version.

numba_cuda/numba/cuda/core/byteflow.py
```diff
@@ -0,0 +1,2346 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: BSD-2-Clause
+
+"""
+Implement python 3.8+ bytecode analysis
```
brandon-b-miller, line 5: I noticed that numba main has some specific [python 3.14 handling](https://github.com/numba/numba/blob/338b0a2a98a3ede6fba7888a03726804d2059a61/numba/core/byteflow.py#L296) that isn't included in the version of this file being vendored. Should we add that in just to avoid having to do so later?

numba_cuda/numba/cuda/core/controlflow.py
```diff
@@ -0,0 +1,989 @@
...
+        self.jump(inst.get_jump_target())
+        self._force_new_block = True
+
+    op_JUMP_BACKWARD = op_JUMP_FORWARD
+
+    def op_RETURN_VALUE(self, inst):
+        self._curblock.terminating = True
+        self._force_new_block = True
+
+    if PYVERSION in ((3, 12), (3, 13)):
```
brandon-b-miller, line 973: Same question here about 3.14 updates

numba_cuda/numba/cuda/tests/test_flow_control.py
```diff
@@ -0,0 +1,1452 @@
...
+    def test_for_loop6_npm(self):
+        self.test_for_loop6()
+
+    def test_for_loop7(self, flags=enable_pyobj_flags):
+        self.run_test(for_loop_usecase7, [5], [0], [0], [0], flags=flags)
+
+    def test_for_loop7_npm(self):
+        self.test_for_loop7()
+
+    # TODO: this test fails with numba-cuda, we need to check if the step in range is non-zero
```
brandon-b-miller: Is this resolved? If not, can we make it some kind of xfail?

numba_cuda/numba/cuda/tests/test_analysis.py
```diff
@@ -0,0 +1,1169 @@
...
+        cout = np.zeros(1)
+        args += (out,)
+        dargs += (cout,)
+        cres.py_func(*args)
+        cres[1, 1](*dargs)
+        self.assertPreciseEqual(out[0], cout[0])
+
+
+class TestBranchPrune(TestBranchPruneBase):
+    @skip_on_cudasim("Analysis passes not done in simulator")
```
brandon-b-miller: AFAICT this whole file is skipped under the simulator so we might just use a global conditional skip here

numba_cuda/numba/cuda/core/byteflow.py
atmnp: We're pinned to v0.61, I'd prefer to avoid partial vendorings of main to make the break cleaner, and to make future rebases cleaner.

numba_cuda/numba/cuda/core/controlflow.py
atmnp: Same as above.

numba_cuda/numba/cuda/tests/test_analysis.py
atmnp: Fixed

brandon-b-miller (approved):
LGTM

PR #434, opened 2025-08-28
[Refactor][NFC] Vendor-in Inline Closure Call Pass for future CUDA-specific changes

atmnp:
This change vendors in the InlineClosureCallPass to allow for future CUDA-specific customizations.


atmnp:
This is a subset of the changes in #439, closing this for now.

PR #436, opened 2025-08-28
Regenerate float16 bindings

lakshayg:
This PR is based on #265 and merely regenerates the bindings
after all the features requested from Numbast were implemented.

### DIFF
--- a/numba_cuda/numba/cuda/_internal/cuda_fp16.py
+++ b/numba_cuda/numba/cuda/_internal/cuda_fp16.py
@@ -19,7 +19,9 @@ import operator
 import numba
 from llvmlite import ir
 from numba import types
+from numba.cuda.cudadrv.driver import _have_nvjitlink
 from numba.core.datamodel import PrimitiveModel, StructModel
+from numba.core.errors import NumbaPerformanceWarning
 from numba.core.extending import (
     lower_cast,
     make_attribute_wrapper,
@@ -56,6 +58,7 @@ from numba.types import (
     uint64,
     void,
 )
+from warnings import warn
 
 float32x2 = vector_types["float32x2"]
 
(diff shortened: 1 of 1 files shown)


numba_cuda/numba/cuda/fp16.py
```diff
@@ -73,7 +73,6 @@
     __hadd,
     __hadd_rn,
     __hadd_sat,
-    __hcmadd,
```
isVoid, line 76: This is an intrinsics that requires half2 type. We removed the bindings for half2 in this PR because we don't expose anything nor documented them. The reduction of LOC is also due to this.

numba_cuda/numba/cuda/_internal/cuda_fp16.py
```diff
@@ -216,14 +229,14 @@ class _ctor_template_unnamed1302366(ConcreteTemplate):
 def _lower__ZN6__halfC1Ev(shim_stream, shim_obj):
     shim_raw_str = """
     extern "C" __device__ int
-    ____half___ZN6__halfC1Ev_1(int &ignore, __half *self ) {
```
isVoid, line 219: This is an interesting shim function naming change. I think it's because the prior version generates after the name mangling change and before dropping the prefixes of struct methods. `C1` in the mangled name is sufficient to indicate that this is a constructor of `__half` struct, so no prefix is needed.

configs/cuda_fp16.yml
```diff
@@ -13,3 +16,6 @@ Data Models:
     __half2: StructModel
 Shim Include Override: "\"cuda_fp16.h\""
 Require Pynvjitlink: False
+Use Separate Registry: True
+GPU Arch:
+    - sm_86
```
isVoid, line 21: GPU Arch defines the macro __CUDA_ARCH__ when clang is used to parse fp16 headers. We don't have a good guidance of this value yet. For float16, after sm70 it is feature complete. So any value after 70 should give us all of the API that's in the header.

isVoid: Ideally - Numbast generated bindings should generate different bindings for different arches based on supported arches the library is generating for. Then use difftools to conditionally enable/disable these features.

PR #437, opened 2025-08-28
Ensure NUMBA can manipulate memory from CUDA graphs before the graph is launched

caugonnet:
When using CUDA graphs, memory nodes correspond to memory that is reserved in virtual memory, but only allocated when the graph is launched. This mean we cannot check whether a pointer is a valid device address by the time we create memory objects that we will pass to kernels.


gmarkall:
Thanks for the PR! Prior to running the tests, I suspect that the general idea of the change should be safe as long as we have a unified virtual address space, which I think is true for Pascal (CC 6.0) devices and later on Linux. I'm not quite sure if it's the case on Windows, and I'll have to check.

caugonnet:
@gmarkall should we merge that PR ? If i understand the only remaining question could be windows ?

gmarkall:
Our CI now includes Windows, so I'm giving it a run to see what happens.

kkraus14 (approved):

PR #439, opened 2025-08-28
Draft: Vendor in the IR module

ashermancinelli:
The plan-of-record as discussed offline with @gmarkall is to vendor this module piecemeal; modules in numba-cuda will continue to use numba until we have vendored in additional modules (such as types and typing).

---

The IR module is near the root of the import graph of numba-cuda, so it has many dependencies. Several other modules must be vendored in before this can be merged, so this will remain a draft for now. The biggest dependency is currently #373. The purpose of this PR being open prior to being ready for review is to give the team a sense of how close we are to having this module vendored in.

Depends-on:
- `numba/core/types`
- #444 
- #447 
- #443
- #442 
- #417 
- #438

PR #440, opened 2025-08-28
[Refactor][NFC] Vendor-in cpu_options for future CUDA-specific changes

atmnp:
Numba-CUDA relies on the upstream Numba cpu_options to configure itself, this change vendors in the class, so we can later remove the ParallelOptions once #439 is merged.

### DIFF
--- a/numba_cuda/numba/cuda/cgutils.py
+++ b/numba_cuda/numba/cuda/cgutils.py
@@ -11,8 +11,8 @@ import functools
 
 from llvmlite import ir
 
-from numba.core import types, debuginfo
-from numba.cuda import config, utils
+from numba.core import types
+from numba.cuda import config, utils, debuginfo
 import numba.core.datamodel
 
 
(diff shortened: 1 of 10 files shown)


atmnp:
There are 30+ tests that use InlineOptions to test inlining for the CPU target, this is a significant effort to change all of them to be viable for the CUDA target, since many of them rely on function returns, unsupported features like parfor, etc. I am unsure if these tests are absolutely necessary to be imported with this PR, we should move it with another more relevant one.

gmarkall:
> There are 30+ tests that use InlineOptions to test inlining for the CPU target, this is a significant effort to change all of them to be viable for the CUDA target, since many of them rely on function returns, unsupported features like parfor, etc. I am unsure if these tests are absolutely necessary to be imported with this PR, we should move it with another more relevant one.

We have existing tests for specific inlining (and fastmath) choices in the CUDA target already so I'm not too concerned here. Even with the upstream inline tests, there were certain aspects of inlining that were broken until recently for CUDA anyway (when they were then fixed and the CUDA target tests added).

PR #443, opened 2025-08-29
Vendor the inline-closure-call module

ashermancinelli:
This module has no standalone tests. All imports have been updated.

### DIFF
--- a/numba_cuda/numba/cuda/compiler.py
+++ b/numba_cuda/numba/cuda/compiler.py
@@ -17,6 +17,7 @@ from numba.core import (
 from numba.core.compiler_lock import global_compiler_lock
 from numba.core.errors import NumbaWarning, NumbaInvalidConfigWarning
 from numba.cuda.core.interpreter import Interpreter
+from numba.cuda.core import inline_closurecall
 
 from numba.cuda import cgutils, typing, lowering, nvvmutils, utils
 from numba.cuda.api import get_current_device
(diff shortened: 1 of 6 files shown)


ashermancinelli:
/ok to test [06586b8](https://github.com/NVIDIA/numba-cuda/pull/443/commits/06586b803ae56141da23f30cfca995cc632226ab)

ashermancinelli:
/ok to test [2e6847f](https://github.com/NVIDIA/numba-cuda/pull/443/commits/2e6847ff15f230e3fcebd3829450874917c3c193)

gmarkall (approved):
> This module has no standalone tests.

Indeed, none were added along with it when it was first added. I had expected that making the following change:

```diff
iff --git a/numba_cuda/numba/cuda/core/ir_utils.py b/numba_cuda/numba/cuda/core/ir_utils.py
index a2ea996..d48e2dd 100644
--- a/numba_cuda/numba/cuda/core/ir_utils.py
+++ b/numba_cuda/numba/cuda/core/ir_utils.py
@@ -1990,7 +1990,7 @@ def get_ir_of_code(glbls, fcode):
     inline_pass = numba.core.inline_closurecall.InlineClosureCallPass(
         ir, numba.cuda.core.options.ParallelOptions(False), swapped
     )
-    inline_pass.run()
+    #inline_pass.run()
 
     # TODO: DO NOT ADD MORE THINGS HERE!
     # If adding more things here is being contemplated, it really is time to
```

would cause an example like this to fail to run:

```python
@cuda.jit()
def f():
    def k(x):
        return x + 1
    print(k(4))

f[1, 1]()
```

but it didn't. It seems like the main reason this was added was for parfors, and for stencil support. I feel 90% sure this isn't doing anything for the CUDA target, but would want to check a bit more thoroughly before we remove this entirely (sometime in the future).

There are some imports from `numba.core.stencil` in `inline_closurecall.py` - for dealing with these in future I'd suggest we just delete / make a no-op the code referencing it, rather than trying to bring that in.

PR #445, opened 2025-08-29
Vendor the base module for CUDA-specific refactoring

ashermancinelli:
There are no standalone tests for this module. It is imported from the cpu and caching modules upstream, but we have not pulled in the cpu module and the caching module has already been vendored but BaseContext was not (as it was unused by the upstream caching module anyways).

There are some tests which import this module, but we have versions of the same tests and they do not use the base module. I am working through them to make sure this module should not be imported there.

* numba/numba/tests/test_target_overloadselector.py
* numba/numba/tests/test_ufuncs.py
* numba/numba/tests/test_target_extension.py


ashermancinelli:
/ok to test [4143571](https://github.com/NVIDIA/numba-cuda/pull/445/commits/414357188b840152cdd71ecb02aea10e2e54afc5)

numba_cuda/numba/cuda/core/base.py
```diff
@@ -0,0 +1,1305 @@
...
+    def enter_fn():
+        event.start_event("numba:llvm_lock")
+
+    def exit_fn():
+        event.end_event("numba:llvm_lock")
+
+    ll.ffi.register_lock_callback(enter_fn, exit_fn)
+
+
+_initialize_llvm_lock_event()
```
gmarkall: I need to check whether duplicating Numba doing this will be a risk / issue.

gmarkall: I've come to the conclusion that we shouldn't register this callback, and avoid having to think about the consequences of having two callbacks registered. We're unlikely to spend any significant amount of time in the llvmlite binding because we only use it to get the size and alignment of LLVM IR types anyway (rather than using it for any actual optimization, compilation, or jitting).

```suggestion
```

gmarkall (changes requested):
Looks good in general - there is one suggestion on the diff, which is that we shouldn't register callbacks for the LLVM lock. Does this seem to make sense?

numba_cuda/numba/cuda/core/base.py
```diff
@@ -0,0 +1,1305 @@
...
+    datamodel,
+    funcdesc,
+    config,
+    imputils,
+)
+from numba.cuda import cgutils, debuginfo
+from numba.core import event, errors, targetconfig
+from numba import _dynfunc, _helperlib
+from numba.core.compiler_lock import global_compiler_lock
+from numba.core.pythonapi import PythonAPI
```
gmarkall: The vendoring of this is now merged, so I think it can be

```suggestion
from numba.cuda.core.pythonapi import PythonAPI
```

ashermancinelli:
> Looks good in general - there is one suggestion on the diff, which is that we shouldn't register callbacks for the LLVM lock. Does this seem to make sense?

This makes sense because of your other comment:

> We're unlikely to spend any significant amount of time in the llvmlite binding because we only use it to get the size and alignment of LLVM IR types

I would be more worried if we were modifying or destroying things, in which case the locks might be more important. I will run our tests with your suggestions to see if anything comes up. Thank you!

ashermancinelli:
Local testing with your suggestions looks good, I'll start the CI.

ashermancinelli:
/ok to test [0d49d63](https://github.com/NVIDIA/numba-cuda/pull/445/commits/0d49d633cbd85eb87d53c17ed169e29a8650cac9)

ashermancinelli:
/ok to test [5cd9daf](https://github.com/NVIDIA/numba-cuda/pull/445/commits/5cd9daf01f9cb703f276aa6740ff0c205b741ebb)

gmarkall:
> I would be more worried if we were modifying or destroying things, in which case the locks might be more important. I will run our tests with your suggestions to see if anything comes up. Thank you!

These aren't doing anything with a lock - they're just setting up a callback to record when the LLVM lock is acquired and released. The LLVM lock will still be obtained when necessary.

gmarkall (approved):

PR #446, opened 2025-08-29
Vendor in the pythonapi module

ashermancinelli:
This module is also used in the base module, so if #445 is merged I will update the imports in this PR. Otherwise, it looks like they can co-exist and I can fix the imports in a follow-up pr. There seems to be a standalone test for this module, but it appears to only use strings, and I'm not sure the test makes sense on the GPU as-is.


ashermancinelli:
/ok to test [723971a](https://github.com/NVIDIA/numba-cuda/pull/446/commits/723971a47cc446a26fa3da90f8435704d4430c0c)

gmarkall (approved):

gmarkall:
> There seems to be a standalone test for this module, but it appears to only use strings, and I'm not sure the test makes sense on the GPU as-is.

This module isn't really doing anything for the CUDA target, but there's a few places to change to get it completely unreferenced, so I've gone with merging this for now and can remove it once we have all the references to it later.

PR #450, opened 2025-08-29
[Refactor] Vendor in numba.core.config for future CUDA-specific changes

VijayKandiah:
This change vendors in `numba.core.config` for future CUDA-specific changes. Also updated a dangling reference to `numba.core.utils`.


VijayKandiah:
cuDF CI tests fail here because cuDF still uses `numba.config` instead of the vendored in `numba.cuda.config` for `_CUDFNumbaConfig`.   https://github.com/rapidsai/cudf/blob/branch-25.10/python/cudf/cudf/utils/_numba.py#L10
So numba-cuda does not see cuDF setting `numba.config.CUDA_LOW_OCCUPANCY_WARNINGS = 0` and we still emit the low occupancy warning which cuDF treats as a kernel execution failure here https://github.com/rapidsai/cudf/blob/branch-25.10/python/cudf/cudf/core/indexed_frame.py#L3446. @gmarkall can we change the default for `CUDA_LOW_OCCUPANCY_WARNINGS` in `numba.cuda.core.config` from 1 to 0? This should alleviate this CI failure without needing to modify cuDF for the time being. The other option is to merge this PR without changing this default, and modify cuDF to make `_CUDFNumbaConfig` use `numba.cuda.config` to fix CI for future PRs.

VijayKandiah:
@brandon-b-miller @gmarkall My recent commits add in an optional import of `numba.config` in `numba.cuda.config`'s `_readenv` and emits a warning about deprecation of `numba.config` in `numba-cuda `if any of the configuration parameters differ. However, as expected, this does not solve the problem with cuDF CI as the `_readenv` is only done during compiler initialization with `reload_config` and any later overwrites to `numba.config` like in cuDF at https://github.com/rapidsai/cudf/blob/branch-25.10/python/cudf/cudf/utils/_numba.py#L10 are not accounted for with my changes. Do you have any suggestions on how to proceed for cuDF, [Awkward Array](https://github.com/scikit-hep/awkward/blob/fd5fa80d94ed47b3030e0d44a805f9e19fd5c7d8/tests-cuda/test_3115_array_typed_cuda_jit.py#L15), and any other such cases where the user explicitly overwrites `numba.config` and expects numba-cuda to respect the changes?

ashermancinelli:
This is great! Just like we talked about offline, I don't think this pattern will work in the same way as before:
```python
from numba.cuda.core.config import NUMBA_FOO
```
because `config.NUMBA_FOO` is a field of instance `config` and not a variable inside the module `config`. Does that sound right? Do we think this will cause any issues for our users?

VijayKandiah:
> ```python
> from numba.cuda.core.config import NUMBA_FOO
> ```

@ashermancinelli, Doing things like `python -c "from numba.cuda.core.config import CUDA_LOW_OCCUPANCY_WARNINGS"` works fine for me. This still invokes the module-level `__getattr__` and returns the boolean value 0 after checking numba.config for any mismatches.

VijayKandiah:
The remaining failures in cuDF CI are to do with configuration variable `CUDA_ENABLE_NRT` not being present in `numba.config`. numba does not set it anywhere so when cuDF tries to fetch `numba.config.CUDA_ENABLE_NRT` [here](https://github.com/rapidsai/cudf/blob/branch-25.10/python/cudf/cudf/core/udf/nrt_utils.py#L40C35-L40C50), it gets an `AttributeError`. With this PR, numba-cuda sets it for `numba.cuda.config` instead of `numba.config` [here](https://github.com/NVIDIA/numba-cuda/blob/main/numba_cuda/numba/cuda/memory_management/nrt.py#L39). 

@brandon-b-miller can we modify cuDF https://github.com/rapidsai/cudf/blob/branch-25.10/python/cudf/cudf/core/udf/nrt_utils.py#L40C35-L40C50 to be `original_value = getattr(numba_config, "CUDA_ENABLE_NRT", False)`?

ci/test_thirdparty.sh
```diff
@@ -39,13 +39,13 @@ rapids-logger "Show Numba system info"
 python -m numba --sysinfo
 
 rapids-logger "Run Scalar UDF tests"
-python -m pytest python/cudf/cudf/tests/dataframe/methods/test_apply.py -W ignore::UserWarning
+python -m pytest python/cudf/cudf/tests/dataframe/methods/test_apply.py -W ignore::UserWarning -W ignore::DeprecationWarning
```
brandon-b-miller: can this and below match the one deprecation warning with the particular string we're expecting?

VijayKandiah: My recent change made it ignore deprecation warnings only from `numba.cuda.core.config`, there should be only one such warning that's ignored.

brandon-b-miller (approved):
Spoke extensively about this PR offline with @VijayKandiah , LGTM.

Issue #453, opened 2025-09-02
[BUG] Intermittent reference errors in test suite


gmarkall:
**Describe the bug**

Sometimes when running locally I'm seeing failures like:

```
===================================== short test summary info =====================================
FAILED numba_cuda/numba/cuda/tests/cudapy/test_ufuncs.py::TestLogUFuncs::test_log10_ufunc - ReferenceError: weakly-referenced object no longer exists
FAILED numba_cuda/numba/cuda/tests/cudapy/test_ufuncs.py::TestLogUFuncs::test_log2_ufunc - ReferenceError: weakly-referenced object no longer exists
FAILED numba_cuda/numba/cuda/tests/cudapy/test_ufuncs.py::TestLogUFuncs::test_log_ufunc - ReferenceError: weakly-referenced object no longer exists
============== 3 failed, 1657 passed, 39 skipped, 11 xfailed, 114 warnings in 37.04s ==============
```

A complete output from one of the failures:

```
__________________________________ TestLogUFuncs.test_log_ufunc ___________________________________
[gw15] linux -- Python 3.12.11 /home/gmarkall/miniforge3/envs/numbadev/bin/python3.12

self = <numba.cuda.tests.cudapy.test_ufuncs.TestLogUFuncs testMethod=test_log_ufunc>

    def setUp(self):
>       BasicUFuncTest.setUp(self)

numba_cuda/numba/cuda/tests/cudapy/test_ufuncs.py:294: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
numba_cuda/numba/cuda/tests/cudapy/test_ufuncs.py:17: in setUp
    super(BaseUFuncTest, self).setUp()
numba_cuda/numba/cuda/tests/support.py:798: in setUp
    super(MemoryLeakMixin, self).setUp()
numba_cuda/numba/cuda/tests/support.py:41: in setUp
    rtsys.memsys_enable_stats()
numba_cuda/numba/cuda/memory_management/nrt.py:55: in wrapper
    return method(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
numba_cuda/numba/cuda/memory_management/nrt.py:184: in memsys_enable_stats
    self._single_thread_launch(
[...]

PR #455, opened 2025-09-02
Close `rtsys` singleton when `cuda.close()` is called

brandon-b-miller:
This PR fixes an issue where `cuda.close` invalidated the context containing the current memsys module without also invalidating the `_Runtime` instance that ostensibly owned that module. This caused reference errors later on when trying to use that module. 

Closes https://github.com/NVIDIA/numba-cuda/issues/453

### DIFF
--- a/numba_cuda/numba/cuda/api.py
+++ b/numba_cuda/numba/cuda/api.py
@@ -11,7 +11,6 @@ import os
 import numpy as np
 
 from .cudadrv import devicearray, devices, driver
-
 from numba.core import config
 from numba.cuda.api_util import prepare_shape_strides_dtype
 
--- a/numba_cuda/numba/cuda/api.py
+++ b/numba_cuda/numba/cuda/api.py
@@ -11,6 +11,7 @@ import os
 import numpy as np
 
 from .cudadrv import devicearray, devices, driver
+
 from numba.core import config
 from numba.cuda.api_util import prepare_shape_strides_dtype
 
@@ -509,6 +510,9 @@ def close():
     contexts if the current thread is the main thread.
     """
     devices.reset()
+    from .memory_management import rtsys
+
+    rtsys.close()
 
 
 def _auto_device(ary, stream=0, copy=True):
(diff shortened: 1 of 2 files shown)


numba_cuda/numba/cuda/api.py
```diff
@@ -509,6 +509,9 @@ def close():
     contexts if the current thread is the main thread.
     """
     devices.reset()
+    from .memory_management import rtsys
```
brandon-b-miller, line 512: This local import breaks a circular import issue. Normally I'd shy away from this for perf reasons but since we're `close`-ing, I'm not sure it matters

numba_cuda/numba/cuda/api.py
```diff
@@ -509,6 +509,9 @@ def close():
...
+
+    rtsys.close()
```
rparolin, line 514: Nit: I'd include a comment describing why this `rtsys.close()` was added here.

rparolin (approved):
lgtm, minor comment added to consider.

numba_cuda/numba/cuda/api.py
gmarkall: Indeed - it's only going to have a big impact on the first `close()` anyway. Since this is usually used at interpreter shutdown or in the test suite, I don't think it's a concern either.

numba_cuda/numba/cuda/api.py
gmarkall: `rtsys.close()` should probably come before `devices.reset()` - in the unlikely event of some exception occurring in `devices.reset()` but execution continuing (it being caught somewhere), the `rtsys` will remain "open".

rparolin: Is that correct? I thought that the devices was using the memory from the rtsys memory system. It would follow that we would want to shutdown the users of that memory before shutting down the memory system or is my mental model incorrect here?

gmarkall: This is not the case - things are unintuitively named here, but what really happens in `devices.reset()` is that all active CUDA contexts are destroyed. The rtsys holds on to a module and allocations made in those contexts, so the correct ordering is to reset the rtsys first.

rparolin: ty!

gmarkall:
@brandon-b-miller Thanks for this - I'm pretty sure this is as-needed, but I'll give it a quick local test tomorrow to make sure I don't see the same error (or variants thereof) popping up locally when testing in parallel tomorrow morning, and then I'll merge it.

PR #456, opened 2025-09-02
Vendor in targetconfig, event, funcdesc, descriptors for CUDA-specific changes

VijayKandiah:
This PR vendors in `targetconfig`, `event`, `funcdesc`, `descriptors` from `numba.core` for CUDA-Specific customization.

### DIFF
--- a/numba_cuda/numba/cuda/core/event.py
+++ b/numba_cuda/numba/cuda/core/event.py
@@ -7,7 +7,7 @@ to register callbacks to listen to specific compiler events.
 
 The following events are built in:
 
-- ``"numba:compile"`` is broadcast when a dispatcher is compiling. Events of
+- ``"numba-cuda:compile"`` is broadcast when a dispatcher is compiling. Events of
   this kind have ``data`` defined to be a ``dict`` with the following
   key-values:
 
@@ -15,14 +15,11 @@ The following events are built in:
   - ``"args"``: the argument types.
   - ``"return_type"``: the return type.
 
-- ``"numba:compiler_lock"`` is broadcast when the internal compiler-lock is
+- ``"numba-cuda:compiler_lock"`` is broadcast when the internal compiler-lock is
   acquired. This is mostly used internally to measure time spent with the lock
   acquired.
 
-- ``"numba:llvm_lock"`` is broadcast when the internal LLVM-lock is acquired.
-  This is used internally to measure time spent with the lock acquired.
-
-- ``"numba:run_pass"`` is broadcast when a compiler pass is running.
+- ``"numba-cuda:run_pass"`` is broadcast when a compiler pass is running.
 
     - ``"name"``: pass name.
     - ``"qualname"``: qualified name of the function being compiled.
@@ -36,9 +33,18 @@ Applications can register callbacks that are listening for specific events using
 of ``Listener`` that defines custom actions on occurrence of the specific event.
 """
 
+import os
+import json
+import atexit
+import abc
 import enum
+import time
+import threading
+from timeit import default_timer as timer
 from contextlib import contextmanager, ExitStack
 from collections import defaultdict
+from numba.core import config
+from numba.cuda import utils
 
 
 class EventStatus(enum.Enum):
@@ -51,10 +57,9 @@ class EventStatus(enum.Enum):
 # Builtin event kinds.
 _builtin_kinds = frozenset(
     [
-        "numba:compiler_lock",
-        "numba:compile",
-        "numba:llvm_lock",
-        "numba:run_pass",
+        "numba-cuda:compiler_lock",
+        "numba-cuda:compile",
+        "numba-cuda:run_pass",
     ]
 )
 
@@ -62,7 +67,7 @@ _builtin_kinds = frozenset(
 def _guard_kind(kind):
     """Guard to ensure that an event kind is valid.
 
-    All event kinds with a "numba:" prefix must be defined in the pre-defined
+    All event kinds with a "numba-cuda:" prefix must be defined in the pre-defined
     ``numba.cuda.core.event._builtin_kinds``.
     Custom event kinds are allowed by not using the above prefix.
 
@@ -74,10 +79,17 @@ def _guard_kind(kind):
     ------
     res : str
     """
-    if kind.startswith("numba:") and kind not in _builtin_kinds:
+    if kind.startswith("numba-cuda:") and kind not in _builtin_kinds:
         msg = (
             f"{kind} is not a valid event kind, "
-            "it starts with the reserved prefix 'numba:'"
+            "it starts with the reserved prefix 'numba-cuda:'"
+        )
+        raise ValueError(msg)
+    if kind.startswith("numba:"):
+        msg = (
+            f"{kind} is using invalid reserved prefix 'numba:' "
+            "it should either start with the reserved prefix 'numba-cuda:'"
+            ", or be a custom event kind"
         )
         raise ValueError(msg)
     return kind
@@ -183,6 +195,33 @@ class Event:
 _registered = defaultdict(list)
 
 
+def register(kind, listener):
+    """Register a listener for a given event kind.
+
+    Parameters
+    ----------
+    kind : str
+    listener : Listener
+    """
+    assert isinstance(listener, Listener)
+    kind = _guard_kind(kind)
+    _registered[kind].append(listener)
+
+
+def unregister(kind, listener):
+    """Unregister a listener for a given event kind.
+
+    Parameters
+    ----------
+    kind : str
+    listener : Listener
+    """
+    assert isinstance(listener, Listener)
+    kind = _guard_kind(kind)
+    lst = _registered[kind]
+    lst.remove(listener)
+
+
 def broadcast(event):
     """Broadcast an event to all registered listeners.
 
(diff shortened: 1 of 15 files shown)


numba_cuda/numba/cuda/core/funcdesc.py
```diff
@@ -0,0 +1,332 @@
...
+        return getattr(self.lookup_module(), self.qualname)
+
+    @property
+    def llvm_func_name(self):
+        """
+        The LLVM-registered name for the raw function.
+        """
+        return self.mangled_name
+
+    # XXX refactor this
```
rparolin: Is this an old comment?  No sure what it relates to.

rparolin (approved):
lgtm, minor comment on a todo.

numba_cuda/numba/cuda/core/funcdesc.py
VijayKandiah: Yes it's an old comment carried over from [numba](https://github.com/numba/numba/blob/main/numba/core/funcdesc.py#L115). In general, I tend to keep around comments from Numba but this seems pretty vague. I'll remove it.

rparolin: ty!

numba_cuda/numba/cuda/core/event.py
```diff
@@ -0,0 +1,252 @@
...
+    END = enum.auto()
+
+
+# Builtin event kinds.
+_builtin_kinds = frozenset(
+    [
+        "numba:compiler_lock",
+        "numba:compile",
+        "numba:llvm_lock",
+        "numba:run_pass",
```
gmarkall: Since this is not coming from Numba itself anymore, it's probably best that we rename these:

```suggestion
        "numba-cuda:compiler_lock",
        "numba-cuda:compile",
        "numba-cuda:llvm_lock",
        "numba-cuda:run_pass",
```

This might have a "downstream" effect where we need to rename the usage of events in the rest of the Numba-CUDA codebase so they begin with `numba-cuda` as well

numba_cuda/numba/cuda/core/event.py
```diff
@@ -0,0 +1,252 @@
...
+
+    Parameters
+    ----------
+    kind : str
+
+    Return
+    ------
+    res : str
+    """
+    if kind.startswith("numba:") and kind not in _builtin_kinds:
```
gmarkall:
```suggestion
    if kind.startswith("numba-cuda:") and kind not in _builtin_kinds:
```

numba_cuda/numba/cuda/core/event.py
gmarkall: (So far I only see one, in `compiler_machinery.py`)

numba_cuda/numba/cuda/core/event.py
gmarkall: Maybe we should also have another guard, which prevents events prefixed `"numba:"` from being created, so that we ensure they're all renamed to `"numba-cuda:"`.

numba_cuda/numba/cuda/core/event.py
```diff
@@ -0,0 +1,252 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: BSD-2-Clause
+
+"""
+The ``numba.cuda.core.event`` module provides a simple event system for applications
+to register callbacks to listen to specific compiler events.
+
+The following events are built in:
+
+- ``"numba:compile"`` is broadcast when a dispatcher is compiling. Events of
```
gmarkall: Updates to the names here (see below) should also be made.

numba_cuda/numba/cuda/core/event.py
```diff
@@ -0,0 +1,252 @@
...
+  - ``"args"``: the argument types.
+  - ``"return_type"``: the return type.
+
+- ``"numba:compiler_lock"`` is broadcast when the internal compiler-lock is
+  acquired. This is mostly used internally to measure time spent with the lock
+  acquired.
+
+- ``"numba:llvm_lock"`` is broadcast when the internal LLVM-lock is acquired.
+  This is used internally to measure time spent with the lock acquired.
+
```
gmarkall: We're not going to be using this to any significant degree, it can be removed. (See https://github.com/NVIDIA/numba-cuda/pull/445#discussion_r2314197179)

numba_cuda/numba/cuda/core/event.py
```diff
@@ -0,0 +1,252 @@
...
+    START = enum.auto()
+    END = enum.auto()
+
+
+# Builtin event kinds.
+_builtin_kinds = frozenset(
+    [
+        "numba:compiler_lock",
+        "numba:compile",
+        "numba:llvm_lock",
```
gmarkall: The llvm_lock event can be removed.

numba_cuda/numba/cuda/core/event.py
gmarkall, line 1: Just a general observation - this vendoring brings in the infrastructure necessary to support the usage of the module in Numba-CUDA, but doesn't bring in any of the APIs that a user would use to observe or record events.

I think it's OK not to bring these in now, but a comment should be added to the top of this file (e.g. a "TODO") that these need bringing in if they're to be used in Numba-CUDA for recording / observing events and measuring performance.

numba_cuda/numba/cuda/descriptor.py
```diff
@@ -1,24 +1,24 @@
...
+from numba.cuda.core.options import TargetOptions
 from .target import CUDATargetContext, CUDATypingContext
 
 
 class CUDATargetOptions(TargetOptions):
     pass
 
 
-class CUDATarget(TargetDescriptor):
+class CUDATarget(metaclass=ABCMeta):
```
gmarkall: I think there's no need for this to be an ABC if it has no abstract methods.

```suggestion
class CUDATarget:
```

numba_cuda/numba/cuda/core/targetconfig.py
```diff
@@ -0,0 +1,326 @@
...
+    @property
+    def default(self):
+        return self._default
+
+    @property
+    def doc(self):
+        return self._doc
+
+
+class _FlagsStack(utils.ThreadLocalStack, stack_name="flags"):
```
gmarkall: For this vendoring to be safe, we need to be sure that the use of `ConfigStack` is going to be consistently all within Numba-CUDA (and not split across Numba and Numba-CUDA, which will have two different thread-local config stacks). I'm going to try to verify this.

gmarkall (changes requested):
I think this mostly looks good - there are some small comments on the diff.

There is one major issue which is with the `_FlagsStack` and `ConfigStack` - because these use TLS for storing the flags, we now have two separate TLS stores for the stacks between Numba and Numba-CUDA. This would be OK if compilation always referred to the Numba-CUDA `ConfigStack`, but unfortunately because overload templates refer to the config stack (see https://github.com/numba/numba/blob/main/numba/core/typing/templates.py#L711), this won't be safe to vendor until all overloads used by the CUDA target have been vendored. We may not see much of an issue in the test suite now, but I can see this going wrong in larger / more complex real world codes where the stack will be much larger and end up with a mix of Flags and CUDA Flags objects in it.

For now for this PR to proceed, can you remove `_FlagsStack` and `ConfigStack`, and keep all references to `ConfigStack` to be the Numba one please?

numba_cuda/numba/cuda/core/event.py
VijayKandiah: Added the additional guard to prevent `"numba:"` prefixed events.

numba_cuda/numba/cuda/core/event.py
VijayKandiah: I brought over the remaining APIs rather than leaving a TODO here.

VijayKandiah:
Thanks for the review Graham! I reverted back to using `numba.core.ConfigStack`. I will revisit this after vendoring in all components that use `numba.core.ConfigStack`, and switch them all at that point to use`numba.cuda.core.ConfigStack`.

gmarkall (changes requested):
Thanks for the updates! I think that importing `ConfigStack` from `numba.core` into `targetconfig` is good.

However, I think the changes in the following files that switched back to importing `targetconfig` from `numba.core` were a move in the wrong direction, because `numba.cuda.core.targetconfig` is safe to use as a result of it importing `ConfigStack` rather than redefining it:

- `numba_cuda/numba/cuda/typing/templates.py`
- `numba_cuda/numba/cuda/target.py`
- `numba_cuda/numba/cuda/nvvmutils.py`
- `numba_cuda/numba/cuda/mathimpl.py`
- `numba_cuda/numba/cuda/lowering.py`

Can you keep these importing `targetconfig` from `numba.cuda.core` so that it's clearer that `ConfigStack` is the only remaining thing to vendor from `targetconfig`, instead of having other imports back to `numba.core.targetconfig`?

The other changes look good.

gmarkall:
> Does this make sense?

We will move the `ConfigStack` over, it's just that we can't do it whilst upstream overloads are in the CUDA target's registry (because they will look at the Numba config stack during compilation). So there will never be a point at which that import will both exist, and fail. That said, I don't think the code in its present form really presents an issue.

PR #457, opened 2025-09-03
[Refactor][NFC] Vendor in numba.misc for CUDA-specific changes

VijayKandiah:
This PR vendors in the necessary submodules from `numba.misc` for future CUDA-specific customizations.


numba_cuda/numba/cuda/misc/appdirs.py
```diff
@@ -0,0 +1,592 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: BSD-2-Clause
```
gmarkall: There was a copyright notice at the top of this file that has been removed - I don't think we should be removing existing copyrights, only prepending SPDX headers.

gmarkall: Also I note the original seems to be MIT licensed: https://github.com/ActiveState/appdirs/blob/master/LICENSE.txt

numba_cuda/numba/cuda/misc/appdirs.py
gmarkall, line 1: All the imports that are not at the module level seem to be missing from this file - did the formatter do something odd here? Or am I misunderstanding something?

numba_cuda/numba/cuda/misc/gdb_hook.py
gmarkall, line 1: The overloads in this file should be for the CUDA target (i.e. `@overload(<name>, target="CUDA")`) to avoid a duplicate registration to Numba's.

numba_cuda/numba/cuda/misc/literal.py
gmarkall, line 1: These overloads also need to be for the CUDA target.

gmarkall (dismissed):
This looks good on the whole - there are a couple of issues noted on the diff.

The AppDirs issues are likely to only show up in use in installations that are a bit different to our CI, like where system-wide installs are used or running in Jupyter notebooks, so I'm cautious to avoid changing things in it without having a strong motivation to do so.

numba_cuda/numba/cuda/misc/appdirs.py
VijayKandiah: The `pre-commit` formatter did not like three of the imports in this file as is, specifically in this code block below because they are not being used anywhere. They exist only to catch import errors but are not used anywhere, while the actual imports being used are done inside the methods `_get_win_folder_with_*`.
```
if system == "win32":
    try:
        import win32com.shell
        _get_win_folder = _get_win_folder_with_pywin32
    except ImportError:
        try:
            from ctypes import windll
            _get_win_folder = _get_win_folder_with_ctypes
        except ImportError:
            try:
                import com.sun.jna
                _get_win_folder = _get_win_folder_with_jna
            except ImportError:
                _get_win_folder = _get_win_folder_from_registry
```
 I modified the above by lifting some imports from `_get_win_folder_with_*` to the above code block to get past formatter errors while maintaining the guards to catch import errors. 
 
 ```
 if system == "win32":
    try:
        from win32com.shell import shellcon, shell

        _get_win_folder = _get_win_folder_with_pywin32
    except ImportError:
        try:
            import ctypes
            from ctypes import windll

            _get_win_folder = _get_win_folder_with_ctypes
        except ImportError:
            try:
                from com.sun import jna
                from com.sun.jna.platform import win32

                _get_win_folder = _get_win_folder_with_jna
            except ImportError:
                _get_win_folder = _get_win_folder_from_registry
```
 
Please let me know if this approach makes sense and if there's any other imports that are missing that caught your eye. I do not see any other differences in imports.

File: numba_cuda/numba/cuda/misc/appdirs.py (hunk 1 above)
Comment (VijayKandiah) [reply numba_cuda/numba/cuda/misc/appdirs.py]: Ah I did not realize we are just pre-prending SPDX headers. Made the change.

Comment (gmarkall) [reply numba_cuda/numba/cuda/misc/appdirs.py]: The original copyright is now present but I think it's still probably not right to put `BSD-2-Clause` as the identifier when the original file was MIT-licensed (just double checking on the correct thing to do here, will follow up)

File: numba_cuda/numba/cuda/misc/appdirs.py
Comment (gmarkall) [reply numba_cuda/numba/cuda/misc/appdirs.py:1]: That makes sense - I totally missed that modification because I'm not used to looking for imports at the bottom of the file (though I recognise this is the way the file is and always has been).

File: numba_cuda/numba/cuda/misc/appdirs.py [hunk 2]
```diff
@@ -0,0 +1,594 @@
...
+        "site_data_dir",
+        "user_config_dir",
+        "site_config_dir",
+        "user_cache_dir",
+        "user_log_dir",
+    )
+
+    print("-- app dirs %s --" % __version__)
+
+    print("-- app dirs (with optional 'version')")
```
rparolin, line 576: In a future PR, I think we should migrate these to unit test checks that are part of the suite.

PR #460, opened 2025-09-03
[Refactor][NFC] Vendor-in numba.core.unsafe for future CUDA-specific changes

atmnp:
We primarily use the exception handling mechanisms in numba.core.unsafe right now, but we will also use it in cpython.hashing once we've ported that over. I've removed the nrt support since we won't end up porting over it's uses from upstream Numba.

### DIFF
--- a/numba_cuda/numba/cuda/compiler.py
+++ b/numba_cuda/numba/cuda/compiler.py
@@ -12,8 +12,8 @@ from numba.core import (
     funcdesc,
     config,
     bytecode,
-    cpu,
 )
+from numba.cuda.core.options import ParallelOptions
 from numba.core.compiler_lock import global_compiler_lock
 from numba.core.errors import NumbaWarning, NumbaInvalidConfigWarning
 from numba.cuda.core.interpreter import Interpreter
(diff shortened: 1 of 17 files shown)


numba_cuda/numba/cuda/core/unsafe/refcount.py
```diff
@@ -0,0 +1,87 @@
...
+Helpers to see the refcount information of an object
+"""
+
+from llvmlite import ir
+
+from numba.core import types
+from numba.cuda import cgutils
+from numba.core.extending import intrinsic
+
+from numba.core.runtime.nrtdynmod import _meminfo_struct_type
```
gmarkall:
```suggestion
_word_type = ir.IntType(64)
_pointer_type = ir.PointerType(ir.IntType(8))

_meminfo_struct_type = ir.LiteralStructType(
    [
        _word_type,  # size_t refct
        _pointer_type,  # dtor_function dtor
        _pointer_type,  # void *dtor_info
        _pointer_type,  # void *data
        _word_type,  # size_t size
    ]
)
```

gmarkall (approved):

PR #461, opened 2025-09-03
Improve debug value range coverage

jiel-nv:
The current kernel debug support tracks debug value of variables by inserting dbg.value immediate after a store instruction to a user variable. However, in some specific scenario the debug value info could be missing range, see details in nvbug#5321343, when a basic block has a load instruction loading from a user variable and saving to a temparary, then store the temporary to an internal variable without any other assignment instruction saving value to that user variable in the whole basic block. The back end libnvvm can not fully track the variable value in this scenario.

This PR adds the enhanced handling of debug value, detecting the not yet covered user variable and inserting the dbg.value for the missing range per basic block.

A test is also added to guard this change.

### DIFF
--- a/numba_cuda/numba/cuda/lowering.py
+++ b/numba_cuda/numba/cuda/lowering.py
@@ -1833,33 +1833,67 @@ class CUDALower(Lower):
             self.context.enable_debuginfo
             # Conditions used to elide stores in parent method
             and self.store_var_needed(name)
-            # No emission of debuginfo for internal names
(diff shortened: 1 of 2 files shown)


rparolin:
@gmarkall Nudging this to the top of your inbox. Please review this PR when are you back. Thank you!

jiel-nv:
Soft ping. @gmarkall @rparolin

gmarkall (approved):

PR #464, opened 2025-09-04
Refactor cuda_paths.py, support CUDA 13 conda packages better

### LINKED ISSUE #452: [BUG] Windows: `numba-cuda` can’t find NVVM/NVRTC with conda-forge `cuda-version=13` layout (DLLs in `Library\bin\x64`); `libs.test()` reports missing `nvvm.dll`/`nvrtc.dll`
**Describe the bug**
On Windows 11, after installing `numba-cuda` from conda-forge with `cuda-version=13`, importing CUDA initially fails with `nvJitLink*.dll` not found. Adding `…\Library\bin\x64` to `PATH` makes the import succeed, but `cuda.cudadrv.libs.test()` still cannot locate **NVVM**, **NVRTC**, **cudadevrt**, and **libdevice**. The same conda-forge environment works fine on WSL. This appears to be a Windows path discovery mismatch: conda-forge places versioned DLLs under `Library\bin\x64` and `Library\nvvm\bin\x64`, whereas `numba-cuda` seems to look only in `Library\bin` / `Library\nvvm\bin` for unversioned `nvvm.dll` / `nvrtc.dll` and expects `cudadevrt.lib` elsewhere.

---

1. On Windows 11, create/activate a conda env and install:

   ```bash
   conda install -c conda-forge numba-cuda "cuda-version=13"
   ```
2. In Python:

   ```python
   from numba import cuda
   ```

   Observe:

   ```
   cuda.pathfinder._dynamic_libs.load_dl_common.DynamicLibNotFoundError:
   Failure finding "nvJitLink*.dll"
   ```
3. Add `C:\Users\<USER>\miniconda3\Library\bin\x64` to system `PATH`. Re-run:

   ```python
   from numba import cuda
   cuda.cudadrv.libs.test()
   ```

   Abridged output:

   ```
            nvcuda.dll
            \windows\system32\nvcuda.dll
    Using loader <class 'ctypes.WinDLL'>
            Trying to load driver...        ok
                    Loaded from nvcuda.dll
    Finding nvvm from <unknown>
            Located at nvvm.dll
            Trying to open library...       ERROR: failed to open nvvm:
    Could not find module 'nvvm.dll' (or one of its dependencies). Try using the full path with constructor syntax.
    Finding nvrtc from <unknown>
            Located at nvrtc.dll
            Trying to open library...       ERROR: failed to open nvrtc:
    Could not find module 'nvrtc.dll' (or one of its dependencies). Try using the full path with constructor syntax.
    Finding cudadevrt from <unknown>
            Located at cudadevrt.lib
            Checking library...     ERROR: failed to find cudadevrt:
    cudadevrt.lib not found
    Finding libdevice from <unknown>
            Located at None
      File "<python-input-0>", line 1, in <module>
        from numba import cuda; cuda.cudadrv.libs.test()
      File "C:\Users\torch\miniconda3\Lib\site-packages\numba_cuda\numba\cuda\cudadrv\libs.py", line 197, in test
        check_static_lib(path)
      File "C:\Users\torch\miniconda3\Lib\site-packages\numba_cuda\numba\cuda\cudadrv\libs.py", line 90, in check_static_lib
        if not os.path.isfile(path):
    TypeError: _path_isfile: path should be string, bytes, os.PathLike or integer, not NoneType
   ```
4. Verify files exist but under `x64` subdirs with versioned names (examples):

   ```
   ...\Library\nvvm\bin\x64\nvvm64_40_0.dll
   ...\Library\bin\x64\nvrtc64_130_0.dll
   ...\Library\bin\x64\nvJitLink_130_0.dll
   ...\Library\bin\x64\cudadevrt.lib
   ```
5. On WSL with the same `conda-forge` install, run:

   ```python
   from numba import cuda
   cuda.cudadrv.libs.test()
   ```

   → All checks pass (`nvvm`/`nvrtc`/`cudadevrt`/`libdevice` are found).
  ```
           libcuda.so
           libcuda.so.1
           /usr/lib/libcuda.so
           /usr/lib/libcuda.so.1
           /usr/lib64/libcuda.so
           /usr/lib64/libcuda.so.1
   Using loader <class 'ctypes.CDLL'>
           Trying to load driver...        ok
                   Loaded from libcuda.so.1
           Mapped libcuda.so paths:
                   /usr/lib/wsl/drivers/nvamsi.inf_amd64_5f371da94dfbc04c/libcuda.so.1.1
                   /usr/lib/wsl/lib/libcuda.so.1
   Finding nvvm from Conda environment (NVIDIA package)
           Located at /home/torch/miniconda3/nvvm/lib64/libnvvm.so.4.0.0
           Trying to open library...       ok
   Finding nvrtc from Conda environment (NVIDIA package)
           Located at /home/torch/miniconda3/lib/libnvrtc.so.13.0.48
           Trying to open library...       ok
   Finding cudadevrt from Conda environment (NVIDIA package)
           Located at /home/torch/miniconda3/lib/libcudadevrt.a
           Checking library...     ok
   Finding libdevice from Conda environment (NVIDIA package)
           Located at /home/torch/miniconda3/nvvm/libdevice/libdevice.10.bc
           Checking library...     ok
   Include directory configuration variable:
           CUDA_INCLUDE_PATH=/usr/local/cuda/include
   Finding include directory from Conda environment (NVIDIA package)
           Located at /home/torch/miniconda3/targets/x86_64-linux/include
           Checking include directory...   ok
   True
  ```

* **Environment location:** Bare-metal Windows 11 (Laptop)
* **GPU:** NVIDIA GeForce RTX 4060 (Laptop)
* **Driver:** 580.97
* **CUDA (from `nvidia-smi`):** 13.0
* **Method of numba-cuda install:** conda (conda-forge)

  ```bash
  conda install -c conda-forge numba-cuda "cuda-version=13"
  ```
* **Python/Numba/Numba-CUDA versions:** python=3.13.5, numba=0.61.2, numba-cuda=0.19.1, conda 25.7.0
* **WSL comparison:** Same conda-forge install passes all checks on WSL.

---

**Additional context**

* Full `cuda-toolkit` via conda did **not** resolve discovery in this Windows conda env.

I’m not sure whether this is an environment-variable/packaging-layout issue on my side or a discovery bug in `numba-cuda`. Could anyone advise on the recommended fix or path forward?

kkraus14:
Fixes #452

Refactors a large chunk of `cuda_paths.py` in order to properly support all of the package types that we support. Removes some outdated functionality for things like when the NVIDIA channel conda packages diverged from conda-forge, debian packages, etc.

It was extremely difficult to follow the existing logic of when directory paths vs file paths were returned from the `get_cuda_paths` function and it seemed to differ based on the source of the package, so everything should now be unified to either always return a file path (suffixed by `_path`) or a directory path (suffixed by `_dir`).


kkraus14:
/ok to test [5a26b9a](https://github.com/NVIDIA/numba-cuda/pull/464/commits/5a26b9a56c209ac8a5deb94797670abee98fc019)

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -62,78 +58,80 @@ def _find_valid_path(options):
...
-        nvvm_path = Path(sp, "nvidia", "cuda_nvcc", "nvvm", dso_dir, dso_path)
-        if nvvm_path.exists():
-            return str(nvvm_path.parent)
+        cuda13_path = nvidia.cu13.__path__[0]
+        nvvm_lib_dir = os.path.join(
+            cuda13_path,
+            "bin" if IS_WIN32 else "lib",
+            "x86_64" if IS_WIN32 else "",
+        )
+        dso_path = os.path.join(nvvm_lib_dir, "libnvvm.so.4")
```
kkraus14: Forgot to handle windows here

numba_cuda/numba/cuda/misc/findlib.py
```diff
@@ -11,9 +11,13 @@ def get_lib_dirs():
...
-        # on windows, historically `DLLs` has been used for CUDA libraries,
-        # since approximately CUDA 9.2, `Library\bin` has been used.
-        dirnames = ["DLLs", os.path.join("Library", "bin")]
+        # CUDA 12 puts in "bin" directory, whereas CUDA 13 puts in "bin\x64" directory
+        dirnames = [
+            os.path.join("Library", "bin"),
+            os.path.join("Library", "bin", "x64"),
+            os.path.join("Library", "nvvm", "bin"),
+            os.path.join("Library", "nvvm", "bin", "x64"),
+        ]
```
kkraus14, line 20: This probably isn't needed and I think all `find_lib` calls now pass a `libdir` explicitly and don't rely on this implicit mechanism

numba_cuda/numba/cuda/tests/nocuda/test_library_lookup.py
```diff
@@ -199,9 +204,14 @@ def test_cudalib_path_decision(self):
...
         self.assertFalse(warns)
         if IS_WIN32:
-            self.assertEqual(info, os.path.join("mycudahome", "bin"))
-        elif IS_OSX:
-            self.assertEqual(info, os.path.join("mycudahome", "lib"))
+            # I think only wheels don't have the "Library" directory?
+            self.assertTrue(
+                info
+                in (
+                    os.path.join("mycudahome", "bin"),
```
kkraus14, line 212: This may have been from older no longer supported CUDA layouts, but as far as I can tell it's always "Library/bin" now

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -253,118 +258,200 @@ def get_system_ctk(*subdirs):
...
-        # The path used on Windows
-        libdir = os.path.join(sys.prefix, "Library", "nvvm", _cudalib_path())
-    else:
-        # The path used on Linux is different to that on Windows
-        libdir = os.path.join(sys.prefix, "nvvm", _cudalib_path())
+    nvvm_path = os.path.join(
+        nvvm_dir, "nvvm64_40_0.dll" if IS_WIN32 else "libnvvm.so.4"
+    )
+    # if os.path.exists(nvvm_path) and os.path.isfile(nvvm_path):
+    #     return nvvm_path
```
kkraus14: It would be good to have this in, but it makes tests fail, since they just check that it resolves the path as expected

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -380,26 +467,81 @@ def get_cuda_home(*subdirs):
...
+        and os.path.exists(os.path.join(nvvm_lib_dir, "x64"))
+        and os.path.isdir(os.path.join(nvvm_lib_dir, "x64"))
+    ):
+        nvvm_lib_dir = os.path.join(nvvm_lib_dir, "x64")
+
+    nvvm_path = os.path.join(
+        nvvm_lib_dir, "nvvm64_40_0.dll" if IS_WIN32 else "libnvvm.so.4"
+    )
+    # if os.path.exists(nvvm_path) and os.path.isfile(nvvm_path):
+    #     return nvvm_path
```
kkraus14: +1 makes tests fail

numba_cuda/numba/cuda/misc/findlib.py
kkraus14: There's one test where apparently there's external users that use `find_lib` and rely on implicit `libdir` handling, so leaving this in.

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -62,78 +58,82 @@ def _find_valid_path(options):
...
+            "x86_64" if IS_WIN32 else "",
+        )
+        dso_path = os.path.join(
+            nvvm_lib_dir, "nvvm64_40_0.dll" if IS_WIN32 else "libnvvm.so.4"
+        )
+    except ImportError:
+        pass
 
-    return None
+    if dso_path and os.path.exists(dso_path) and os.path.isfile(dso_path):
```
rwgk: Here and in other places:

`and os.path.exists(dso_path)`

is redundant in combination with `os.path.isfile`|`isdir` and can be removed. It'll still work exactly the same.

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -164,50 +164,61 @@ def _get_nvrtc_wheel():
...
 
 
-def _cudalib_path():
+def _cuda_static_libdir():
     if IS_WIN32:
-        return "bin"
+        return os.path.join("lib", "x64")
     else:
-        return "lib64"
+        return ("lib64",)
```
rwgk, line 203: Above:

```
>>> os.path.join("lib", "x64")
'lib/x64'
```
I.e. that's a `str`.

While this is a `tuple` of length 1. Is that intentional?

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -253,118 +260,201 @@ def get_system_ctk(*subdirs):
...
-        # If the path doesn't exist, we didn't find the NVIDIA conda package
-        return
-
-    paths = find_lib("nvvm", libdir=libdir)
+    libdir = os.path.join(
+        sys.prefix,
+        "Library" if IS_WIN32 else "lib",
+        "bin" if IS_WIN32 else "",
+    )
+    # Windows CUDA 13.0 uses "bin\x64" directory
```
rwgk: Maybe it's more accurate to change the comment to
```
    # Windows CUDA 13.0.0 uses "bin\x64" directory
```
?

I just got the super-fresh `13.0.1` packages (I don't know why they show as `13.0.88`):

```
(nvvm) PS C:\Users\rgrossekunst> conda list
# packages in environment at C:\Users\rgrossekunst\AppData\Local\miniforge3\envs\nvvm:
#
# Name                    Version                   Build  Channel
...
cuda-nvrtc                13.0.88              hac47afa_0    conda-forge
cuda-nvvm-impl            13.0.88              h2466b09_0    conda-forge
cuda-nvvm-tools           13.0.88              h2466b09_0    conda-forge
...
```

I see this:

```
rwgk-win11.localdomain:/mnt/c/Users/rgrossekunst/AppData/Local/miniforge3/envs/nvvm $ find . -name '*.dll' | grep -i nvrtc
./Library/bin/nvrtc-builtins64_130.dll
./Library/bin/nvrtc64_130_0.alt.dll
./Library/bin/nvrtc64_130_0.dll
rwgk-win11.localdomain:/mnt/c/Users/rgrossekunst/AppData/Local/miniforge3/envs/nvvm $ find . -name '*.dll' | grep -i nvvm
./Library/nvvm/bin/x64/nvvm64_40_0.dll
rwgk-win11.localdomain:/mnt/c/Users/rgrossekunst/AppData/Local/miniforge3/envs/nvvm $
```

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -62,78 +58,82 @@ def _find_valid_path(options):
...
 
-def _get_nvvm_wheel():
-    platform_map = {
-        "linux": ("lib64", "libnvvm.so"),
-        "win32": ("bin", "nvvm64_40_0.dll"),
-    }
+def _get_nvvm_wheel_path():
+    dso_path = None
+    # CUDA 12
+    try:
```
rwgk: `try`-`except ImportError` are notorious for masking errors elsewhere. (I know from several real-life cases that this can be extremely confusing.)

Maybe it's good enough for a stop-gap, but I'd try hard to use `importlib.metadata. distribution` instead for anything more permanent:

https://docs.python.org/3/library/importlib.metadata.html#distributions

E.g. for CTK 12:

```
>>> from importlib.metadata import distribution
>>> distribution("nvidia-cuda-nvcc-cu12")
<importlib.metadata.PathDistribution object at 0x7f423a8002f0>
>>>
```

It's `"nvidia-nvvm"` for CTK 13.

I think it'll be less code here with that approach.

rwgk (commented):
The code is indeed very difficult to follow. I made only one pass through, picking out a few things that stood out to me. Please let me know if you want me to spend time looking more carefully.

numba_cuda/numba/cuda/cuda_paths.py
kkraus14: Fixed this, both should have been tuples, thanks!

numba_cuda/numba/cuda/cuda_paths.py
kkraus14: Updated the comments, thanks!

numba_cuda/numba/cuda/cuda_paths.py
kkraus14: Moved everything to use `importlib.metadata.distribution`. As far as I can tell there isn't a great way to get to the specific paths of the packages without either iterating over all of the files from the distribution, or using it just to get to the site packages root and doing normal path construction from there. I took the route of getting the site packages root and constructing the path, but let me know what you think.

rwgk (commented):
Looks good to me, except I think it's worth the effort to work on enabling these checks:

```diff
diff --git a/numba_cuda/numba/cuda/cuda_paths.py b/numba_cuda/numba/cuda/cuda_paths.py
index 44abc3a..d566d58 100644
--- a/numba_cuda/numba/cuda/cuda_paths.py
+++ b/numba_cuda/numba/cuda/cuda_paths.py
@@ -325,9 +325,8 @@ def _get_nvvm_system_path():
     nvvm_path = os.path.join(
         nvvm_lib_dir, "nvvm64_40_0.dll" if IS_WIN32 else "libnvvm.so.4"
     )
-    # if os.path.isfile(nvvm_path):
-    #     return nvvm_path
-    return nvvm_path
+    if os.path.isfile(nvvm_path):
+        return nvvm_path


 def get_conda_ctk_libdir():
@@ -388,9 +387,8 @@ def _get_nvvm_conda_path():
     nvvm_path = os.path.join(
         nvvm_dir, "nvvm64_40_0.dll" if IS_WIN32 else "libnvvm.so.4"
     )
-    # if os.path.isfile(nvvm_path):
-    #     return nvvm_path
-    return nvvm_path
+    if os.path.isfile(nvvm_path):
+        return nvvm_path


 def get_wheel_static_libdir():
@@ -528,9 +526,8 @@ def _get_nvvm_cuda_home_path():
     nvvm_path = os.path.join(
         nvvm_lib_dir, "nvvm64_40_0.dll" if IS_WIN32 else "libnvvm.so.4"
     )
-    # if os.path.isfile(nvvm_path):
-    #     return nvvm_path
-    return nvvm_path
+    if os.path.isfile(nvvm_path):
+        return nvvm_path


 def _get_nvvm_path():
```

I approached this review as explained in the first prompt here:

https://chatgpt.com/share/68bcb5be-83fc-8008-832f-95dc66489c40

Copying out the relevant section for there for easy reference:

> My current time budget is 30-60 minutes, which isn't sufficient to really understand the very convoluted code (it has a very long history of adjusting to evolving environments).
>
> My goal is to identify high-level potential issues. I'm trusting that CI testing and general usage will shake out more subtle problems.
>
> Also, the code is meant to be replaced with a modern rewrite asap

After reviewing the full AI response, I decided the only actionable item is the diff above.

I cloned this PR and ran the tests, with the idea to then figure out an approach for enabling the `isfile` tests. I got a bit surprised by:

1. There is one failing test with your PR as-is. — I'm using native (**not** WSL) Ubuntu 24.04. I'll attach the full log.
2. After applying the diff, there is still only that one failing test. I'll attach the 2nd log, too.

Maybe I'm overlooking or missing something? — If there are test failures triggered by the `isfile` diffs, I think it'll be straightforward to mock out `os.path.isfile` (or similar) to make the tests pass without compromising production code quality.

I also reviewed `cuda_paths.py` again, reading through top-down, without trying to understand every single detail. Nothing major stood out to me, only sytle-kind of things; mainly I'd systematically explicitly `return None` for clarity. Concretely, I'd convert bare `return` to `return None`, or if a `return` is missing completely, `return None` explicitly, too.

rwgk:
This is how I ran the tests (`fresh_venv` and `now` are simple bash functions, essentially `python -m venv` and `date` with bells and whistles):

```
git clean -fdx
fresh_venv TestVenv
pip install -v -e .[test] cuda-bindings cuda-core
pytest -n auto --pyargs numba.cuda.tests -v |& tee /wrk/logs/numba.cuda.tests_log_`now`.txt
```

Between the two tests I stashed, stash-popped the `isfile` diff.

Log for PR as-is:

[numba.cuda.tests_log_2025-09-06+151724.txt](https://github.com/user-attachments/files/22192130/numba.cuda.tests_log_2025-09-06%2B151724.txt)

Log with `isfile` diff:

[numba.cuda.tests_log_2025-09-06+152001.txt](https://github.com/user-attachments/files/22192133/numba.cuda.tests_log_2025-09-06%2B152001.txt)

Both have this (or very similar) at the end:

```
=========================== short test summary info ============================
FAILED numba_cuda/numba/cuda/tests/cudapy/test_ipc.py::TestIpcStaged::test_ipc_array
= 1 failed, 1657 passed, 62 skipped, 11 xfailed, 173 warnings in 94.26s (0:01:34) =
```

kkraus14:
> =========================== short test summary info ============================
> FAILED numba_cuda/numba/cuda/tests/cudapy/test_ipc.py::TestIpcStaged::test_ipc_array
> = 1 failed, 1657 passed, 62 skipped, 11 xfailed, 173 warnings in 94.26s (0:01:34) =

Did this pass on `main` for you? I figured this was purely related to WSL and driver limitations that we don't account for in testing.

rwgk:
> > =========================== short test summary info ============================
> > FAILED numba_cuda/numba/cuda/tests/cudapy/test_ipc.py::TestIpcStaged::test_ipc_array
> > = 1 failed, 1657 passed, 62 skipped, 11 xfailed, 173 warnings in 94.26s (0:01:34) =
> 
> Did this pass on `main` for you? I figured this was purely related to WSL and driver limitations that we don't account for in testing.

It's the same failure on `main`:

[numba.cuda.tests_log_2025-09-07+192817.txt](https://github.com/user-attachments/files/22200317/numba.cuda.tests_log_2025-09-07%2B192817.txt)

I hadn't tried before, but did just now, exact same machine, venv, repo state:

```
commit 98a9213144be410f52f79802cfa38c4c5a0378c1 (HEAD -> main, upstream/main, upstream/HEAD)
Author: Vijay Kandiah <vkandiah@nvidia.com>
Date:   Thu Sep 4 17:48:18 2025 -0500

    Vendor in targetconfig, event, funcdesc, descriptors for CUDA-specific changes (#456)

    This PR vendors in `targetconfig`, `event`, `funcdesc`, `descriptors`
    from `numba.core` for CUDA-Specific customization.
```

This is on a native Ubuntu 24.04 machine. (There is at least one extra failure under WSL, and it deadlocked under WSL in one test; I didn't try to understand but simply moved to a colossus machine.)

The colossus machine:

```
:::::::::::::
/etc/lsb-release
::::::::::::::
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=24.04
DISTRIB_CODENAME=noble
DISTRIB_DESCRIPTION="Ubuntu 24.04.2 LTS"
::::::::::::::
/etc/os-release
::::::::::::::
PRETTY_NAME="Ubuntu 24.04.2 LTS"
NAME="Ubuntu"
VERSION_ID="24.04"
VERSION="24.04.2 LTS (Noble Numbat)"
VERSION_CODENAME=noble
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=noble
LOGO=ubuntu-logo
```

```
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 580.65.01              Driver Version: 580.65.01      CUDA Version: 13.0     |
+-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  Quadro RTX 8000                Off |   00000000:02:00.0 Off |                  Off |
| 34%   34C    P8              8W /  260W |       6MiB /  49152MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+
|   1  Quadro RTX 8000                Off |   00000000:21:00.0 Off |                  Off |
| 33%   41C    P8             18W /  260W |      48MiB /  49152MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|    0   N/A  N/A            2596      G   /usr/lib/xorg/Xorg                        4MiB |
|    1   N/A  N/A            2596      G   /usr/lib/xorg/Xorg                       25MiB |
|    1   N/A  N/A            2812      G   /usr/bin/gnome-shell                      7MiB |
+-----------------------------------------------------------------------------------------+
```

kkraus14:
> Looks good to me, except I think it's worth the effort to work on enabling these checks:
> 
> ```diff
> diff --git a/numba_cuda/numba/cuda/cuda_paths.py b/numba_cuda/numba/cuda/cuda_paths.py
> index 44abc3a..d566d58 100644
> --- a/numba_cuda/numba/cuda/cuda_paths.py
> +++ b/numba_cuda/numba/cuda/cuda_paths.py
> @@ -325,9 +325,8 @@ def _get_nvvm_system_path():
>      nvvm_path = os.path.join(
>          nvvm_lib_dir, "nvvm64_40_0.dll" if IS_WIN32 else "libnvvm.so.4"
>      )
> -    # if os.path.isfile(nvvm_path):
> -    #     return nvvm_path
> -    return nvvm_path
> +    if os.path.isfile(nvvm_path):
> +        return nvvm_path
> 
> 
>  def get_conda_ctk_libdir():
> @@ -388,9 +387,8 @@ def _get_nvvm_conda_path():
>      nvvm_path = os.path.join(
>          nvvm_dir, "nvvm64_40_0.dll" if IS_WIN32 else "libnvvm.so.4"
>      )
> -    # if os.path.isfile(nvvm_path):
> -    #     return nvvm_path
> -    return nvvm_path
> +    if os.path.isfile(nvvm_path):
> +        return nvvm_path
> 
> 
>  def get_wheel_static_libdir():
> @@ -528,9 +526,8 @@ def _get_nvvm_cuda_home_path():
>      nvvm_path = os.path.join(
>          nvvm_lib_dir, "nvvm64_40_0.dll" if IS_WIN32 else "libnvvm.so.4"
>      )
> -    # if os.path.isfile(nvvm_path):
> -    #     return nvvm_path
> -    return nvvm_path
> +    if os.path.isfile(nvvm_path):
> +        return nvvm_path
> 
> 
>  def _get_nvvm_path():
> ```

I've been looking at this and I don't see a clean way of doing this without either removing the existing tests or investing a lot of effort to essentially mock a filesystem for it. The tests currently test `nvvm`, `libdevice`, and the general CUDA lib directory. `libdevice` passes because the file isn't checked: https://github.com/NVIDIA/numba-cuda/pull/464/files#diff-682ac5ffbee833445b5101e759557a35163bda936522251370e29f9625147554R73-R80, similar for the CUDA lib directory we don't check for the existence of a specific folder or path: https://github.com/NVIDIA/numba-cuda/pull/464/files#diff-682ac5ffbee833445b5101e759557a35163bda936522251370e29f9625147554R482-R495.

In theory I could refactor the functions in `cuda_paths.py` to split them between path resolution and file / directory existence checking and then only use the path resolution functions in the tests, but I suspect there's still some overlap needed and the testing wouldn't really mirror real world usage.

This is only meant to be a stopgap until we rip all of this out and move to `cuda.pathfinder` anyway, so I would propose that we punt on this and just bring the learnings to continuing to improve and refine `cuda.pathfinder`.

rwgk:
> lot of effort to essentially mock a filesystem for it.

There is this very easy alternative (assumes you're using `pytest`):

```
def test_get_some_file(monkeypatch):
    monkeypatch.setattr("os.path.isfile", lambda _: True)
    assert get_some_file() == "/got/it/somehow.txt"
```

(I was wanting to give you a working patch like that, but since none of the tests failed I didn't know what to patch.)

The `monkeypatch` is specific to the test function and cleanly removed after the test finishes.

ZzEeKkAa:
Dumb question: is not it easier to switch to `cuda.pathfinder` at this point?

rwgk:
I still have to add `find_nvidia_header_dir` for CTK libs to cuda-pathfinder, and also some API for finding `libdevice`.

kkraus14:
> Dumb question: is not it easier to switch to `cuda.pathfinder` at this point?

I was concerned about detecting the distribution and finding things like `libdevice.10.bc` / `libcudadevrt.a` / header directories would have included as much or more hackery than what I did here.

kkraus14:
> > lot of effort to essentially mock a filesystem for it.
> 
> There is this very easy alternative (assumes you're using `pytest`):
> 
> ```
> def test_get_some_file(monkeypatch):
>     monkeypatch.setattr("os.path.isfile", lambda _: True)
>     assert get_some_file() == "/got/it/somehow.txt"
> ```
> 
> (I was wanting to give you a working patch like that, but since none of the tests failed I didn't know what to patch.)
> 
> The `monkeypatch` is specific to the test function and cleanly removed after the test finishes.

The test is spawning processes and running on these spawned processed so it's not straightforward to inject the monkeypatching into the spawned process without further refactoring unfortunately. I think my choices are unfortunately to either keep the filecheck removed or add the filecheck back and remove the `$CUDA_HOME` testing that currently exists (it isn't particularly compelling regardless).

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -251,120 +284,196 @@ def get_system_ctk(*subdirs):
         result = os.path.join("/usr/local/cuda", *subdirs)
         if os.path.exists(result):
             return result
+        return None
+    return None
 
 
-def get_conda_ctk():
```
brandon-b-miller, line 256: this is technically public facing, though likely used minimally, do we want to add some kind of deprecation warning around this?

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -251,120 +284,196 @@ def get_system_ctk(*subdirs):
...
+        nvvm_lib_dir = os.path.join(nvvm_lib_dir, "x64")
 
-def get_wheel_include():
-    try:
-        import nvidia
+    nvvm_path = os.path.join(
+        nvvm_lib_dir, "nvvm64_40_0.dll" if IS_WIN32 else "libnvvm.so.4"
+    )
+    # if os.path.isfile(nvvm_path):
+    #     return nvvm_path
```
brandon-b-miller, line 336:
```suggestion

```

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -251,120 +284,196 @@ def get_system_ctk(*subdirs):
...
-        return
 
-    paths = find_lib("nvvm", libdir=libdir)
-    if not paths:
-        return
-    # Use the directory name of the max path
-    return os.path.dirname(max(paths))
+def _get_nvvm_conda_path():
+    """Return path to directory containing the nvvm library."""
+    is_conda_env = os.path.isdir(os.path.join(sys.prefix, "conda-meta"))
```
brandon-b-miller, line 383: Does this require the env to be activated?

brandon-b-miller: Never mind I think this just depends on what the active `python` is

brandon-b-miller:
Broadly speaking is it worth adding any CI jobs here for testing installs with system packages only, or whatever conda edge cases we might not already cover?

numba_cuda/numba/cuda/cuda_paths.py
kkraus14: Yea, it only requires using the python from the environment. Do you have concerns here? In `cuda-pathfinder` I think we were planning to use the `CONDA_PREFIX` environment variable which is only set on environment activation.

brandon-b-miller: I was initially worried this might grab an nvvm from an unactivated environment but I don't think it's a concern.

numba_cuda/numba/cuda/cudadrv/libs.py
```diff
@@ -54,9 +54,9 @@ def get_cudalib(lib, static=False):
...
     if lib in {"nvrtc", "nvvm"}:
         return get_cuda_paths()[lib].info or _dllnamepattern % lib
-    else:
-        dir_type = "static_cudalib_dir" if static else "cudalib_dir"
-        libdir = get_cuda_paths()[dir_type].info
+
+    dir_type = "static_cudalib_dir" if static else "cudalib_dir"
+    libdir = get_cuda_paths()[dir_type].info
 
     candidates = find_lib(lib, libdir, static=static)
```
brandon-b-miller, line 61: Do we still need to go through `find_lib` here?

numba_cuda/numba/cuda/cuda_paths.py
kkraus14: I intended to keep this comment as a reminder for when we refactor this to use `cuda.pathfinder` and extend `cuda.pathfinder` testing to handle any differences.

numba_cuda/numba/cuda/cuda_paths.py
kkraus14: I did a GitHub code search for this and the only usages I could find were in this codebase or vendors of the `numba-cuda` / `numba` codebases. In general we don't have the best separation of public vs private API right now where I kinda figured this was buried enough that no one is using it.

kkraus14:
> Broadly speaking is it worth adding any CI jobs here for testing installs with system packages only, or whatever conda edge cases we might not already cover?

I was hoping for this code to be short-lived and rip it all out when we move to using `cuda.pathfinder` and we can invest the effort in CI environments for `cuda.pathfinder` instead?

numba_cuda/numba/cuda/cudadrv/libs.py
kkraus14: Yes because we only resolve the directory via `libdir` and then `find_lib` uses regex to find possible versioned candidates.

brandon-b-miller (approved):

PR #466, opened 2025-09-09
[Refactor][NFC] Vendor-in numba.extending for future CUDA-specific changes

atmnp:
This change vendors in numba.extending for future CUDA-specific changes. As part of the vendoring, many tests have been removed that test boxing/unboxing, rely on returning arrays allocated during a kernel execution, and other unsupported features on the GPU target. Additionally, there were tests that use the `EXTENDED_ARG` python opcode that were also removed since it's unclear if that feature is actually used.

### DIFF
--- a/numba_cuda/numba/cuda/extending.py
+++ b/numba_cuda/numba/cuda/extending.py
@@ -5,11 +5,36 @@
 Added for symmetry with the core API
 """
 
-from numba.core.extending import intrinsic as _intrinsic
+import os
+import uuid
+import weakref
+import collections
+import functools
+
+import numba
+from numba.core import types, errors, utils, config
+
+# # Exported symbols
+from numba.core.typing.typeof import typeof_impl  # noqa: F401
+from numba.core.typing.asnumbatype import as_numba_type  # noqa: F401
+from numba.core.typing.templates import infer, infer_getattr  # noqa: F401
+
+from numba.core.imputils import (  # noqa: F401
+    lower_builtin,
+    lower_getattr,
+    lower_getattr_generic,  # noqa: F401
+    lower_setattr,
+    lower_setattr_generic,
+    lower_cast,
+)  # noqa: F401
+from numba.cuda.core.pythonapi import box, unbox, reflect, NativeValue  # noqa: F401
+from numba._helperlib import _import_cython_function  # noqa: F401
+from numba.cuda.serialize import ReduceMixin
+from numba.core.datamodel import models as core_models  # noqa: F401
+
+
 from numba.cuda.models import register_model  # noqa: F401
 from numba.cuda import models  # noqa: F401
-
-intrinsic = _intrinsic(target="cuda")
 
 
 def make_attribute_wrapper(typeclass, struct_attr, python_attr):
@@ -18,7 +43,7 @@ def make_attribute_wrapper(typeclass, struct_attr, python_attr):
     as a read-only attribute named *python_attr*.
     The given *typeclass*'s model must be a StructModel subclass.
 
-    Vendored from numba.core.extending with a change to consider the CUDA data
+    Vendored from numba.cuda.extending with a change to consider the CUDA data
     model manager.
     """
     from numba.cuda.typing.templates import AttributeTemplate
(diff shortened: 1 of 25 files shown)


numba_cuda/numba/cuda/extending.py
```diff
@@ -63,3 +89,543 @@ def struct_getattr_impl(context, builder, typ, val):
...
+    be tried first or last. The default (`False`) is to use non-literal types.
+    Implementations that can specialize based on literal values should set the
+    option to `True`. Note, this option maybe expanded in the near future to
+    allow for more control (e.g. disabling non-literal types).
+
+    **kwargs prescribes additional arguments passed through to the overload
+    template. The only accepted key at present is 'target' which is a string
+    corresponding to the target that this overload should be bound against.
+    """
+    from numba.core.typing.templates import make_overload_template, infer_global
```
brandon-b-miller, line 199: are these available through `numba.cuda.typing` now? https://github.com/NVIDIA/numba-cuda/blob/ed5d280b96d0443df981e4140f02d6f648d5b6c5/numba_cuda/numba/cuda/typing/templates.py#L934

brandon-b-miller:
Are any of the tests that return arrays in upstream possibly portable now that NRT is enabled for the cuda target?

numba_cuda/numba/cuda/extending.py
```diff
@@ -63,3 +89,543 @@ def struct_getattr_impl(context, builder, typ, val):
...
+
+        @overload_attribute(types.Array, "nbytes")
+        def array_nbytes(arr):
+            def get(arr):
+                return arr.size * arr.itemsize
+
+            return get
+    """
+    # TODO implement setters
+    from numba.core.typing.templates import make_overload_attribute_template
```
brandon-b-miller, line 271: same comment as above https://github.com/NVIDIA/numba-cuda/blob/ed5d280b96d0443df981e4140f02d6f648d5b6c5/numba_cuda/numba/cuda/typing/templates.py#L1263

numba_cuda/numba/cuda/extending.py
```diff
@@ -63,3 +89,543 @@ def struct_getattr_impl(context, builder, typ, val):
...
+        def wrapper(func):
+            return _intrinsic(func)
+
+        return wrapper
+
+
+intrinsic = _intrinsic(target="cuda")
+
+
+def include_path():
```
brandon-b-miller: Fairly sure we don't need this, I think this just returns the path for the NRT CPU headers. Our equivalent for the cuda target is at https://github.com/NVIDIA/numba-cuda/blob/ed5d280b96d0443df981e4140f02d6f648d5b6c5/numba_cuda/numba/cuda/memory_management/nrt.py#L30.

numba_cuda/numba/cuda/tests/pdlike_usecase.py
```diff
@@ -0,0 +1,348 @@
...
+    NativeValue,
+    overload,
+    overload_attribute,
+    overload_method,
+    make_attribute_wrapper,
+)
+from numba.core.imputils import impl_ret_borrowed
+
+
+class Index(object):
```
brandon-b-miller: For this file generally I think if we're not going to have the tests from [upstream test_extending.py](https://github.com/numba/numba/blob/2e8a8ba1c3385c45ccc7a8bed0d141297f4265e1/numba/tests/test_extending.py) (which I don't see how we could with all the python c api interaction) I think we might as well not bring over the code

numba_cuda/numba/cuda/extending.py
atmnp: Partially. The definition for `make_overload_template` exists, but the `infer_global` had to be removed since we couldn't vendor in the builtins template registry (to prevent duplication). Using the helpers results in subclass instance check assertion failures from upstream, and we could hackily do inheritance to pass these (naive methods fail), but if we can put off using the problematic methods/classes/objects from templates until the end, it would be substantially cleaner. There are comments in `typing.templates` that Graham added that briefly talk about this hack. I'd be a fan of waiting to add the templates registry back in after we vendor in the rest of typing, so we don't escape into numba.core for these checks. Trying to add the registry in at this point, still introduces circular imports.

numba_cuda/numba/cuda/extending.py
```diff
@@ -5,20 +5,44 @@
...
+import uuid
+import weakref
+import collections
+import functools
+
+from numba.core import types, errors, config
+from numba.cuda import utils
+
+# # Exported symbols
+from numba.core.typing.typeof import typeof_impl  # noqa: F401
```
rparolin, line 17: non-blocking: What are these comments for?

atmnp: For development ease, we often re-export symbols at different locations to make the paths shorter/easier to write at use (despite not using them at the import location like above), and the noqa stops our formatter from yelling that it's unused. There's probably like a hundred of these and in the future, ideally, I'd like to see these removed but I can see the utility for now.

numba_cuda/numba/cuda/extending.py
brandon-b-miller: Thanks for this context, @atmnp .

brandon-b-miller (approved):

rparolin (approved):

rparolin:
lgtm!

brandon-b-miller:
Thanks @atmnp !

PR #467, opened 2025-09-09
[Refactor][NFC] Vendor-in numba.np components for future CUDA-specific changes

atmnp:
This change vendors in part of the numba.np components that can be vendored in at this time. The remaining components are linalg, npdatetime, or rely on those, which require a bit more work to hook up properly.

This PR has been temporarily abandoned in favor of https://github.com/NVIDIA/numba-cuda/pull/494 which is a bulk change.


rparolin (dismissed):
lgtm.

PR #468, opened 2025-09-10
[Refactor][NFC] Vendor-in cpython components for future CUDA-specific changes 

atmnp:
This change vendors in parts of the cpython components for future CUDA-specific changes. In particular, this contains the implementations that do not explicitly require additional tests from Numba to verify behavior. Additionally, there are a few small cleanup fixes such as removing the defunct CPUTarget that is unnecessary in our implementation of the CUDA target.

This PR has been temporarily abandoned in favor of https://github.com/NVIDIA/numba-cuda/pull/493 which is a bulk change.


rparolin (dismissed):
lgtm

numba_cuda/numba/cuda/cpython/unsafe/numbers.py
```diff
@@ -0,0 +1,63 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: BSD-2-Clause
+
+"""This module provides the unsafe things for targets/numbers.py"""
+
+from numba.core import types, errors
+from numba.core.extending import intrinsic
```
brandon-b-miller, line 7: Does this work with `numba.cuda.extending.intrinsic` now?

numba_cuda/numba/cuda/cpython/unsafe/tuple.py
```diff
@@ -0,0 +1,91 @@
...
+# SPDX-License-Identifier: BSD-2-Clause
+
+"""
+This file provides internal compiler utilities that support certain special
+operations with tuple and workarounds for limitations enforced in userland.
+"""
+
+from numba.core import types, typing, errors
+from numba.cuda.cgutils import alloca_once
+from numba.core.extending import intrinsic
```
brandon-b-miller, line 11: Same question about `intrinsic`

numba_cuda/numba/cuda/cpython/enumimpl.py
```diff
@@ -0,0 +1,103 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: BSD-2-Clause
+
+"""
+Implementation of enums.
```
brandon-b-miller, line 5: Could we vendor a few of the enum related tests? Things appear to work with the CUDA target:
```python

from enum import IntEnum
from numba import cuda
import numpy as np

class Color(IntEnum):
    RED = 1
    GREEN = 2
    BLUE = 3

@cuda.jit(device=True)
def is_primary(color):
    return int(color) in (Color.RED, Color.BLUE)

@cuda.jit
def kernel(input, output):
    idx = cuda.grid(1)
    if idx < input.size:
        output[idx] = is_primary(input[idx])

input = cuda.to_device([1, 2, 3])
output = cuda.to_device([0, 0, 0])

kernel[1, 3](input, output)
print(output.copy_to_host())
```
```
[1 0 1]
```

We don't need everything but things might map reasonably well to simple kernels that test the CPU implementations of `Enum` properties on perhaps a single cuda thread

numba_cuda/numba/cuda/cpython/rangeobj.py
```diff
@@ -0,0 +1,289 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: BSD-2-Clause
+
+"""
+Implementation of the range object for fixed-size integers.
```
brandon-b-miller, line 5: These also appear to be supported so it would be nice to include so basic tests 

```python
@cuda.jit
def kernel(output):
    idx = cuda.grid(1)
    if idx < output.size:
        r = range(10)
        sum = 0
        for i in r:
            sum += 1
        output[idx] = sum

output = cuda.to_device([0, 0, 0])

kernel[1, 3](output)
print(output.copy_to_host())
```
```
[10 10 10]
```

numba_cuda/numba/cuda/cpython/tupleobj.py
```diff
@@ -0,0 +1,456 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: BSD-2-Clause
+
+"""
+Implementation of tuple objects
```
brandon-b-miller, line 5: could use a few test cases since this ports to cuda. Things like iteration, possibly `__getitem__`, etc. 
```python

@cuda.jit
def kernel(output):
    idx = cuda.grid(1)
    if idx < output.size:
        t = (1, 2, 3)
        sum = 0
        for i in t:
            sum += 1
        output[idx] = sum

output = cuda.to_device([0, 0, 0])

kernel[1, 3](output)
print(output.copy_to_host())
```
```
[3 3 3]
```

PR #474, opened 2025-09-16
Mmason nvidia/bugfix/lto linking

mmason-nvidia:
Update linker options to only set link_time_optimization=True if LTO-IR code is being linked.

Move linker option creation to a separate function to reduce duplicate code,
and remove self.options in favor of computing the options on the fly.

### DIFF
--- a/numba_cuda/numba/cuda/cudadrv/driver.py
+++ b/numba_cuda/numba/cuda/cudadrv/driver.py
@@ -3065,7 +3065,9 @@ class _Linker(_LinkerBase):
         )
 
     def get_linked_ptx(self):
-        self.linker = Linker(*self._object_codes, options=self._get_linker_options(ptx=True))
+        self.linker = Linker(
+            *self._object_codes, options=self._get_linker_options(ptx=True)
+        )
 
         result = self.linker.link("ptx")
         self.close()
@@ -3078,7 +3080,9 @@ class _Linker(_LinkerBase):
         self.linker.close()
 
     def complete(self):
-        self.linker = Linker(*self._object_codes, options=self._get_linker_options(ptx=False))
+        self.linker = Linker(
+            *self._object_codes, options=self._get_linker_options(ptx=False)
+        )
         result = self.linker.link("cubin")
         self.close()
         self._complete = True
--- a/numba_cuda/numba/cuda/cudadrv/driver.py
+++ b/numba_cuda/numba/cuda/cudadrv/driver.py
@@ -2968,13 +2968,6 @@ class _Linker(_LinkerBase):
             lto = None
         self.lto = lto
         self.additional_flags = additional_flags
-
-        self.options = LinkerOptions(
-            max_register_count=self.max_registers,
-            lineinfo=lineinfo,
-            arch=arch,
-            link_time_optimization=lto,
-        )
         self._complete = False
         self._object_codes = []
         self.linker = None  # need at least one program
@@ -3058,16 +3051,21 @@ class _Linker(_LinkerBase):
 
         fn(data, name)
 
-    def get_linked_ptx(self):
-        options = LinkerOptions(
+    def _get_linker_options(self, ptx=False):
+        # Enable link time optimization if there is an LTO-IR object in the
+        # _object_codes list. This has to be deferred until now as it requires
+        # the full set of objects to be available.
+        has_ltoir = any(obj._code_type == "ltoir" for obj in self._object_codes)
+        return LinkerOptions(
             max_register_count=self.max_registers,
             lineinfo=self.lineinfo,
             arch=self.arch,
-            link_time_optimization=True,
-            ptx=True,
+            link_time_optimization=has_ltoir,
+            ptx=ptx and has_ltoir,
         )
 
-        self.linker = Linker(*self._object_codes, options=options)
+    def get_linked_ptx(self):
+        self.linker = Linker(*self._object_codes, options=self._get_linker_options(ptx=True))
 
         result = self.linker.link("ptx")
         self.close()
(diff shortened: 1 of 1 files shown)


numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -3058,16 +3051,23 @@ def add_data(self, data, kind, name):
 
         fn(data, name)
 
-    def get_linked_ptx(self):
-        options = LinkerOptions(
+    def _get_linker_options(self, ptx=False):
+        # Enable link time optimization if there is an LTO-IR object in the
+        # _object_codes list. This has to be deferred until now as it requires
+        # the full set of objects to be available.
+        has_ltoir = any(obj._code_type == "ltoir" for obj in self._object_codes)
```
brandon-b-miller: Hm, it would be nice if `ObjectCode._code_type` was exposed as a public property. @leofang ?

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -3058,16 +3051,23 @@ def add_data(self, data, kind, name):
...
+        return LinkerOptions(
             max_register_count=self.max_registers,
             lineinfo=self.lineinfo,
             arch=self.arch,
-            link_time_optimization=True,
-            ptx=True,
+            link_time_optimization=has_ltoir,
```
brandon-b-miller: would we still be able to force LTO even if just a bunch of PTX files were passed?

brandon-b-miller (commented):
Couple Q's otherwise LGTM

numba_cuda/numba/cuda/cudadrv/driver.py
mmason-nvidia: No, it's not possible. LTO happens at the NVVM level before the PTX is generated.

numba_cuda/numba/cuda/cudadrv/driver.py
leofang: Yeah why isn't it already exposed, I wonder? 🤔

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -3058,16 +3051,23 @@ def add_data(self, data, kind, name):
...
+            ptx=ptx and has_ltoir,
```
leofang: Q: Why don't we emit PTX if `has_ltoir` is False?

mmason-nvidia: According to the linker manual (https://docs.nvidia.com/cuda//pdf/nvjitlink.pdf page 14 section 2.3) the -ptx command line option is only valid when used with -lto:

▶ -lto Do link time optimization.
▶ -ptx Emit ptx after linking instead of cubin; only supported with -lto

brandon-b-miller:
Is this https://github.com/NVIDIA/numba-cuda/blob/97ce4b1525de474efbc93b45eea0668167fcce3b/numba_cuda/numba/cuda/cudadrv/driver.py#L2966-L2968 removable with this PR?

numba_cuda/numba/cuda/cudadrv/driver.py
leofang: Addressing this in https://github.com/NVIDIA/cuda-python/pull/890.

gmarkall:
What bug does this fix?

mmason-nvidia:
> What bug does this fix?

There was an issue with which flags we were passing to nvJitLink in the case of linking a JITed kernel with LTO-IR code where the kernel was compiled with debugging enabled. In particular: the ltoir flag should only be used if there is at least one LTO-IR object to be linked, and ptx=True is only supported if ltoir=True. Not all unsupported linker option combinations are checked for, so without this change undefined behavior will occur (usually in the form of an internal error from the linker).

mmason-nvidia:
> Is this
> 
> https://github.com/NVIDIA/numba-cuda/blob/97ce4b1525de474efbc93b45eea0668167fcce3b/numba_cuda/numba/cuda/cudadrv/driver.py#L2966-L2968
> 
> removable with this PR?

I think so. I'll give it a try.

gmarkall:
I get the impression that this PR is now causing us to silently accept an incorrect set of flags, instead of rejecting it, but it's hard to tell without a test case - can you add a test case representing the bug that is fixed here please?

numba_cuda/numba/cuda/cudadrv/driver.py
kkraus14: This is now a publicly exposed property under `.code_type`

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -3036,16 +3029,27 @@ def add_data(self, data, kind, name):
...
-        options = LinkerOptions(
+    def _get_linker_options(self, ptx=False):
+        # Enable link time optimization if there is an LTO-IR object in the
+        # _object_codes list. This has to be deferred until now as it requires
+        # the full set of objects to be available.
+        has_ltoir = any(obj._code_type == "ltoir" for obj in self._object_codes)
+
+        # Due to a bug in cuda.core linker flag creation, we need to pass in None
+        # instead of False for boolean values. Once cuda_core is fixed, we can
+        # remove this workaround.
```
kkraus14: This was fixed in https://github.com/NVIDIA/cuda-python/pull/989 and is included in the recently released `cuda.core 0.4.0`

numba_cuda/numba/cuda/cudadrv/driver.py
leofang: If we use the public attribute now, we'll need to bump the lower bound on cuda.core to 0.4.0.

If we don't use the public attribute, at one point when we cythonize `ObjectCode` (https://github.com/NVIDIA/cuda-python/issues/1081) this will likely break.

Just wanna call this out to make an informed decision. Whatever works for me 😛

mmason-nvidia:
> I get the impression that this PR is now causing us to silently accept an incorrect set of flags, instead of rejecting it, but it's hard to tell without a test case - can you add a test case representing the bug that is fixed here please?

Yes, I will update the PR in response to this and Keith's comment about https://github.com/NVIDIA/cuda-python/pull/989 being merged and released.

PR #475, opened 2025-09-16
Change dangling imports of numba.core.lowering to numba.cuda.lowering

VijayKandiah:
This is a small PR to change dangling imports of `numba.core.lowering` to `numba.cuda.lowering`.

### DIFF
--- a/numba_cuda/numba/cuda/testing.py
+++ b/numba_cuda/numba/cuda/testing.py
@@ -17,6 +17,7 @@ from pathlib import Path
 from typing import Iterable, Union
 from io import StringIO
 import unittest
+import numpy as np
 
 if PYVERSION >= (3, 10):
     from filecheck.matcher import Matcher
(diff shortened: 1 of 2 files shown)


brandon-b-miller (approved):

brandon-b-miller (changes requested):
This fails tests due to a circular import issue with the recent changes from https://github.com/NVIDIA/numba-cuda/pull/466 (I likely should have merged this PR first). AFAICT this issue also technically exists in the dependency graph of upstream numba as well, but the way numba's imports are set up we always happen to hit `ir_utils.py` before we hit `extending.py`, whereas with this PR the reverse will be true for `numba-cuda` exposing the issue. 

For now to fix this I think we may need to explicitly load `ir_utils` first or do a local import of `from numba.cuda.extending import _Intrinsic` inside `ir_utils.py` where needed.

brandon-b-miller:
cc @atmnp

VijayKandiah:
> This fails tests due to a circular import issue with the recent changes from #466 (I likely should have merged this PR first). AFAICT this issue also technically exists in the dependency graph of upstream numba as well, but the way numba's imports are set up we always happen to hit `ir_utils.py` before we hit `extending.py`, whereas with this PR the reverse will be true for `numba-cuda` exposing the issue.
> 
> For now to fix this I think we may need to explicitly load `ir_utils` first or do a local import of `from numba.cuda.extending import _Intrinsic` inside `ir_utils.py` where needed.

Thanks! I switched to a local import of `from numba.cuda.extending import _Intrinsic`.

brandon-b-miller (approved):
CI failure unrelated, approving

PR #476, opened 2025-09-17
Vendor in optional, boxing for CUDA-specific changes, fix dangling imports

VijayKandiah:
This change vendors in `numba.core.optional` and `numba.core.boxing` for future CUDA-specific customizations. Also changes a few dangling imports of submodules from `numba.core` to be of `numba.cuda.core`.


brandon-b-miller:
AFAICT this is a straight move of `optional.py` and `boxing.py` along with the usual trick with registry installation and fixing all the imports. The one question I had is I noticed in 0.61.0 the upstream `numba` `boxing.py` was moved to `old_boxing.py`, is this material to the goal here or just a refactor on the numba side?

brandon-b-miller (dismissed):

VijayKandiah:
> AFAICT this is a straight move of `optional.py` and `boxing.py` along with the usual trick with registry installation and fixing all the imports. The one question I had is I noticed in 0.61.0 the upstream `numba` `boxing.py` was moved to `old_boxing.py`, is this material to the goal here or just a refactor on the numba side?

Upstream numba has some modules like boxing named with prefix `old_` and `new_` and a `config.USE_LEGACY_TYPE_SYSTEM` in `boxing.py` to choose between the two. @gmarkall said it's OK to assume this config option always returns `True` and avoid vendoring in any code for the case when it's `False`.  https://github.com/NVIDIA/numba-cuda/pull/374#discussion_r2302196064

rparolin:
LGTM. Ideally, there would be tested that validate the boxing/unboxing state but we can bring those over in a future PR.

rparolin (approved):
lgtm

rparolin (approved):

brandon-b-miller (approved):

brandon-b-miller:
@VijayKandiah this should be ready to merge once conflicts are resolved

PR #478, opened 2025-09-17
[Refactor][NFC] Vendor-in more cpython registries for future CUDA-specific changes

atmnp:
This PR vendors in more of the cpython registries from upstream numba - charseq, iterators, listobj, slicing, and unicode. There are some components from cpython remaining and will be included in a future PR. Some dangling imports are also fixed.


numba_cuda/numba/cuda/core/inline_closurecall.py
```diff
@@ -34,7 +34,7 @@
     compute_live_variables,
 )
 from numba.core.imputils import impl_ret_untracked
-from numba.core.extending import intrinsic
+from numba.cuda.extending import intrinsic
```
brandon-b-miller, line 37: Just to confirm my understanding, this is leftover from previous efforts and should be cleaned up, but isn't directly related to the other changes in this PR correct?

numba_cuda/numba/cuda/cpython/charseq.py
```diff
@@ -0,0 +1,1214 @@
...
+        # return bytes(s).upper()  # TODO: implement upper for Bytes
+        return s._to_str().upper()._to_bytes()
+
+    return impl
+
+
+@overload_method(types.UnicodeCharSeq, "find")
+@overload_method(types.CharSeq, "find")
+@overload_method(types.Bytes, "find")
+def unicode_charseq_find(a, b):
```
brandon-b-miller, line 756: A lot of these lowerings seems to provide operations on some kind of string dtype. I'm not aware of support for strings in general for `numba-cuda`. Indeed, removing a bunch of these locally on my clone of this branch seems to still result in a passing state. Are all of these needed for the CUDA target?

numba_cuda/numba/cuda/cpython/unicode.py
```diff
@@ -0,0 +1,2863 @@
...
+    def __init__(self, dmm, fe_type):
+        members = [
+            ("data", types.voidptr),
+            ("length", types.intp),
+            ("kind", types.int32),
+            ("is_ascii", types.uint32),
+            ("hash", _Py_hash_t),
+            ("meminfo", types.MemInfoPointer(types.voidptr)),
+            # A pointer to the owner python str/unicode object
+            ("parent", types.pyobject),
```
brandon-b-miller, line 106: This sticks out to me as well as potentially not being needed for the GPU. Without bonafide python objects on the GPU, I'm guessing there's no currently supported python expressions that lead to a valid variable backed by this type.

numba_cuda/numba/cuda/core/inline_closurecall.py
atmnp: Yep, our policy so far has been to fix dangling imports as we see them.

brandon-b-miller:
Connected with @atmnp offline around the unicode/string generation pieces of things, leaving them in for now.

Issue #154, opened 2025-03-10
[MNT] Drop `CUDA_USE_NVIDIA_BINDING` and always use `cuda.core` (and fallback to `cuda.bindings`)


leofang:
Before cuda-bindings (the old cuda-python) was a thing, Numba already had ctype-based driver/runtime API bindings for many years. After cuda-python was out, there is a `CUDA_USE_NVIDIA_BINDING` that allows users to switch to the official bindings, but the default has always been off and needs users' explicit opt-in.

As part of the RFE #128 we should move to drop the ctypes-based bindings entirely. We always use `cuda.core` whenever possible, and use `cuda.bindings` as a fallback in case `cuda.core` has not caught up.

PR #479, opened 2025-09-19
[MNT] Drop NUMBA_CUDA_USE_NVIDIA_BINDING; always use cuda.core and cuda.bindings as fallback

rparolin:
### Summary

Historically, `numba-cuda` supported two binding layers (ctypes and cuda-python). This change simplifies maintenance and aligns with the long-term direction to use cuda.core while leveraging cuda.bindings to fill gaps.

### What changed?
- Remove the NUMBA_CUDA_USE_NVIDIA_BINDING environment variable and all ctypes-based CUDA paths.
- Require CUDA Python package at import.
- Always prefer cuda.core APIs; use cuda.bindings where required

### Breaking changes
- The NUMBA_CUDA_USE_NVIDIA_BINDING env var no longer exists.
- Legacy ctypes-based NVRTC/driver paths are removed.
- cuda/cuda.bindings must be installed for Numba-CUDA to import.

Fixes: https://github.com/NVIDIA/numba-cuda/issues/154

### DIFF
--- a/ci/test_conda_ctypes_binding.sh
+++ b/ci/test_conda_ctypes_binding.sh
@@ -76,7 +76,7 @@ popd
 
 
 rapids-logger "Run Tests"
-NUMBA_CUDA_USE_NVIDIA_BINDING=0 NUMBA_CUDA_TEST_BIN_DIR=$NUMBA_CUDA_TEST_BIN_DIR pytest --pyargs numba.cuda.tests -v
+NUMBA_CUDA_TEST_BIN_DIR=$NUMBA_CUDA_TEST_BIN_DIR pytest --pyargs numba.cuda.tests -v
 
 popd
 
--- a/docs/source/user/bindings.rst
+++ b/docs/source/user/bindings.rst
@@ -5,25 +5,19 @@
 CUDA Bindings
 =============
 
-Numba supports two bindings to the CUDA Driver APIs: its own internal bindings
-based on ctypes, and the official `NVIDIA CUDA Python bindings
-<https://nvidia.github.io/cuda-python/>`_. Functionality is equivalent between
-the two bindings.
-
-The internal bindings are used by default. If the NVIDIA bindings are installed,
-then they can be used by setting the environment variable
-``NUMBA_CUDA_USE_NVIDIA_BINDING`` to ``1`` prior to the import of Numba. Once
-Numba has been imported, the selected binding cannot be changed.
+Numba-CUDA uses the official `NVIDIA CUDA Python bindings
+<https://nvidia.github.io/cuda-python/>`_ for all CUDA Driver interactions.
+The legacy internal ctypes bindings have been removed and the
+``NUMBA_CUDA_USE_NVIDIA_BINDING`` environment variable no longer has any effect.
 
 
 Per-Thread Default Streams
 --------------------------
 
 Responsibility for handling Per-Thread Default Streams (PTDS) is delegated to
-the NVIDIA bindings when they are in use. To use PTDS with the NVIDIA bindings,
-set the environment variable ``CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM`` to
-``1`` instead of Numba's environmnent variable
-:envvar:`NUMBA_CUDA_PER_THREAD_DEFAULT_STREAM`.
+the NVIDIA bindings. To use PTDS, set the environment variable
+``CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM`` to ``1`` instead of Numba's
+environment variable :envvar:`NUMBA_CUDA_PER_THREAD_DEFAULT_STREAM`.
 
 .. seealso::
 
@@ -35,13 +29,5 @@ set the environment variable ``CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM`` to
 Roadmap
 -------
 
-In Numba 0.56, the NVIDIA Bindings will be used by default, if they are
-installed.
-
-In future versions of Numba:
-
-- The internal bindings will be deprecated.
-- The internal bindings will be removed.
-
-At present, no specific release is planned for the deprecation or removal of
-the internal bindings.
+The ctypes-based internal bindings have been removed in favor of the NVIDIA
+bindings. Future work focuses on expanding usage of ``cuda.core`` APIs.
--- a/docs/source/user/installation.rst
+++ b/docs/source/user/installation.rst
@@ -61,14 +61,9 @@ Configuration
 CUDA Bindings
 -------------
 
-Numba supports interacting with the CUDA Driver API via either the `NVIDIA CUDA
-Python bindings <https://nvidia.github.io/cuda-python/>`_ or its own ctypes-based
-bindings. Functionality is equivalent between the two binding choices. The
-NVIDIA bindings are the default, and the ctypes bindings are now deprecated.
-
-If you do not want to use the NVIDIA bindings, the (deprecated) ctypes bindings
-can be enabled by setting the environment variable
-:envvar:`NUMBA_CUDA_USE_NVIDIA_BINDING` to ``"0"``.
+Numba-CUDA uses the `NVIDIA CUDA Python bindings <https://nvidia.github.io/cuda-python/>`_
+for interacting with the CUDA Driver API. The legacy ctypes bindings and the
+``NUMBA_CUDA_USE_NVIDIA_BINDING`` environment variable have been removed.
 
 
 .. _cudatoolkit-lookup:
@@ -79,28 +74,13 @@ CUDA Driver and Toolkit search paths
 Default behavior
 ~~~~~~~~~~~~~~~~
 
-When using the NVIDIA bindings, searches for the CUDA driver and toolkit
-libraries use its `built-in path-finding logic <https://github.com/NVIDIA/cuda-python/tree/main/cuda_bindings/cuda/bindings/_path_finder>`_.
-
-Ctypes bindings (deprecated) behavior
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-When using the ctypes bindings, Numba searches for a CUDA toolkit installation
-in the following order:
-
-1. Conda-installed CUDA Toolkit packages
-2. Pip-installed CUDA Toolkit packages
-3. The environment variable ``CUDA_HOME``, which points to the directory of the
-   installed CUDA toolkit (i.e. ``/home/user/cuda-12``)
-4. System-wide installation at exactly ``/usr/local/cuda`` on Linux platforms.
-   Versioned installation paths (i.e. ``/usr/local/cuda-12.0``) are intentionally
-   ignored. Users can use ``CUDA_HOME`` to select specific versions.
+Searches for the CUDA driver and toolkit libraries use the NVIDIA bindings'
+`built-in path-finding logic <https://github.com/NVIDIA/cuda-python/tree/main/cuda_bindings/cuda/bindings/_path_finder>`_.
 
 In addition to the CUDA toolkit libraries, which can be installed by conda into
 an environment or installed system-wide by the `CUDA SDK installer
 <https://developer.nvidia.com/cuda-downloads>`_, the CUDA target in Numba also
-requires an up-to-date NVIDIA driver.  Updated NVIDIA drivers are also installed
-by the CUDA SDK installer, so there is no need to do both. If the ``libcuda``
-library is in a non-standard location, users can set environment variable
+requires an up-to-date NVIDIA driver. If the ``libcuda`` library is in a
+non-standard location, users can set environment variable
 :envvar:`NUMBA_CUDA_DRIVER` to the file path (not the directory path) of the
 shared library file.
--- a/numba_cuda/numba/cuda/__init__.py
+++ b/numba_cuda/numba/cuda/__init__.py
@@ -27,45 +27,22 @@ if getattr(config, "CUDA_ENABLE_PYNVJITLINK", None) is None:
         )
         config.CUDA_ENABLE_PYNVJITLINK = pynvjitlink_auto_enabled
 
-# Upstream numba sets CUDA_USE_NVIDIA_BINDING to 0 by default, so it always
-# exists. Override, but not if explicitly set to 0 in the envioronment.
-_nvidia_binding_enabled_in_env = _readenv(
-    "NUMBA_CUDA_USE_NVIDIA_BINDING", bool, None
-)
-if _nvidia_binding_enabled_in_env is False:
-    USE_NV_BINDING = False
-else:
-    USE_NV_BINDING = True
-    config.CUDA_USE_NVIDIA_BINDING = USE_NV_BINDING
-if config.CUDA_USE_NVIDIA_BINDING:
-    if not (
-        importlib.util.find_spec("cuda")
-        and importlib.util.find_spec("cuda.bindings")
-    ):
-        raise ImportError(
-            "CUDA bindings not found. Please pip install the "
-            "cuda-bindings package. Alternatively, install "
-            "numba-cuda[cuXY], where XY is the required CUDA "
-            "version, to install the binding automatically. "
-            "If no CUDA bindings are desired, set the env var "
-            "NUMBA_CUDA_USE_NVIDIA_BINDING=0 to enable ctypes "
-            "bindings."
-        )
+# Require NVIDIA CUDA bindings at import time
+if not (
+    importlib.util.find_spec("cuda")
+    and importlib.util.find_spec("cuda.bindings")
+):
+    raise ImportError(
+        "NVIDIA CUDA Python bindings not found. Install the 'cuda' package "
+        "(e.g. pip install nvidia-cuda-python or numba-cuda[cuXY])."
+    )
 
 if config.CUDA_ENABLE_PYNVJITLINK:
-    if USE_NV_BINDING and not pynvjitlink_auto_enabled:
+    if not pynvjitlink_auto_enabled:
         warnings.warn(
-            "Explicitly enabling pynvjitlink is no longer necessary. "
-            "NVIDIA bindings are enabled. cuda.core will be used "
-            "in place of pynvjitlink."
+            "Explicit pynvjitlink enablement is unnecessary; cuda.core will be "
+            "used where available."
         )
-    elif pynvjitlink_auto_enabled:
-        # Ignore the fact that pynvjitlink is enabled, because that was an
-        # automatic decision based on discovering pynvjitlink was present; the
-        # user didn't ask for it
-        pass
-    else:
-        raise RuntimeError("nvJitLink requires the NVIDIA CUDA bindings. ")
 
 if config.ENABLE_CUDASIM:
     from .simulator_init import *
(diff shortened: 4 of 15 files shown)


ci/test_conda_ctypes_binding.sh
```diff
@@ -76,7 +76,7 @@ popd
 
 
 rapids-logger "Run Tests"
-NUMBA_CUDA_USE_NVIDIA_BINDING=0 NUMBA_CUDA_TEST_BIN_DIR=$NUMBA_CUDA_TEST_BIN_DIR pytest --pyargs numba.cuda.tests -v
```
brandon-b-miller: If we remove the ctypes bindings I think we don't need this job at all anymore right?

rparolin: That's what I was suspected but I want to confirm with people more familar with the build configuration than I am currently.

rparolin: Deleted ci script.

docs/source/user/installation.rst
```diff
@@ -79,28 +74,13 @@ CUDA Driver and Toolkit search paths
...
 
 In addition to the CUDA toolkit libraries, which can be installed by conda into
 an environment or installed system-wide by the `CUDA SDK installer
 <https://developer.nvidia.com/cuda-downloads>`_, the CUDA target in Numba also
-requires an up-to-date NVIDIA driver.  Updated NVIDIA drivers are also installed
-by the CUDA SDK installer, so there is no need to do both. If the ``libcuda``
-library is in a non-standard location, users can set environment variable
+requires an up-to-date NVIDIA driver. If the ``libcuda`` library is in a
+non-standard location, users can set environment variable
 :envvar:`NUMBA_CUDA_DRIVER` to the file path (not the directory path) of the
```
gmarkall: I wonder if we should also remove this environment variable at the same time and the logic that uses it (`locate_driver_and_loader()` and `load_driver()` in driver.py) as I think it could interfere with the cuda.bindings / cuda.pathfinder mechanisms to locate the driver.

rparolin: That's a good point.  I can try to include it as part of this PR.

cpcloud:
@gmarkall @rparolin Anything blocking this PR other than merge conflicts?

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -1,292 +1,52 @@
...
+    from cuda.bindings import nvrtc as bindings_nvrtc  # type: ignore
 
-class NVRTC:
-    """
-    Provides a Pythonic interface to the NVRTC APIs, abstracting away the C API
-    calls.
+    _HAVE_BINDINGS = True
+except Exception:  # pragma: no cover - environment dependent
+    bindings_nvrtc = None
+    _HAVE_BINDINGS = False
```
cpcloud: This seems unnecessary, given that `cuda.bindings` was imported before along with `cuda.core` only if `USE_NV_BINDING` was `True`. Since that's always `True` after this PR, why are we adding additional import guards?

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -1,292 +1,52 @@
...
-        self.nvrtcGetLTOIRSize(program.handle, byref(lto_size))
-
-        lto = b" " * lto_size.value
-        self.nvrtcGetLTOIR(program.handle, lto)
-
-        return lto
+@functools.cache
+def _get_nvrtc_version():
+    if not _HAVE_BINDINGS:
+        raise RuntimeError(
```
cpcloud: This is an anti-pattern, since it allows broken installations. Let's just make `cuda.bindings` and `cuda.core` hard requirements by adding it as a dependency to `pyproject.toml`. We can then dispense with the UX of allowing broken installations.

docs/source/user/bindings.rst
```diff
@@ -5,25 +5,19 @@
...
-the two bindings.
-
-The internal bindings are used by default. If the NVIDIA bindings are installed,
-then they can be used by setting the environment variable
-``NUMBA_CUDA_USE_NVIDIA_BINDING`` to ``1`` prior to the import of Numba. Once
-Numba has been imported, the selected binding cannot be changed.
+Numba-CUDA uses the official `NVIDIA CUDA Python bindings
+<https://nvidia.github.io/cuda-python/>`_ for all CUDA Driver interactions.
+The legacy internal ctypes bindings have been removed and the
+``NUMBA_CUDA_USE_NVIDIA_BINDING`` environment variable no longer has any effect.
```
kkraus14: Does it make sense to reference an environment variable that is no longer functional in docs? I would understand if it was release notes, but I don't think it makes sense here.

Maybe update this to keep the first sentence about using the official bindings and add a second sentence along the lines of "numba-cuda used to have its own internal bindings and those APIs are kept for compatibility reasons but we recommend users to use cuda-python directly if needing to interact with the CUDA Driver or other CUDA libraries"?

docs/source/user/bindings.rst
```diff
@@ -5,25 +5,19 @@
...
 --------------------------
 
 Responsibility for handling Per-Thread Default Streams (PTDS) is delegated to
-the NVIDIA bindings when they are in use. To use PTDS with the NVIDIA bindings,
-set the environment variable ``CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM`` to
-``1`` instead of Numba's environmnent variable
-:envvar:`NUMBA_CUDA_PER_THREAD_DEFAULT_STREAM`.
+the NVIDIA bindings. To use PTDS, set the environment variable
+``CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM`` to ``1`` instead of Numba's
+environment variable :envvar:`NUMBA_CUDA_PER_THREAD_DEFAULT_STREAM`.
```
kkraus14, line 23: +1 about referencing a no longer relevant environment variable

docs/source/user/installation.rst
```diff
@@ -61,14 +61,9 @@ Configuration
...
-Python bindings <https://nvidia.github.io/cuda-python/>`_ or its own ctypes-based
-bindings. Functionality is equivalent between the two binding choices. The
-NVIDIA bindings are the default, and the ctypes bindings are now deprecated.
-
-If you do not want to use the NVIDIA bindings, the (deprecated) ctypes bindings
-can be enabled by setting the environment variable
-:envvar:`NUMBA_CUDA_USE_NVIDIA_BINDING` to ``"0"``.
+Numba-CUDA uses the `NVIDIA CUDA Python bindings <https://nvidia.github.io/cuda-python/>`_
+for interacting with the CUDA Driver API. The legacy ctypes bindings and the
+``NUMBA_CUDA_USE_NVIDIA_BINDING`` environment variable have been removed.
```
kkraus14: +1

numba_cuda/numba/cuda/codegen.py
```diff
@@ -26,10 +26,11 @@ def run_nvdisasm(cubin, flags):
     try:
         fd, fname = tempfile.mkstemp()
         with open(fname, "wb") as f:
-            if config.CUDA_USE_NVIDIA_BINDING:
+            # Support both ObjectCode (bindings/core) and raw bytes
+            if hasattr(cubin, "code"):
```
cpcloud: This seems unnecessary. Why not just remove the branch instead of changing the code?

docs/source/user/installation.rst
```diff
@@ -79,22 +74,8 @@ CUDA Driver and Toolkit search paths
...
-
-1. Conda-installed CUDA Toolkit packages
-2. Pip-installed CUDA Toolkit packages
-3. The environment variable ``CUDA_HOME``, which points to the directory of the
-   installed CUDA toolkit (i.e. ``/home/user/cuda-12``)
-4. System-wide installation at exactly ``/usr/local/cuda`` on Linux platforms.
-   Versioned installation paths (i.e. ``/usr/local/cuda-12.0``) are intentionally
-   ignored. Users can use ``CUDA_HOME`` to select specific versions.
+Searches for the CUDA driver and toolkit libraries use the NVIDIA bindings'
+`built-in path-finding logic <https://github.com/NVIDIA/cuda-python/tree/main/cuda_bindings/cuda/bindings/_path_finder>`_.
```
kkraus14, line 81: Maybe use the docs: https://nvidia.github.io/cuda-python/cuda-pathfinder/latest/generated/cuda.pathfinder.load_nvidia_dynamic_lib.html?

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -51,26 +49,27 @@
...
 from .drvapi import API_PROTOTYPES
-from .drvapi import cu_occupancy_b2d_size, cu_stream_callback_pyobj, cu_uuid
+from .drvapi import cu_occupancy_b2d_size, cu_stream_callback_pyobj
 from .mappings import FILE_EXTENSION_MAP
 from .linkable_code import LinkableCode, LTOIR, Fatbin, Object
 from numba.cuda.utils import cached_file_read
 from numba.cuda.cudadrv import enums, drvapi, nvrtc
 
-USE_NV_BINDING = config.CUDA_USE_NVIDIA_BINDING
+from cuda.bindings import driver as binding
```
kkraus14, line 59: nitpick: `binding` is ambiguous given we're binding to multiple libraries nowadays

numba_cuda/numba/cuda/__init__.py
```diff
@@ -27,45 +27,22 @@
...
 
 if config.CUDA_ENABLE_PYNVJITLINK:
-    if USE_NV_BINDING and not pynvjitlink_auto_enabled:
+    if not pynvjitlink_auto_enabled:
         warnings.warn(
-            "Explicitly enabling pynvjitlink is no longer necessary. "
-            "NVIDIA bindings are enabled. cuda.core will be used "
-            "in place of pynvjitlink."
+            "Explicit pynvjitlink enablement is unnecessary; cuda.core will be "
+            "used where available."
```
kkraus14: pynvjitlink has been removed so we can remove referencing it entirely

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -1,292 +1,52 @@
...
+try:
+    # Prefer cuda.core for compilation
+    from cuda.core.experimental import Program, ProgramOptions  # type: ignore
 
-    def __del__(self):
-        if self._handle:
-            self._nvrtc.destroy_program(self)
+    _HAVE_CORE = True
+except Exception:  # pragma: no cover - environment dependent
+    _HAVE_CORE = False
```
kkraus14: `cuda.core` is a required dependency so I don't think we need to do this?

numba_cuda/numba/cuda/cudadrv/nvrtc.py
kkraus14: Same here where `cuda.bindings` is now a required dependency so we don't need to do this?

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -488,12 +204,13 @@ def get_lowest_supported_cc():
...
-            raise RuntimeError(
-                f"{retcode.name} when calling nvrtcGetSupportedArchs()"
-            )
-        return [(arch // 10, arch % 10) for arch in archs]
-    else:
-        return NVRTC().get_supported_archs()
+    if not _HAVE_BINDINGS:
+        raise RuntimeError(
+            "NVIDIA CUDA bindings are required to query supported archs"
+        )
```
kkraus14:
```suggestion
```

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -377,70 +124,39 @@ def compile(src, name, cc, ltoir=False):
...
-            arch=arch,
-            include_path=includes,
-            relocatable_device_code=True,
-            link_time_optimization=ltoir,
-            name=name,
+    if not _HAVE_CORE:
+        raise NvrtcError(
+            "cuda.core.experimental.Program is required for compilation; "
+            "please install a recent NVIDIA CUDA Python package."
         )
```
kkraus14:
```suggestion
```

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -1,292 +1,52 @@
...
-        lto = b" " * lto_size.value
-        self.nvrtcGetLTOIR(program.handle, lto)
-
-        return lto
+@functools.cache
+def _get_nvrtc_version():
+    if not _HAVE_BINDINGS:
+        raise RuntimeError(
+            "NVIDIA CUDA bindings are required to determine NVRTC version"
+        )
```
kkraus14:
```suggestion
```

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -381,76 +109,33 @@ def compile(src, name, cc, ltoir=False, lineinfo=False, debug=False):
...
-        )
-
-        class Logger:
-            def __init__(self):
-                self.log = []
+    options = ProgramOptions(
+        arch=arch,
+        include_path=includes,
+        relocatable_device_code=True,
+        link_time_optimization=ltoir,
```
gmarkall, line 116: `debug` and `lineinfo` seem to be missing here.

gmarkall: (We are probably missing a test that debug and lineinfo are present for C++ code linked into Numba kernels - cc @isVoid)

numba_cuda/numba/cuda/cudadrv/driver.py
gmarkall: The heritage of this is that `driver` was ambiguous as this is the `driver` module with a global called `driver` (a `Driver` instance). `binding` seemed like the best choice at the time, but I'd probably go with
```suggestion
from cuda.bindings import driver as cuda_driver
```

now. I also find it confusing when I find references to `numba.cuda.cudadrv.binding` in other parts of the Numba-CUDA, along the lines of "what binding?" ambiguity as suggested by Keith.

I don't feel that we need to fix this in this PR, would be happy with it remaining for now and being changed next time there's some rework of this module.

rparolin:
> @gmarkall @rparolin Anything blocking this PR other than merge conflicts?

Just fixing the CI issues and addressing the feedback.   I want to get this sorted today.

numba_cuda/numba/cuda/cudadrv/driver.py
rparolin: I generally agree. I don't like to mix renames with functional changes because  if we ever have to revert more code ripples occur than necessary.

gmarkall:
I think if the CI issues and the lineinfo / debug options for NVRTC are resolved, then this will be good to go in.

numba_cuda/numba/cuda/cudadrv/nvrtc.py
isVoid: In the other PR I was hoping that [these tests](https://github.com/NVIDIA/numba-cuda/blob/5667cf857147857a36f720c62f2119e059a50163/numba_cuda/numba/cuda/tests/cudapy/test_compiler.py#L569-L612) can verify the arguments.

But agreed it's better to have a more direct `nvrtc.compile` test on these arguments.

numba_cuda/numba/cuda/cudadrv/driver.py
isVoid: My own suggestion: can we make it an one-off enforcement that for usage of `cuda.bindings` inside `numba.cuda.cudadrv`, always use the fully-qualified name? In the example below:

```python
binding.CUdevice_attribute
```
becomes 
```python
cuda.bindings.driver.CUdevice_attribute
```

numba_cuda/numba/cuda/tests/cudapy/test_compiler.py
```diff
@@ -557,12 +557,9 @@ def f(z, x, y):
                 link_obj = LinkableCode.from_path(link)
                 if link_obj.kind == "cu":
                     # if link is a cu file, result contains a compiled object code
-                    if cuda.config.CUDA_USE_NVIDIA_BINDING:
-                        from cuda.core.experimental import ObjectCode
+                    from cuda.core.experimental import ObjectCode
```
isVoid, line 560: Suggest moving this import to top level if we use cuda.core universally.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -120,20 +109,27 @@ def make_logger():
...
             nvjitlink_internal._inspect_function_pointer("__nvJitLinkVersion")
-            != 0
-        )
+            == 0
+        ):
+            return False
+        try:
+            ver = nvjitlink_internal.nvJitLinkVersion()
+            major, minor = ver
+            if (major, minor) < (12, 3):
```
gmarkall: This seems to be raising

```
AttributeError: module 'cuda.bindings._internal.nvjitlink' has no attribute 'nvJitLinkVersion'
```

I think you want something like

```suggestion
            from cuda.bindings import nvjitlink
            if nvjitlink.version() < (12, 3):
```

brandon-b-miller: I think https://github.com/NVIDIA/cuda-python/pull/907#discussion_r2356514889 might be a better way of doing this at this point

numba_cuda/numba/cuda/__init__.py
```diff
@@ -8,65 +8,15 @@
...
-        # Ignore the fact that pynvjitlink is enabled, because that was an
-        # automatic decision based on discovering pynvjitlink was present; the
-        # user didn't ask for it
-        pass
-    else:
-        raise RuntimeError("nvJitLink requires the NVIDIA CUDA bindings. ")
+# Require NVIDIA CUDA bindings at import time
+if not (
+    importlib.util.find_spec("cuda")
+    and importlib.util.find_spec("cuda.bindings")
```
cpcloud, line 14: This should be an impossible code path without actively trying to create a broken installation, so we can remove these checks.

gmarkall:
I'm intermittently seeing #517 with this PR as well. I'm still looking to understand if that's an existing bug or a side-effect of this PR.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -52,26 +50,27 @@
...
-if USE_NV_BINDING:
-    from cuda.bindings import driver as binding
-    from cuda.core.experimental import (
-        Linker,
-        LinkerOptions,
-        ObjectCode,
-    )
+# For backwards compatibility: indicate that the NVIDIA CUDA Python bindings are
+# in use. Older code checks this flag to branch on binding-specific behavior.
+USE_NV_BINDING = True
```
gmarkall, line 68: I'm not aware of code that checks `driver.USE_NV_BINDING`. The code I was thinking of that checks whether the NVIDIA bindings are in use checks `config.CUDA_USE_NVIDIA_BINDING`. E.g.: https://github.com/rapidsai/rmm/blob/branch-25.12/python/rmm/rmm/allocators/numba.py#L76

numba_cuda/numba/cuda/__init__.py
cpcloud: I will change this in a follow-up.

PR #485, opened 2025-09-23
Vendor in core.registry for CUDA-specific changes

VijayKandiah:
This PR vendors in `numba.core.registry` for future CUDA-specific customizations.


rparolin (dismissed):

rparolin:
Please add a PR description before committing.

numba_cuda/numba/cuda/core/inline_closurecall.py
```diff
@@ -325,7 +325,9 @@ def check_reduce_func(func_ir, func_var):
                             analysis"
         )
     if isinstance(reduce_func, (ir.FreeVar, ir.Global)):
-        if not isinstance(reduce_func.value, numba.core.registry.CPUDispatcher):
+        if not isinstance(
```
VijayKandiah: Can we remove this instance check that raises a ValueError if found True? It would help remove registry's dependency on `numba.cuda.core.cpu`

numba_cuda/numba/cuda/core/ir_utils.py
```diff
@@ -829,7 +829,7 @@ def has_no_side_effect(rhs, lives, call_table):
             or call_list[0]._name == "unsafe_empty_inferred"
         ):
             return True
-        from numba.core.registry import CPUDispatcher
+        from numba.cuda.core.registry import CPUDispatcher
         from numba.np.linalg import dot_3_mv_check_args
 
         if isinstance(call_list[0], CPUDispatcher):
```
VijayKandiah: Another instance check that I do not think is needed here.

numba_cuda/numba/cuda/cudadrv/ndarray.py
```diff
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: BSD-2-Clause
 
 from numba.cuda.cudadrv import devices, driver
-from numba.core.registry import cpu_target
+from numba.cuda.core.registry import cpu_target
 
 
 def _calc_array_sizeof(ndim):
```
VijayKandiah: `numba.cuda.core.registry.cpu_target` is only used in `_calc_array_sizeof` which does not seem to be used anywhere in numba-cuda. Could we remove `_calc_array_sizeof` entirely? This would help remove `registry`'s dependency on `numba.cuda.core.cpu`

numba_cuda/numba/cuda/core/inline_closurecall.py
brandon-b-miller: I _think_ that what this code is doing is ensuring that the variable `reduce_func` is essentially a CPU jitted function. a `CPUDispatcher` is what we get when using normal numba jit on a python func:

```python
>>> from numba import jit
>>> @jit
... def f():
...     pass
... 
>>> f
CPUDispatcher(<function f at 0x7f2949833910>)
```

I think we need to define what we want to support here. I think having this code might essentially add to a world where we have our _own_ `@jit` decorator from `numba-cuda`, except it works for CPU functions. This raises a question. The following is currently supported where we can use a numba CPU jit function within a kernel:

```python
>>> from numba import cuda
>>> from numba import jit
>>> @jit
... def f():
...     print('hello world!')
...     
>>> @cuda.jit
... def kernel():
...     f()
...     
>>> kernel[1,1]()
hello world!
```
I'm not sure we want to go down the road of having our own `@jit` and I think we need to think about the tradeoffs of having that vs just supporting actual upstream `CPUDispatcher` functions.

numba_cuda/numba/cuda/core/ir_utils.py
```diff
@@ -829,7 +829,7 @@ def has_no_side_effect(rhs, lives, call_table):
             or call_list[0]._name == "unsafe_empty_inferred"
         ):
             return True
-        from numba.core.registry import CPUDispatcher
+        from numba.cuda.core.registry import CPUDispatcher
```
brandon-b-miller: Instead of having this class internally at all could we make it so that we try/except around an upstream `numba` import and check against the class if it exists? Stuff related to numba CPU works if numba CPU is present rather than having our own copies of the CPU stuff.

VijayKandiah: I agree with this approach after discussing with @gmarkall. In my latest commit, I removed copies of the CPU stuff in `core.registry`, and removed `core.cpu` entirely.

brandon-b-miller (approved):
LGTM

PR #481, opened 2025-09-19
[DO NOT MERGE] Debug llvmlite Windows wheel

leofang:

PR #487, opened 2025-09-23
Restrict `llvmlite` to `<0.45` for windows only 

brandon-b-miller:
xref https://github.com/NVIDIA/numba-cuda/pull/481, https://github.com/numba/llvmlite/issues/1297

### DIFF
--- a/ci/test_wheel.ps1
+++ b/ci/test_wheel.ps1
@@ -32,7 +32,7 @@ rapids-logger "Install wheel with test dependencies"
 $package = Resolve-Path wheel\numba_cuda*.whl | Select-Object -ExpandProperty Path
 echo "Package path: $package"
 python -m pip install "${package}[cu${CUDA_VER_MAJOR},test-cu${CUDA_VER_MAJOR}]"
-python -m pip install "llvmlite<0.45" # WAR for https://github.com/numba/llvmlite/issues/1297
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -20,7 +20,10 @@ authors = [
 license = "BSD-2-clause"
 license-files = ["LICENSE", "LICENSE.numba"]
 requires-python = ">=3.9"
-dependencies = ["numba>=0.60.0"]
+dependencies = [
+    "numba>=0.60.0",
+    "llvmlite<0.45; platform_system == 'Windows'"
```
kkraus14: I don't think we want to do this in `pyproject.toml`. The issue only surfaces if you have an environment with an older `msvcp140.dll` which isn't particularly common, but our CI system happens to have.

I think we want to just temporarily restrict the version in CI, but not in the package itself.

brandon-b-miller: Ah that makes sense, thanks @kkraus14 will push some updates here

kkraus14:
@brandon-b-miller is there something we should do to address the `test-thirdparty` job in this PR?

brandon-b-miller:
> @brandon-b-miller is there something we should do to address the `test-thirdparty` job in this PR?

This was just due to a broken cuDF nightly that should be fixed now. With the 25.10 cuDF release we'll be switching back to testing against a stable release here which should reduce a lot of the CI noise.

PR #488, opened 2025-09-23
[MNT] Managed UM memset fallback and skip CUDA IPC tests on WSL2

rparolin:
Summary
- Resolves unit tests failures on WSL2.
- Make device_memset robust for managed memory: on CUDA_ERROR_INVALID_VALUE for host-attached UM, fall back to a host-side fill.
- Add skip_on_wsl2 test decorator and skip CUDA IPC tests on WSL2, where IPC is unreliable despite R510+ drivers.


numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -3452,7 +3455,24 @@ def device_memset(dst, val, size, stream=0):
     else:
         fn = driver.cuMemsetD8
 
-    fn(device_pointer(dst), val, size, *varargs)
+    try:
+        fn(ptr, val, size, *varargs)
+    except CudaAPIError as e:
+        invalid = (
```
gmarkall, line 3461: How do you create a situation which this code is intended to handle? Is there an example that could be turned into a test?

rparolin: The test suite was already catching it.  I'm was running into unit test failures running on WSL.  One was lack of support for IPC and the other was memset throwing errors when attempting to access memory accessible by host and device.  This fallback enables the test suite to complete successfully.

PR #490, opened 2025-09-25
[test] Remove dependency on cpu_target

ashermancinelli:
Parts of test_ir_utils was vendored in from numba and repurposed for the cuda target, however it was still using the CPU target. We may not need the CPU target at all; try to remove dependency on it.

PR #491, opened 2025-09-29
[test] Use numpy's tolerance for float16

ashermancinelli:
Some CI runs were spuriously failing due to overly stringent tolerance for fp16 tests (#490). Use numpy's epsilon for 16b floats instead of the default relative tolerance for np.testing.assert_allclose().

This PR passes `CUDATestCase.FLOAT16_RTOL` to relevant calls to `np.testing.assert_allclose`.
Alternatively, we could add `assertAllClose` to `CUDATestCase` that infers the appropriate relative tolerance from the data types passed in.
Please let me know if we'd like to go that route instead. Thanks in advance!

### DIFF
--- a/numba_cuda/numba/cuda/testing.py
+++ b/numba_cuda/numba/cuda/testing.py
@@ -17,6 +17,7 @@ from pathlib import Path
 from typing import Iterable, Union
 from io import StringIO
 import unittest
+import numpy as np
 
 if PYVERSION >= (3, 10):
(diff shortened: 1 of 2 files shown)


ashermancinelli:
/ok to test [213e8bf](https://github.com/NVIDIA/numba-cuda/pull/491/commits/213e8bf35ce31d9e58278e484d948cd617c96cde)

PR #492, opened 2025-09-30
[test] Move tests outside package directory

ashermancinelli:
Existing tests are part of the `numba_cuda` package, which is then redirected from `numba.cuda`. For running tests against an installed numba-cuda, we either need to run tests from the installed package or move the tests outside of the package, but running tests from the source directory when the package has been installed without being editable (as is done in CI) is problematic.

When we run tests from the source directory, some components are imported from the source directory even though the package was installed in a non-editable way. This is particularly problematic with the C extensions, which end up as DSOs in the installation prefix, not the source directory.

In summary, `numba.cuda.tests` becomes `numba_cuda_tests` and `numba.cuda.testing.support` becomes `numba.cuda.testing_support`. The latter must remain in the numba-cuda package to the same extent as `numba.cuda.testing`. They both ought to be in or out of the package because they are coupled; I chose to have them both _in_ the package. If reviewers would like them both outside the package, please let me know.

TODO:
- [x] move tests to `tests/` directory
- [ ] (maybe) move `numba_cuda` under `src` (this will involve other changes, eg to wheels/pyproject.tom/etc)
- [ ] The data directory may be moved under the test directory
- [ ] Support modules may be moved into the test directory

### DIFF
--- a/docs/source/user/cuda_ffi.rst
+++ b/docs/source/user/cuda_ffi.rst
@@ -255,9 +255,9 @@ The definitions of the C++ template APIs are in two different locations:
    :caption: ``tests/doc_examples/ffi/include/mul.cuh``
    :linenos:
 
-.. literalinclude:: ../../../tests/data/include/add.cuh
+.. literalinclude:: ../../../numba_cuda/numba/cuda/test_data/include/add.cuh
    :language: C
-   :caption: ``tests/data/include/add.cuh``
+   :caption: ``numba_cuda/numba/cuda/test_data/include/add.cuh``
    :linenos:
 
 Neither of the headers are in the default search paths of Numba-CUDA, but the
@@ -265,7 +265,7 @@ foreign device function ``saxpy`` depends on them:
 
 .. literalinclude:: ../../../tests/doc_examples/ffi/saxpy.cu
    :language: C
-   :caption: ``tests/data/doc_examples/ffi/saxpy.cu``
+   :caption: ``tests/doc_examples/ffi/saxpy.cu``
    :linenos:
 
 In the Python code, assume that ``mul_dir`` and ``add_dir`` are set to the


ashermancinelli:
/ok to test [c22a77e](https://github.com/NVIDIA/numba-cuda/pull/492/commits/c22a77e30472c8ecac620f4e99f994c07d080fc7)

ashermancinelli:
/ok to test [f1a80db](https://github.com/NVIDIA/numba-cuda/pull/492/commits/f1a80dbec15bd7fc9d21232e8857f7e682e55ece)

ashermancinelli:
/ok to test [d22e099](https://github.com/NVIDIA/numba-cuda/pull/492/commits/d22e0999701c5a4a107aa489f0c7163a064ff487)

ashermancinelli:
/ok to test [64b71a9](https://github.com/NVIDIA/numba-cuda/pull/492/commits/64b71a9a5a8b742919470078606b1ee27500e14c)

ashermancinelli:
/ok to test [898e427](https://github.com/NVIDIA/numba-cuda/pull/492/commits/898e4277ac0acf4c3d22793e762429ada36e88b2)

leofang:
I am not an active Numba developer so take my words with a grain of salt, but I am supportive for this move and here's my 2c 🙂 I understand the desire of shipping tests together with a package, so that after installation one could validate the package immediately. However, the same can be done too with a carefully crafted CLI utility (ex: `python -m numba.test ...`) that downloads and executes the tests from the public git tag. I have been bothered by projects shipping tests with the package, for same reasons outlined in the PR description, appearance as importable modules, as well as binary size concerns.

That said, I think it's currently documented (at least in upstream) that tests are shipped as part of the package:
https://numba.readthedocs.io/en/stable/developer/repomap.html
It'd be nice to check with stakeholders whether this would be too disruptive to downstream (I don't think so).

ashermancinelli:
@leofang thank you for your input!

Those docs are for upstream numba, [ours are here](https://nvidia.github.io/numba-cuda/) IIUC. Maybe we should conform to upstream numba, but at least we have not committed to/documented shipping our test code, I think. I will check with other stakeholders. Cheers!

ashermancinelli:
/ok to test [1a3b504](https://github.com/NVIDIA/numba-cuda/pull/492/commits/1a3b504d7b90a70077f847a6586077a4a212f6ce)

ashermancinelli:
/ok to test [c65d59c](https://github.com/NVIDIA/numba-cuda/pull/492/commits/c65d59cdb658c9426fca7caf590328c5a7f0f47e)

ashermancinelli:
/ok to test [ef2ccbc](https://github.com/NVIDIA/numba-cuda/pull/492/commits/ef2ccbc7f3111b246b28a0783e28bf543d7bb416)

ashermancinelli:
/ok to test [2d61b2d](https://github.com/NVIDIA/numba-cuda/pull/492/commits/2d61b2d4a66cc3c6c2954e2f557dd438b9dc6b89)

ashermancinelli:
/ok to test [1521222](https://github.com/NVIDIA/numba-cuda/pull/492/commits/152122221f0d6d559c0f1a8f9d141a7d12c79d69)

ashermancinelli:
/ok to test [a6d862f](https://github.com/NVIDIA/numba-cuda/pull/492/commits/a6d862f6acdf4de044310ea1947cb6967ff33c0c)

ashermancinelli:
/ok to test [0f8ecea](https://github.com/NVIDIA/numba-cuda/pull/492/commits/0f8ecea6d15ad73fefafde353f46c4d142cb9d4b)

ashermancinelli:
Note that I had to add the tests directory to the python path for some of the suites. This is because some tests import from files local to the test directory and it's still treated like its own package. These may be changed to relative imports if it proves problematic (eg if deveopers need to `PYTHONPATH=$PYTHONPATH:$PWD/tests pytest ./tests` or similar).

ZzEeKkAa:
> need to run tests from the installed package or move the tests outside of the package, but running tests from the source directory when the package has been installed without being editable (as is done in CI) is problematic.

Why do you need to run tests from the source directory once package has been installed without editable mode on?
I usually run tests `pytest --pyargs numba.cuda.tests` or so (not sure if I tested it but this should work) to run them from installed directory.

ashermancinelli:
That works in the top of tree, but our problems were with the C extensions PR. Thanks for the suggestions though, it seems Vijay and Graham have another workaround working for the C extensions PR. Thank you @ZzEeKkAa!

PR #493, opened 2025-09-30
[Refactor][NFC] Vendor-in numba.cpython modules for future CUDA-specific changes

atmnp:
This is a bulk change that vendors in all of the modules in numba.cpython without tests to prevent blocking of other components that need to be vendored in. All of the relevant tests will be vendored in gradually in other PRs once it becomes a higher priority to ensure comprehensive test coverage.

### DIFF
--- a/numba_cuda/numba/cuda/core/base.py
+++ b/numba_cuda/numba/cuda/core/base.py
@@ -1040,7 +1040,7 @@ class BaseContext(object):
         return self._make_helper(builder, typ, ref=ref, kind="data")
 
     def make_array(self, typ):
-        from numba.np import arrayobj
+        from numba.cuda.np import arrayobj
 
         return arrayobj.make_array(typ)
 
@@ -1048,7 +1048,7 @@ class BaseContext(object):
         """
         Populate array structure.
         """
-        from numba.np import arrayobj
+        from numba.cuda.np import arrayobj
 
         return arrayobj.populate_array(arr, **kwargs)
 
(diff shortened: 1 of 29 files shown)


numba_cuda/numba/cuda/cpython/builtins.py
gmarkall, line 1: Overloads in this file should use `target="cuda"`, e.g.:

```python
@overload(operator.truth, target="cuda")
```

numba_cuda/numba/cuda/cpython/charseq.py
gmarkall, line 1: `overload` and `overload_method` in this file should use `target="cuda"`.

gmarkall (changes requested):
In general the vendoring looks good. The only change I would suggest is that the overloads are re-registering everything for the CPU and CUDA targets - the `@overload` and `@overload_method` decorators should use `target="cuda"` so that we're not getting in front of overloads built in to Numba.

gmarkall:
(I just noticed I left a comment about overloads on a couple of files before I submitted the review - however, all files that use overloads should make sure they're only overloading for the CUDA target)

numba_cuda/numba/cuda/cpython/builtins.py
atmnp: Done, but once we use the vendored in the @overload decorator and use it from numba-cuda, do we plan on removing this kwarg?

atmnp: Nevermind, just saw https://github.com/NVIDIA/numba-cuda/pull/511

PR #494, opened 2025-09-30
[Refactor][NFC] Vendor-in numba.np modules for future CUDA-specific changes

atmnp:
This is a bulk change that vendors in all of the modules in numba.np without tests to prevent blocking of other components that need to be vendored in. All of the relevant tests will be vendored in gradually in other PRs once it becomes a higher priority to ensure comprehensive test coverage.

### DIFF
--- a/numba_cuda/numba/cuda/extending.py
+++ b/numba_cuda/numba/cuda/extending.py
@@ -136,6 +136,7 @@ def overload(
     strict=True,
     inline="never",
     prefer_literal=False,
+    target="cuda",
     **kwargs,
 ):
     """
@@ -205,6 +206,8 @@ def overload(
     # TODO: abort now if the kwarg 'target' relates to an unregistered target,
     # this requires sorting out the circular imports first.
 
+    kwargs["target"] = target
+
     def decorate(overload_func):
         template = make_overload_template(
             func, overload_func, opts, strict, inline, prefer_literal, **kwargs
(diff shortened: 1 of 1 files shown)


numba_cuda/numba/cuda/np/arrayobj.py
```diff
@@ -0,0 +1,7690 @@
...
+
+
+def raise_with_shape_context(src_shapes, index_shape):
+    """Targets should implement this if they wish to specialize the error
+    handling/messages. The overload implementation takes two tuples as arguments
+    and should raise a ValueError."""
+    raise NotImplementedError
+
+
+@overload(raise_with_shape_context, target="generic")
```
atmnp: @gmarkall Is it correct to replace these "generic" targets with "cuda"? I imagine at some point we'd want to define our own overloads, but for now, is this a valid option to avoid duplicate registrations while maintaining functionality?

numba_cuda/numba/cuda/np/arrayobj.py
```diff
@@ -0,0 +1,7690 @@
...
+        return llintp(cgctx.get_abi_sizeof(llty))
+
+    return sig, codegen
+
+
+def _compatible_view(a, dtype):
+    pass
+
+
+@overload(_compatible_view, target="generic")
```
atmnp: Same as above.

numba_cuda/numba/cuda/np/arrayobj.py
gmarkall: "generic" is the default, so I think we should just remove the target kwarg, and merge #511 so that "cuda" is the default anyway.

gmarkall:
I'm re-running the tests post-#511 and will set it to auto-merge.

PR #497, opened 2025-10-01
Use numba.core.config when available, otherwise use numba.cuda.core.config

VijayKandiah:
Following discussions with @gmarkall, this PR removes the dynamic checking of configuration variables from `numba.core.config` in `numba.cuda.core.config`. We instead redirect `numba.cuda.core.config` module directly to `numba.core.config` if available, and use the local implementation only when it's not.

### DIFF
--- a/numba_cuda/numba/cuda/core/config.py
+++ b/numba_cuda/numba/cuda/core/config.py
@@ -124,49 +124,6 @@ def _process_opt_level(opt_level):
         return _OptLevel(opt_level)
 
 
-class _EnvVar(object):
-    """Descriptor for configuration values that checks numba.config on access."""
-
-    def __init__(self, value, name):
-        self.name = name
-        if isinstance(value, _EnvVar):
-            self.value = value.__get__()
-        else:
-            self.value = value
-        self.check_numba_config()
-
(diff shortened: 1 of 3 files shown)


numba_cuda/numba/cuda/core/config.py
```diff
@@ -348,12 +289,6 @@ def optional_str(x):
         # Enable NRT statistics counters
         NRT_STATS = _readenv("NUMBA_NRT_STATS", int, 0)
 
-        # Enable NRT statistics
-        CUDA_NRT_STATS = _readenv("NUMBA_CUDA_NRT_STATS", int, 0)
```
brandon-b-miller, line 352: Is it possible for these two to stay here? It'd be nice to have them with the rest of the config options.

VijayKandiah: I am inclined to leaving these two out in `nrt.py` where it was initially present before I vendored in config. This is primarily because none of those configuration variables in `numba.cuda.core.config` are exposed when `numba.config` is present, as we simply redirect the module to `numba.config`, and these two variables are not initialized in `numba.config`.

PR #499, opened 2025-10-02
Vendor in typeconv for future CUDA-specific changes

VijayKandiah:
This PR vendors in `numba.core.typeconv` for CUDA-specific customizations. The `_typeconv` C extension can be vendored in after PR https://github.com/NVIDIA/numba-cuda/pull/373 gets merged in; it already contains the necessary cpp and hpp files for `_typeconv`.

### DIFF
--- a/numba_cuda/numba/cuda/typeconv/typeconv.py
+++ b/numba_cuda/numba/cuda/typeconv/typeconv.py
@@ -1,0 +1,139 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: BSD-2-Clause
+
+try:
+    # This is usually the the first C extension import performed when importing
+    # Numba, if it fails to import, provide some feedback
+    from numba.core.typeconv import _typeconv
+except ImportError as e:
+    base_url = "https://numba.readthedocs.io/en/stable"
+    dev_url = f"{base_url}/developer/contributing.html"
+    user_url = f"{base_url}/user/faq.html#numba-could-not-be-imported"
+    dashes = "-" * 80
+    msg = (
+        f"Numba could not be imported.\n{dashes}\nIf you are seeing this "
+        "message and are undertaking Numba development work, you may need "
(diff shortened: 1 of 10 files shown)


VijayKandiah:
With my latest commit, this PR also vendors in the `_typeconv` C extension for CUDA-specific changes.

numba_cuda/numba/cuda/typeconv/typeconv.py
```diff
@@ -0,0 +1,139 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: BSD-2-Clause
+
+try:
+    # This is usually the the first C extension import performed when importing
+    # Numba, if it fails to import, provide some feedback
+    from numba.cuda.cext import _typeconv
+except ImportError as e:
+    base_url = "https://numba.readthedocs.io/en/stable"
```
VijayKandiah: Do we have equivalent doc urls for numba-cuda?

gmarkall: We don't - I think given all the changes and the way we package numba-cuda, it's not worth having the `try` and `except ImportError` block.

numba_cuda/numba/cuda/typeconv/typeconv.py
```diff
@@ -0,0 +1,139 @@
...
+    dashes = "-" * 80
+    msg = (
+        f"Numba could not be imported.\n{dashes}\nIf you are seeing this "
+        "message and are undertaking Numba development work, you may need "
+        "to rebuild Numba.\nPlease see the development set up guide:\n\n"
+        f"{dev_url}.\n\n{dashes}\nIf you are not working on Numba "
+        f"development, the original error was: '{str(e)}'.\nFor help, "
+        f"please visit:\n\n{user_url}\n"
+    )
+    raise ImportError(msg)
```
gmarkall: Let's just have:

```suggestion
from numba.cuda.cext import _typeconv
```

gmarkall:
A quick note - I pushed the change I suggested so I could run it through CI an set it to auto-merge, rather than us messing about with a cycle of feedback to iterate.

PR #501, opened 2025-10-03
Persist Compile Options via LinkableCode, augment declare_device API

isVoid:


gmarkall:
It's not clear to me that mixing debug and lineinfo options across different components of a single kernel is a good idea - it feels more like debug and lineinfo should be enabled for all "translation units" or none of them. Was there some requirement to mix them for a specific purpose?

mmason-nvidia:
> It's not clear to me that mixing debug and lineinfo options across different components of a single kernel is a good idea - it feels more like debug and lineinfo should be enabled for all "translation units" or none of them. Was there some requirement to mix them for a specific purpose?

The general thought I had was:
- If debug/lineinfo/opt/lto/etc is specified in device_declare(), then those are the options to use.
- If not specified in device_declare(), then "inherit" the settings from the calling kernel.

The idea being that if you're debugging kernel "my_kernel", you want to enable debug generation for that kernel and for the code that it calls into. I was concerned that having to update both locations to get debugging to work might have a couple of user traps/issues. For example, having to change both the @cuda.jit decorator and the device_declare() call(s) and then forgetting to switch both of them back when done debugging (esp if the device_declare() call is "somewhere else". I thought it preferable to avoid that sort of user error. Don't want to ship code that had a debug=True or other such attribute left around.

isVoid:
> The idea being that if you're debugging kernel "my_kernel", you want to enable debug generation for that kernel and for the code that it calls into. I was concerned that having to update both locations to get debugging to work might have a couple of user traps/issues.

So, it sounds like you are alluding to a single source of truth for `debug` argument? Wouldn't that contradict with setting it at both the `declare_device` and `cuda.jit` argument list? From a debugging perspective, IMO it makes sense to have a single `debug` argument controlling every thing in the TU, user can control stepping into / over details when needed.

mmason-nvidia:
> > The idea being that if you're debugging kernel "my_kernel", you want to enable debug generation for that kernel and for the code that it calls into. I was concerned that having to update both locations to get debugging to work might have a couple of user traps/issues.
> 
> So, it sounds like you are alluding to a single source of truth for `debug` argument? Wouldn't that contradict with setting it at both the `declare_device` and `cuda.jit` argument list? From a debugging perspective, IMO it makes sense to have a single `debug` argument controlling every thing in the TU, user can control stepping into / over details when needed.

Are you proposing that the TU compile CUDA C/C++ code based based on the `cuda.jit()` arguments only and not have debug/lineinfo/opt specifiable in `declare_device()`? (ie. we use `cuda.jit() as the single source of truth). I was thinking that finer-grained control at the `device_declare()` level would also be useful, but I realize that might be introducing too much complication.

As an aside: would it be useful/desirable to add an environment variable which globally sets debug/lineinfo/etc.? I could see that complimenting the break-on-launch capability of the debug tools.

Any approach I think will have implications for the caching code as the current implementation uses the compute capability as the key without including the compilation flags (assuming the scope of the cache spans multiple TUs - which I'm not certain of).

isVoid:
>  based on the cuda.jit() arguments only and not have debug/lineinfo/opt specifiable in declare_device()

Yes. I now tend to think that `cuda.jit(debug=True)` currently not enabled for external function as a bug.

> would it be useful/desirable to add an environment variable which globally sets debug/lineinfo/etc.?

If we agree on the above, this implies `jit(debug=True)` is propagated globally. So the environment variable is changing the default on this kwarg. I think it can be useful when developer don't want to modify their kernel in place for debugging purposes.

> implications for the caching code as the current implementation uses the compute capability as the key

Right, specifically `CUDACodeLibrary` cache keys needs to go from `cc` to a 4-tuple `(cc, debug, lineinfo, opt)`.

isVoid:
Based on the above - we need to close this PR and rethink how to propagate `jit(debug)` argument universally.

PR #502, opened 2025-10-03
Vendor in dispatcher, entrypoints, pretty_annotate for CUDA-specific changes

VijayKandiah:
This PR vendors in the rest of `dispatcher` along with a couple other modules: `core.entrypoints` and `core.annotations.pretty_annotate` from numba for future CUDA-specific customizations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,10 +37,17 @@ they are compiled for the appropriate compute capability.
 ```
 cd testing
 # Optionally, build test binaries and point to their location for the test suite
-make
+make -j $(nproc)
 export NUMBA_CUDA_TEST_BIN_DIR=`pwd`
 # Execute tests
 pytest -n auto -v
+```
+
+Alternatively, you can use [pixi](https://pixi.sh/latest/installation/) to wrap all of that up for you:
+
+```
+# run tests against CUDA 13
+pixi run -e cu13 test -n auto -v
 ```
 
 
--- a/ci/coverage_report.sh
+++ b/ci/coverage_report.sh
@@ -20,7 +20,7 @@ python -m pip install \
 rapids-logger "Build test binaries"
 export NUMBA_CUDA_TEST_BIN_DIR=`pwd`/testing
 pushd $NUMBA_CUDA_TEST_BIN_DIR
-make
+make -j $(nproc)
 
 rapids-logger "Check GPU usage"
 nvidia-smi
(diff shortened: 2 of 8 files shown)


numba_cuda/numba/cuda/core/entrypoints.py
gmarkall, line 1: I think we should not vendor this, because it could result in initializing Numba extensions twice.

Instead, we can just remove the use of `entrypoints` from `numba.cuda.dispatcher`.

gmarkall: Alternatively, and perhaps better - we should try to import Numba and run its `init_all()` function instead.

numba_cuda/numba/cuda/core/entrypoints.py
```diff
@@ -0,0 +1,50 @@
...
+import warnings
+
+from importlib import metadata as importlib_metadata
+
+
+_already_initialized = False
+logger = logging.getLogger(__name__)
+
+
+def init_all():
```
gmarkall, line 14: For a safe alternative that allows us to keep entry point support and have support for Numba-CUDA-specific entry points, I'm thinking something like

```suggestion
def init_all():
    try:
        from numba.core import entrypoints
        entrypoints.init_all()
    except ImportError:
        pass
```

followed by a change in the name of the entry points later (see comment below)

numba_cuda/numba/cuda/core/entrypoints.py
```diff
@@ -0,0 +1,50 @@
...
+            )
+            warnings.warn(msg, stacklevel=3)
+            logger.debug("Extension loading failed for: %s", entry_point)
+
+    eps = importlib_metadata.entry_points()
+    # Split, Python 3.10+ and importlib_metadata 3.6+ have the "selectable"
+    # interface, versions prior to that do not. See "compatibility note" in:
+    # https://docs.python.org/3.10/library/importlib.metadata.html#entry-points
+    if hasattr(eps, "select"):
+        for entry_point in eps.select(group="numba_extensions", name="init"):
```
gmarkall:
```suggestion
        for entry_point in eps.select(group="numba_cuda_extensions", name="init"):
```

numba_cuda/numba/cuda/core/entrypoints.py
```diff
@@ -0,0 +1,50 @@
...
+            load_ep(entry_point)
+    else:
+        for entry_point in eps.get("numba_extensions", ()):
```
gmarkall:
```suggestion
        for entry_point in eps.get("numba_cuda_extensions", ()):
```

VijayKandiah:
Thanks! I addressed the feedback in my latest commit.

PR #503, opened 2025-10-03
Correct `NUMBA_CUDA_DEVICEARRAY_IMPORT_NAME`

gmarkall:
This addresses one of the fixups required following the merge of #373.

The package name is `numba.cuda.cext`, not `numba_cuda`. However, fixing this results in a circular import during `PyCapsule_Import` when running something as simple as:

```python
from numba import cuda
```

which gives:

```
AttributeError: cannot access submodule 'cuda' of module 'numba' (most likely due to a circular import)
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/__init__.py", line 73, in <module>
    from .device_init import *
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/device_init.py", line 66, in <module>
    from .decorators import jit, declare_device
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/decorators.py", line 9, in <module>
    from numba.cuda.dispatcher import CUDADispatcher
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 50, in <module>
    from numba.cuda.cext import _dispatcher
ImportError: numba.cuda.cext._devicearray failed to import
```

This is because when `import_devicearray()` is called, we're partway through importing `numba.cuda`. Therefore, the `PyCapsule_Import()` fails because it tries to access packages under `numba.cuda` during its initialization, which then fails due to this circularity. This was not a problem in upstream Numba because `_devicearray` was not in the `numba.cuda` package.

In order to work around this, we can get the `_DEVICEARRAY_API` attribute of the `_devicearray` module directly from its module dict, and then use `PyCapsule_GetPointer()` to set the `DeviceArray_API` global.

### DIFF
--- a/numba_cuda/numba/cuda/cext/_devicearray.h
+++ b/numba_cuda/numba/cuda/cext/_devicearray.h
@@ -8,7 +8,7 @@
     extern "C" {
 #endif
 
-#define NUMBA_DEVICEARRAY_IMPORT_NAME "numba_cuda._devicearray"
+#define NUMBA_DEVICEARRAY_IMPORT_NAME "numba.cuda.cext._devicearray"
 /* These definitions should only be used by consumers of the Device Array API.
  * Consumers access the API through the opaque pointer stored in
  * _devicearray._DEVICEARRAY_API.  We don't want these definitions in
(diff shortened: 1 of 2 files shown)


kkraus14:
@gmarkall fix LGTM, is there a test gap here that didn't reveal this issue?

gmarkall:
Closing as subsumed by #506.

PR #504, opened 2025-10-03
Allow registering an ArgHandler

brandon-b-miller:
This PR allows a user to register extra arg handling for any type they wish to pass to a kernel. Arg handlers must inherit from `ArgHandlerBase` and implement the required handling. Then, a user may pass it to `register_arg_handler`, potentially upon the import of a library that expects to pass a custom object to a numba kernel.

### DIFF
--- a/numba_cuda/numba/cuda/dispatcher.py
+++ b/numba_cuda/numba/cuda/dispatcher.py
@@ -536,12 +536,29 @@ class _Kernel(serialize.ReduceMixin):
         """
 
         # map the arguments using any extension you've registered
+        handler = _arg_handlers.get(type(val))
+        mutated = False
+
         if self.extensions:
             for extension in reversed(self.extensions):
+                if not mutated:
+                    _ty, _val = ty, val
+
                 ty, val = extension.prepare_args(
                     ty, val, stream=stream, retr=retr
                 )
-        elif handler := _arg_handlers.get(type(val)):
+
+                if not mutated and not (_ty is ty and _val is val):
+                    mutated = True
+                    if handler is not None:
+                        raise RuntimeError(
+                            "Argument type %s and value %s were mutated by "
+                            "extension; cannot also be handled by registered "
+                            "argument handler" % (ty, val)
+                        )
+
+        # only run global handler if untouched by extensions
+        if not mutated and handler is not None:
             ty, val = handler(ty, val, stream=stream, retr=retr)
 
         if isinstance(ty, types.Array):
@@ -2122,6 +2139,8 @@ def register_arg_handler(handler, handled_types):
             raise ValueError(
                 f"A handler for args of type {ty} is already registered."
             )
+
+    for ty in handled_types:
         _arg_handlers[ty] = handler
 
 
--- a/numba_cuda/numba/cuda/tests/cudapy/test_extending.py
+++ b/numba_cuda/numba/cuda/tests/cudapy/test_extending.py
@@ -496,6 +496,8 @@ class TestArgHandlerRegistration(CUDATestCase):
 
         class PassedArgHandler_float32:
             def prepare_args(self, ty, val, **kwargs):
+                if val.arr.dtype != np.float32:
+                    return ty, val
                 return types.float32[::1], val.arr
 
         @cuda.jit(


gmarkall:
Adding to the pre-populated list of extensions will increase the launch time for all kernels whether they use the extension or not - I'm not sure how great the impact will be, but I am concerned it could increase kernel launch time by a noticeable percentage. If we go with this route (as opposed to globally registered extensions being tried right before the `NotImplementedError` in `_prepare_args`) can we make some measurements to determine the impact on launch time, so we can understand the tradeoff we're making in ease of use vs. speed?

ZzEeKkAa:
Regarding performance - could we add support through the type's method instead of global registry? It should not impact existing kernel launches and will have O(1) overhead for those types that actually need custom support.

UPD: I looked at the implementation - it is map lookup, so should be also O(1) overhead per argument. We should greatly *improve* performance by switching existing logic to `register_arg_handler`

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -533,8 +537,13 @@ def _prepare_args(self, ty, val, stream, retr, kernelargs):
...
 
         # map the arguments using any extension you've registered
-        for extension in reversed(self.extensions):
-            ty, val = extension.prepare_args(ty, val, stream=stream, retr=retr)
+        if self.extensions:
+            for extension in reversed(self.extensions):
+                ty, val = extension.prepare_args(
+                    ty, val, stream=stream, retr=retr
+                )
+        elif handler := _arg_handlers.get(type(val)):
```
ZzEeKkAa: We can put this check at very bottom of `if-elif-..-else` logic to avoid any performance downgrade for now

brandon-b-miller:
> Regarding performance - could we add support through the type's method instead of global registry? It should not impact existing kernel launches and will have O(1) overhead for those types that actually need custom support.
> 
> UPD: I looked at the implementation - it is map lookup, so should be also O(1) overhead per argument. We should greatly _improve_ performance by switching existing logic to `register_arg_handler`

Agree. @gmarkall does the lookup approach alleviate your concern about perf?

numba_cuda/numba/cuda/dispatcher.py
brandon-b-miller: ah looks like we need to keep it at the top to avoid the provided signature causing other branches of this if-else structure to handle the processing instead of the handler. We still need the lookup first, but, it enforces that the handler definitely does the handling instead of something else accidentally.

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -2122,6 +2129,20 @@ def _reduce_states(self):
         return dict(py_func=self.py_func, targetoptions=self.targetoptions)
 
 
+class ArgHandlerBase(ABC):
+    @abstractmethod
+    def prepare_args(self, ty: Any, val: Any, **kwargs) -> Any:
+        raise NotImplementedError
```
gmarkall: Could the handler just be a function:

```python
def handler(ty, val, strem=None, retr=None):
    ...
```

rather than using a class + ABC?

Although arg handlers have been classes before, I don't think this pattern actually makes any sense - no state has ever been used by one, nor any inheritance.

numba_cuda/numba/cuda/tests/cudapy/test_extending.py
```diff
@@ -303,5 +303,35 @@ def foo(r, x):
...
+
+        class NumpyArrayWrapperArgHandler(ArgHandlerBase):
+            def prepare_args(self, ty, val, **kwargs):
+                return types.int32[::1], val.arr
+
+        register_arg_handler(
+            NumpyArrayWrapperArgHandler(), (NumpyArrayWrapper,)
+        )
+
+        @cuda.jit("void(int32[::1])")
```
gmarkall, line 431: The test only works with the signature. As the common case is not to provide signatures, it would be good to test without one as well.

To make this work, Numba will need to be able to determine the type of the `NumpyArrayWrapper` from the perspective of the kernel. This will need a typeof implementation:

```python
from numba.cuda.typing.typeof import typeof

@typeof_impl.register(NumpyArrayWrapper)
def typeof_numpy_array_wrapper(val, c):
    return typeof(val.arr)
```

numba_cuda/numba/cuda/tests/cudapy/test_extending.py
```diff
@@ -303,5 +303,35 @@ def foo(r, x):
         np.testing.assert_equal(r, x * 2)
 
 
+@skip_on_cudasim("Extensions not supported in the simulator")
+class TestArgHandlerRegistration(CUDATestCase):
+    def test_register_arg_handler(self):
+        from numba.cuda.dispatcher import register_arg_handler, ArgHandlerBase
```
gmarkall: Any reason for these imports not to be with the others in this file?

numba_cuda/numba/cuda/tests/cudapy/test_extending.py
```diff
@@ -303,5 +303,35 @@ def foo(r, x):
         np.testing.assert_equal(r, x * 2)
 
 
+@skip_on_cudasim("Extensions not supported in the simulator")
+class TestArgHandlerRegistration(CUDATestCase):
```
gmarkall, line 391: I think it would be wise to test with some other cases:

* Two arg handlers registered for different types - the test uses at least one argument of each type.
* One arg handler registered, one passed in `extensions=[...]` - does it behave as expected?
* Two arg handlers registered for the same type. What happens? An error? One displaces the other?
* Other things that could be thought of.

numba_cuda/numba/cuda/tests/cudapy/test_extending.py
```diff
@@ -385,6 +388,137 @@ def foo(r, x):
...
+    def test_register_arg_handler(self):
+        class NumpyArrayWrapper_int32:
+            def __init__(self, arr):
+                self.arr = arr
+
+        def numpy_array_wrapper_int32_arg_handler(ty, val, **kwargs):
```
gmarkall: I think here `ty` will be the Numba type from `typeof`, which will be `types.int32[::1]` in this case, but would also match any other type that comes out of `numpy_array_wrapper_int32_typeof_impl`. Would it make a more exemplar test to just return `ty, val.arr`, here?

(If there's a flaw in my logic / understanding, please do let me know)

gmarkall: In fact, why make these specific to `int32` types at all? Apart from the name, and the hardcoded `types.int32[::1]`, it looks like these could be fully generic for a wrapper holding any `dtype` / shape NumPy array?

numba_cuda/numba/cuda/tests/cudapy/test_extending.py
```diff
@@ -385,6 +388,137 @@ def foo(r, x):
...
+
+        arr_f = np.zeros(10, dtype=np.float32)
+        arr_i = np.zeros(10, dtype=np.int32)
+        wrapped_arr_f = NumpyArrayWrapper_float32(arr_f)
+        wrapped_arr_i = NumpyArrayWrapper_int32(arr_i)
+        kernel.forall(len(arr_f))(wrapped_arr_f, wrapped_arr_i)
+        np.testing.assert_equal(arr_f, np.ones(10, dtype=np.float32))
+        np.testing.assert_equal(arr_i, np.ones(10, dtype=np.int32) * 2)
+
+        # multiple handlers for the same type - last one wins
```
gmarkall: I think raising an error for multiple handlers would be a less surprising policy. Are there reasons to prefer "last one wins" instead?

numba_cuda/numba/cuda/tests/cudapy/test_extending.py
```diff
@@ -385,6 +388,137 @@ def foo(r, x):
         np.testing.assert_equal(r, x * 2)
 
 
+@skip_on_cudasim("Extensions not supported in the simulator")
+class TestArgHandlerRegistration(CUDATestCase):
+    def test_register_arg_handler(self):
```
gmarkall: This test looks like it's about four tests in one method. It would be nice to break it up into distinct test cases.

numba_cuda/numba/cuda/tests/cudapy/test_extending.py
brandon-b-miller: Since this is test code, I prefer to spell things out in some cases. Here we're testing the effect of one or more handlers on one or more classes, it makes sense to me to write them separately and somewhat verbosely for readability as to what we're testing, even if things are repeated.

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -2119,6 +2125,18 @@ def _reduce_states(self):
         return dict(py_func=self.py_func, targetoptions=self.targetoptions)
 
 
+def register_arg_handler(handler, handled_types, impl):
+    global _arg_handlers
+
+    for ty in handled_types:
+        typeof_impl.register(ty)(impl)
+        if _arg_handlers.get(ty, None):
```
gmarkall, line 2138: This should come before registering the `typeof_impl`, because the registration can't be backed out afterwards if something goes wrong here.

brandon-b-miller: thanks, nice catch.

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -535,8 +537,13 @@ def _prepare_args(self, ty, val, stream, retr, kernelargs):
...
 
         # map the arguments using any extension you've registered
-        for extension in reversed(self.extensions):
-            ty, val = extension.prepare_args(ty, val, stream=stream, retr=retr)
+        if self.extensions:
+            for extension in reversed(self.extensions):
+                ty, val = extension.prepare_args(
+                    ty, val, stream=stream, retr=retr
+                )
+        elif handler := _arg_handlers.get(type(val)):
```
gmarkall: I'm not sure the if / elif logic here is correct. It seems to implement "If there are any extensions registered for this kernel, ignore all globally-registered extensions". But I could imagine a scenario where there is a globally-registered extension for one argument, and a "locally"-registered one for another argument - I think this situation is precluded from working in the current implementation.

brandon-b-miller: I think this section of the code might have been what pushed me towards having a class that the user supplies with a particular interface in the original implementation. The existing API requires we pass a class with a `_prepare_args` method. If we expect the same thing from the new API, it lets us assemble a single container of handlers we end up passing along the existing codepath, and we don't have to have deal with two kinds / can error if we find a duplicate along the way.

brandon-b-miller:
@gmarkall should we instead be changing the existing `extensions` interface to be more like the one we're moving towards in this PR now? I note that the existing `extensions` API does not require a `typeof_impl`, meaning it only works for the case where a signature is provided as well. 

Either way we go, I think these two APIs should expect to be passed the same things, and there may be some flexibility due to the lack of existing docs. We can go with the function vs the class if we desire, but the more I think about it the more I think the `typeof_impl` requirement should be optional. It doesn't buy the person registering the handler much over just calling `typeof_impl.register` on import in their own library, but it does make it hard to normalize the `extensions` api with the `register_arg_handler` API. What do you think?

brandon-b-miller:
@gmarkall I've made some simplifications and would like to circle back to this if possible. I have:
- For now removed the no-signature cases. The current registration API doesn't support this either, since it adds a significant number of test cases and complexity I would like to peel this off into a follow up.
- I'd like to make a decision about if we want to register using a function or a class with `_prepare_args`. I agree with your previous comment that there's not been any state or inheritance used historically with these, but I do think it's important to either always pass classes or always pass functions. In the API we're adding, we now pass functions. Should the old API change? Should it be deprecated? Should we just be consistent and go back to passing classes in this PR?

PR #505, opened 2025-10-03
chore(dev-deps): add pixi

cpcloud:
This PR adds a pixi environment setup to pyproject.toml so that developers have a working environment out the box with which to develop `numba-cuda`. This is far from complete but it is enough to get started with `pixi run -e cu13 pytest numba_cuda/numba/cuda/tests`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,6 +46,7 @@ cu13 = [
     "nvidia-cuda-cccl==13.*",
 ]
 
+[dependency-groups]
 test = [
     "pre-commit",
     "psutil",
@@ -56,14 +57,12 @@ test = [
     "ml_dtypes",
 ]
 test-cu12 = [
-    "numba-cuda[cu12]",
-    "numba-cuda[test]",
     "nvidia-curand-cu12",
+    { include-group = "test" }
 ]
 test-cu13 = [
-    "numba-cuda[cu13]",
-    "numba-cuda[test]",
     "nvidia-curand==10.4.*",
+    { include-group = "test" }
 ]
 
 [project.urls]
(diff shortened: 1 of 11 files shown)


pyproject.toml
```diff
@@ -54,16 +55,15 @@ test = [
     "pytest-xdist",
     "filecheck",
     "ml_dtypes",
+    "pyinstrument",
```
cpcloud: I added this to make it easier to profile things.

pyproject.toml
```diff
@@ -153,3 +153,18 @@ exclude = [
 "numba_cuda/numba/cuda/tests/doc_examples/test_cg.py" = ["E501"]
 "numba_cuda/numba/cuda/tests/doc_examples/test_matmul.py" = ["E501"]
 "numba_cuda/numba/tests/doc_examples/test_interval_example.py" = ["E501"]
+
+[tool.pixi.workspace]
+channels = ["conda-forge"]
+platforms = ["linux-64"]
```
kkraus14: We need to support windows as well

pyproject.toml
kkraus14: my 2c: we should keep this minimal to what is needed to run tests

In general for optional dependency groups I think we should only include the reasonably minimal set of what is needed to run all of the optional tests / benchmarks / etc.

Is there a standard practice for bringing in user specific tooling in pixi while working with an existing `pyproject.toml` / `pixi.toml`?

pyproject.toml
```diff
@@ -55,15 +56,14 @@ test = [
     "filecheck",
     "ml_dtypes",
 ]
+prof = ["pyinstrument"]
```
kkraus14: Why do we need to add this? Someone could use `py-spy` or `memray` or any number of other python profiling / introspection tools. If someone wants to use `pyinstrument` for their own usage that's perfectly reasonable, but do we need it in the project level `pyproject.toml` / `pixi.toml`?

cpcloud: I guess we don't, but it's also easier to collaborate and share profiling output if I can simply say `pixi run -e prof pyinstrument script.py` than to figure out what other weird setup someone has.

That said, it's not necessary for this PR so I'll remove it.

PR #506, opened 2025-10-06
Remove C extension loading hacks

gmarkall:
## Summary

The goal is to remove the C extension loading hacks and allow them to be discovered and loaded by the Python interpreter in the normal way. This should resolve issues where the extensions are not found in certain users' setups.

This is also a step in the right direction for testing, because it ensures that we're actually testing code loaded from the built package, and not a hybrid of some of the code from the package and some of the code accidentally discovered in the source repo. The overall change to run tests from a subfolder of the repo, `testing`, avoids the `numba_cuda` folder in the root of the repo being found and used at test time in CI.

## Changes

A number of distinct changes accomplish this:

* Correct the name used in `NUMBA_DEVICEARRAY_IMPORT_NAME` (subsumes #503).
* Delete the import hacks from `numba_cuda/numba/cuda/cext/__init__.py`.
* Move the test binary generation files out of the package and into a subfolder of the repo called `testing`.
* Move the pytest configuration (pytest.ini, conftest.py) to testing folder. `conftest.py` is moved verbatim; `pytest.ini` contains the options previously in `pyproject.toml`. `pytest.ini` also adds `--pyargs numba.cuda.tests` so that running `pytest` with no further options does the right thing.

* The creation of `$RAPIDS_TEST_DIR` and `cd`ing into it has been removed from the CI scripts. Whilst not directly related to this PR, it served no purpose and was confounding when the new correct location to run tests from is the `testing` folder.
* README updates to reflect how to set up and run tests following these changes.

## Additional info on the change of `NUMBA_DEVICEARRAY_IMPORT_NAME` (repeated from #503):

The package name is `numba.cuda.cext`, not `numba_cuda`. However, fixing this results in a circular import during `PyCapsule_Import` when running something as simple as:

```python
from numba import cuda
```

which gives:

```
AttributeError: cannot access submodule 'cuda' of module 'numba' (most likely due to a circular import)
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/__init__.py", line 73, in <module>
    from .device_init import *
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/device_init.py", line 66, in <module>
    from .decorators import jit, declare_device
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/decorators.py", line 9, in <module>
    from numba.cuda.dispatcher import CUDADispatcher
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/dispatcher.py", line 50, in <module>
    from numba.cuda.cext import _dispatcher
ImportError: numba.cuda.cext._devicearray failed to import
```

This is because when `import_devicearray()` is called, we're partway through importing `numba.cuda`. Therefore, the `PyCapsule_Import()` fails because it tries to access packages under `numba.cuda` during its initialization, which then fails due to this circularity. This was not a problem in upstream Numba because `_devicearray` was not in the `numba.cuda` package.

In order to work around this, we can get the `_DEVICEARRAY_API` attribute of the `_devicearray` module directly from its module dict, and then use `PyCapsule_GetPointer()` to set the `DeviceArray_API` global.


pyproject.toml
```diff
@@ -82,13 +82,6 @@ include = ["numba_cuda*"]
 [tool.setuptools.package-data]
 "*" = ["*.cu", "*.h", "*.hpp", "*.ptx", "*.cuh", "VERSION", "Makefile"]
 
-[tool.pytest.ini_options]
-minversion = "8.0"
-testpaths = ["numba_cuda/numba/cuda/tests"]
-consider_namespace_packages = true
-# loadscope ensures the grouping required by CUDATestCase
-addopts = "--dist loadscope"
```
kkraus14, line 90: Why did we pull these out into a separate `ini` file?

gmarkall: I was trying to get everything related to testing in the `testing` subfolder, because running from the root of the repository causes the `numba_cuda` package in the source repo to be discovered when we're trying to test the installed wheel / conda package (which was the root of the problems requiring the C extension loading hacks). I was concerned also that relying on `pyproject.toml` in the root might lead to the accidental discovery of `numba_cuda` in the root of the repo too.

rparolin (approved):
lgtm! Happy to see unit tests getting simpler to run via the terminal.

PR #507, opened 2025-10-07
Remove some unnecessary uses of ContextResettingTestCase

gmarkall:
Some cases that I think look safe to remove; a few more complicated ones still remain.

Testing on CI for now - seems to pass locally for me.

### DIFF
--- a/numba_cuda/numba/cuda/tests/cudadrv/test_reset_device.py
+++ b/numba_cuda/numba/cuda/tests/cudadrv/test_reset_device.py
@@ -4,11 +4,11 @@
 import threading
 from numba import cuda
 from numba.cuda.cudadrv.driver import driver
-from numba.cuda.testing import unittest, ContextResettingTestCase
+from numba.cuda.testing import unittest, CUDATestCase
 from queue import Queue
 
 
-class TestResetDevice(ContextResettingTestCase):
+class TestResetDevice(CUDATestCase):
     def test_reset_device(self):
         def newthread(exception_queue):
             try:
--- a/numba_cuda/numba/cuda/tests/cudadrv/test_select_device.py
+++ b/numba_cuda/numba/cuda/tests/cudadrv/test_select_device.py
@@ -9,7 +9,7 @@ from queue import Queue
 
 import numpy as np
 from numba import cuda
-from numba.cuda.testing import unittest, ContextResettingTestCase
+from numba.cuda.testing import unittest, CUDATestCase
 
 
 def newthread(exception_queue):
@@ -21,12 +21,12 @@ def newthread(exception_queue):
         stream.synchronize()
         del dA
         del stream
-        cuda.close()
+        cuda.synchronize()
     except Exception as e:
         exception_queue.put(e)
 
 
-class TestSelectDevice(ContextResettingTestCase):
+class TestSelectDevice(CUDATestCase):
     def test_select_device(self):
         exception_queue = Queue()
         for i in range(10):
--- a/numba_cuda/numba/cuda/tests/cudapy/test_cuda_array_interface.py
+++ b/numba_cuda/numba/cuda/tests/cudapy/test_cuda_array_interface.py
@@ -5,14 +5,14 @@ import numpy as np
 
 from numba import vectorize, guvectorize
 from numba import cuda
-from numba.cuda.testing import unittest, ContextResettingTestCase, ForeignArray
+from numba.cuda.testing import unittest, CUDATestCase, ForeignArray
 from numba.cuda.testing import skip_on_cudasim, skip_if_external_memmgr
 from numba.cuda.tests.support import linux_only, override_config
 from unittest.mock import call, patch
 
 
 @skip_on_cudasim("CUDA Array Interface is not supported in the simulator")
-class TestCudaArrayInterface(ContextResettingTestCase):
+class TestCudaArrayInterface(CUDATestCase):
     def assertPointersEqual(self, a, b):
         self.assertEqual(
             a.device_ctypes_pointer.value, b.device_ctypes_pointer.value
--- a/numba_cuda/numba/cuda/tests/cudadrv/test_inline_ptx.py
+++ b/numba_cuda/numba/cuda/tests/cudadrv/test_inline_ptx.py
@@ -4,12 +4,12 @@
 from llvmlite import ir
 
 from numba.cuda.cudadrv import nvvm
-from numba.cuda.testing import unittest, ContextResettingTestCase
+from numba.cuda.testing import unittest, CUDATestCase
 from numba.cuda.testing import skip_on_cudasim
 
 
 @skip_on_cudasim("Inline PTX cannot be used in the simulator")
-class TestCudaInlineAsm(ContextResettingTestCase):
+class TestCudaInlineAsm(CUDATestCase):
     def test_inline_rsqrt(self):
         mod = ir.Module(__name__)
         mod.triple = "nvptx64-nvidia-cuda"
--- a/numba_cuda/numba/cuda/tests/cudadrv/test_pinned.py
+++ b/numba_cuda/numba/cuda/tests/cudadrv/test_pinned.py
@@ -5,10 +5,10 @@ import numpy as np
 import platform
 
 from numba import cuda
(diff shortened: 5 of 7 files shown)


numba_cuda/numba/cuda/tests/cudadrv/test_select_device.py
```diff
@@ -21,12 +21,12 @@ def newthread(exception_queue):
         stream.synchronize()
         del dA
         del stream
-        cuda.close()
+        cuda.synchronize()
```
gmarkall, line 24: I think the intention of `cuda.close()` was to make sure any errors from asynchronous operations are detected; I think `cuda.synchronize()` should be sufficient for this purpose.

numba_cuda/numba/cuda/tests/cudadrv/test_reset_device.py
```diff
@@ -4,11 +4,11 @@
 import threading
 from numba import cuda
 from numba.cuda.cudadrv.driver import driver
-from numba.cuda.testing import unittest, ContextResettingTestCase
+from numba.cuda.testing import unittest, CUDATestCase
 from queue import Queue
 
 
-class TestResetDevice(ContextResettingTestCase):
+class TestResetDevice(CUDATestCase):
```
gmarkall, line 11: The comment (and code) below suggests that the context on the main thread is unaffected (it creates new threads for the tests) so I think there should be no need to reset the context again after the test has run.

numba_cuda/numba/cuda/tests/cudapy/test_cuda_array_interface.py
```diff
@@ -5,14 +5,14 @@
...
-from numba.cuda.testing import unittest, ContextResettingTestCase, ForeignArray
+from numba.cuda.testing import unittest, CUDATestCase, ForeignArray
 from numba.cuda.testing import skip_on_cudasim, skip_if_external_memmgr
 from numba.cuda.tests.support import linux_only, override_config
 from unittest.mock import call, patch
 
 
 @skip_on_cudasim("CUDA Array Interface is not supported in the simulator")
-class TestCudaArrayInterface(ContextResettingTestCase):
+class TestCudaArrayInterface(CUDATestCase):
```
gmarkall, line 15: I don't think context resets are needed here - some tests check the list of pending deallocations and flush it, but that should not require a context reset.

kkraus14 (dismissed):

numba_cuda/numba/cuda/tests/cudadrv/test_module_callbacks.py
```diff
@@ -42,7 +41,7 @@ def get_hashable_handle_value(handle):
 
 
 @skip_on_cudasim("Module loading not implemented in the simulator")
-class TestModuleCallbacksBasic(ContextResettingTestCase):
+class TestModuleCallbacksBasic(CUDATestCase):
```
brandon-b-miller, line 44: Connected with @isVoid about these and they should be safe to remove.

numba_cuda/numba/cuda/tests/cudadrv/test_host_alloc.py
```diff
@@ -4,10 +4,10 @@
 import numpy as np
 from numba.cuda.cudadrv import driver
 from numba import cuda
-from numba.cuda.testing import unittest, ContextResettingTestCase
+from numba.cuda.testing import unittest, CUDATestCase
 
 
-class TestHostAlloc(ContextResettingTestCase):
+class TestHostAlloc(CUDATestCase):
```
brandon-b-miller, line 10: A lot of these are fairly old tests from the upstream numba code base where it looks like a lot of tests may have simply reset the device every time in an effort to be extra safe and start with a clean slate every time. This may have made sense at the time but might not anymore.

kkraus14 (approved):

numba_cuda/numba/cuda/tests/cudadrv/test_cuda_memory.py
```diff
@@ -6,17 +6,18 @@
...
 
 @skip_on_cudasim("CUDA Memory API unsupported in the simulator")
-class TestCudaMemory(ContextResettingTestCase):
+class TestCudaMemory(CUDATestCase):
     def setUp(self):
         super().setUp()
         self.context = devices.get_context()
 
     def tearDown(self):
+        self.context.reset()
```
brandon-b-miller, line 20: cc @gmarkall I think this should clear things safely, this appears to free the resources within the context but is less scorched earth than a full `reset`

gmarkall: Does this mean it will no longer invalidate assumptions made by cuda-core?

numba_cuda/numba/cuda/tests/cudadrv/test_module_callbacks.py
gmarkall: One of the reasons I left this was because I hadn't dug into why it was used, and it wasn't obvious to me - can you share the reasoning please?

numba_cuda/numba/cuda/tests/cudadrv/test_cuda_memory.py
kkraus14: @gmarkall are you asking if `cuda.core` protects against the context being reset in some kind of way, the answer is no. Resetting the context would 100% lead to any resources that existed being dangling and probably result in segfaults.

brandon-b-miller: The `reset` name of this [context method ](https://github.com/NVIDIA/numba-cuda/blob/main/numba_cuda/numba/cuda/cudadrv/driver.py#L1288-L1297)is perhaps a little misleading. 

Step by step this method:
1. Calls `reset` on the context's `MemoryManager`, in this case a `NumbaCUDAMemoryManager` instance, whose implementation is inherited from HostOnlyCUDAMemoryManager [here](https://github.com/NVIDIA/numba-cuda/blob/main/numba_cuda/numba/cuda/cudadrv/driver.py#L1026-L1033). This in turn calls `clear` on two of objects attached to the instance: First the `allocations` and then the `deallocations`. TL;DR, clearing the `allocations` member has the effect of adding everything the context allocated to its pending deallocs, and then `deallocations.clear()` eventually calls the dtor which maps to a true cuda free through the driver. 
2. Does the same for all modules (same thing but terminating in cuModuleUnload)
3. Does the same for anything left in its own `_PendingDeallocs` list

My reading of this is that the net effect is that all resources owned by _this context object_ are released - but it only releases resources _it itself allocated_. 

This is distinct from the [full gpu reset](https://github.com/NVIDIA/numba-cuda/blob/9666a185cedaaeb388d7a1cc7c00940cd99e9c44/numba_cuda/numba/cuda/cudadrv/devices.py#L196-L211) function called by `ContextResettingTestCase` which destroys all the contexts on the device. If `cuda.core` has cached contexts it interacts with internally after this is called, I can imagine things going wrong... but this particular kind of reset should just be cleaning up objects that are being carefully kept track of.

numba_cuda/numba/cuda/tests/cudadrv/test_module_callbacks.py
brandon-b-miller: I believe it is superfluous. `context.reset` should have the effect of calling the finalizer on any modules.

numba_cuda/numba/cuda/tests/cudadrv/test_cuda_memory.py
gmarkall: Thanks for clarifying - this makes sense.

numba_cuda/numba/cuda/tests/cudadrv/test_context_stack.py
```diff
@@ -14,12 +14,7 @@ class TestContextStack(CUDATestCase):
     def setUp(self):
         super().setUp()
         # Reset before testing
-        cuda.close()
-
-    def test_gpus_current(self):
```
brandon-b-miller, line 19: This test seems to rely on there being no active context at all which is hard to guarantee without the hard `cuda.close`.

brandon-b-miller:
@gmarkall do these latest changes seem reasonable?

Issue #477, opened 2025-09-17
Reduce kernel launch overhead


leofang:
One known overhead in numba-cuda is kernel launch. We should profile it carefully and come up with a comprehensive improvement plan, but one obvious source of latency is preparing the kernel args, which is currently done using ctypes. We should switch to use `cuda.core.launch()`.
https://github.com/NVIDIA/numba-cuda/blob/97ce4b1525de474efbc93b45eea0668167fcce3b/numba_cuda/numba/cuda/dispatcher.py#L454-L487

leofang:
xref: #98

leofang:
possible benchmark: https://github.com/NVIDIA/numba-cuda/pull/288#issuecomment-2964416055

PR #510, opened 2025-10-07
perf: speed up kernel launch

cpcloud:
This PR speeds up kernel launching by ~10-25%.

To figure out where some of the hotter paths were, I used the script that's in
the top-level of the repo `prof.py`.

The script makes a call to a cuda-jitted function that takes a single argument
and does nothing in the body of the function.

I don't plan to include this script if this PR is merged, but it's there so
people can pull the PR down and run the code if they'd like.

I also don't plan to include the changes to pixi that add the `prof` group.

Most of the changes here fall under the category of removing dynamism.

1. Replacing `__getattr__` calls/implementations by concretizing frequently accessed attributes.
1. Avoiding `isinstance` checks of `abc.ABCMeta`-based objects.
1. Accessing attributes by direct lookup instead of using `getattr(x, string)`.

### DIFF
--- a/numba_cuda/numba/cuda/cudadrv/devices.py
+++ b/numba_cuda/numba/cuda/cudadrv/devices.py
@@ -70,10 +70,9 @@ class _DeviceContextManager(object):
 
     def __init__(self, device):
         self._device = device
-
-    def get_primary_context(self, *args, **kwargs):
-        """This attribute is forwarded directly, to avoid the performance overhead of `__getattr__`."""
-        return self._device.get_primary_context(*args, **kwargs)
+        # Forwarded directly, to avoid the performance overhead of
+        # `__getattr__` and method lookup for a commonly accessed method
+        self.get_primary_context = self._device.get_primary_context
 
     def __getattr__(self, item):
         return getattr(self._device, item)
@@ -83,7 +82,7 @@ class _DeviceContextManager(object):
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         # this will verify that we are popping the right device context.
-        self._device.get_primary_context().pop()
+        self.get_primary_context().pop()
 
     def __str__(self):
         return f"<Managed Device {self.id}>"


numba_cuda/numba/cuda/cudadrv/devicearray.py
```diff
@@ -204,10 +204,11 @@ def _numba_type_(self):
     @property
     def device_ctypes_pointer(self):
         """Returns the ctypes pointer to the GPU data buffer"""
-        if self.gpu_data is None:
-            return c_void_p(0)
-        else:
+        try:
+            # apparently faster in the non-exceptional case
             return self.gpu_data.device_ctypes_pointer
```
cpcloud, line 204: This is faster because after Python 3.11, there's no cost to the non-exceptional case, and this case appears to be more common in the kernel launching path than the case of `self.gpu_data is None` being `True`.

numba_cuda/numba/cuda/core/config.py
```diff
@@ -709,15 +710,8 @@ def which_gdb(path_or_bin):
 
 def __getattr__(name):
     """Module-level __getattr__ provides dynamic behavior for _EnvVar descriptors."""
-    # Fetch non-descriptor globals directly
-    if name in globals():
```
cpcloud: I removed this because this is the default behavior of module-level attribute access, so it's pointless to do it twice. `__getattr__` is only called if `name` isn't found using the normal attribute lookup.

numba_cuda/numba/cuda/cudadrv/devices.py
```diff
@@ -22,22 +22,13 @@
...
-            ]
-            # Define "lst" to avoid re-initialization
-            self.lst = gpus
-            return gpus
-
-        # Other attributes
-        return super(_DeviceList, self).__getattr__(attr)
+    @property
+    @functools.cache
+    def lst(self):
```
cpcloud, line 27: This is faster for all instances, because there's no longer any dynamic attribute lookup happening in Python, it's all happening in native code with the exception of this attribute.

numba_cuda/numba/cuda/cudadrv/devices.py
```diff
@@ -61,8 +52,9 @@ def current(self):
         """Returns the active device or None if there's no active device"""
         with driver.get_active_context() as ac:
             devnum = ac.devnum
-            if devnum is not None:
-                return self[devnum]
+        if devnum is not None:
+            return self[devnum]
+        return None
```
cpcloud: This change probably doesn't affect performance, so I am happy to remove it.

numba_cuda/numba/cuda/cudadrv/devices.py
```diff
@@ -167,28 +165,29 @@ def _get_or_create_context_uncached(self, devnum):
...
-                            " CUDA context {:x}"
+                            f" CUDA context {ac_ctx_handle:x}"
                         )
-                        raise RuntimeError(msg.format(ac_ctx_handle))
                     # Ensure the context is ready
                     ctx.prepare_for_use()
                 return ctx
 
     def _activate_context_for(self, devnum):
-        with self._lock:
```
cpcloud, line 179: This lock (while reentrant, so not incorrect) was held inside of another section, so I just inlined the lock-holding to the one place this method was being called without a lock, and was able to remove one acquire and release operation.

numba_cuda/numba/cuda/args.py
```diff
@@ -11,11 +11,10 @@
 from numba.core.typing.typeof import typeof, Purpose
 
 
-class ArgHint(metaclass=abc.ABCMeta):
```
cpcloud, line 14: This is probably the biggest bang for the buck here because `isinstance` on `abc.ABCMeta` users is wildly more expensive than without it.

Since this API is only defining 3 classes (and I assume it's not actually public), enforcing the API seems overkill given the performance cost.

prof.py
leofang, line 1: btw another good benchmark is this https://github.com/NVIDIA/numba-cuda/pull/288#issuecomment-2964416055. Varying the number of kernel arguments (or the dimension of a single input array) might help reveal additional hot paths. For every array argument, last time I checked (which was a while ago) numba-cuda would unpack it to (1 + ndim*2) arguments (ptr/shape/strides).

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -3346,7 +3351,11 @@ def is_device_memory(obj):
     "device_pointer" which value is an int object carrying the pointer
     value of the device memory address.  This is not tested in this method.
     """
-    return getattr(obj, "__cuda_memory__", False)
+    try:
+        # This is cheaper than getattr in the non-exceptional case
```
leofang, line 3081: 👍

numba_cuda/numba/cuda/cudadrv/devices.py
```diff
@@ -21,30 +21,19 @@
...
+        self.lst = [
+            _DeviceContextManager(driver.get_device(devid))
+            for devid in range(driver.get_device_count())
+        ]
+
+    def __getitem__(self, devnum: int | slice):
         """
         Returns the context manager for device *devnum*.
         """
-        if not isinstance(devnum, (int, slice)) and USE_NV_BINDING:
```
gmarkall: I have a feeling that this code path is to support someone using a cuda-python device object, that can occur sometimes. I'll see if I can find the history of it.

gmarkall: Ah - this was added in #319 to allow indexing using `CUdevice` objects. I'll check if it is necessary at any point when the CUDA Python bindings are in use (because they are now mandatory).

gmarkall: I think the change here is breaking what was fixed in #319, where a user might have a `CUdevice` obtained from the device list API. I think CI hasn't run in a while so I'm not sure if this would also fail the test suite.

An example of execution before this PR:

```python
In [1]: from numba import cuda

In [2]: dev0 = cuda.devices.gpus[0]

# Passing a CUdevice works (dev0._device.id is a way to get the underlying
# CUdevicefor the sake of this example)
In [3]: cuda.devices.gpus[dev0._device.id]
Out[3]: <numba.cuda.cudadrv.devices._DeviceContextManager at 0x7f09d24fac90>
```


After this PR is applied:

```python
In [1]: from numba import cuda

In [2]: dev0 = cuda.devices.gpus[0]

# Passing a CUdevice no longer works
In [3]: cuda.devices.gpus[dev0._device.id]
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[3], line 1
----> 1 cuda.devices.gpus[dev0._device.id]

File ~/numbadev/numba-cuda/numba_cuda/numba/cuda/cudadrv/devices.py:37, in _DeviceList.__getitem__(self, devnum)
     33 def __getitem__(self, devnum: int | slice):
     34     """
     35     Returns the context manager for device *devnum*.
     36     """
---> 37     return self.lst[devnum]

TypeError: list indices must be integers or slices, not cuda.bindings.driver.CUdevice

# User must cast to integer
In [4]: cuda.devices.gpus[int(dev0._device.id)]
Out[4]: <numba.cuda.cudadrv.devices._DeviceContextManager at 0x71b8650255b0>
```

numba_cuda/numba/cuda/tests/cudadrv/test_context_stack.py
```diff
@@ -30,7 +30,7 @@ def test_gpus_cudevice_indexing(self):
         # usable as an index into cuda.gpus.
         device_ids = [device.id for device in cuda.list_devices()]
         for device_id in device_ids:
-            with cuda.gpus[device_id]:
+            with cuda.gpus[int(device_id)]:
```
gmarkall: I saw this change after writing https://github.com/NVIDIA/numba-cuda/pull/510/files#r2428469695 above. This line is the key part of this test, so changing it is adapting it to the fact that this PR changed the API.

numba_cuda/numba/cuda/cudadrv/devices.py
cpcloud: Reverted this bit. It's not a huge amount of improvement, but it is annoying that it's noticeable.

numba_cuda/numba/cuda/tests/cudadrv/test_context_stack.py
cpcloud: Reverted.

numba_cuda/numba/cuda/cudadrv/devices.py
gmarkall: Can we recover the performance but keep the API the same by holding a dict that maps from integers and `CUdevice` objects to the devices, and look up through that in `__getitem__`?

cpcloud: I think it's probably not worth more time on it for now.

cpcloud:
I'm going to push up a small PR to add a couple benchmarks, so that it's easy to use `pytest-benchmark` to run them and I don't have to keep using `pyinstrument` to verify results.

cpcloud:
Changed the title to reflect the variance in speedup. It's somewhere between 10-25%.

cpcloud:
This is ready for review.

cpcloud:
This PR does *not* close #477.

There's still some significant difference in our code path that automatically converts torch tensors into cuda arrays

cpcloud:
@gmarkall This is ready for review. I am still working on isolating the `as_cuda_array` differences, which I will tackle in a follow-up. I already have a couple wins there, but I'm still trying to track down some more of the performance issues there.

cpcloud:
Ping @gmarkall, would you mind reviewing this? Also happy to give this over to another reviewer if you don't have time!

PR #511, opened 2025-10-08
Make the CUDA target the default for CUDA overload decorators

gmarkall:
This should save manually setting the target to CUDA for vendored overloads.

### DIFF
--- a/numba_cuda/numba/cuda/extending.py
+++ b/numba_cuda/numba/cuda/extending.py
@@ -136,6 +136,7 @@ def overload(
     strict=True,
     inline="never",
     prefer_literal=False,
+    target="cuda",
     **kwargs,
 ):
     """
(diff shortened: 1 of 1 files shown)


numba_cuda/numba/cuda/extending.py
```diff
@@ -352,6 +360,9 @@ def impl(cls, nitems):
         def foo(n):
             return types.Array.make(n)
     """
+
+    kwargs["target"] = "cuda"
```
ashermancinelli: Is this what we intended here?
```suggestion
    kwargs["target"] = target
```

gmarkall: Whoops!

gmarkall:
> @gmarkall is this something we should test in any way?

It's already exercised by many of the `@overload` decorators we already have that are presently registering things with upstream Numba. It's not apparent that we were re-registering things before because it'd only come out if there's a discrepancy between upstream Numba and our vendored implementations, so it's hard to catch the error that this fixes in our CI or test suite.

Additionally it's going to be exercised by the many overloads added in #493 and #494.

PR #513, opened 2025-10-09
[Refactor][NFC] Vendor-in vectorize decorators for future CUDA-specific changes

atmnp:
This PR vendors in the vectorize and guvectorize decorators from Numba, which will allow for future CUDA-specific changes. The CPU-related functionality has been stripped so it no longer relies on a C extension.


brandon-b-miller (approved):
looks good by my eye.

PR #522, opened 2025-10-13
test(pixi): update pixi testing command to work with the new `testing` directory

cpcloud:
Fixes the pixi `test` task to work with the new `testing` directory.

Based off of #521.

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -3917,7 +3917,7 @@ packages:
 - pypi: ./
   name: numba-cuda
   version: 0.20.0
-  sha256: 828152f05744ec6ebb7bb0f92a39380bcb7932e21d05a259e557bf9693f30545
+  sha256: fa8baff5d01a9f506b51e0f41091e6db084e976f7ea9a53ef8cab1df2fe96d41
   requires_dist:
   - numba>=0.60.0
   - cuda-bindings>=12.9.1,<14.0.0


pyproject.toml
```diff
@@ -153,10 +153,25 @@ platforms = ["linux-64", "linux-aarch64", "win-64"]
 [tool.pixi.pypi-dependencies]
 numba-cuda = { path = ".", editable = true }
 
+[tool.pixi.build-dependencies]
+gxx = "*"
```
kkraus14: Is the intention to use this on Windows as well? If not, we should maybe consider switching to using `c-compiler` or otherwise handle making it platform specific with Windows using msvc.

pyproject.toml
```diff
@@ -153,10 +153,25 @@ platforms = ["linux-64", "linux-aarch64", "win-64"]
 [tool.pixi.pypi-dependencies]
 numba-cuda = { path = ".", editable = true }
 
+[tool.pixi.build-dependencies]
```
kkraus14, line 162: I think we're missing `make` from here as well.

kkraus14, line 162: Are we using pixi build to build a conda package? I don't think we are as of yet where `build-dependencies` is probably not the right key?

kkraus14 (dismissed):

pyproject.toml
cpcloud: Hm, this seems to bring in the dependencies at the exact place I need them. Where should I put them instead?

pyproject.toml
cpcloud: Ah, okay. Yep, I'd like this to work on Windows too.

cpcloud: `c-compiler` doesn't give access to `g++`. Is there some other package I need to include?

pyproject.toml
cpcloud: Added.

pyproject.toml
cpcloud: I just moved gxx to a `target.linux` build-dependencies section.

cpcloud: Will figure out the Windows story in a follow-up.

pyproject.toml
kkraus14: Maybe just the `dependencies` section? According to the docs, https://pixi.sh/latest/reference/pixi_manifest/#build-dependencies, it's part of the `pixi-build` backend.

cpcloud: `dependencies` is for runtime dependencies though, so I don't think that would make sense. Since `build-dependencies` is working as I have it, can we just leave it like that?

I can open an issue upstream to ask for clarification on the topic and adjust according to the response in a follow-up PR here.

pyproject.toml
cpcloud: Figured it out in the meantime: it was `cxx-compiler`!

cpcloud:
I think @kkraus14 needs to approve before this can be merge, since changes were requested.

pyproject.toml
cpcloud: https://github.com/prefix-dev/pixi/issues/4762

PR #523, opened 2025-10-14
[WIP] Port numpy reduction tests to CUDA

brandon-b-miller:
This works towards a reimplementation of upstream [test_array_reductions.py](https://github.com/numba/numba/blob/4c37ce138ef7ecbf04474cc706885a6d4312b279/numba/tests/test_array_reductions.py#L4) with the goal being allocating a numpy array on a single thread and performing the CPU check using the result. 

cc @atmnp

### DIFF
--- a/numba_cuda/numba/cuda/device_init.py
+++ b/numba_cuda/numba/cuda/device_init.py
@@ -87,6 +87,9 @@ from numba.cuda.cudadrv.linkable_code import (
     PTXSource,
 )
 
+from numba.cuda.misc.special import literal_unroll
+from numba.cuda.misc import literal
+
 reduce = Reduce = reduction.Reduce
 
 # Expose vector type constructors and aliases as module level attributes.
--- a/numba_cuda/numba/cuda/__init__.py
+++ b/numba_cuda/numba/cuda/__init__.py
@@ -66,33 +66,3 @@ if sys.platform.startswith("linux") and (sys.getdlopenflags() & 0x100) != 0:
     )
 
 from numba.cuda.np.ufunc import vectorize, guvectorize
-
-
-_lazy_exports = {
-    "literal_unroll": ("numba.cuda.misc.special", "literal_unroll"),
-    "literal": ("numba.cuda.misc", "literal"),
-}
-
-__all__ = list(globals().get("__all__", [])) + list(_lazy_exports.keys())
-
-
-def __getattr__(name):
-    """
-    Lazily import a few attrs that might import lowering
-    """
-    if name in _lazy_exports:
-        mod_name, attr_name = _lazy_exports[name]
-        try:
-            module = importlib.import_module(mod_name)
-            value = getattr(module, attr_name)
-        except Exception as exc:
-            raise ImportError(
-                f"could not import name {attr_name!r} from {mod_name!r}"
-            ) from exc
-        globals()[name] = value
-        return value
-    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
-
-
-def __dir__():
-    return sorted(set(list(globals().keys()) + list(_lazy_exports.keys())))
--- a/numba_cuda/numba/cuda/tests/nocuda/test_import.py
+++ b/numba_cuda/numba/cuda/tests/nocuda/test_import.py
@@ -38,7 +38,6 @@ class TestImport(unittest.TestCase):
             "numba.cuda.python.printimpl",
             "numba.cuda.core.optional",
             "numba.cuda.misc.gdb_hook",
-            "numba.cuda.misc.literal",
             "numba.cuda.misc.cffiimpl",
             "numba.np.linalg",
             "numba.np.polynomial",


numba_cuda/numba/cuda/cpython/listobj.py
```diff
@@ -300,7 +300,7 @@ def define_dtor(self):
         # Declare dtor
         fnty = ir.FunctionType(ir.VoidType(), [cgutils.voidptr_t])
         fn = cgutils.get_or_insert_function(
-            mod, fnty, ".dtor.list.{}".format(self.dtype)
+            mod, fnty, "numba_cuda_dtor_list_{}".format(self.dtype)
```
brandon-b-miller: NVVM has special rules about what things can be named, and apparently the name of a variable can't start with a period (or contain one). 
```
numba.cuda.cudadrv.error.NvvmError: Failed to verify                                                                                                                                                                                             
                                                                                                                                                                                                                                                 
error: Error: : Global Value `.dtor.list.float64': Invalid identifier name: .dtor.list.float64  Must match [a-zA-Z$_][a-zA-Z$_0-9]*                                                                                                              
                                                                                                                                         
```

numba_cuda/numba/cuda/target.py
```diff
@@ -225,6 +223,14 @@ def target_data(self):
             self._target_data = ll.create_target_data(nvvm.NVVM().data_layout)
         return self._target_data
 
+    def build_list(self, builder, list_type, items):
```
brandon-b-miller, line 242: this is vendored from the CPUContext. It is needed to construct an literal list object using python bracket notation, as is used in the tests.

numba_cuda/numba/cuda/tests/test_array_reductions.py
```diff
@@ -0,0 +1,438 @@
...
+        ]
+
+        case_0 = cases[0]
+        case_1 = cases[1]
+        case_2 = cases[2]
+        case_3 = cases[3]
+        case_4 = cases[4]
+
+        @cuda.jit
+        def kernel(out):
```
brandon-b-miller, line 36: Without this these test cases are pretty slow with a kernel for each case. This hurts my eyes but is faster. Should we do this?

numba_cuda/numba/cuda/cpython/listobj.py
```diff
@@ -299,8 +299,12 @@ def define_dtor(self):
         mod = builder.module
         # Declare dtor
         fnty = ir.FunctionType(ir.VoidType(), [cgutils.voidptr_t])
+        if isinstance(self.dtype, types.containers.List):
+            dtypestr = f"list_{self.dtype.dtype}"
+        else:
+            dtypestr = str(self.dtype)
         fn = cgutils.get_or_insert_function(
-            mod, fnty, ".dtor.list.{}".format(self.dtype)
+            mod, fnty, "numba_cuda_dtor_list_{}".format(dtypestr)
```
cpcloud:
```suggestion
            mod, fnty, f"numba_cuda_dtor_list_{dtypestr}"
```

Might as well move it to f-strings if we're changing the code.

numba_cuda/numba/cuda/memory_management/nrt.cu
```diff
@@ -78,6 +78,39 @@ extern "C" __device__ void NRT_MemInfo_call_dtor(NRT_MemInfo* mi)
...
+
+__device__ NRT_MemInfo* NRT_MemInfo_new_varsize(size_t size)
+{
+    NRT_MemInfo *mi = NULL;
+    void *data = NRT_Allocate(size);
+    if (data == NULL) {
+        return NULL; /* return early as allocation failed */
+    }
+
+    mi = NRT_MemInfo_new(data, size, nrt_varsize_dtor, NULL);
```
cpcloud:
```suggestion
    void *data = NRT_Allocate(size);
    if (data == NULL) {
        return NULL; /* return early as allocation failed */
    }

    NRT_MemInfo *mi = NRT_MemInfo_new(data, size, nrt_varsize_dtor, NULL);
```

More readable IMO due to variables being declared closer to their use site(s).

numba_cuda/numba/cuda/memory_management/nrt.cu
```diff
@@ -78,6 +78,39 @@ extern "C" __device__ void NRT_MemInfo_call_dtor(NRT_MemInfo* mi)
...
+    if (info) {
+        /* call element dtor */
+        typedef void dtor_fn_t(void *ptr);
+        dtor_fn_t *dtor = (dtor_fn_t *)info;
+        dtor(ptr);
+    }
+    NRT_Free(ptr);
+}
+
+__device__ NRT_MemInfo* NRT_MemInfo_new_varsize(size_t size)
```
cpcloud, line 93:
```suggestion
static NRT_MemInfo* __device__ NRT_MemInfo_new_varsize(size_t size)
```

Can this also be `static` since it's only called by `NRT_MemInfo_new_varsize_dtor`? It's also not declared in the corresponding `nrt.cuh` header file, so it looks like a private function.

brandon-b-miller: That is a good question. So far I think I've mostly exported things into the header, the NRT CUDA API is technically user facing so maybe it's best to continue exposing it.

gmarkall: I guess it should technically be declared in the header, because it can be called from generated code.

numba_cuda/numba/cuda/tests/test_array_reductions.py
gmarkall: I think the test time can really balloon when you start compiling a lot of kernels, so this is probably a good idea even if it doesn't look great. I wonder if it can be cleaned up by making cases a tuple and using `@literal_unroll` to loop over the cases though.

numba_cuda/numba/cuda/tests/test_array_reductions.py
gmarkall, line 1: This should probably be in `cudapy`. I just realised from this, that we have a bunch of other uncategorized tests (`test_byteflow`, etc.) that have accidentally crept in too.

numba_cuda/numba/cuda/tests/test_array_reductions.py
gmarkall: This fixes up `literal_unroll` for CUDA and reduces a bit of the boilerplate / repetition - what do you think of it?

```diff
diff --git a/numba_cuda/numba/cuda/device_init.py b/numba_cuda/numba/cuda/device_init.py
index b4258578..ed3f5249 100644
--- a/numba_cuda/numba/cuda/device_init.py
+++ b/numba_cuda/numba/cuda/device_init.py
@@ -87,6 +87,9 @@ from numba.cuda.cudadrv.linkable_code import (
     PTXSource,
 )
 
+from numba.cuda.misc.special import literal_unroll
+from numba.cuda.misc import literal
+
 reduce = Reduce = reduction.Reduce
 
 # Expose vector type constructors and aliases as module level attributes.
diff --git a/numba_cuda/numba/cuda/tests/test_array_reductions.py b/numba_cuda/numba/cuda/tests/test_array_reductions.py
index 44dc671a..5717a9e5 100644
--- a/numba_cuda/numba/cuda/tests/test_array_reductions.py
+++ b/numba_cuda/numba/cuda/tests/test_array_reductions.py
@@ -4,6 +4,7 @@ import numpy as np
 
 from numba.tests.support import TestCase, MemoryLeakMixin
 from numba import cuda
+from numba.cuda import literal_unroll
 from numba.cuda import config
 
 
@@ -23,38 +24,24 @@ class TestArrayReductions(MemoryLeakMixin, TestCase):
         super(TestArrayReductions, self).tearDown()
 
     def test_all_basic(self):
-        cases = [
+        cases = (
             np.float64([1.0, 0.0, float("inf"), float("nan")]),
             np.float64([1.0, -0.0, float("inf"), float("nan")]),
             np.float64([1.0, 1.5, float("inf"), float("nan")]),
             np.float64([[1.0, 1.5], [float("inf"), float("nan")]]),
             np.float64([[1.0, 1.5], [1.5, 1.0]]),
-        ]
-
-        case_0 = cases[0]
-        case_1 = cases[1]
-        case_2 = cases[2]
-        case_3 = cases[3]
-        case_4 = cases[4]
+            )
 
         @cuda.jit
         def kernel(out):
-            gid = cuda.grid(1)
-            if gid == 0:
-                ans = np.all(case_0)
-            if gid == 1:
-                ans = np.all(case_1)
-            if gid == 2:
-                ans = np.all(case_2)
-            if gid == 3:
-                ans = np.all(case_3)
-            if gid == 4:
-                ans = np.all(case_4)
-            out[gid] = ans
+            i = 0
+            for case in literal_unroll(cases):
+                out[i] = np.all(case)
+                i += 1
 
         expected = np.array([np.all(a) for a in cases], dtype=np.bool_)
         out = cuda.to_device(np.zeros(len(cases), dtype=np.bool_))
-        kernel[1, len(cases)](out)
+        kernel[1, 1](out)
         got = out.copy_to_host()
 
         self.assertPreciseEqual(expected, got)
```

numba_cuda/numba/cuda/tests/test_array_reductions.py
brandon-b-miller: Yeah, I noticed that too. Should everything frontend related be tested under `cudapy` then? Happy to move the rest of them before/after this PR.

numba_cuda/numba/cuda/device_init.py
```diff
@@ -87,8 +87,8 @@
     PTXSource,
 )
 
-from numba.cuda.misc.special import literal_unroll
-from numba.cuda.misc import literal
+# from numba.cuda.misc.special import literal_unroll
```
gmarkall: The purpose of importing this is to provide `numba.cuda.literal_unroll()` as a public API, so if we're going to use / support `literal_unroll()` (which I think is a good thing, and we should do it), we need this import.

numba_cuda/numba/cuda/device_init.py
```diff
@@ -87,8 +87,8 @@
...
+# from numba.cuda.misc import literal
```
gmarkall: I added this import so that the overload for `literal_unroll()` would get registered. Without it, the implementation won't be found. It doesn't necessarily need to be imported here (and if we want to avoid exposing it, we could do `del literal` afterwards) but it needs to get imported somehow when everything is initialized / imported.

numba_cuda/numba/cuda/typing/context.py
```diff
@@ -522,7 +522,7 @@ def is_external(obj):
             else:
                 # A type was already inserted, see if we can add to it
                 newty = existing.augment(gty)
-                if newty is None:
+                if newty is None and existing != gty:
```
gmarkall: I'm concerned that this change hides some logic error we have in our typing implementation - looking into it now.

gmarkall: I think we have an erroneous double-registration of the number classes in our codebase - avoiding the second registration in npydecl seems to avoid it, and allows us to avoid making a more architectural modification of the typing context. I have this commit locally at the moment, but would like to pinpoint why the double-registration occurs:

```diff
commit f7111fbf4980a1e5d2b041f1537ed72df1f522a6 (HEAD -> vendor-test-array-reductions)
Author: Graham Markall <gmarkall@nvidia.com>
Date:   Tue Oct 28 12:16:34 2025 +0000

    Fix double-registration of number classes
    
    Undo workaround involving modifying the typing context.

diff --git a/numba_cuda/numba/cuda/target.py b/numba_cuda/numba/cuda/target.py
index ac227669..af80a7e3 100644
--- a/numba_cuda/numba/cuda/target.py
+++ b/numba_cuda/numba/cuda/target.py
@@ -36,7 +36,7 @@ from numba.cuda.core import config, targetconfig
 # Typing
 
 
-class CUDATypingContext(typing.Context):
+class CUDATypingContext(typing.BaseContext):
     def load_additional_registries(self):
         from . import (
             cudadecl,
@@ -46,18 +46,18 @@ class CUDATypingContext(typing.Context):
             libdevicedecl,
             vector_types,
         )
-        from numba.cuda.typing import enumdecl, cffi_utils
+        from numba.cuda.typing import enumdecl, cffi_utils, npydecl
 
         self.install_registry(cudadecl.registry)
         self.install_registry(cffi_utils.registry)
         self.install_registry(cudamath.registry)
         self.install_registry(cmathdecl.registry)
         self.install_registry(libdevicedecl.registry)
+        self.install_registry(npydecl.registry)
         self.install_registry(enumdecl.registry)
         self.install_registry(vector_types.typing_registry)
         self.install_registry(fp16.typing_registry)
         self.install_registry(bf16.typing_registry)
-        super().load_additional_registries()
 
     def resolve_value_type(self, val):
         # treat other dispatcher object as another device function
diff --git a/numba_cuda/numba/cuda/typing/context.py b/numba_cuda/numba/cuda/typing/context.py
index 88f134ed..e6ac0ad8 100644
--- a/numba_cuda/numba/cuda/typing/context.py
+++ b/numba_cuda/numba/cuda/typing/context.py
@@ -522,7 +522,7 @@ class BaseContext(object):
             else:
                 # A type was already inserted, see if we can add to it
                 newty = existing.augment(gty)
-                if newty is None and existing != gty:
+                if newty is None:
                     raise TypeError(
                         "cannot augment %s with %s" % (existing, gty)
                     )
diff --git a/numba_cuda/numba/cuda/typing/npydecl.py b/numba_cuda/numba/cuda/typing/npydecl.py
index 1c20849d..faa34dd5 100644
--- a/numba_cuda/numba/cuda/typing/npydecl.py
+++ b/numba_cuda/numba/cuda/typing/npydecl.py
@@ -558,7 +558,7 @@ def register_number_classes(register_global):
         register_global(np_type, types.NumberClass(nb_type))
 
 
-register_number_classes(infer_global)
+# register_number_classes(infer_global)
 
 # -----------------------------------------------------------------------------
 # Numpy array constructors
```

gmarkall: Forgot to mention, with the above patch all tests pass locally for me, in conjunction with putting back the import of `literal_unroll()` so that the API is available for the tests.

gmarkall: I found it - in `cudadecl`, we call `register_number_classes`, which we did for very partial support of the NumPy API early on. Now that the "real" npydecl" is included, it is no longer needed in `cudadecl`. We should probably remove the NumPy ufunc registrations in `cudadecl` if we're getting them from `npydecl` too.

gmarkall: Here's what the real fix looks like:

```diff
commit 04f7cfc65ac82aff17747dbedaf3270950a093cb
Author: Graham Markall <gmarkall@nvidia.com>
Date:   Tue Oct 28 12:16:34 2025 +0000

    Fix double-registration of items from npydecl
    
    Historically cudadecl registered a few things from npydecl to support
    NumPy dtypes in kernels and a small set of ufuncs. Now that the real
    `npydecl` is registered, these items should not also be registered in
    `cudadecl`, because this leads to an erroneous double-registration.
    
    Also undo the workaround involving modifying the typing context.

diff --git a/numba_cuda/numba/cuda/cudadecl.py b/numba_cuda/numba/cuda/cudadecl.py
index 6b805514..6f1d7ed9 100644
--- a/numba_cuda/numba/cuda/cudadecl.py
+++ b/numba_cuda/numba/cuda/cudadecl.py
@@ -6,12 +6,6 @@ from numba.cuda import types
 from numba.cuda.typing.npydecl import (
     parse_dtype,
     parse_shape,
-    register_number_classes,
-    register_numpy_ufunc,
-    trigonometric_functions,
-    comparison_functions,
-    math_operations,
-    bit_twiddling_functions,
 )
 from numba.cuda.typing.templates import (
     AttributeTemplate,
@@ -29,8 +23,6 @@ register = registry.register
 register_attr = registry.register_attr
 register_global = registry.register_global
 
-register_number_classes(register_global)
-
 
 class Cuda_array_decl(CallableTemplate):
     def generic(self):
@@ -562,19 +554,3 @@ class CudaModuleTemplate(AttributeTemplate):
 
 
 register_global(cuda, types.Module(cuda))
-
-
-# NumPy
-
-for func in trigonometric_functions:
-    register_numpy_ufunc(func, register_global)
-
-for func in comparison_functions:
-    register_numpy_ufunc(func, register_global)
-
-for func in bit_twiddling_functions:
-    register_numpy_ufunc(func, register_global)
-
-for func in math_operations:
-    if func in ("log", "log2", "log10"):
-        register_numpy_ufunc(func, register_global)
diff --git a/numba_cuda/numba/cuda/target.py b/numba_cuda/numba/cuda/target.py
index ac227669..af80a7e3 100644
--- a/numba_cuda/numba/cuda/target.py
+++ b/numba_cuda/numba/cuda/target.py
@@ -36,7 +36,7 @@ from numba.cuda.core import config, targetconfig
 # Typing
 
 
-class CUDATypingContext(typing.Context):
+class CUDATypingContext(typing.BaseContext):
     def load_additional_registries(self):
         from . import (
             cudadecl,
@@ -46,18 +46,18 @@ class CUDATypingContext(typing.Context):
             libdevicedecl,
             vector_types,
         )
-        from numba.cuda.typing import enumdecl, cffi_utils
+        from numba.cuda.typing import enumdecl, cffi_utils, npydecl
 
         self.install_registry(cudadecl.registry)
         self.install_registry(cffi_utils.registry)
         self.install_registry(cudamath.registry)
         self.install_registry(cmathdecl.registry)
         self.install_registry(libdevicedecl.registry)
+        self.install_registry(npydecl.registry)
         self.install_registry(enumdecl.registry)
         self.install_registry(vector_types.typing_registry)
         self.install_registry(fp16.typing_registry)
         self.install_registry(bf16.typing_registry)
-        super().load_additional_registries()
 
     def resolve_value_type(self, val):
         # treat other dispatcher object as another device function
diff --git a/numba_cuda/numba/cuda/typing/context.py b/numba_cuda/numba/cuda/typing/context.py
index 88f134ed..e6ac0ad8 100644
--- a/numba_cuda/numba/cuda/typing/context.py
+++ b/numba_cuda/numba/cuda/typing/context.py
@@ -522,7 +522,7 @@ class BaseContext(object):
             else:
                 # A type was already inserted, see if we can add to it
                 newty = existing.augment(gty)
-                if newty is None and existing != gty:
+                if newty is None:
                     raise TypeError(
                         "cannot augment %s with %s" % (existing, gty)
                     )
```

numba_cuda/numba/cuda/ufuncs.py
```diff
@@ -682,4 +682,65 @@ def np_real_atanh_impl(context, builder, sig, args):
...
+        "L->L": numbers.real_conjugate_impl,
+        "q->q": numbers.real_conjugate_impl,
+        "Q->Q": numbers.real_conjugate_impl,
+        "f->f": numbers.real_conjugate_impl,
+        "d->d": numbers.real_conjugate_impl,
+        "F->F": numbers.complex_conjugate_impl,
+        "D->D": numbers.complex_conjugate_impl,
+    }
+
+    db[np.divide] = {
```
gmarkall, line 729: I think this comes from a branch in Numba which is followed when `np.divide != np.true`, which doesn't seem to be the case for any supported version of NumPy. From the docs at https://numpy.org/doc/1.23/reference/generated/numpy.divide.html:

> The `true_divide(x1, x2)` function is an alias for `divide(x1, x2)`.

These integer implementations should probably be using `np_int_truediv_impl` instead.

numba_cuda/numba/cuda/tests/test_array_reductions.py
gmarkall: Yeah, I think `cudapy` makes sense for all frontend-related stuff.

numba_cuda/numba/cuda/device_init.py
brandon-b-miller: This is partially solvable easily - I believe the imports themselves can be added to the target context `load_additional_registries` function with the effect of having import delayed until a compilation is invoked. However the public availability of the API as `from numba.cuda import literal` violates the [test that checks that we don't register lowerings upon import](https://github.com/NVIDIA/numba-cuda/blob/58b0897e4ae1ccc390fe57619c048f0ac6194a12/numba_cuda/numba/cuda/tests/nocuda/test_import.py#L41). 

To me there are two solutions, either refactor the `literal` module such that it registers lowering in a lazy way (and remove it from the banlist), or implement a pep 562 style module level gettattr that delays the actual import of `literal` until a user attempts to access it (or the context has already imported it).

brandon-b-miller: Something to get us off the ground at https://github.com/NVIDIA/numba-cuda/pull/523/commits/f210837cec34715408207ab0a8c5928febe8e234

gmarkall: The import of

```python
from numba.cuda.misc.special import literal_unroll
```

needs to be here so that we provide the API for public use.

I think we can remove `numba.cuda.misc` from the banlist, as it doesn't overload anything that affects other targets - I would suggest doing that for this PR.

There are other CUDA modules (e.g. `numba.cuda.cg`) that register overloads for the CUDA target on import, so I don't think there's a problem with doing it. The banlist in that test has grown as we've vendored code, but it's protecting against the risk of us polluting the CPU target, so a lot of it is unnecessary.

PR #526, opened 2025-10-15
dev: allow running tests from the root of the repo

cpcloud:
The hassle of running tests from the `testing` directory does not appear to be necessary, and it breaks running specific tests with pytest. This PR moves the configuration to the top level of the repo, in `pyproject.toml` and allows running tests in the usual way without anything special.


pyproject.toml
```diff
@@ -156,10 +156,11 @@ numba-cuda = { path = ".", editable = true }
 [tool.pixi.environments]
 cu12 = { features = ["cu12", "test"], solve-group = "cu12" }
 cu13 = { features = ["cu13", "test"], solve-group = "cu13" }
-test = { features = ["test"], solve-group = "test" }
-
-[tool.pixi.tasks]
-test = { cmd = ["pytest", "$PIXI_PROJECT_ROOT/numba_cuda/numba/cuda/tests"] }
```
cpcloud, line 162: This seems overkill given that the previous behavior specified here is now the default behavior in this PR.

cpcloud:
Abandoning this for now.

PR #527, opened 2025-10-15
test: speed up ipc tests by ~6.5x

cpcloud:
This PR speeds up multiprocessing-based IPC tests by using the concurrent futures interface.

This is faster because processes are reused from a pool as opposed to creating all the machinery
to run a thing in new process in every single test or `subTest` invocation.

While this is less isolated, I would argue that it's a much more common use case to use pools and/or
the concurrent futures interface (it's much less noisy, for example).

### DIFF
--- a/numba_cuda/numba/cuda/tests/cudapy/test_ipc.py
+++ b/numba_cuda/numba/cuda/tests/cudapy/test_ipc.py
@@ -1,6 +1,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: BSD-2-Clause
 
+import pytest
 import concurrent.futures
 import multiprocessing as mp
 import os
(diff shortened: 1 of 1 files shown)


numba_cuda/numba/cuda/tests/cudapy/test_ipc.py
```diff
@@ -21,80 +22,69 @@
...
-    def the_work():
-        dtype = np.dtype(np.intp)
-        darr = handle.open_array(
-            cuda.current_context(),
-            shape=handle.size // dtype.itemsize,
-            dtype=dtype,
-        )
+def base_ipc_handle_test(handle, size, parent_pid):
+    pid = os.getpid()
+    assert pid != parent_pid
```
cpcloud, line 28: Passing the `parent_pid` here isn't strictly necessary and I can remove it if desired, I was just verifying my understanding of what was happening with `ProcessPoolExecutor`.

cpcloud:
@gmarkall Would you mind reviewing this?

gmarkall (approved):
> While this is less isolated, I would argue that it's a much more common use case to use pools and/or
> the concurrent futures interface (it's much less noisy, for example).

I think this is fine - it's more that we're testing that our API works here anyway, than verifying the correctness / robustness of the underlying driver APIs.

PR #529, opened 2025-10-16
test: enable fail-on-warn and clean up resulting failures

cpcloud:
This PR cleans up all warnings that are currently being dumped out of test runs.

### DIFF
--- a/numba_cuda/numba/cuda/simulator/cudadrv/devicearray.py
+++ b/numba_cuda/numba/cuda/simulator/cudadrv/devicearray.py
@@ -8,6 +8,8 @@ stubs to allow tests to import correctly.
 
 from contextlib import contextmanager
 from numba.cuda.np.numpy_support import numpy_version
+from numba.cuda.np import numpy_support
+from numba.cuda import types
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


numba_cuda/numba/cuda/tests/test_extending.py
```diff
@@ -92,12 +93,14 @@ def test_func1(self):
         pyfunc = call_func1_nullary
         cfunc = jit(pyfunc)
         res = np.zeros(1)
-        cfunc[1, 1](res)
+        with override_config("DISABLE_PERFORMANCE_WARNINGS", 1):
```
kkraus14, line 102: Should we do this at the class or module level instead of doing it at each launch point?

numba_cuda/numba/cuda/tests/cudapy/test_compiler.py
```diff
@@ -584,7 +584,13 @@ def f(z, x, y):
...
         args = (float32[::1], float32, float32)
         code_list, resty = compile_all(
-            f, args, debug=True, output="ptx", device=False, abi="numba"
+            f,
+            args,
+            debug=True,
+            output="ptx",
+            device=False,
+            abi="numba",
+            opt=False,
```
kkraus14, line 593: Was it required to explicitly disable optimizations here? Is that expected?

cpcloud: Yes, to prevent the subsequent warning.

numba_cuda/numba/cuda/tests/test_extending.py
cpcloud: I suppose we can. I was going to do a slightly deeper refactor later of some of this stuff to use pytest autouse fixtures and I think leaving this as is might make that effort less churny.

cpcloud: Actually, doing it in `setUp`/`tearDown` might be the least churny.

kkraus14 (approved):

PR #530, opened 2025-10-16
refactor: remove unnecessary custom map and set implementations

cpcloud:
This PR removes some custom mapping and set implementations that I believe can be supplanted by `dict`s (and judicious use of `sorted` where applicable).


numba_cuda/numba/cuda/lowering.py
```diff
@@ -292,7 +292,7 @@ def lower_function_body(self):
         )
 
         # Lower all blocks
-        for offset, block in sorted(self.blocks.items()):
```
gmarkall, line 295: If I'm reading this right, `self.blocks` was a `SortedMap`, and then we sorted it again before lowering. 😆

PR #533, opened 2025-10-16
Vendor in types and datamodel for CUDA-specific changes

VijayKandiah:
This is a large PR that vendors in `types` and `datamodel`  from Numba for future CUDA-specific customizations.  This PR includes a `convert_to_cuda_type` API in `numba.cuda.core.sigutils` which is called by default inside `normalize_signature` to map `numba.core.types` (if any) to equivalent `numba.cuda.types` if possible. This mapping enables the typing system in numba-cuda to work on `numba.core.types`.

### DIFF
--- a/numba_cuda/numba/cuda/api.py
+++ b/numba_cuda/numba/cuda/api.py
@@ -39,10 +39,9 @@ def from_cuda_array_interface(desc, owner=None, sync=True):
 
     shape = desc["shape"]
     strides = desc.get("strides")
-    dtype = np.dtype(desc["typestr"])
 
     shape, strides, dtype = prepare_shape_strides_dtype(
-        shape, strides, dtype, order="C"
+        shape, strides, desc["typestr"], order="C"
     )
     size = driver.memory_size_from_info(shape, strides, dtype.itemsize)
 
--- a/numba_cuda/numba/cuda/api_util.py
+++ b/numba_cuda/numba/cuda/api_util.py
@@ -3,6 +3,8 @@
 
 import numpy as np
 
+import functools
+
 
 def prepare_shape_strides_dtype(shape, strides, dtype, order):
     dtype = np.dtype(dtype)
@@ -14,25 +16,33 @@ def prepare_shape_strides_dtype(shape, strides, dtype, order):
         raise TypeError("shape must be an integer or tuple of integers")
     if isinstance(shape, int):
         shape = (shape,)
+    else:
+        shape = tuple(shape)
     if isinstance(strides, int):
         strides = (strides,)
     else:
-        strides = strides or _fill_stride_by_order(shape, dtype, order)
+        if not strides:
+            strides = _fill_stride_by_order(shape, dtype, order)
+        else:
+            strides = tuple(strides)
     return shape, strides, dtype
 
 
+@functools.cache
 def _fill_stride_by_order(shape, dtype, order):
-    nd = len(shape)
-    if nd == 0:
+    ndims = len(shape)
+    if not ndims:
         return ()
-    strides = [0] * nd
+    strides = [0] * ndims
     if order == "C":
         strides[-1] = dtype.itemsize
-        for d in reversed(range(nd - 1)):
+        # -2 because we subtract one for zero-based indexing and another one
+        # for skipping the already-filled-in last element
+        for d in range(ndims - 2, -1, -1):
             strides[d] = strides[d + 1] * shape[d + 1]
     elif order == "F":
         strides[0] = dtype.itemsize
-        for d in range(1, nd):
+        for d in range(1, ndims):
             strides[d] = strides[d - 1] * shape[d - 1]
     else:
         raise ValueError("must be either C/F order")
--- a/numba_cuda/numba/cuda/cudadrv/devicearray.py
+++ b/numba_cuda/numba/cuda/cudadrv/devicearray.py
@@ -86,8 +86,13 @@ class DeviceNDArrayBase(_devicearray.DeviceArray):
         """
         if isinstance(shape, int):
             shape = (shape,)
+        else:
+            shape = tuple(shape)
         if isinstance(strides, int):
             strides = (strides,)
+        else:
+            if strides:
+                strides = tuple(strides)
         dtype = np.dtype(dtype)
         itemsize = dtype.itemsize
         self.ndim = ndim = len(shape)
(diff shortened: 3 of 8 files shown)


VijayKandiah:
Working on the cuDF failures.

numba_cuda/numba/cuda/typing/arraydecl.py
```diff
@@ -30,6 +30,10 @@
 
 numpy_version = tuple(map(int, np.__version__.split(".")[:2]))
 
+registry = Registry()
+infer = registry.register
+infer_global = registry.register_global
+infer_getattr = registry.register_attr
```
gmarkall, line 36: This seems to accidentally duplicate the lines below:
```suggestion
```

numba_cuda/numba/cuda/typing/collections.py
```diff
@@ -18,6 +19,12 @@
 infer_getattr = registry.register_attr
 
 
+registry = Registry()
+infer = registry.register
+infer_global = registry.register_global
+infer_getattr = registry.register_attr
```
gmarkall: Also a duplication of the above code:
```suggestion
```

numba_cuda/numba/cuda/typing/npdatetime.py
```diff
@@ -18,6 +19,9 @@
 
 # timedelta64-only operations
 
+registry = Registry()
+infer_global = registry.register_global
+
```
gmarkall:
```suggestion
```

numba_cuda/numba/cuda/core/sigutils.py
```diff
@@ -1,15 +1,124 @@
...
+
+This module also provides type mapping between numba.core types
+and numba.cuda types to ensure cross-compatibility.
+"""
+
+from numba.cuda import types, typing
 
 try:
     from numba.core.typing import Signature as CoreSignature
+    from numba.core import types as core_types
```
brandon-b-miller: Not blocking for this PR but have we considered determining the presence of numba in a central way up front that we can just reference instead of proliferating try/catches? `if HAVE_NUMBA:` etc

VijayKandiah: I like that idea. Yes, we should be doing that. A centralized `HAVE_NUMBA` is much cleaner than all the try imports we have in `typing` and other modules. This can be in a future PR.

VijayKandiah: Also, since we have the redirector in place, this type conversion/mapping in sigutils is not really needed. If we have `numba` in the env, the redirector will ensure that numba.core.types are used everywhere. So we would not encounter a case where there'd be a mix of numba.cuda.types and numba.core.types needing this conversion. Should we just remove this?

VijayKandiah: I removed the type mapping.

numba_cuda/numba/cuda/types.py
```diff
@@ -1,99 +1,13 @@
...
-        if isinstance(other, (types.Float, types.Integer)):
-            return typingctx.unify_pairs(self, other)
-
-
-bfloat16 = Bfloat16()
+import importlib
+import sys
+from numba.cuda.utils import _RedirectSubpackage
+
+if importlib.util.find_spec("numba.core.types"):
```
gmarkall: This redirection method does not work - if I change this to simulate what would happen if Numba were not found, like:

```diff
diff --git a/numba_cuda/numba/cuda/types.py b/numba_cuda/numba/cuda/types.py
index 94e30c17..ba81757e 100644
--- a/numba_cuda/numba/cuda/types.py
+++ b/numba_cuda/numba/cuda/types.py
@@ -5,7 +5,7 @@ import importlib
 import sys
 from numba.cuda.utils import _RedirectSubpackage
 
-if importlib.util.find_spec("numba.core.types"):
+if importlib.util.find_spec("numba.core.types") and False:
     sys.modules[__name__] = _RedirectSubpackage(locals(), "numba.core.types")
 else:
     sys.modules[__name__] = _RedirectSubpackage(
```

then importing `cuda` from `numba` no longer works:

```
$ python -c "from numba import cuda"
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/__init__.py", line 11, in <module>
    import numba.cuda.types as types
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/types.py", line 11, in <module>
    sys.modules[__name__] = _RedirectSubpackage(
                            ^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/utils.py", line 592, in __init__
    new_mod_obj = import_module(new_module)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/miniforge3/envs/numbadev/lib/python3.12/importlib/__init__.py", line 98, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/cuda_types/__init__.py", line 10, in <module>
    from .containers import *
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/cuda_types/containers.py", line 26, in <module>
    from .misc import Undefined, unliteral, Optional, NoneType
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/cuda_types/misc.py", line 4, in <module>
    from numba.cuda.types.abstract import Callable, Literal, Type, Hashable
ModuleNotFoundError: No module named 'numba.cuda.types.abstract'; 'numba.cuda.types' is not a package
```

gmarkall: Just realised I forgot to clarify: the problem is you can't use a module to redirect a package, only another module - this is why when I prototyped this I had individual redirectors like `cuda_abstract.py` -> `abstract.py`, etc.

numba_cuda/numba/cuda/ext_types.py
gmarkall, line 1: I don't think any of the globals in here (e.g. `Dim3`, `GridGroup`), etc., have been used like public API under `numba.cuda.types` before, so this should be a safe move. If this turns out to be incorrect we could always move the ones that have been used publicly into `types/__init__.py`.

PR #534, opened 2025-10-17
[Refactor][NFC] Vendor-in errors for future CUDA-specific changes

atmnp:
This change vendors in numba.core.errors for future CUDA-specific changes, but it's blocked on PR #533. PR#533 adds types and datamodels which raise errors, and without that change, there will be discrepancies between the errors thrown/caught based on the module it originates from. 

This change also incorporates some minor import cleanups to use already vendored in modules instead of upstream ones.

I'm pushing this now to test with CI.

### DIFF
--- a/numba_cuda/numba/cuda/core/ir_utils.py
+++ b/numba_cuda/numba/cuda/core/ir_utils.py
@@ -529,7 +529,7 @@ def add_offset_to_labels(blocks, offset):
             for inst in b.body:
                 for T, f in add_offset_to_labels_extensions.items():
                     if isinstance(inst, T):
-                        f_max = f(inst, offset)  # noqa: F841
+                        f(inst, offset)
         if isinstance(term, ir.Jump):
             b.body[-1] = ir.Jump(term.target + offset, term.loc)
         if isinstance(term, ir.Branch):
@@ -546,9 +546,7 @@ find_max_label_extensions = {}
 def find_max_label(blocks):
     max_label = 0
     for l, b in blocks.items():
-        term = None
         if b.body:
-            term = b.body[-1]  # noqa: F841
             for inst in b.body:
                 for T, f in find_max_label_extensions.items():
                     if isinstance(inst, T):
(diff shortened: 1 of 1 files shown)


numba_cuda/numba/cuda/core/analysis.py
```diff
@@ -3,7 +3,8 @@
 
 from collections import namedtuple, defaultdict
 from numba.cuda import types
-from numba.core import ir, errors
+from numba.cuda.core import ir
+from numba.core import errors
```
gmarkall, line 7: Is there a reason this still comes from `numba.core`?

numba_cuda/numba/cuda/core/bytecode.py
```diff
@@ -8,7 +8,8 @@
 
 from types import CodeType, ModuleType
 
-from numba.core import errors, serialize
+from numba.core import errors
```
gmarkall, line 11: Should these imports also come from numba.cuda?

numba_cuda/numba/cuda/core/ir_utils.py
gmarkall, line 1: I see the commit that changed this says "fix ir_utils that was vendored in incorrectly". What does that mean? Why these changes?

numba_cuda/numba/cuda/core/typeinfer.py
```diff
@@ -49,6 +49,8 @@
     CompilerError,
     NumbaValueError,
 )
+
+# from numba.core import errors as cudaErrors
```
gmarkall:
```suggestion
```

numba_cuda/numba/cuda/core/ir_utils.py
```diff
@@ -2162,7 +2165,7 @@ def is_namedtuple_class(c):
         return False
     # should have only tuple as superclass
     bases = c.__bases__
-    if len(bases) != 1 or bases[0] is not tuple:
+    if len(bases) != 1 or isinstance(bases[0], tuple):
```
gmarkall: The other changes appear to have no side effects, but this effectively inverts the condition. I can't find a version like this in Numba - it appears unchanged since it was first added. What is this change for?

gmarkall:
In the interests of getting this merged, I'm checking what happens with the `ir_utils` changes undone - if it passes, I'm inclined to merge this. Then, if there are still issues with `ir_utils`, we can address that in a follow-up.

PR #536, opened 2025-10-21
perf: remove context threading in various pointer abstractions

cpcloud:
This PR removes the `context` member of the various pointer abstractions in the
codebase.

None of these `context`s were ever being used, except as arguments to
constructors of these types.

The main implication is that `from_cuda_array_interface` no longer
needs either `require_context` nor `current_context()` (since that seemed to be
the reason for having the decorator). Over thousands of calls to `numba.cuda.as_cuda_array()`, this
repeated context acquisition is expensive.

xref: https://github.com/numba/numba/issues/9835

This is not the only overhead for that function that we can address, but this is a set of self-contained changes.

I will open separate PRs for the others.

### DIFF
--- a/numba_cuda/numba/cuda/api.py
+++ b/numba_cuda/numba/cuda/api.py
@@ -21,7 +21,6 @@ current_context = devices.get_context
 gpus = devices.gpus
 
 
-@require_context
 def from_cuda_array_interface(desc, owner=None, sync=True):
     """Create a DeviceNDArray from a cuda-array-interface description.
     The ``owner`` is the owner of the underlying memory.


cpcloud:
Perhaps I am misunderstanding where `require_context` is needed, because I see lots of functions with this decorator that never explicitly use the context, but use it somewhere deep in the call stack.

One thing that would help me understand is if there were a test that fails when I remove its `require_context` decorator. I haven't yet found one.

cpcloud:
~It seems like `require_context` may still be needed, but the `current_context()` calls still look unnecessary.~

EDIT: I looked through the calls in `from_cuda_array_interface` and all of the corresponding APIs that require a context have `@require_context`, so I think we can keep it removed from `from_cuda_array_interface` if I am understanding things correctly.

rparolin:
The code changes look good to me.  I'd just confirm with @leofang and @gmarkall to confirm that the context is not required in a future design.

rparolin (approved):

gmarkall:
> Perhaps I am misunderstanding where `require_context` is needed, because I see lots of functions with this decorator that never explicitly use the context, but use it somewhere deep in the call stack.

I suspect they were added out of an abundance of caution. If a function really requires a context, I think it should place the requirement as close to the usage as possible.

> One thing that would help me understand is if there were a test that fails when I remove its `require_context` decorator. I haven't yet found one.

I imagine the use of the decorators are mostly superfluous.

numba_cuda/numba/cuda/tests/cudadrv/test_emm_plugins.py
```diff
@@ -58,10 +57,9 @@ def finalizer():
 
             # We use an AutoFreePointer so that the finalizer will be run when
             # the reference count drops to zero.
-            ctx = weakref.proxy(self.context)
```
gmarkall, line 61: I think this is also a nice reduction in complexity / cognitive load.

gmarkall (approved):

PR #537, opened 2025-10-21
perf: remove duplicated size computation

cpcloud:
This PR removes some overhead caused by repeated computation of array `size` by forwarding attributes from the underlying dummy array.

### DIFF
--- a/numba_cuda/numba/cuda/cudadrv/devicearray.py
+++ b/numba_cuda/numba/cuda/cudadrv/devicearray.py
@@ -90,26 +90,27 @@ class DeviceNDArrayBase(_devicearray.DeviceArray):
         if isinstance(strides, int):
             strides = (strides,)
         dtype = np.dtype(dtype)
-        self.ndim = len(shape)
-        if len(strides) != self.ndim:
+        itemsize = dtype.itemsize
+        self.ndim = ndim = len(shape)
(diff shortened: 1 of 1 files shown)


numba_cuda/numba/cuda/cudadrv/devicearray.py
```diff
@@ -90,26 +90,27 @@ def __init__(self, shape, strides, dtype, stream=0, gpu_data=None):
...
         )
         # confirm that all elements of shape are ints
         if not all(isinstance(dim, (int, np.integer)) for dim in shape):
             raise TypeError("all elements of shape must be ints")
-        self.shape = tuple(shape)
-        self.strides = tuple(strides)
+        self.shape = shape = dummy.shape
+        self.strides = strides = dummy.strides
         self.dtype = dtype
-        self.size = int(functools.reduce(operator.mul, self.shape, 1))
```
cpcloud, line 105: This is the line that is redundant with `dummy.size`.

gmarkall (approved):

cpcloud:
> Nit: Is there a test that we could create where if a perf regression is introduced in the future that we could catch it?

Not without more work to setup some basic statistical benchmarking and roll it into CI. I'd like to do this at some point soon, but it was out of scope for this PR.

PR #538, opened 2025-10-21
perf: reduce the number of `__cuda_array_interface__` accesses

cpcloud:
This PR reduces the number of accesses of `__cuda_array_interface__`, which shows up when calling `as_cuda_array` in a loop with a large number of iterations.

### DIFF
--- a/numba_cuda/numba/cuda/cudadrv/devicearray.py
+++ b/numba_cuda/numba/cuda/cudadrv/devicearray.py
@@ -15,7 +15,6 @@ from ctypes import c_void_p
 
 import numpy as np
 
-import numba
 from numba.cuda.cext import _devicearray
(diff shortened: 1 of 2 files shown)


numba_cuda/numba/cuda/cudadrv/devicearray.py
```diff
@@ -901,8 +900,12 @@ def auto_device(obj, stream=0, copy=True, user_explicit=False):
     """
     if _driver.is_device_memory(obj):
         return obj, False
-    elif hasattr(obj, "__cuda_array_interface__"):
-        return numba.cuda.as_cuda_array(obj), False
```
cpcloud, line 905: Before this PR, there were total of 3 (!) accesses of `__cuda_array_interface__` per call to `auto_device`.

PR #539, opened 2025-10-21
chore(perf): add torch to benchmark

cpcloud:
This PR adds pytorch to the benchmarking suite, in an effort to track related performance improvments to kernel launching and device array conversion.

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -7,8 +7,8 @@ environments:
     - https://pypi.org/simple
     packages:
       linux-64:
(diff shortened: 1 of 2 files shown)


cpcloud:
Current baseline:

image

cpcloud:
Just to clarify what's happening in that table:

- `device_array` is testing arrays allocated directly on the device
- `torch` is testing arrays allocated with `torch` (in theory already on the device as well)

The torch arrays are 3-7x slower than device arrays, to pass into a function that does nothing, because of the overhead of calling `numba.cuda.as_cuda_array` for the same array N-iterations times.

gmarkall (approved):

PR #541, opened 2025-10-22
feat: add support for `math.exp2`

kaeun97:
Adding support for `math.exp2`. Follow up of https://github.com/numba/numba/pull/10276.

### DIFF
--- a/numba_cuda/numba/cuda/tests/cudapy/test_fastmath.py
+++ b/numba_cuda/numba/cuda/tests/cudapy/test_fastmath.py
@@ -146,6 +146,7 @@ class TestFastMathOption(CUDATestCase):
     @unittest.skipUnless(sys.version_info >= (3, 11), "Python 3.11+ required")
     def test_exp2f(self):
         from math import exp2
+
         self._test_fast_math_unary(
             exp2,
             FastMathCriterion(


kaeun97:
@gmarkall can you re-run CI please?

kaeun97:
@gmarkall is there a way to re-run the failed tests? I don't think it is relevant to the PR

gmarkall (changes requested):
The code changes look good, and I'm expecting CI to pass. There's one other bit of test coverage - could you add a test for the exp2 function to `test_fastmath` as well please? This checks that the PTX is as-expected when fastmath is enabled.

I think you can use the existing test for `exp` in that file as a guide, but the pattern to expect when fastmath would be enabled would contain `ex2.approx.ftz.f32`.

For background, I found that pattern by running this slightly cumbersome one-liner:

```bash
NUMBA_DUMP_ASSEMBLY=1 python -c "from numba import cuda, types; import math; print(cuda.compile_ptx(lambda x: math.exp2(x), (types.float32,), fastmath=True, device=True))"
```

to get PTX output containing (abridged):

```assembly
.visible .func  (.param .b32 func_retval0) lambda_func(
	.param .b64 param_0,
	.param .b32 param_1
)
{
	.reg .f32 	%f<3>;
	.reg .b32 	%r<2>;
	.reg .b64 	%rd<2>;


	ld.param.u64 	%rd1, [param_0];
	ld.param.f32 	%f1, [param_1];
	ex2.approx.ftz.f32 	%f2, %f1;
	st.f32 	[%rd1], %f2;
	mov.u32 	%r1, 0;
	st.param.b32 	[func_retval0+0], %r1;
	ret;

}
```

kaeun97:
@gmarkall Hey Graham, thanks for the comment regarding when fastmath is enabled. ~~I could not find `__nv_fast_exp2f` [here](https://docs.nvidia.com/cuda/libdevice-users-guide/), so I have implemented `exp2` with inline assembly~~. Realized the compiler takes care of it already 😅

Added test in `test_fastmath` as you suggested. Would appreciate if you can take a look!

gmarkall:
@kaeun97 Thanks for fixing that so fast! I was just fixing it myself because I wanted to get it through CI and merge it without making us wait / go through another iteration, but you beat me to it :slightly_smiling_face:

gmarkall (approved):

gmarkall:
I've set it to auto-merge, so assuming CI passes this time it should go in.

Many thanks @kaeun97!

PR #542, opened 2025-10-22
perf: cache dimension computations

cpcloud:
This PR adds some `functools.cache` decorations to a few functions for
computing dimensions, which are repeatedly computed for the same inputs many
times.

This does trade some space to reduce duration of conversion. That space will
scale with the number of unique combinations of shape, strides, dtypes, and
memory ordering.

If someone has a lot of arrays that have different shapes,
strides, dtypes, and memory ordering, the cache can potentially use a lot of memory.

I'm not sure that's a likely scenario, but it's worth pointing out as
a potential issue.

Another question is whether we can assume that `Array`s (dummy arrays) are immutable.

If we can't, then we shouldn't be caching them.

### DIFF
--- a/numba_cuda/numba/cuda/cudadrv/devicearray.py
+++ b/numba_cuda/numba/cuda/cudadrv/devicearray.py
@@ -123,17 +123,17 @@ class DeviceNDArrayBase(_devicearray.DeviceArray):
 
     @property
     def __cuda_array_interface__(self):
-        if self.device_ctypes_pointer.value is not None:
-            ptr = self.device_ctypes_pointer.value
+        if (value := self.device_ctypes_pointer.value) is not None:
+            ptr = value
         else:
             ptr = 0
 
         return {
-            "shape": tuple(self.shape),
+            "shape": self.shape,
             "strides": None if is_contiguous(self) else tuple(self.strides),
             "data": (ptr, False),
             "typestr": self.dtype.str,
-            "stream": int(self.stream) if self.stream != 0 else None,
+            "stream": int(stream) if (stream := self.stream) != 0 else None,
             "version": 3,
         }
 


cpcloud:
Here are the performance improvements on top of `main`:

image

So, there's another ~66% improvement for the many-arg torch case and another ~34% improvement for the one-arg case torch case.

That would bring us to:

- the torch case is now about 2.5x slower than the device array case (down from 7x) in the many-arg scenario
- the torch case is now about 1.8x slower than the device array case (down from 3x) in the one-arg scenario

image

pyproject.toml
```diff
@@ -198,7 +198,7 @@ benchcmp = { cmd = [
     "numba.cuda.tests.benchmarks",
     "--benchmark-only",
     "--benchmark-enable",
-    "--benchmark-group-by=func",
+    "--benchmark-group-by=name",
```
cpcloud, line 201: This is a more useful grouping when comparing benchmarks.

numba_cuda/numba/cuda/api_util.py
```diff
@@ -14,13 +16,19 @@ def prepare_shape_strides_dtype(shape, strides, dtype, order):
...
     if isinstance(shape, int):
         shape = (shape,)
+    else:
+        shape = tuple(shape)
     if isinstance(strides, int):
         strides = (strides,)
     else:
-        strides = strides or _fill_stride_by_order(shape, dtype, order)
+        if not strides:
+            strides = _fill_stride_by_order(tuple(shape), dtype, order)
```
cpcloud: This `tuple` conversion is unnecessary here.

numba_cuda/numba/cuda/cudadrv/devicearray.py
```diff
@@ -96,9 +101,6 @@ def __init__(self, shape, strides, dtype, stream=0, gpu_data=None):
         self._dummy = dummy = dummyarray.Array.from_desc(
             0, shape, strides, itemsize
         )
-        # confirm that all elements of shape are ints
-        if not all(isinstance(dim, (int, np.integer)) for dim in shape):
```
cpcloud, line 100: I moved this because caching avoids having to run this code every time this constructor is called, saving a good chunk of time in the constructor.

Issue #520, opened 2025-10-13
[FEA] DWARF discriminated union support for Python dynamic types in Numba-CUDA


jiel-nv:
Summary:
Add support for emitting DWARF discriminated union metadata (DW_TAG_variant_part) in the Numba-CUDA compiler to enable proper debugging of Python's dynamically-typed variables in CUDA kernels.

Motivation:
Python's dynamic type system allows variables to hold different types at runtime. When compiled to CUDA, these are represented as tagged unions (discriminated unions/variants). However, current debug info lacks the discriminant metadata that allows debuggers to automatically determine which union member is active, resulting in a less ideal debugging experience where developers see garbage values for inactive union members, and have to mentally choose the correct type and value.

Expected solution:
Generate proper DWARF discriminated union LLVM IR metadata with typed discriminant values.

One dependancy idenfied:
Needs llvmlite enhancement: Support typed constants in debug info metadata.
A PR is already submitted [(PR#1326](https://github.com/numba/llvmlite/pull/1326))

Some references are:
1. LLVM DebugInfoMetadata.h: DIDerivedType::extraData field
2. DWARF 5 Specification: Section 5.7.10 (Variant Entries)

PR #544, opened 2025-10-23
Add DWARF variant part support for polymorphic variables in CUDA debug info

jiel-nv:
### Overview
This PR implements proper DWARF debug information generation for polymorphic variables in Numba CUDA kernels using `DW_TAG_variant_part` (discriminated unions). This enables debuggers like cuda-gdb to correctly inspect polymorphic variables that can hold different types throughout their lifetime.

#### 1. Enhanced Polymorphic Variable Lowering (`lowering.py`)
- With `CUDA_DEBUG_POLY` enabled:
  - Allocates storage for discriminant + data
  - Stores discriminant at offset 0 and data at offset `sizeof(fetype)`
  - Retrieves data pointer with proper offset calculation
  - Memory layout: 
   ```[discriminant: 8 bits][padding][data: varies by type]```

#### 2. DWARF Variant Part Metadata Generation (`debuginfo.py`)
- Generate proper debug info metadata for variant types:
  - Creates `DW_TAG_variant_part` with discriminator field
  - Emits variant members with `extraData` field containing discriminant values
  - Wraps variant part in a structure type with proper size calculations
  - Each variant member includes correct offset (equals element's width in bits)
  The final DWARF structure: after this change would look like the following e.g.,
  ```
  variant_wrapper_struct (DW_TAG_structure_type)
  ├── discriminator (DW_TAG_member, 8 bits)
  └── variant_part (DW_TAG_variant_part)
      ├── _bool (DW_TAG_member, extraData: 0, offset: 8)
      ├── _float64 (DW_TAG_member, extraData: 1, offset: 64)
      ├── _int32 (DW_TAG_member, extraData: 2, offset: 32)
      └── _int64 (DW_TAG_member, extraData: 3, offset: 64)
  ```
- Added llvmlite version-dependent logic for `extraData` emission:
  - Typed constant (`i8 N`) for newer llvmlite versions (> 0.45)
  - Metadata node reference for older versions (<= 0.45), requires an in progress nvvm patch
  - Controlled by `CUDA_DEBUG_POLY_USE_TYPED_CONST` config flag
- Added CTK version threshold check:
  - Feature enabled: When CTK > 13.1
  - Feature disabled: When CTK <= 13.1

#### 3. Test Coverage (`test_debuginfo.py`)
- Added a new test `test_poly_variant_part()` to verify DWARF output:
  - Conditionally turns on, opt-in via `CUDA_DEBUG_POLY` configuration flag
  - Validates complete hierarchical structure: DILocalVariable → wrapper struct → elements list → discriminator + variant_part → variant members
  - Dynamically adapts FileCheck patterns based on `CUDA_DEBUG_POLY_USE_TYPED_CONST`
  - Verifies all variant members (_bool, _float64, _int32, _int64) with correct offsets

#### 4. Backward Compatibility
- Without `CUDA_DEBUG_POLY`: Falls back to existing behavior, i.e. user mentally choose runtime type from a union like types wrapper in a debugging session.
- Older test (`test_union_poly_types`) preserved and conditionally skipped

This change implements [issue#520](https://github.com/NVIDIA/numba-cuda/issues/520).

## Summary by CodeRabbit

* **New Features**
  * Added support for polymorphic debug information for CUDA union types, enabling enhanced debugging capabilities with structured variant metadata.
  * Introduced new configuration flags to control polymorphic debug behavior and data representation modes.

* **Tests**
  * Added test coverage for polymorphic variant-part debugging information under various configuration scenarios.

### DIFF
--- a/numba_cuda/numba/cuda/debuginfo.py
+++ b/numba_cuda/numba/cuda/debuginfo.py
@@ -5,6 +5,7 @@ import abc
 import os
 from contextlib import contextmanager
 
+import llvmlite
 from llvmlite import ir
 from numba.cuda import types
 from numba.cuda.core import config
@@ -18,8 +19,6 @@ from cuda.bindings import runtime
 def _get_llvmlite_version():
     """Get llvmlite version as tuple (major, minor)."""
     try:
-        import llvmlite
-
         version_str = llvmlite.__version__
         # Parse version string like "0.46.0" or "0.46.0dev"
         parts = version_str.split(".")
@@ -28,7 +27,7 @@ def _get_llvmlite_version():
         major = int(parts[0])
         minor = int(parts[1])
         return (major, minor)
-    except (ImportError, AttributeError, ValueError):
+    except (IndexError, AttributeError, ValueError):
         return (0, 0)
 
 


gmarkall:
@coderabbitai review

numba_cuda/numba/cuda/debuginfo.py
```diff
@@ -13,6 +13,62 @@
...
+def _check_polymorphic_debug_info_support():
+    """Check if CTK and llvmlite support polymorphic debug info.
+
+    Returns:
+        tuple: (supported: bool, use_typed_const: bool)
+        - supported: Whether feature is supported at all
+        - use_typed_const: True for typed constant,
+                           False for node reference
+    """
+    try:
```
gmarkall: We should use

```
from cuda.bindings import runtime
```

at top-level scope

and call `runtime.getLocalRuntimeVersion()` to get the runtime version.

I don't think a `try` / `except` should be necessary here.

jiel-nv: Thank you! Resolved via commit 7ad4646f9e2b56847b0b265a149dc62696a96085

numba_cuda/numba/cuda/debuginfo.py
```diff
@@ -11,6 +12,51 @@
...
 from numba.core.datamodel.models import ComplexModel, UnionModel, UniTupleModel
 from numba.cuda.types import GridGroup
+from cuda.bindings import runtime
+
+
+# Check if CUDA Toolkit and llvmlite support polymorphic debug info
+def _get_llvmlite_version():
+    """Get llvmlite version as tuple (major, minor)."""
+    # Parse version string like "0.46.0" or "0.46.0dev"
+    parts = llvmlite.__version__.split(".")
```
gmarkall: Sorry, I was a little unclear in my previous comment. I had only intended to convey that a `try ... except` would not be needed for checking the CUDA runtime version. Because llvmlite uses versioneer, it's quite possible for it to mess up determining its own version. 

For example, I can create a situation where the following occurs:

```
$ python -c "import llvmlite; print(llvmlite.__version__)"
unknown
```

I did this by creating a tag in the repo that didn't fit versioneer's expectations, but I imagine versioneer has other failure modes too (the general recommendation these days seems to be to avoid using it, but llvmlite was set up with it over a decade ago now).

So, it's likely still necessary to check that unpacking the version number is actually possible. I think if llvmlite's version number can't be determined, then it's still appropriate to assume that it doesn't support typed constants.

jiel-nv: Thank you @gmarkall for the detailed explanation. Resolved with 2641d3341d5a0ee6a2a624e53c27a47674740dec

jiel-nv:
@gmarkall Could you please take another look at this PR?

numba_cuda/numba/cuda/debuginfo.py
```diff
@@ -11,6 +11,65 @@
...
 from numba.cuda.datamodel.models import ComplexModel, UnionModel, UniTupleModel
 from numba.cuda.types.ext_types import GridGroup
+from cuda.bindings import runtime
+
+
+# Check if CUDA Toolkit and llvmlite support polymorphic debug info
+def _get_llvmlite_version():
+    """Get llvmlite version as tuple (major, minor)."""
+    try:
+        import llvmlite
```
gmarkall: This import can be at the top level - llvmlite will always be available in a working Numba-CUDA installation.

numba_cuda/numba/cuda/debuginfo.py
```diff
@@ -11,6 +11,65 @@
...
+
+        version_str = llvmlite.__version__
+        # Parse version string like "0.46.0" or "0.46.0dev"
+        parts = version_str.split(".")
+        if len(parts) < 2:
+            return (0, 0)
+        major = int(parts[0])
+        minor = int(parts[1])
+        return (major, minor)
+    except (ImportError, AttributeError, ValueError):
```
gmarkall: With the import at the top level an `ImportError` should not be expected here. If the version splits into less than two parts (e.g. a single integer for some reason), we could also get an `IndexError`:

```suggestion
    except (IndexError, AttributeError, ValueError):
```

gmarkall:
(Set to auto merge)

PR #545, opened 2025-10-24
[CI TEST] Modified redirector for #533

gmarkall:
Testing a modified redirection method (redirecting modules individually, not a package as a module) for #533.


numba_cuda/numba/cuda/ext_types.py
gmarkall, line 1: I don't think any of the globals in here (e.g. `Dim3`, `GridGroup`), etc., have been used like public API under `numba.cuda.types` before, so this should be a safe move. If this turns out to be incorrect we could always move the ones that have been used publicly into `types/__init__.py`.

Issue #471, opened 2025-09-12
RFC: Drop host-side array APIs


leofang:
Over the years, we have been recommending users to use CuPy if they need a GPU array in Python (ex: https://github.com/numba/numba/issues/7624). Users are encouraged to use CuPy arrays to hold their data on GPUs, and pass CuPy arrays to a Numba kernel. This has been working for very long time.

As of today, CuPy offers a more performant and full-fledged array container, covering all numba-cuda's device array features and beyond. There is no reason we want to maintain the legacy constructs in numba-cuda, nor do we have the intent to grow numba-cuda into an array library. 

This RFC states our intent to deprecate and remove numba-cuda's host-side device array APIs, i.e. those called on the host side to allocate and/or initialize memory. Specifically,
- all array constructors, including `to_device()`
- `DeviceNDArray`
- all device/pinned/managed memory variants

as shown in [this page](https://nvidia.github.io/numba-cuda/reference/memory.html) and [this page](https://nvidia.github.io/numba-cuda/user/memory.html), are considered deprecated immediately and subject to future removal (which, however, is expected to be a very slow process, giving users ample time to migrate their code).

For new libraries, tutorials, demos, including those from numba-cuda's downstream libraries, we encourage you to switch to use CuPy asap. 

The device-side array APIs (i.e. those that are called inside a numba-cuda kernel, such as shared and local arrays) are not considered by this RFC.

leofang:
cc @nv-kriehl @samaid @jacobtomlinson for vis

PR #546, opened 2025-10-24
Deprecate the `DeviceNDArray` class and public APIs that return instances

brandon-b-miller:
Part of https://github.com/NVIDIA/numba-cuda/issues/471

- Adds a `DeprecatedNDArrayAPIWarning` emitted from all user facing functions for moving data around (cuda.to_device, driver.host_to_device, device_to_host, also as_cuda_array, is_cuda_array, etc
- Separates existing now deprecated APIs into internal non-warning versions and external warning versions
- Adds a deprecation warning to the `DeviceNDArray` ctor
- Adds `DeviceNDArray._create_nowarn`
- Removes as many usages of the deprecated APIs as possible from the test suite in favor of cupy arrays
- Catches warnings for tests of the currently exposed and now deprecated APIs
- Where absolutely necessary, tests calls internal non-warning versions of the deprecated APIs
- Rework tests to not use these apis as much as possible

### DIFF
--- a/numba_cuda/numba/cuda/tests/cudadrv/test_cuda_ndarray.py
+++ b/numba_cuda/numba/cuda/tests/cudadrv/test_cuda_ndarray.py
@@ -627,9 +627,10 @@ class TestCoreContiguous(DeprecatedDeviceArrayApiTest):
         view = np.zeros(shape, order="F")[::2, ::2]
         self._test_against_array_core(view)
 
-    @skip_if_cupy_unavailable
     def test_kernel_with_buffer(self):
         from cuda.core import Buffer
+
+        cp = pytest.importorskip("cupy")
 
         @cuda.jit
         def kernel(buf, n):
@@ -643,7 +644,6 @@ class TestCoreContiguous(DeprecatedDeviceArrayApiTest):
         func(buf, out)
         assert out[0] == n
 
-    @skip_if_cupy_unavailable
     def test_kernel_with_strided_memory_view(self):
         from cuda.core.utils import StridedMemoryView
 


numba_cuda/numba/cuda/cudadrv/devicearray.py
```diff
@@ -35,6 +37,18 @@ def lru_cache(func):
         return func
 
 
+def deprecated_array_api(func):
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        warnings.warn(
+            f"{func.__name__} api is deprecated. Please prefer cupy for array functions",
```
cpcloud, line 42: `cupy` arrays are much slower than `DeviceNDArray` because they require creating an external (i.e., non-numba-cuda-created) stream, so I'm not sure a recommendation for that is what we should do right now.

I was thinking that we can keep the top-level APIs (`device_array` etc.) and replace their internals with StridedMemoryView or something similar, in an effort to allow folks to as-cheaply-as-possible construct arrays.

cpcloud: Here's the current state of the art:

image

brandon-b-miller: I concur that a light weight device array like container should exist, I'm just not sure that `numba-cuda` should necessarily be the library providing it publicly. I think we should nudge users away from using `numba-cuda` as such, like for moving data from host to device. That said, I'm open to suggestions on what we _should_ recommend.

numba_cuda/numba/cuda/tests/cudapy/test_gufunc.py
```diff
@@ -119,10 +122,12 @@ def test_gufunc_stream(self):
         )
 
         stream = cuda.stream()
-        dA = cuda.to_device(A, stream)
-        dB = cuda.to_device(B, stream)
+        with warnings.catch_warnings():
+            warnings.simplefilter('ignore', DeprecatedDeviceArrayApiWarning)
+            dA = cuda.to_device(A, stream)
+            dB = cuda.to_device(B, stream)
```
cpcloud:
```suggestion
        with pytest.warns(DeprecatedDeviceArrayApiWarning):
            dA = cuda.to_device(A, stream)
        with pytest.warns(DeprecatedDeviceArrayApiWarning):
            dB = cuda.to_device(B, stream)
```

numba_cuda/numba/cuda/tests/cudapy/test_gufunc_scalar.py
```diff
@@ -119,7 +120,12 @@ def foo(a, b, out):
 
         # test error
         a = np.array(a)
-        da = cuda.to_device(a)
+
+        # As this test specifically tests the behavior of passing a DeviceNDArray,
+        # we'll catch the expected warning explicitly here. 
+        with warnings.catch_warnings():
+            warnings.simplefilter('ignore', DeprecatedDeviceArrayApiWarning)
+            da = cuda.to_device(a)
```
cpcloud:
```suggestion
        with pytest.warns(DeprecatedDeviceArrayApiWarning):
            da = cuda.to_device(a)
```

numba_cuda/numba/cuda/tests/cudapy/test_random.py
```diff
@@ -19,7 +19,7 @@
     xoroshiro128p_uniform_float64,
     xoroshiro128p_normal_float64,
 )
-
+import cupy as cp
```
cpcloud: Why is this import needed here? It doesn't appear to be used and nothing else was changed.

numba_cuda/numba/cuda/tests/cudapy/test_vectorize.py
```diff
@@ -256,56 +261,58 @@ def bar(x):
...
+            raise CudaAPIError(999, "Transfer not allowed")
 
-            # A mock of a CUDA function that always raises a CudaAPIError
+        self.old_HtoD = getattr(driver, "cuMemcpyHtoD", None)
+        self.old_DtoH = getattr(driver, "cuMemcpyDtoH", None)
 
-            def raising_transfer(*args, **kwargs):
-                raise CudaAPIError(999, "Transfer not allowed")
+        setattr(driver, "cuMemcpyHtoD", raising_transfer)
+        setattr(driver, "cuMemcpyDtoH", raising_transfer)
```
cpcloud: Why are we using `setattr` here? We know the attribute name, so there's no obvious reason to set this attribute dynamically.

```suggestion
        driver.cuMemcpyHtoD = driver.cuMemcpyDtoH = raising_transfer
```

numba_cuda/numba/cuda/tests/cudapy/test_vectorize.py
```diff
@@ -256,56 +261,58 @@ def bar(x):
...
 
-            old_HtoD = getattr(driver, "cuMemcpyHtoD", None)
-            old_DtoH = getattr(driver, "cuMemcpyDtoH", None)
+    def tearDown(self):
+        if self.old_HtoD is not None:
+            setattr(driver, "cuMemcpyHtoD", self.old_HtoD)
+        else:
+            del driver.cuMemcpyHtoD
+        if self.old_DtoH is not None:
+            setattr(driver, "cuMemcpyDtoH", self.old_DtoH)
```
cpcloud:
```suggestion
            driver.cuMemcpyDtoH = self.old_DtoH
```

numba_cuda/numba/cuda/tests/cudapy/test_vectorize.py
```diff
@@ -256,56 +261,58 @@ def bar(x):
...
+        setattr(driver, "cuMemcpyDtoH", raising_transfer)
 
-            # Use the mock for transfers between the host and device
+        super().setUp()
 
-            old_HtoD = getattr(driver, "cuMemcpyHtoD", None)
-            old_DtoH = getattr(driver, "cuMemcpyDtoH", None)
+    def tearDown(self):
+        if self.old_HtoD is not None:
+            setattr(driver, "cuMemcpyHtoD", self.old_HtoD)
```
cpcloud:
```suggestion
            driver.cuMemcpyHtoD = self.old_HtoD
```

numba_cuda/numba/cuda/tests/cudadrv/test_pinned.py
```diff
@@ -5,10 +5,11 @@
 import platform
 
 from numba import cuda
-from numba.cuda.testing import unittest, CUDATestCase
+from numba.cuda.testing import unittest, DeprecatedDeviceArrayApiTest
 
 
-class TestPinned(CUDATestCase):
+# TODO
```
cpcloud, line 11: non-blocking: is there a specific todo here?

PR #547, opened 2025-10-24
bench: add cupy to array constructor kernel launch benchmarks

cpcloud:
Add cupy to kernel launch benchmarks

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -161,13 +161,14 @@ cuda = "12"
 [tool.pixi.feature.cu13.system-requirements]
 cuda = "13"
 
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -161,13 +161,14 @@ cuda = "12"
 [tool.pixi.feature.cu13.system-requirements]
 cuda = "13"
 
-[tool.pixi.build-dependencies]
+[tool.pixi.dependencies]
```
cpcloud, line 164: Unrelated change, technically, but `build-dependencies` is deprecated and throwing warnings, so I thought I would just clean that up here.

PR #551, opened 2025-10-27
chore(deps): move to self-contained pixi.toml to avoid mixed-pypi-pixi environments

cpcloud:
This PR moves our pixi environment to be self-contained in order to avoid mixing cuda dependencies between pypi and conda(-forge).


pyproject.toml
```diff
@@ -3,41 +3,34 @@
...
-    { name = "Anaconda Inc." },
-    { name = "NVIDIA Corporation" }
-]
+authors = [{ name = "Anaconda Inc." }, { name = "NVIDIA Corporation" }]
 license = "BSD-2-clause"
 license-files = ["LICENSE", "LICENSE.numba"]
-requires-python = ">=3.9"
-dependencies = ["numba>=0.60.0", "cuda-bindings>=12.9.1,<14.0.0", "cuda-core>=0.3.2,<0.4.0dev0"]
+requires-python = ">=3.9,<3.14"
+dependencies = [
```
cpcloud: Fixing this in another PR.

gmarkall: What needs fixing?

cpcloud: We have duplicated packages in `dependencies` and `optional-dependencies` (`cuda-core` and `cuda-bindings` are in both).

Fix is in https://github.com/NVIDIA/numba-cuda/pull/552

cpcloud: Closed #552 for now, and reverted these changes.

pixi.toml
```diff
@@ -0,0 +1,106 @@
...
+channels = ["conda-forge"]
+platforms = ["linux-64", "linux-aarch64", "win-64"]
+
+[pypi-dependencies]
+numba-cuda = { path = ".", editable = true }
+
+[dependencies]
+cuda-compiler = "*"
+make = "*"
+cxx-compiler = "*"
```
kkraus14, line 13: nitpick: the `cxx-compiler` / `c-compiler` metapackages touch a bunch of compiler / linker flags related to things like RPATHs and whatnot that can really mess with a local development flow. I personally make it a point to avoid them in my local environments for this reason.

I.E. when building test executables it sets RPATH to point into the conda environment before `$ORIGIN` which makes it find an installed conda package before the built DSO in the working dir.

Not as much of a concern for numba-cuda but some food for thought.

pixi.toml
```diff
@@ -0,0 +1,106 @@
...
+
+[feature.cu12.dependencies]
+cuda-python = "12.9.*"
+cuda-nvcc = "12.*"
+cuda-nvvm = "12.*"
+cuda-runtime = "12.*"
+cuda-nvrtc = "*"
+libnvjitlink = "12.*"
+cuda-cccl = "12.*"
+libcurand = "*"
```
kkraus14, line 30: Instead of specifying `12.*` for all of these dependencies we should be able to use `cuda-version=12` which all of these packages constrain themselves to.

pixi.toml
```diff
@@ -0,0 +1,106 @@
...
+cuda = "13"
+
+[feature.cu13.dependencies]
+cuda-python = "13.*"
+cuda-nvvm = "13.*"
+cuda-runtime = "13.*"
+cuda-nvrtc = "*"
+libnvjitlink = "13.*"
+cuda-cccl = "13.*"
+libcurand = "10.4.*"
```
kkraus14, line 42: +1 to `cuda-version`

pixi.toml
```diff
@@ -0,0 +1,106 @@
...
+libtorch = ">=2.8"
+cupy = "*"
+
+[feature.test.dependencies]
+pre-commit = "*"
+psutil = "*"
+cffi = "*"
+pytest = "*"
+pytest-xdist = "*"
+pytest-benchmark = "*"
```
kkraus14, line 56: Do we not want to version these and then use `pixi update` or whatever command to upgrade them?

pixi.toml
```diff
@@ -0,0 +1,106 @@
...
+
+[feature.test.pypi-dependencies]
+ml_dtypes = "*"
+filecheck = "*"
+
+[environments]
+cu12 = { features = ["cu12", "test", "bench"], solve-group = "cu12" }
+cu13 = { features = ["cu13", "test"], solve-group = "cu13" }
```
kkraus14, line 64: I guess there isn't cuda 13 builds of pytorch in conda-forge yet?

pixi.toml
cpcloud: Yeah, the `RPATH` behavior isn't my favorite, but I understand why it works that way.

Is the main concern development workflows for pseudo-monorepos like `cuda-python` where dependencies are native *and* they live in the same repo?

In the case of numba-cuda I think it's worth it to bring as much in to the env as possible to ensure that devs have a good day 1 (and day N 😄!) experience.

pixi.toml
cpcloud: I will try to make this change in the follow-up I have to replace conda ci with pixi (#554), to avoid lots of conflicts in that PR.

pixi.toml
cpcloud: We can, I was just replicating what was there before. I'll make that change in a follow-up.

pixi.toml
cpcloud: No, unfortunately there are not.

pixi.toml
cpcloud: Actually, it turns out I already did that in that PR 👌🏻

pixi.toml
kkraus14: I've mostly run into issues with C++ projects where you're building a DSO, say `mylib.so` and then building various test executables, but like to have a stable version of `mylib` installed in the environment as well. Given how the state of tooling has improved so vastly, maybe this is less of a concern and these should just be separate environments nowadays.

I'm more than happy to move forward as is and we can always course correct if we find pain / issues.

cpcloud:
Not entirely sure why that job is hanging. This PR doesn't touch anything CI related.

PR #552, opened 2025-10-27
deps: remove now-required dependencies from optional dependencies `pyproject.toml` table

cpcloud:


pyproject.toml
```diff
@@ -21,12 +21,10 @@ authors = [
 license = "BSD-2-clause"
 license-files = ["LICENSE", "LICENSE.numba"]
 requires-python = ">=3.9"
-dependencies = ["numba>=0.60.0", "cuda-bindings>=12.9.1,<14.0.0", "cuda-core>=0.3.2,<0.4.0dev0"]
+dependencies = ["numba>=0.60.0", "cuda-bindings>=12.9.1,<14.0.0", "cuda-core>=0.3,<0.4.0dev0"]
```
gmarkall, line 24: I think there was a reason that this can't be 0.3.0 or 0.3.1 - will have to see if I can find it.

cpcloud: Hm, okay. I guess the extra optional dependencies aren't causing any harm, and they may be necessary. It would be good to know, but I'm going to close this PR and revert the corresponding changes in #551

cpcloud:
Closing as not needed for now.

PR #554, opened 2025-10-27
ci: replace conda testing with pixi

cpcloud:
This PR moves our conda CI to use pixi in order to speed it up. Based off of #551.

### DIFF
--- a/ci/test_conda.sh
+++ b/ci/test_conda.sh
@@ -4,16 +4,6 @@
 
 set -euo pipefail
 
-. /opt/conda/etc/profile.d/conda.sh
-
-CTK_PACKAGE_DEPENDENCIES=(
-    "cuda-cccl"
-    "cuda-nvcc-impl"
-    "cuda-nvrtc"
-    "libcurand-dev"
-    "cuda-cuobjdump"
-)
-
 DISTRO=`cat /etc/os-release | grep "^ID=" | awk 'BEGIN {FS="="} { print $2 }'`
 
 if [ "$DISTRO" = "ubuntu" ]; then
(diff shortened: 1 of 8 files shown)


pixi.toml
```diff
@@ -3,43 +3,58 @@
 [workspace]
 channels = ["conda-forge"]
 platforms = ["linux-64", "linux-aarch64", "win-64"]
+preview = ["pixi-build"]
 
-[pypi-dependencies]
-numba-cuda = { path = ".", editable = true }
+[workspace.build-variants]
+python = ["3.10.*", "3.11.*", "3.12.*", "3.13.*"]
```
cpcloud, line 9: These can't quite be used yet because of https://github.com/prefix-dev/pixi/issues/4877.

pixi.toml
```diff
@@ -107,8 +162,33 @@ depends-on = [{ task = "build-tests" }]
...
 cmd = ["pytest", "$PIXI_PROJECT_ROOT/testing"]
-depends-on = [{ task = "build-tests" }]
 
 [target.linux.tasks.simtest.env]
 NUMBA_CUDA_TEST_BIN_DIR = "$PIXI_PROJECT_ROOT/testing"
 NUMBA_ENABLE_CUDASIM = "1"
+
+[package]
+name = "numba-cuda"
+version = "0.20.0"
```
cpcloud: I'd like to automatically compute this, but I'm not sure if that's easily possible with pixi.

cpcloud:
Looking at a random PR build without this change: https://github.com/NVIDIA/numba-cuda/actions/runs/19040099227/job/54375099989

I see test-conda job durations of >= 6 minutes, whereas after this PR jobs are 2-3 minutes (which includes the pixi setup time), so there's a 2-3x speed up here by moving to pixi, likely from avoiding the solve on every job.

pixi.toml
```diff
@@ -3,43 +3,58 @@
...
+python = ["3.10.*", "3.11.*", "3.12.*", "3.13.*"]
 
 [dependencies]
-cuda-compiler = "*"
-make = "*"
-cxx-compiler = "*"
-numba = ">=0.60.0"
-cuda-bindings = ">=12.9.1,<14"
-cuda-core = ">=0.3,<0.4.0dev0"
-python = ">=3.9,<3.14"
```
cpcloud, line 17: None of these are necessary anymore because they are covered by the `package` section below, which includes `numba-cuda` in the environment (because of `numba-cuda = ...` in the `dependencies` section.

ci/test_conda.sh
```diff
@@ -22,67 +12,30 @@ if [ "$DISTRO" = "ubuntu" ]; then
...
-    "pytest-benchmark"
-    "cffi"
-    "ml_dtypes"
-    "python=${RAPIDS_PY_VERSION}"
-)
 # Constrain oldest supported dependencies for testing
 if [ "${RAPIDS_DEPENDENCIES:-}" = "oldest" ]; then
-    DEPENDENCIES+=("numba==0.60.0")
+    # add to the default environment's dependencies
+    pixi add "numba=0.60.0"
```
kkraus14, line 18: Is there a way we can have this version somewhere in the `pixi.toml` or something else not buried in a `ci` script to make it more easily testable / reproducible outside of CI?

numba_cuda/numba/cuda/tests/cudadrv/test_deallocations.py
```diff
@@ -17,8 +17,8 @@
 
 
 @skip_on_cudasim("not supported on CUDASIM")
-@skip_if_external_memmgr("Deallocation specific to Numba memory management")
```
kkraus14, line 20: Is it just not supported to skip on classes as opposed to individual tests?

pixi.toml
```diff
@@ -3,43 +3,58 @@
...
 
-[feature.cu13.system-requirements]
+[feature.cu-12.system-requirements]
+cuda = "12"
+
+[feature.cu-12.dependencies]
+cuda-version = "12.*"
+
+[feature.cu-12-0.dependencies]
+cuda-version = "12.0.*"
```
kkraus14, line 31: I don't know if using `cuda-version` can pin to specific minor versions. I.E. `libnvjitlink` allows floating forward intentionally for better environment compatibility: https://github.com/conda-forge/libnvjitlink-feedstock/blob/main/recipe/meta.yaml#L54

May need to pin to specific minor versions of the various libraries.

pixi.toml
kkraus14: That's a bit of a shame. We just released `0.20.1` so in the least this needs to be bumped.

numba_cuda/numba/cuda/tests/cudadrv/test_deallocations.py
cpcloud: No, these tests are stateful across the test classes when run in parallel with pytest-xdist and cause accounting to be incorrect in that setting, so I moved them to a single class which when run with `--dist loadscope` all the tests in a class run in the same process.

ci/test_conda.sh
cpcloud: I suppose I could make another environment with this as a dependency.

kkraus14: I think that makes the most sense. Something like `numba_oldest`.

cpcloud: Actually, since it's in the matrix I would have to make one for every cuda + python verison marked `'oldest'` in `matrix.yml`.

cpcloud: That's possible for sure. It might be a pain to maintain. Can we do that in a follow-up?

pixi.toml
cpcloud: Fixed

pixi.toml
cpcloud: This is no different than the status quo, so can we do it in a follow-up?

The previous conda setup did the same kind of variable substitution of `CUDA_VER` to work around the lack of specific versions on conda-forge that are listed in `matrix.yml`, e.g., 12.0.1.

cpcloud:
I'll remove the dependency of the `test-conda` job on `build-conda` in a follow-up, since after this PR we're effectively building a conda package with each `pixi run` that requires it.

cpcloud:
Just realized that this is all testing only python 3.13.

Fix incoming.

PR #555, opened 2025-10-27
Remove dependencies on target_extension for CUDA target

VijayKandiah:
This PR removes dependency on `numba.core.target_extension`. This import was primarily used to get the local target which is `CUDA` in our case unless the CPU dispatcher is used. There are a couple numba.core APIs that I had to monkey patch in `descriptor.py` to handle our default being CUDA target: `numba.core.utils.order_by_target_specificity`, and `numba.core.target_extension.get_local_target`. @gmarkall please take a look and let me know if there's a way around doing this sort of monkey patching.

### DIFF
--- a/numba_cuda/numba/cuda/compiler.py
+++ b/numba_cuda/numba/cuda/compiler.py
@@ -741,16 +741,17 @@ def compile_cuda(
     flags.max_registers = max_registers
     flags.lto = lto
 
-    cres = compile_extra(
-        typingctx=typingctx,
-        targetctx=targetctx,
-        func=pyfunc,
-        args=args,
-        return_type=return_type,
-        flags=flags,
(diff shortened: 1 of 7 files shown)


gmarkall:
>  @gmarkall please take a look and let me know if there's a way around doing this sort of monkey patching.

Having a look at this now.

I think there's a mixup between the `CUDATarget` class and the `cuda_target` instance somewhere, which is confusing Numba's `order_by_target_specificity` and maybe driving the need for a monkey patch, but I'm still a little fuzzy about what's going on here. Will update once I've worked out a bit more.

gmarkall:
(Other than that, everything looks good with the PR)

gmarkall (approved):

PR #558, opened 2025-10-28
Move frontend tests to `cudapy` namespace

brandon-b-miller:
Follow up for https://github.com/NVIDIA/numba-cuda/pull/523#discussion_r2466469275

This moves some test files to the cudapy namespace and merges `tests/test_extending` and `cudapy/test_extending.py`.


leofang:
Once upon a time, cuda.core was called cuda.py... 😆 Thank god we renamed it. So confusing!

gmarkall (dismissed):

gmarkall:
Just needs the moved test fixing up - otherwise, looks good.

brandon-b-miller:
@gmarkall should be fixed up now :)

PR #559, opened 2025-10-28
Relax the pinning to `cuda-core` to allow it floating across minor releases

isVoid:
Requested by nvshmem4py release.

### DIFF
--- a/ci/test_wheel.sh
+++ b/ci/test_wheel.sh
@@ -13,7 +13,7 @@ echo "Package path: ${package}"
 DEPENDENCIES=(
     "${package}"
     "cuda-python==${CUDA_VER_MAJOR_MINOR%.*}.*"
-    "cuda-core==0.4.*"
+    "cuda-core>=0.3.0,<1.0.0"
     "--group"
     "test"
 )
--- a/conda/recipes/numba-cuda/meta.yaml
+++ b/conda/recipes/numba-cuda/meta.yaml
@@ -30,7 +30,7 @@ requirements:
     - python
     - numba >=0.59.1
     - cuda-bindings >=12.9.1
-    - cuda-core ==0.4.*
+    - cuda-core >=0.3.0,<1.0.0
 
 about:
   home: {{ project_urls["Homepage"] }}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,12 +21,12 @@ authors = [
 license = "BSD-2-clause"
 license-files = ["LICENSE", "LICENSE.numba"]
 requires-python = ">=3.9"
-dependencies = ["numba>=0.60.0", "cuda-bindings>=12.9.1,<14.0.0", "cuda-core>=0.4.0,<0.5.0dev0"]
+dependencies = ["numba>=0.60.0", "cuda-bindings>=12.9.1,<14.0.0", "cuda-core>=0.3.2,<1.0.0"]
 
 [project.optional-dependencies]
 cu12 = [
     "cuda-bindings>=12.9.1,<13.0.0",
-    "cuda-core==0.4.*",
+    "cuda-core>=0.3.0,<1.0.0",
     "cuda-python==12.9.*",  # supports all CTK 12.x
     "nvidia-cuda-nvcc-cu12",  # for libNVVM
     "nvidia-cuda-runtime-cu12",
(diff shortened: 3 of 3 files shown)


leofang (changes requested):
If nvshmem4py needs 0.4.0+, it needs to pin it explicitly. Keeping the pinning too tight is a major problem that our product managers are not happy about, and this impacts everyone depending on ~~cuda-core~~ numba-cuda. I am drafting a version pinning policy that we need to adopt before GA.

isVoid:
Would it make it easier to relax to >=0.3.0, <0.5.0?

leofang:
I doubt that any breakage we introduce to cuda.core would impact much to any numba-cuda users. Perhaps `cuda-core~=0.3` (equivalent to `>=0.3.0,<1.0.0`) is better. 

IIRC @brandon-b-miller and/or @mmason-nvidia encountered situations where using a newer cuda-core is preferred (but not required). Please remind me the conclusion we reached elsewhere, if any?

isVoid:
I'll wait for more feedbacks.

> Perhaps cuda-core~=0.3 (equivalent to >=0.3.0,<1.0.0) is better.

Nit: this actually means `>=0.3.0,<0.4.0`.

benhg:
Hi @leofang , just chiming in to say NVSHMEM4Py does in fact have cuda-core pinned to >= 0.4.0. The only thing I'm trying to accomplish is not to have users see warnings when they `pip install nvshmem4py numba-cuda` when we have done QA against those versions.

Something like >=0.3,<1.0 would be fine from our perspective

leofang:
Sounds good! 
> > Perhaps cuda-core~=0.3 (equivalent to >=0.3.0,<1.0.0) is better.
> 
> Nit: this actually means `>=0.3.0,<0.4.0`.

Yes, Michael you're right. I was juggling across multiple things in the late evening. I meant the latter (`>=0.3.0,<1.0.0`).

> not to have users see warnings when they `pip install nvshmem4py numba-cuda`

What warnings?

isVoid:
> What warnings?

Nvshmem4py pins to 0.4.0 because it requires the VMM feature. So there's a version mismatch between Numba-CUDA and nvshmem4py.

leofang:
Why are Windows tests canceled? 🤔

leofang (approved):
LGTM but we should fix the CI issue if possible.

brandon-b-miller:
> I doubt that any breakage we introduce to cuda.core would impact much to any numba-cuda users. Perhaps `cuda-core~=0.3` (equivalent to `>=0.3.0,<1.0.0`) is better.
> 
> IIRC @brandon-b-miller and/or @mmason-nvidia encountered situations where using a newer cuda-core is preferred (but not required). Please remind me the conclusion we reached elsewhere, if any?

Perhaps you're referring to [this conversation](https://github.com/NVIDIA/numba-cuda/pull/474/files#r2355430235)? As written that PR seems like it would need 0.4.0 to be safe, so there's some benefit for us to have it.

isVoid:
> As written that PR seems like it would need 0.4.0 to be safe, so there's some benefit for us to have it.

@brandon-b-miller I looked at that PR and it seems like you meant it's good to pin the lower bound to 0.4.0 because of the `.code_type` public attribute and linker flag improvement. However, per @leofang's note that bumping the lower bound require wider sync across downstream users. Is it ok if we proceed with the more permissive change in this PR and sync up later about bumping the lower bound?

isVoid:
> Why are Windows tests canceled? 🤔

Looks like an one-off CI hiccup. :)

leofang (approved):

leofang:
Many thanks!

ci/test_wheel.sh
```diff
@@ -13,7 +13,7 @@ echo "Package path: ${package}"
 DEPENDENCIES=(
     "${package}"
     "cuda-python==${CUDA_VER_MAJOR_MINOR%.*}.*"
-    "cuda-core==0.3.*"
+    "cuda-core>=0.3.0,<1.0.0"
```
kkraus14, line 16: I didn't get a chance to review this PR, but the reason we pinned to `0.3.*` previously is because we historically had API breaking changes in minor releases of `cuda-core` thus far. I fear if someone goes to pull an older version of `numba-cuda` with this pinning that they'll eventually get `cuda-core 0.X` where X had breaking changes relative to 0.3 / 0.4 that breaks numba-cuda.

leofang: see https://github.com/NVIDIA/numba-cuda/pull/559#issuecomment-3459311972

PR #561, opened 2025-10-29
[Refactor][NFC][Cleanups] Update imports to upstream numba to use the numba.cuda modules

atmnp:
This PR updates the vast majority of imports to upstream numba modules, that had already been vendored in for future CUDA-specific changes. After this PR, there should be exactly 3 (not strictly necessary) upstream numba imports remaining - one in testing vectorization for in test_math, one in random.py, and the global compiler_lock.

It seems from the existing comments that removing the CPU jit in random.py would result in considerable performance drops, so I've just added a guard for numba being available.

There are modules such as numba.core.misc.{mergesort,quicksort} that had not yet been vendored in, so this PR also vendors those modules in. They had to be modified to not use the CPU jit, since these are not necessarily supported within numba-cuda. It's still unclear to me if this is necessary or is done correctly, since we don't have tests for these yet. My impression is that these are to support jitting of mergesort/quicksort on arrays, which I am fairly confident we cannot support on the CUDA target until more development is done on the Numba-CUDA runtime. I am unsure if we can remove these uses but I would prefer if we did.

There are many tests that rely on CPU jitting, these tests are now guarded to not run in environments where the numba package is not available. This is done using a global importlib check in numba-cuda's `__init__.py`. This cannot be moved into numba.cuda.config since the upstream numba doesn't have this flag, and we intend on forwarding calls to upstream numba when it is available. All previously sporadic guards for this *should* have been updated to use this global flag now.

Support for prange, pndindex, and stencils has been removed, likewise for typed.Dict, and typed.List. There was a pass that relied on this, which was the MakeFunctionIntoJitFunction pass, which implicitly converts functions to jitted functions, which is also a use case that is not currently tested. Also, this removes all mention of the legacy type system flag since we implicitly assume the legacy type system in place.

### DIFF
--- a/numba_cuda/numba/cuda/tests/cudadrv/test_init.py
+++ b/numba_cuda/numba/cuda/tests/cudadrv/test_init.py
@@ -1,6 +1,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: BSD-2-Clause
 
+import concurrent.futures
 import multiprocessing as mp
 import os
 
@@ -20,11 +21,8 @@ def cuInit_raising(arg):
 # not assigned until we attempt to initialize - mock.patch.object cannot locate
 # the non-existent original method, and so fails. Instead we patch
 # driver.cuInit with our raising version prior to any attempt to initialize.
-def cuInit_raising_test(result_queue):
+def cuInit_raising_test():
     driver.cuInit = cuInit_raising
-
-    success = False
-    msg = None
 
     try:
         # A CUDA operation that forces initialization of the device
@@ -32,76 +30,74 @@ def cuInit_raising_test(result_queue):
     except CudaSupportError as e:
         success = True
         msg = e.msg
+    else:
+        success = False
+        msg = None
 
-    result_queue.put((success, msg))
+    return success, msg
 
 
 # Similar to cuInit_raising_test above, but for testing that the string
 # returned by cuda_error() is as expected.
-def initialization_error_test(result_queue):
+def initialization_error_test():
     driver.cuInit = cuInit_raising
-
-    success = False
-    msg = None
 
     try:
         # A CUDA operation that forces initialization of the device
         cuda.device_array(1)
     except CudaSupportError:
         success = True
+    else:
+        success = False
 
-    msg = cuda.cuda_error()
-    result_queue.put((success, msg))
+    return success, cuda.cuda_error()
 
 
 # For testing the path where Driver.__init__() catches a CudaSupportError
-def cuda_disabled_test(result_queue):
-    success = False
-    msg = None
-
+def cuda_disabled_test():
     try:
         # A CUDA operation that forces initialization of the device
         cuda.device_array(1)
     except CudaSupportError as e:
         success = True
         msg = e.msg
+    else:
+        success = False
+        msg = None
 
-    result_queue.put((success, msg))
+    return success, msg
 
 
 # Similar to cuda_disabled_test, but checks cuda.cuda_error() instead of the
 # exception raised on initialization
-def cuda_disabled_error_test(result_queue):
-    success = False
-    msg = None
-
+def cuda_disabled_error_test():
     try:
         # A CUDA operation that forces initialization of the device
         cuda.device_array(1)
     except CudaSupportError:
         success = True
+    else:
+        success = False
 
-    msg = cuda.cuda_error()
-    result_queue.put((success, msg))
+    return success, cuda.cuda_error()
 
 
 @skip_on_cudasim("CUDA Simulator does not initialize driver")
 class TestInit(CUDATestCase):
     def _test_init_failure(self, target, expected):
         # Run the initialization failure test in a separate subprocess
-        ctx = mp.get_context("spawn")
-        result_queue = ctx.Queue()
-        proc = ctx.Process(target=target, args=(result_queue,))
-        proc.start()
-        proc.join(30)  # should complete within 30s
-        success, msg = result_queue.get()
+        with concurrent.futures.ProcessPoolExecutor(
+            mp_context=mp.get_context("spawn")
+        ) as exe:
+            # should complete within 30s
+            success, msg = exe.submit(target).result(timeout=30)
 
         # Ensure the child process raised an exception during initialization
         # before checking the message
         if not success:
-            self.fail("CudaSupportError not raised")
+            assert "CudaSupportError not raised" in msg
 
-        self.assertIn(expected, msg)
+        assert expected in msg
 
     def test_init_failure_raising(self):
         expected = "Error at driver init: CUDA_ERROR_UNKNOWN (999)"
--- a/numba_cuda/numba/cuda/tests/cudadrv/test_runtime.py
+++ b/numba_cuda/numba/cuda/tests/cudadrv/test_runtime.py
@@ -1,21 +1,18 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: BSD-2-Clause
 
+import concurrent.futures
 import multiprocessing
 import os
 from numba.cuda.testing import unittest
 
 
-def set_visible_devices_and_check(q):
-    try:
-        from numba import cuda
-        import os
+def set_visible_devices_and_check():
+    from numba import cuda
+    import os
 
-        os.environ["CUDA_VISIBLE_DEVICES"] = "0"
-        q.put(len(cuda.gpus.lst))
-    except:  # noqa: E722
-        # Sentinel value for error executing test code
-        q.put(-1)
+    os.environ["CUDA_VISIBLE_DEVICES"] = "0"
+    return len(cuda.gpus.lst)
 
 
 class TestVisibleDevices(unittest.TestCase):
@@ -38,22 +35,13 @@ class TestVisibleDevices(unittest.TestCase):
             msg = "Cannot test when CUDA_VISIBLE_DEVICES already set"
             self.skipTest(msg)
 
-        ctx = multiprocessing.get_context("spawn")
-        q = ctx.Queue()
-        p = ctx.Process(target=set_visible_devices_and_check, args=(q,))
-        p.start()
-        try:
-            visible_gpu_count = q.get()
-        finally:
-            p.join()
+        with concurrent.futures.ProcessPoolExecutor(
+            mp_context=multiprocessing.get_context("spawn")
+        ) as exe:
+            future = exe.submit(set_visible_devices_and_check)
 
-        # Make an obvious distinction between an error running the test code
-        # and an incorrect number of GPUs in the list
-        msg = "Error running set_visible_devices_and_check"
-        self.assertNotEqual(visible_gpu_count, -1, msg=msg)
-
-        # The actual check that we see only one GPU
-        self.assertEqual(visible_gpu_count, 1)
+        visible_gpu_count = future.result()
+        assert visible_gpu_count == 1
 
 
 if __name__ == "__main__":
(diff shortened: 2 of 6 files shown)


numba_cuda/_version.py
```diff
@@ -2,10 +2,54 @@
...
+    minor = try_int(parts[1]) if len(parts) >= 2 else None
+    patch = try_int(parts[2]) if len(parts) >= 3 else None
+    short = (major, minor)
+    full = (major, minor, patch)
+    string = version
+    tup = tuple(parts)
+    git_revision = tup[3] if len(tup) >= 4 else None
+    return version_info(
+        major, minor, patch, short, full, string, tup, git_revision
+    )
```
gmarkall: I don't think we should have this in our codebase - it seems to pertain to the Numba version. We maintain the Numba-CUDA version in `__version__`

numba_cuda/numba/cuda/__init__.py
```diff
@@ -7,12 +7,20 @@
 import warnings
 import sys
 
+from numba_cuda._version import __version__
+
+from numba_cuda._version import generate_version_info
+
+version_info = generate_version_info(__version__)
+del generate_version_info
```
gmarkall: I don't think we want to create a `version_info` member here either.

gmarkall:
Whilst looking at the impact of moving `MakeFunctionToJitFunction`, I noticed that the following:

```python
from numba import cuda

@cuda.jit
def kernel(x):
    f = lambda x: x + 2

    def g(x):
        return map(f, x)
    res = g(x)

    print(res[0])
    print(res[1])
    print(res[2])


kernel[1, 1]((1, 2, 3))

cuda.synchronize()
```

results in 

```
...
  File "/home/gmarkall/numbadev/numba/numba/core/types/functions.py", line 311, in get_call_type
    sig = temp.apply(nolitargs, nolitkws)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba/numba/core/typing/templates.py", line 358, in apply
    sig = generic(args, kws)
          ^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba/numba/core/typing/templates.py", line 621, in generic
    disp, new_args = self._get_impl(args, kws)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba/numba/core/typing/templates.py", line 720, in _get_impl
    impl, args = self._build_impl(cache_key, args, kws)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba/numba/core/typing/templates.py", line 781, in _build_impl
    jitter = self._get_jit_decorator()
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba/numba/core/typing/templates.py", line 749, in _get_jit_decorator
    jitter = jit_registry[target_hw]
             ~~~~~~~~~~~~^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba/numba/core/registry.py", line 71, in __getitem__
    return super(DelayedRegistry, self).__getitem__(item)
KeyError: <class 'numba.core.target_extension.CUDA'>
```

so there's a bug impacting some of the code I believe this PR touches that needs fixing. I'm looking into a fix for this.

numba_cuda/_version.py
atmnp: fixed!

numba_cuda/numba/cuda/__init__.py
```diff
@@ -7,12 +7,15 @@
 import warnings
 import sys
 
+from numba_cuda._version import __version__
+
```
gmarkall: I think this is a leftover from earlier changes?
```suggestion
```

numba_cuda/numba/cuda/__init__.py
```diff
@@ -7,12 +7,15 @@
...
 # Re-export types itself
 import numba.cuda.types as types
 
 # Re-export all type names
 from numba.cuda.types import *
 
+HAS_NUMBA = importlib.util.find_spec("numba") is not None
```
gmarkall, line 17: I think we might need to change the implementation of this in future, so having it centralized here is convenient.

numba_cuda/numba/cuda/__init__.py
```diff
@@ -66,3 +69,6 @@
     )
 
 from numba.cuda.np.ufunc import vectorize, guvectorize
+from numba.cuda.misc import quicksort, mergesort
```
gmarkall: I think your observation in the PR text that we might not need these for now is correct. I agree with your estimation that it would be best to remove them. We can always add something like this in future if necessary.

However, I expect it would be better to encourage users to use CCCL for sorting arrays in kernels so that they can benefit from thread/block/warp-level parallelism for sorting rather than single-threaded sorts on separate arrays by individual threads.

numba_cuda/numba/cuda/__init__.py
```diff
@@ -66,3 +69,6 @@
...
+from numba.cuda.misc.special import literal_unroll
```
gmarkall: This should already have been brought in by `device_init` or `simulator_init`. If there's no version of it for `simulator_init`, then a "dummy" version needs adding in the simulator. I expect that would look like:

```python
def literal_unroll(x):
    return x
```

i.e. just a "no-op", because the simulator isn't compiling anything so there is no hard requirement to unroll tuples for a consistent typing.

gmarkall: For clarity, I am suggesting:
```suggestion
```

numba_cuda/numba/cuda/__init__.py
gmarkall:
```suggestion
```

numba_cuda/numba/cuda/__init__.py
```diff
@@ -66,3 +69,6 @@
...
+from numba.cuda.np.numpy_support import *
```
gmarkall: This is only used for `from_dtype` in `random.py`, and I think` `from_dtype` should not have been in the top-level `numba.cuda` namespace. Therefore, we should remove this import:

```suggestion
```

and fix `random.py`:

```diff
diff --git a/numba_cuda/numba/cuda/random.py b/numba_cuda/numba/cuda/random.py
index 8cf7cc07..d22d9d32 100644
--- a/numba_cuda/numba/cuda/random.py
+++ b/numba_cuda/numba/cuda/random.py
@@ -10,11 +10,12 @@ from numba.cuda import (
     uint32,
     int64,
     uint64,
-    from_dtype,
     HAS_NUMBA,
 )
 from numba.cuda import config
 
+from numba.cuda.np.numpy_support import from_dtype
+
 if HAS_NUMBA:
     from numba import jit
 
```

numba_cuda/numba/cuda/core/errors.py
```diff
@@ -454,7 +456,7 @@ def termcolor():
 This should not have happened, a problem has occurred in Numba's internals.
 You are currently using Numba version %s.
 %s
-""" % (numba.__version__, feedback_details)
+""" % (numba_cuda.__version__, feedback_details)
```
gmarkall, line 459: I think this whole message needs a broader modification to indicate that it comes specifically from Numba-CUDA. I will provide some suggested text.

numba_cuda/numba/cuda/core/ir.py
```diff
@@ -25,6 +28,7 @@
     ConstantInferenceError,
 )
 from numba.cuda.core import consts
+from numba.cuda import HAS_NUMBA
```
gmarkall: Already present above:
```suggestion
```

numba_cuda/numba/cuda/core/ir.py
```diff
@@ -776,7 +780,10 @@ def __repr__(self):
 class Del(Stmt):
     def __init__(self, value, loc):
         assert isinstance(value, str)
-        assert isinstance(loc, (Loc, numba.core.ir.Loc))
+        if HAS_NUMBA:
+            assert isinstance(loc, (Loc, numba.core.ir.Loc))
+        else:
+            assert isinstance(loc)
```
gmarkall: `isinstance` takes two arguments.

numba_cuda/numba/cuda/core/untyped_passes.py
```diff
@@ -618,72 +617,6 @@ def run_pass(self, state):
...
-                                    [
-                                        isinstance(getdef(x), ir.Const)
-                                        for x in kw_default.items
-                                    ]
-                                )
-                            if not ok:
-                                continue
-
-                            pyfunc = convert_code_obj_to_function(node, func_ir)
-                            func = njit()(pyfunc)
```
gmarkall, line 672: When this is brought back, this will need to be `cuda.jit`.

numba_cuda/numba/cuda/random.py
```diff
@@ -3,18 +3,21 @@
...
-from numba import (
-    cuda,
+from numba import cuda
+from numba.cuda import (
     float32,
     float64,
     uint32,
     int64,
     uint64,
     from_dtype,
```
gmarkall: This should come from `numba.cuda.np.numpy_support` so that it doesn't need to be brought into the top level `numba.cuda` package.

numba_cuda/numba/cuda/random.py
```diff
@@ -3,18 +3,21 @@
...
-    jit,
+    HAS_NUMBA,
 )
 from numba.cuda import config
 
+if HAS_NUMBA:
+    from numba import jit
```
gmarkall, line 19: I think in the long term we probably need to replace the jitted implementation with a C extension, find a way to do the initialization in CUDA, or to deprecate this module.

numba_cuda/numba/cuda/tests/core/serialize_usecases.py
```diff
@@ -11,6 +11,7 @@
 import numpy as np
 import numpy.random as nprand
 
+# This does not need a guard, it's already guarded at the import site
```
gmarkall, line 14: Good comment :+1:

numba_cuda/numba/cuda/tests/core/test_serialize.py
```diff
@@ -60,13 +59,15 @@ def check_result(func):
             new_func = pickle.loads(pickled)
             check_result(new_func)
 
+    @skip_on_standalone_numba_cuda
```
gmarkall: If you have the skip decorator on a class you don't need it on the individual test cases as well (applies to all tests in this class).

numba_cuda/numba/cuda/typing/templates.py
```diff
@@ -409,7 +407,7 @@ def unpack_opt(x):
 
     def get_template_info(self):
         impl = getattr(self, "generic")
-        basepath = os.path.dirname(os.path.dirname(numba.__file__))
+        basepath = os.path.dirname(os.path.dirname(numba.cuda.__file__))
```
gmarkall: As this is one package deeper than the original, it needs one more call to `os.path.dirname` than the original.

gmarkall: This applies to the following five variations of this as well.

gmarkall:
> Whilst looking at the impact of moving `MakeFunctionToJitFunction`, I noticed that the following:
> 
> ```python
> from numba import cuda
>  # ...
> kernel[1, 1]((1, 2, 3))
> 
> cuda.synchronize()
> ```
> 
> results in
> 
> ```
> ...
>   File "/home/gmarkall/numbadev/numba/numba/core/typing/templates.py", line 749, in _get_jit_decorator
>     jitter = jit_registry[target_hw]
>              ~~~~~~~~~~~~^^^^^^^^^^^
>   File "/home/gmarkall/numbadev/numba/numba/core/registry.py", line 71, in __getitem__
>     return super(DelayedRegistry, self).__getitem__(item)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> KeyError: <class 'numba.core.target_extension.CUDA'>
> ```
> 
> so there's a bug impacting some of the code I believe this PR touches that needs fixing. I'm looking into a fix for this.

Fix for the target registration is in #566

numba_cuda/numba/cuda/core/errors.py
gmarkall: Issue created to track this: https://github.com/NVIDIA/numba-cuda/issues/568

PR #562, opened 2025-10-29
Backporting Features to v0.20, bump to v0.20.1

isVoid:
- Relax the pinning to cuda-core to allow it floating across minor releases (#559) (645e46cf)
- Bump version to 0.20.1 (2e58567e)
- [test] Use numpy's tolerance for float16 (#491) (8bb46bce)


isVoid:
Since test-simulator is also spuriously failing with the same cause of https://github.com/NVIDIA/numba-cuda/pull/491. I'm going to backport this to v0.20 as well.

PR #565, opened 2025-10-30
[Refactor][NFC] Vendor-in compiler_lock for future CUDA-specific changes

atmnp:
This change vendors in the global_compiler_lock that is used to protect the core compiler machinery for future CUDA-specific changes.

### DIFF
--- a/numba_cuda/numba/cuda/core/compiler_lock.py
+++ b/numba_cuda/numba/cuda/core/compiler_lock.py
@@ -33,12 +33,6 @@ class _CompilerLock(object):
     def __exit__(self, exc_val, exc_type, traceback):
         self.release()
 
-    def is_locked(self):
-        is_owned = getattr(self._lock, "_is_owned")
-        if not callable(is_owned):
-            is_owned = self._is_owned
-        return is_owned()
-
     def __call__(self, func):
         @functools.wraps(func)
         def _acquire_compile_lock(*args, **kwargs):
@@ -46,16 +40,6 @@ class _CompilerLock(object):
                 return func(*args, **kwargs)
 
         return _acquire_compile_lock
-
-    def _is_owned(self):
-        # This method is borrowed from threading.Condition.
-        # Return True if lock is owned by current_thread.
-        # This method is called only if _lock doesn't have _is_owned().
-        if self._lock.acquire(0):
-            self._lock.release()
-            return False
-        else:
-            return True
 
 
 _numba_cuda_compiler_lock = _CompilerLock()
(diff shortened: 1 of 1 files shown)


gmarkall (changes requested):
We also need to hold the Numba compiler lock, because we'll likely have a route into Numba somewhere through types and / or the data model that doesn't hold or acquire the Numba compiler lock.

numba_cuda/numba/cuda/core/compiler_lock.py
```diff
@@ -0,0 +1,115 @@
...
+    _numba_compiler_lock = None
+
+
+# Lock for the preventing multiple compiler execution
+class _CompilerLock(object):
+    def __init__(self):
+        self._lock = threading.RLock()
+
+    def acquire(self):
+        ev.start_event("numba.cuda:compiler_lock")
```
gmarkall: We're using `numba-cuda:` as the prefix for events from Numba-CUDA. See

https://github.com/NVIDIA/numba-cuda/blob/5aa6b2eedd1be1af8b0fa118db04aa99fedede47/numba_cuda/numba/cuda/core/event.py#L58-L64

numba_cuda/numba/cuda/core/compiler_lock.py
```diff
@@ -0,0 +1,115 @@
...
+# Wrapper that coordinates both numba and numba-cuda compiler locks
+class _DualCompilerLock(object):
+    """Wrapper that coordinates both the numba-cuda and upstream numba compiler locks."""
+
+    def __init__(self, cuda_lock, numba_lock):
+        self._cuda_lock = cuda_lock
+        self._numba_lock = numba_lock
+
+    def acquire(self):
+        if self._numba_lock:
```
gmarkall: In what circumstance would this ever be `False`?

numba_cuda/numba/cuda/core/compiler_lock.py
```diff
@@ -0,0 +1,115 @@
...
+        _numba_cuda_compiler_lock, _numba_compiler_lock
+    )
+else:
+    global_compiler_lock = _numba_cuda_compiler_lock
+
+
+def require_global_compiler_lock():
+    """Sentry that checks the global_compiler_lock is acquired."""
+    # Use assert to allow turning off this check
+    assert global_compiler_lock.is_locked()
```
gmarkall: We don't use this, and it feels a little bit of a dangerous API to me, in that the name sounds like a function that will block until the global compiler lock is held (I'd probably have named it `assert_global_compiler_lock()` or similar). Can we remove it?
```suggestion
```

PR #569, opened 2025-10-31
Propose Alternative Module Path for `ext_types` and Maintain `numba.cuda.types.bfloat16` Import API

isVoid:
Recent upstream Numba type code vending moved the path of bfloat16 type to `ext_types` submodule.
This PR proposes a different layout of `ext_types` under `types.ext_types`. This layout
avoids the previous circular import issue and reimports `ext_types.bfloat16` to `numba.cuda.types.bfloat16`.

### DIFF
--- a/numba_cuda/numba/cuda/types/__init__.py
+++ b/numba_cuda/numba/cuda/types/__init__.py
@@ -14,7 +14,7 @@ from .misc import *
 from .npytypes import *
 from .scalars import *
 from .function_type import *
-from .ext_types import bfloat16
+from .ext_types import bfloat16, dim3, grid_group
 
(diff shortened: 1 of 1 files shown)


numba_cuda/numba/cuda/types/__init__.py
```diff
@@ -218,6 +219,7 @@
     ffi_forced_object
     ffi
     deferred_type
+    bfloat16
```
VijayKandiah, line 222: Should we also reimport `dim3` and `grid_group` here similar to `bfloat16`?

isVoid: Agreed. 04d674fe

PR #313, opened 2025-07-01
Updates for recent API changes

gmarkall:
This PR updates Numba-CUDA for recent driver API changes. Some related changes and simplifications are required; these are detailed in individual commit messages.

Fixes #281.

PR #571, opened 2025-11-03
Fix the `cuda.is_supported_version()` API

gmarkall:
PR #313 removed the `runtime.is_supported_version()` API, but it is used by the `cuda.is_supported_version()` public API. This commit restores the `cuda.is_supported_version()` API by checking whether the CUDA runtime major version is 12 or 13.

The version number check will need bumping as appropriate when future toolkit major versions are added and existing toolkit major version are dropped. This situation will be caught by the test that is added to exercise this API.

### DIFF
--- a/numba_cuda/numba/cuda/device_init.py
+++ b/numba_cuda/numba/cuda/device_init.py
@@ -133,7 +133,7 @@ def is_supported_version():
     - Generating an error or otherwise preventing the use of CUDA.
     """
 
-    return runtime.get_version()[0] in (12, 13)
+    return runtime.is_supported_version()
 
 
 def cuda_error():
--- a/numba_cuda/numba/cuda/cudadrv/runtime.py
+++ b/numba_cuda/numba/cuda/cudadrv/runtime.py
@@ -10,10 +10,18 @@ to the runtime anymore. This file is provided to maintain the existing API.
 
 from numba.cuda.cudadrv.nvrtc import _get_nvrtc_version
 
+SUPPORTED_TOOLKIT_MAJOR_VERSIONS = (12, 13)
+
 
 class Runtime:
     def get_version(self):
         return _get_nvrtc_version()
+
+    def is_supported_version(self):
+        """
+        Returns True if the CUDA Runtime is a supported version.
+        """
+        return self.get_version()[0] in SUPPORTED_TOOLKIT_MAJOR_VERSIONS
 
 
 runtime = Runtime()


numba_cuda/numba/cuda/tests/cudadrv/test_runtime.py
```diff
@@ -43,6 +43,16 @@ def test_visible_devices_set_after_import(self):
         visible_gpu_count = future.result()
         assert visible_gpu_count == 1
 
+    def test_is_supported_version(self):
+        # Exercise the `cuda.is_supported_version()` API.
+        #
+        # Assume for the purpose of the test that we're running on a supported
+        # toolkit version; if not, there's not much point in running the test
+        # suite.
+        from numba import cuda
```
cpcloud: Just curious: why are all the `numba.cuda` imports in this module happening inside of test methods instead of at module scope?

numba_cuda/numba/cuda/tests/cudadrv/test_runtime.py
```diff
@@ -43,6 +43,16 @@ def test_visible_devices_set_after_import(self):
...
+
+        self.assertTrue(cuda.is_supported_version())
```
cpcloud: Is `is_supported_version()` actually visible in `numba.cuda`? I don't see it either implicitly or explicitly named in `numba/cuda/__init__.py`.

numba_cuda/numba/cuda/device_init.py
```diff
@@ -133,7 +133,7 @@ def is_supported_version():
     - Generating an error or otherwise preventing the use of CUDA.
     """
 
-    return runtime.is_supported_version()
+    return runtime.get_version()[0] in (12, 13)
```
kkraus14: Is there any way to pull these versions into some kind of global in a more centralized space? This is kinda asking to bitrot in the future.

numba_cuda/numba/cuda/tests/cudadrv/test_runtime.py
gmarkall: The reason is because of the other test's logic:

https://github.com/NVIDIA/numba-cuda/blob/9eb2e25d7598da5ee1f31a77d8666393915420e7/numba_cuda/numba/cuda/tests/cudadrv/test_runtime.py#L27-L29

Maybe the test I added in this PR should be in another file. Let me change that.

numba_cuda/numba/cuda/tests/cudadrv/test_runtime.py
gmarkall: Because of the way the simulator works, `__init__.py` either imports everything from `simulator_init.py` or `device_init.py`. The `is_supported_version()` function comes from one of those.

numba_cuda/numba/cuda/device_init.py
gmarkall: We can put it in a global in a more centralized space - I don't think we have anywhere centralized in the code that explicitly lists supported versions anymore, so it would be a new addtion. I did aim to ensure that it would not bitrot in future with the test (it will fail when run on a hypothetical CUDA 14), but I can see that it would be better to embed the versions elsewhere.

gmarkall: I've put the `Runtime.is_supported_version()` method back and created a global in `numba.cuda.cudadrv.runtime` for the supported major versions.

numba_cuda/numba/cuda/tests/cudadrv/test_runtime.py
gmarkall: I've moved the test so it doesn't need to follow this odd pattern anymore.

numba_cuda/numba/cuda/tests/cudadrv/test_runtime.py
gmarkall: Looks like it was never even in `simulator_init()`, so I've added it there.

cpcloud (approved):

PR #572, opened 2025-11-03
[CI] Run PR workflow on merge to main

gmarkall:
Since we don't require branches to be up to date before merge, it would be a good idea to run CI on `main` after a merge. (Issue noted by @cpcloud)

I think this change achieves the intended effect, but someone who knows better than me should take a look :slightly_smiling_face:


cpcloud (approved):
Thanks!

PR #574, opened 2025-11-03
chore(pixi): set up doc builds; remove most `build-conda` dependencies

cpcloud:
Set up pixi for doc builds and remove the `build-conda` dependency in CI for conda tests


cpcloud:
I disabled the pixi cache, because it appears to be slower than using it due to download/upload.

PR #576, opened 2025-11-04
Fix checks on main

gmarkall:
#572 set up running the PR workflow on a merge to `main`, to catch any errors introduced by the merge of out-of-date branches that were otherwise tested and approved.

Parts of the RAPIDS `checks` workflow assume they are running on a PR, which causes them to fail when run on `main`. I think the only parts we are using from it run `ci/check_style.sh`, so this PR aims to fix the checks when run on `main` by bringing the `checks` workflow into Numba-CUDA and keeping only the parts we need.

The commits in this PR include one where I intentionally broke the formatting (https://github.com/NVIDIA/numba-cuda/pull/576/commits/9334737f4389929b9046121501c83264f714c5cd) to ensure that the style checks are still working as expected. They are then fixed in the subsequent commit.

### DIFF
--- a/numba_cuda/__init__.py
+++ b/numba_cuda/__init__.py
@@ -4,5 +4,3 @@
 from numba_cuda._version import __version__
 
 __all__ = ["__version__"]
-
-


.github/workflows/checks.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  check-style:
+    runs-on: ubuntu-latest
+    container:
+      image: rapidsai/ci-conda:latest
```
cpcloud, line 15: Can we avoid using special images and run using `ubuntu-latest` *only*? I'm happy to merge this to get CI green for now and then work on it in a follow-up.

cpcloud (approved):

.github/workflows/checks.yaml
gmarkall: I'm happy for us to do that - I don't quite know what's involved there but would be keen for us to do more standard things.

For background: the whole Numba-CUDA CI setup is derived from RAPIDS CI setups, which is why there are many RAPIDSisms in the setup here. Unfortunately I was not really knowledgeable enough to start afresh with a clean CI setup for Numba-CUDA (and leverage only the relevant parts of the RAPIDS setup) originally, and with the changes and additions over the past year there are quite a lot of anachronisms and oddities that have built up.

PR #577, opened 2025-11-04
ci: move pre-commit checks to pre commit action

cpcloud:
Move pre-commit check away from the rapids container

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,7 +7,6 @@ repos:
   hooks:
     - id: check-added-large-files
     - id: check-ast
-    - id: check-json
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -7,14 +7,12 @@ repos:
   hooks:
     - id: check-added-large-files
     - id: check-ast
-    - id: check-json
```
kkraus14, line 10: Any reason these were removed?

cpcloud: They aren't being used, because there are no JSON files and there's no `requirements.txt` file.

PR #579, opened 2025-11-04
refactor: decouple `Context` from `Stream` and `Event` objects

cpcloud:
Remove some unnecessary weak reference usage by decoupling `Context` objects from `Stream` and `Event`.

### DIFF
--- a/ci/test_conda.sh
+++ b/ci/test_conda.sh
@@ -4,14 +4,6 @@
 
 set -euo pipefail
 
-DISTRO=`cat /etc/os-release | grep "^ID=" | awk 'BEGIN {FS="="} { print $2 }'`
-
-if [ "$DISTRO" = "ubuntu" ]; then
-  apt-get update
-  apt remove --purge `dpkg --get-selections | grep cuda-nvvm | awk '{print $1}'` -y
-  apt remove --purge `dpkg --get-selections | grep cuda-nvrtc | awk '{print $1}'` -y
-fi
-
 # Constrain oldest supported dependencies for testing
 if [ "${NUMBA_VERSION:-*}" != "*" ]; then
     # add to the default environment's dependencies


numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2077,18 +2072,18 @@ def __cuda_stream__(self):
...
         ptr = self.handle.value or drvapi.CU_STREAM_DEFAULT
 
         if ptr in default_streams:
-            return default_streams[ptr] % self.context
+            return default_streams[ptr]
         elif self.external:
-            return "<External CUDA stream %d on %s>" % (ptr, self.context)
+            return f"<External CUDA stream {ptr:d}>"
         else:
-            return "<CUDA stream %d on %s>" % (ptr, self.context)
```
cpcloud, line 2091: The context was only being used for repring. The complexity of weak references doesn't seem justifiable for pretty printing, hence this PR :)

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2190,9 +2185,8 @@ def callback(stream, status, future):
         return future
 
 
-class Event(object):
-    def __init__(self, context, handle, finalizer=None):
-        self.context = context
```
cpcloud, line 2195: This was just never used in the class.

ci/test_conda.sh
```diff
@@ -4,14 +4,6 @@
 
 set -euo pipefail
 
-DISTRO=`cat /etc/os-release | grep "^ID=" | awk 'BEGIN {FS="="} { print $2 }'`
-
-if [ "$DISTRO" = "ubuntu" ]; then
-  apt-get update
-  apt remove --purge `dpkg --get-selections | grep cuda-nvvm | awk '{print $1}'` -y
-  apt remove --purge `dpkg --get-selections | grep cuda-nvrtc | awk '{print $1}'` -y
-fi
```
cpcloud, line 13: This is not related to the other changes, but it was causing problems in CI, and isn't necessary anymore since we're getting all dependencies from `pixi` in this script, so I removed it.

cpcloud:
Interesting, will have to explore this more tomorrow.

cpcloud:
I'll forego the `Context` removal from `Module` for now, and put that up in a separate PR, so that the other removals can remain.

ci/test_conda.sh
gmarkall: I think the reason this remained is that the CUDA pathfinder can mistakenly find system libraries ahead of ones in the environment, so we remove them to make sure we're testing with the right libraries. This was discussed in https://github.com/NVIDIA/cuda-python/issues/839 - note that although the issue is closed, I believe the fix only makes it less likely to find the library from the system install of the toolkit, not impossible.

kkraus14: The behavior now should be correct. pip wheels will be found first, then conda packages, then a default system search using `dlopen` which has its own hierarchy.

As far as I know the only issue that remains is that if you're trying to find system libraries (as opposed to pip wheels or conda packages) that nvvm doesn't end up in `ldconfig` / `LD_LIBRARY_PATH` or any of the standard search paths. We hope to address this in the near future.

cpcloud:
Sounds like we can continue with the review then? @gmarkall @kkraus14

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2058,9 +2054,8 @@ class ManagedOwnedPointer(OwnedPointer, mviewbuf.MemAlloc):
     pass
 
 
-class Stream(object):
-    def __init__(self, context, handle, finalizer, external=False):
-        self.context = context
+class Stream:
+    def __init__(self, handle, finalizer=None, external=False):
```
kkraus14, line 2058: There is unfortunately public code using the `Stream` constructor here with the context argument. It looks like most of it is stale and pinned to old versions of Numba, but we should maybe be a bit mindful.

@gmarkall what are your thoughts?

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2190,9 +2185,8 @@ def callback(stream, status, future):
         return future
 
 
-class Event(object):
-    def __init__(self, context, handle, finalizer=None):
-        self.context = context
+class Event:
+    def __init__(self, handle, finalizer=None):
```
kkraus14, line 2189: +1 on this being something that has been used with the `context` argument @gmarkall

cpcloud: It is currently doing nothing in the class and to my eye wasting space + adding the complexity of a `weakref`. I'm not sure preserving backwards compatibility for something that is not clearly serving any purpose is worth the trouble.

numba_cuda/numba/cuda/cudadrv/driver.py
cpcloud: We should probably lay out somewhere what constructors are public, since hand constructing these doesn't appear to be part of any explicit public API, just the one you get from the default Python conventions.

numba_cuda/numba/cuda/cudadrv/driver.py
kkraus14: I agree about it not serving a purpose currently, but there's something to be said about not breaking existing code using it...

cpcloud: No use of `numba.cuda` and `Event` in the same module across all of searchable rapidsai:

https://github.com/search?q=org%3Arapidsai+numba.cuda+language%3APython++Event&type=code

Same is true for `Stream`: https://github.com/search?q=org%3Arapidsai+numba.cuda+language%3APython++%22Stream%22&type=code

If there's some usage of it that we can point to that would be useful information.

numba_cuda/numba/cuda/cudadrv/driver.py
gmarkall: Can you highlight the code that's using the constructor explicitly? The way it's presented in the documentation is maybe not ideal, but users are directed to use one of:

* `cuda.stream()`
* `cuda.default_stream()`
* `cuda.legacy_default_stream()`
* `cuda.per_thread_default_stream()`
* `cuda.external_stream(ptr)`

to construct a `Stream` object rather than using the constructor directly: https://nvidia.github.io/numba-cuda/reference/host.html#stream-management

numba_cuda/numba/cuda/cudadrv/driver.py
gmarkall: Similarly to streams, users shoudn't be creating `Event` instances directly: https://nvidia.github.io/numba-cuda/reference/host.html#events

gmarkall:
I'm happy for this to be merged, though I'm mindful that it sounds like there could be code using the `Stream` and `Event` constructors directly that I'm unaware of (and can't find with a few minutes of searching and regexes with Sourcegraph).

cpcloud:
Let's merge, and we can revisit these constuctors if something comes up.

Issue #580, opened 2025-11-05
[FEA] Constant lists currently not implemented


brandon-b-miller:
We currently lack the ability to freeze a list in as constant memory:

```python
from numba import cuda
import numpy as np

lst = [1,2,3]

@cuda.jit
def kernel():
    result = np.percentile(lst, 50)
    print(result)

kernel[1,1]()
```

```
    raise errors.NumbaNotImplementedError(f"{self}, {sig}")
numba.cuda.core.errors.NumbaNotImplementedError: Failed in cuda mode pipeline (step: cuda native lowering)
<numba.cuda.core.base.OverloadSelector object at 0x7ff724f9ecf0>, (List(int64, True),)
During: lowering "$34load_global.3 = global(lst: [1, 2, 3])" at /home/nfs/brmiller/repro.py (8)
During: Pass cuda_native_lowering
```

This should work like it does for numpy arrays with the list frozen into the kernel. In the future it may need to be transformed into an persistent 1D device array once we stop moving global arrays of arbitrary length into constant memory, xref https://github.com/numba/numba/issues/9084

PR #584, opened 2025-11-06
Enable global lists within kernels

brandon-b-miller:
Closes https://github.com/NVIDIA/numba-cuda/issues/580

### DIFF
--- a/numba_cuda/numba/cuda/target.py
+++ b/numba_cuda/numba/cuda/target.py
@@ -273,58 +273,6 @@ class CUDATargetContext(BaseContext):
             name, argtypes, abi_tags=abi_tags, uid=uid
         )
 
-    def make_constant_list(self, builder, listty, lst):
-        """
-        Create a list structure with data in constant address space.
-        The list metadata is in normal memory but the data array is in constant memory.
-        """
-        import numpy as np
-        from numba.cuda.cpython.listobj import ListInstance
-
-        lmod = builder.module
-
-        # Convert to array and serialize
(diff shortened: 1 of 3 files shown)


numba_cuda/numba/cuda/typing/typeof.py
```diff
@@ -193,7 +193,7 @@ def _typeof_list(val, c):
     ty = typeof_impl(val[0], c)
     if ty is None:
         raise ValueError(f"Cannot type list element type {type(val[0])}")
-    return types.List(ty, reflected=True)
+    return ty[::1]
```
gmarkall, line 196: This was a sketch of how things work for a 1D list only. A more robust implementation would:

- Check that the purpose of the typeof operation is for a constant. We should not accept lists as arguments to kernels.
- Convert the value to a NumPy array, then determine the Numba type of the NumPy array. This will allow handling of lists-of-lists, e.g. `[[1, 2], [3, 4]]`. It also prevents having to handle what happens if there are different element types (e.g. int and float), whereas the existing implementation assumes that all list elements have the same type, which need not be the case.
- Not check for zero-length, because a zero-length NumPy array is possible and can be typed. The inability to handle a zero-length list is an issue of the reflected list implementation in Numba that doesn't present an issue for freezing global lists as arrays.

numba_cuda/numba/cuda/target.py
gmarkall, line 1: I think the changes here are now superfluous?

PR #586, opened 2025-11-10
Add more thirdparty tests

gmarkall:
This adds tests of additional third party libraries and fixes up issues that were found during testing them. Notes on the changes:

* The latest release, 0.6.0, of nvmath-python is used, and its device API tests are run.
* nvmath-python patches `types.number_domain`, which was not handled by our vendoring. The changes in `cudaimpl.py` resolve this issue.
* nvmath-python also creates extensions that can lead to a Numba-CUDA `Signature` class leaking into Numba, which checks for an instance of a Numba `Signature` only. The change in `templates.py` mitigates this by always constructing core `Signature` instances, if Numba is present.
* I tested the code with my ["Extending Numba CUDA" tutorials](https://github.com/gmarkall/extending-numba-cuda), and noticed that the `Dim3` and `GridGroup` were not in our `types` package, so I've added these back. I haven't added these examples as part of our CI as there would be some extra work involved in converting them to tests, but I may add them in a future PR.
* Awkward 2.8.10 is used for tests, as outlined in https://github.com/scikit-hep/awkward/discussions/3587
* The nvmath-python tests take 45 minutes so I've set them to only run on pushes to `main` rather than slowing down CI for PR testing. This is a little less than ideal, but better than increasing the iteration time for testing in CI. The commit https://github.com/NVIDIA/numba-cuda/pull/586/commits/e85e8d72119a244976d5a1d83b7e24b61ca314d5 demonstrates that they are successful (green tick from CI) before the commit https://github.com/NVIDIA/numba-cuda/pull/586/commits/fa92cb2ab4ec761d8bb1fea7cf4d0323e9e63e66 sets them to only run on `main`.
* It may be possible to run a reduced set of nvmath-python tests for every PR, but I need to work with the nvmath-python team to understand how to do that without sacrificing too much coverage.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,3 +30,7 @@ repos:
       language: python
       additional_dependencies:
         - pathspec==0.12.1
+- repo: https://github.com/sphinx-contrib/sphinx-lint
+  rev: v1.0.2
+  hooks:
+    - id: sphinx-lint
--- a/ci/test_conda.sh
+++ b/ci/test_conda.sh
@@ -4,14 +4,6 @@
 
 set -euo pipefail
 
-DISTRO=`cat /etc/os-release | grep "^ID=" | awk 'BEGIN {FS="="} { print $2 }'`
-
-if [ "$DISTRO" = "ubuntu" ]; then
-  apt-get update
-  apt remove --purge `dpkg --get-selections | grep cuda-nvvm | awk '{print $1}'` -y
-  apt remove --purge `dpkg --get-selections | grep cuda-nvrtc | awk '{print $1}'` -y
-fi
-
 # Constrain oldest supported dependencies for testing
 if [ "${NUMBA_VERSION:-*}" != "*" ]; then
     # add to the default environment's dependencies
--- a/docs/source/reference/types.rst
+++ b/docs/source/reference/types.rst
@@ -119,40 +119,51 @@ A number of math intrinsics that utilizes the device native computing feature
 on ``bfloat16`` are provided:
 
 .. function:: numba.cuda.bf16.htrunc(b)
+
     Round ``b`` to the nearest integer value that does not exceed ``b`` in magnitude.
 
 .. function:: numba.cuda.bf16.hceil(b)
+
     Compute the smallest integer value not less than ``b``.
 
 .. function:: numba.cuda.bf16.hfloor(b)
+
     Calculate the largest integer value which is less than or equal to ``b``.
 
 .. function:: numba.cuda.bf16.hrint(b)
+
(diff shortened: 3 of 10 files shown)


cpcloud:
Can we exercise the pixi build infrastructure for this? It looks like `nvmath-python@0.6.0` is available on conda-forge.

gmarkall:
> Can we exercise the pixi build infrastructure for this? It looks like `nvmath-python@0.6.0` is available on conda-forge.

I'm not sure - what does it mean for us to "exercise the pixi build infrastructure"? Where should I look to get an understanding? (Unfortunately I've not kept up with developments in this area so I'm not sure what to focus on to begin understanding the question)

ci/test_thirdparty_nvmath.sh
```diff
@@ -0,0 +1,55 @@
...
+
+rapids-logger "Show Numba system info"
+python -m numba --sysinfo
+
+rapids-logger "Run nvmath-python device tests"
+pushd tests
+# Required for nvmath-python to locate pip-install MathDx
+export SYS_PREFIX=`python -c "import sys; print(sys.prefix)"`
+export MATHDX_HOME=${SYS_PREFIX}/lib/python3.13/site-packages/nvidia/mathdx
+python -m pytest nvmath_tests/device --tb=native -x
```
brandon-b-miller, line 52: RE: clogging up CI, did you try scaling out with multiple pytest workers?

brandon-b-miller (approved):
One non-blocking Q

ci/test_thirdparty_nvmath.sh
gmarkall: I did, but the tests use a lot of memory so it's not easy to find a sweet spot that speeds things up and doesn't OOM.

rparolin (approved):

ci/test_thirdparty_nvmath.sh
```diff
@@ -0,0 +1,55 @@
...
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: BSD-2-Clause
+
+set -euo pipefail
+
+CUDA_VER_MAJOR_MINOR=${CUDA_VER%.*}
+
+NVMATH_PYTHON_VERSION="0.6.0"
+# The commit on Github corresponding to 0.6.0
+NVMATH_PYTHON_SHA="6bddfa71c39c07804127adeb23f5b0d2168ae38c"
```
rparolin: Non-blocking: these dependency management settings should be hoisted out into a yml file located at the project root or something similar.

gmarkall:
Thanks for the reviews! I'm still needing to resolve the issue with `pr-builder / run` being considered skipped, prior to merging this.

PR #587, opened 2025-11-10
feat: add support for cache-hinted load and store operations

kaeun97:
Follow up of #51 requested by @gmarkall.

- [x] Add documentation.
- ~Test cases for erroneous arguments. It would be good to check for accidental use on shared or local arrays, but this may not be easy to do.~
- [x] Add additional validations as described in comments in ld_cache_operator and st_cache_operator.
- [x] Add test for bitwidth validation
- [x] Refactor the implementation - the load and store implementations contain a lot of common code.
- ~Decide on whether to support complex, and work out why it presently doesn't work.~
- [x] Support `CPointer()`
- [x] Test for support`CPointer()`

Updated items from feedback.

- [x] Small nits on the documentation
- [x] Small nits on the tests
- [x] Add tests with 2D arrays
- [x] Fix the simulator, either by stub implementations and test skips

### DIFF
--- a/numba_cuda/numba/cuda/tests/cudapy/test_cache_hints.py
+++ b/numba_cuda/numba/cuda/tests/cudapy/test_cache_hints.py
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: BSD-2-Clause
 
 from numba import cuda, errors, typeof, types
-from numba.cuda.testing import unittest, CUDATestCase
+from numba.cuda.testing import unittest, CUDATestCase, skip_on_cudasim
 import numpy as np
 
 tested_types = (
@@ -34,6 +34,7 @@ store_operators = (
 )
 
 
+@skip_on_cudasim("Cache hints not supported on simulator")
 class TestCacheHints(CUDATestCase):
     def test_loads(self):
         for operator, modifier in load_operators:
--- a/numba_cuda/numba/cuda/simulator/api.py
+++ b/numba_cuda/numba/cuda/simulator/api.py
@@ -164,3 +164,16 @@ def defer_cleanup():
 
 def is_supported_version():
     return True
+
+
+# cache hints operations
+ldca = None
+ldcg = None
+ldcs = None
+ldlu = None
+ldcv = None
+
+stcg = None
+stcs = None
+stwb = None
+stwt = None


numba_cuda/numba/cuda/cache_hints.py
```diff
@@ -0,0 +1,267 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: BSD-2-Clause
+
+from llvmlite import ir
+from numba import types
+from numba.core import cgutils
+from numba.core.extending import intrinsic, overload
+from numba.core.errors import NumbaTypeError
```
gmarkall: Since we vendored some code from upstream Numba into Numba-CUDA since I made the original PR, these imports also need an update:


```suggestion
from numba.cuda import cgutils
from numba.cuda.extending import intrinsic, overload
from numba.cuda.core.errors import NumbaTypeError
```

gmarkall:
Some relevant discussion also in https://github.com/NVIDIA/numba-cuda/pull/51/files#r2513450704

numba_cuda/numba/cuda/cache_hints.py
kaeun97: Oh wow, thanks a bunch @gmarkall (for both of this suggestion and the comment [here](https://github.com/NVIDIA/numba-cuda/pull/51/files#r2513450704))! It seemed like I was not registering the operations properly as was constantly getting:
```
numba.cuda.core.errors.TypingError: Failed in cuda mode pipeline (step: nopython frontend)
E                   Unknown attribute 'ldcs' of type Module(<module 'numba.cuda' from '../numba-cuda/numba_cuda/numba/cuda/__init__.py'>)
```

This resolves the bug :)

gmarkall:
> Decide on whether to support complex, and work out why it presently doesn't work.

I should mention, I think it's fine if we don't support complex types for this PR, and could add it in a follow-up PR.

gmarkall:
A quick follow up on a couple of items:

> Add documentation.

Do you need any assistance here? I know that adding documentation is often a bit hard to find the right place and wording, so I'm happy to add something here if you want.

> Test cases for erroneous arguments. It would be good to check for accidental use on shared or local arrays, but this may not be easy to do.

I see you have test cases for various user errors already. I think checking for accidental use on shared or local arrays is probably hard / impossible without something like #236, so maybe this item can be considered already done?

kaeun97:
Thanks for the follow up! @gmarkall 

> I see you have test cases for various user errors already. I think checking for accidental use on shared or local arrays is probably hard / impossible without something like https://github.com/NVIDIA/numba-cuda/pull/236, so maybe this item can be considered already done?

I agree. If needed, I can follow up on that later (ideally after https://github.com/NVIDIA/numba-cuda/pull/236 gets merged)

> Do you need any assistance here? I know that adding documentation is often a bit hard to find the right place and wording, so I'm happy to add something here if you want.

I've added a rough draft that needs refinement (in terms of the content, where to put it, etc). Would you be able to take a quick look? Would be happy to iterate on that. If needed, we can hop on a call or feel free to directly make changes to this branch.

docs/source/reference/cache_hints.rst
```diff
@@ -0,0 +1,63 @@
+..
+   SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+   SPDX-License-Identifier: BSD-2-Clause
+
+_cache-hints:
```
gmarkall: Annoyingly, sphinx-lint did not catch this either:

```suggestion
.. _cache-hints:
```

docs/source/reference/cache_hints.rst
```diff
@@ -0,0 +1,63 @@
...
+
+Cache Hints for Memory Operations
+=================================
+
+These functions provide explicit control over caching behavior for memory
+operations. They generate PTX instructions with cache policy hints that can
+optimize specific memory access patterns. All functions support arrays or
+pointers with integer and floating-point types (bitwidths 8, 16, 32, 64, or
+128).
```
gmarkall: I think we don't really support 128-bit values very well in Numba-CUDA, so it'd probably best to just mention up to 64-bit values here to avoid creating the impression that `float128` and `int128` might work elsewhere:

```suggestion
pointers with integer and floating-point types (bitwidths 8, 16, 32, or 64).
```

A side-question: Is your expectation here that these instructions would be expected to work for e.g. `float64x2`? (It doesn't matter for the scope of this PR, I think the work is more than sufficient / complete, but I'm curious).

docs/source/reference/cache_hints.rst
```diff
@@ -0,0 +1,63 @@
...
+   is the default caching behavior.
+
+.. function:: numba.cuda.stwt(array, i, value)
+
+   Store ``value`` to ``array[i]`` with write-through policy (``st.global.wt``).
+   Writes through cache hierarchy to memory.
+
+.. seealso:: `Cache Operators
+   <https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#cache-operators>`_
+   in the PTX ISA documentation.
```
gmarkall: This is a great reference to point to - I think it would be also nice to move this section to just before the descriptions of each of the instructions (i.e. around line 15), because on most screens it will be off the first page of text, and it will likely be handy to reference immediately on a first-time read-through.

docs/source/reference/cache_hints.rst
gmarkall, line 1: This new documentation page is a very nice overview of the instructions, and the descriptions seem clear and accessible for a typical user encountering these instructions for the first time. Thanks for doing this thoughtful write-up!

docs/source/reference/cache_hints.rst
gmarkall: Now I've got to looking at the tests, I realise the mention of 128 was because of the `complex128` support (which was not working in my initial PR). Perhaps it's more straightfoward to refer to "all bitwidths of", like:



```suggestion
pointers with all bitwidths of signed/unsigned integer, floating-point, and
complex types.
```

gmarkall: Agh, I didn't read far enough - there is a reference to complex types in the test file (https://github.com/NVIDIA/numba-cuda/pull/587/files#diff-e1fbbc5bafe8831deb4b0ff82f2b8c21438516e645a9574efef670e71966ea34R22), I see now they're not tested. So maybe:

```suggestion
pointers with all bitwidths of signed/unsigned integer and floating-point
types.
```

docs/source/reference/cache_hints.rst
kaeun97: thanks for catching!

numba_cuda/numba/cuda/tests/cudapy/test_cache_hints.py
```diff
@@ -0,0 +1,173 @@
...
+class TestCacheHints(CUDATestCase):
+    def test_loads(self):
+        for operator, modifier in load_operators:
+
+            @cuda.jit
+            def f(r, x):
+                for i in range(len(r)):
+                    r[i] = operator(x, i)
+
+            @cuda.jit
```
gmarkall: Because `f_ptr` goes through `compile_ptx()`, it should only be a normal Python function; we wouldn't normally JIT a function then also compile it to PTX for use outside of Numba.

```suggestion
```

numba_cuda/numba/cuda/tests/cudapy/test_cache_hints.py
```diff
@@ -0,0 +1,173 @@
...
+                    r = np.zeros_like(x)
+
+                    f[1, 1](r, x)
+                    np.testing.assert_equal(r, x)
+
+                    # Check PTX contains a cache-policy load instruction
+                    numba_type = typeof(x)
+                    bitwidth = numba_type.dtype.bitwidth
+                    sig = (numba_type, numba_type)
+                    ptx, _ = cuda.compile_ptx(f, sig)
```
gmarkall: It would be better to check the assembly of the kernel we've just launched, rather than compiling a new variant:

```suggestion
                    ptx = f.inspect_asm(signature=sig)
```

This way, we are also getting a test with the two pathways to compilation - one for the jit compilation and launch from Python, and one from the compilation API where we produce PTX (or LTO-IR) that gets used somewhere else (e.g. embedded into code used by another, possibly non-Python, library).

numba_cuda/numba/cuda/tests/cudapy/test_cache_hints.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def test_stores(self):
+        for operator, modifier in store_operators:
+
+            @cuda.jit
+            def f(r, x):
+                for i in range(len(r)):
+                    operator(r, i, x[i])
+
+            @cuda.jit
```
gmarkall: As above:

```suggestion
```

numba_cuda/numba/cuda/tests/cudapy/test_cache_hints.py
```diff
@@ -0,0 +1,173 @@
...
+                    r = np.zeros_like(x)
+
+                    f[1, 1](r, x)
+                    np.testing.assert_equal(r, x)
+
+                    # Check PTX contains a cache-policy store instruction
+                    numba_type = typeof(x)
+                    bitwidth = numba_type.dtype.bitwidth
+                    sig = (numba_type, numba_type)
+                    ptx, _ = cuda.compile_ptx(f, sig)
```
gmarkall: As above:

```suggestion
                    ptx = f.inspect_asm(signature=sig)
```

numba_cuda/numba/cuda/tests/cudapy/test_cache_hints.py
```diff
@@ -0,0 +1,173 @@
...
+                    operator=operator, ty=ty, input_type="cpointer"
+                ):
+                    x = np.zeros(5, dtype=ty)
+                    numba_type = typeof(x)
+                    ptr_type = types.CPointer(numba_type.dtype)
+                    sig = (ptr_type, numba_type, types.intp)
+
+                    ptx, _ = cuda.compile_ptx(f_ptr, sig)
+
+                    numba_type = typeof(x)
```
gmarkall: This looks redundant (an accidental copy/paste?):

```suggestion
```

docs/source/reference/cache_hints.rst
kaeun97: Make sense!

numba_cuda/numba/cuda/tests/cudapy/test_cache_hints.py
kaeun97: Sounds good, thank you for the suggestion!

numba_cuda/numba/cuda/tests/cudapy/test_cache_hints.py
kaeun97: Good catch, I've refactored a bit to avoid these redundancy.

docs/source/reference/cache_hints.rst
kaeun97: Thank you for checking. After testing, complex types don't work (as they are structures), and 128-bit types can't be tested (`float128` not supported in numba). Let me update the docs and remove the unused variables.

kaeun97:
@gmarkall Thank you for the thorough feedback. I believe I have addressed all items you have mentioned. Would be great if you can re-visit the PR!

gmarkall:
I'm just waiting to merge this until https://github.com/NVIDIA/numba-cuda/actions/runs/19534738630 completes, as it's a run of a new CI workflow for `main` - I'd like to make sure we have one complete successful run of that before continuing with merging other PRs.

gmarkall:
Many thanks @kaeun97!

PR #589, opened 2025-11-14
Fix freezing in of constant arrays with negative strides

brandon-b-miller:
This PR fixes a bug when freezing certain views of numpy arrays into kernels. The current implementation loops through the bytes of the source array in viewed order but maintains a potentially negative stride without also in that case bumping the pointer to the end of the array. This PR changes it so that the physical array we put in constant memory is just the contiguous logical equivalent of the original array with a positive stride. This should be simpler than actually jumping through the pointer arithmetic given we have to make a copy of the view anyways, so the inherent "viewness" is already going to be lost when moving the data to device.

### DIFF
--- a/numba_cuda/numba/cuda/target.py
+++ b/numba_cuda/numba/cuda/target.py
@@ -8,6 +8,7 @@ from llvmlite import ir
 import warnings
 import importlib.util
 import numpy as np
+
 from numba.cuda import types
 from numba.cuda import HAS_NUMBA
 from numba.cuda.core.compiler_lock import global_compiler_lock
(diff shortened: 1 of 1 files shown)


numba_cuda/numba/cuda/target.py
```diff
@@ -280,6 +280,16 @@ def make_constant_array(self, builder, aryty, arr):
         addrspace.
         """
 
+        # Ensure we have a contiguous buffer with non-negative strides. views with
+        # negative strides must be materialized so that the
+        # constant bytes and the data pointer/strides are consistent.
+        if (
+            not getattr(arr, "flags", None)
```
gmarkall: Is there ever an array that doesn't have a `flags` property?

numba_cuda/numba/cuda/target.py
```diff
@@ -280,6 +280,16 @@ def make_constant_array(self, builder, aryty, arr):
...
+            or (not arr.flags.c_contiguous)
+            or any(s < 0 for s in arr.strides)
+        ):
+            arr = np.ascontiguousarray(arr)
```
gmarkall, line 290: I think `ascontiguousarray` just returns the input array if it was already contiguous, so maybe we could skip all the checks and just always do `arr = np.ascontiguousarray(arr)`? (noted from the examples in https://numpy.org/devdocs/reference/generated/numpy.ascontiguousarray.html)

brandon-b-miller:
Missing a fortran contiguous case here we shouldn't cast, fixing.

Issue #595, opened 2025-11-18
[FEA] Support for Python 3.14


gmarkall:
**Is your feature request related to a problem? Please describe.**

Support Python 3.14

**Describe the solution you'd like**

A similar implementation to the Numba implementation: 

* https://github.com/numba/numba/pull/10150

gmarkall:
cc @rparolin

PR #599, opened 2025-11-19
Support python `3.14`

brandon-b-miller:
This PR vendors the relevant changes from upstream numba required to support python 3.14.

Closes https://github.com/NVIDIA/numba-cuda/issues/595

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -3869,6 +3869,764 @@ environments:
         build: py313he80dd91_0
       - pypi: https://files.pythonhosted.org/packages/5c/40/69ca9ea803303e14301fff9d4931b6d080b9603e134df0419c55e9764df4/filecheck-1.0.3-py3-none-any.whl
       - pypi: https://files.pythonhosted.org/packages/e1/8b/200088c6859d8221454825959df35b5244fa9bdf263fd0249ac5fb75e281/ml_dtypes-0.5.4-cp313-cp313-win_amd64.whl
+  cu-13-0-py314:
+    channels:
+    - url: https://conda.anaconda.org/conda-forge/
+    indexes:
+    - https://pypi.org/simple
+    options:
+      pypi-prerelease-mode: if-necessary-or-explicit
+    packages:
+      linux-64:
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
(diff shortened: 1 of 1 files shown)


brandon-b-miller:
This PR likely needs to expand the CI matrix to include python 3.14 testing / pass before merging.

brandon-b-miller:
Ah, our CI scripts are written to use currently nonexistent rapids python 3.14 containers. I can see what can be done about this.

numba_cuda/numba/cuda/core/interpreter.py
```diff
@@ -2378,6 +2386,16 @@ def op_LOAD_FAST_AND_CLEAR(self, inst, res):
     else:
         raise NotImplementedError(PYVERSION)
```
cpcloud, line 2396: I am hitting this line when running tests in this PR.

If this is directly coming from numba, I'm not sure how this code ever passed 3.14.

brandon-b-miller: probably a missed change, will check.

cpcloud: there's also a failing assert here:

```
    def op_SET_FUNCTION_ATTRIBUTE(self, state, inst):
>       assert PYVERSION in ((3, 13),)
               ^^^^^^^^^^^^^^^^^^^^^^^
E       AssertionError

numba_cuda/numba/cuda/core/byteflow.py:1866: AssertionError
```

numba_cuda/numba/cuda/tests/cudapy/test_vector_type.py
```diff
@@ -187,41 +187,43 @@ def kernel(res):
...
-            f4_26,
-            f4_27,
-            f4_28,
-            f4_29,
-            f4_30,
-            f4_31,
-            f4_32,
-            f4_33,
-            f4_34,
+        for v in tuple(
```
cpcloud, line 190: While this is redundant, without the additional `tuple` here numba is unable to type the inner `tuple` correct (it ends up as a `list` somehow).

cpcloud:
@greptile-apps The only valid comment there is the typo. Where are `or` and `and` on any of those lines?

cpcloud:
@greptile-apps That's not the typo you originally mentioned. The original typo you mentioned was `keept -> kept`.

PR #606, opened 2025-11-22
fix: warp vote operations must use a constant int for the `mode` parameter

### LINKED ISSUE #592: [BUG] Warp vote operations must use a constant int for the `mode`
**Describe the bug**

NVVM IR generation for warp sync intrinsics was fixed by #231 (for Issue #228). This only fixes the issue for some intrinsics; the warp vote intrinsics, documented at https://docs.nvidia.com/cuda/nvvm-ir-spec/index.html?highlight=data%2520movement#vote also require the `mode` parameter to be a constant int.

These are used in the implementation of the following functions in Numba-CUDA:

- `cuda.all_sync()`
- `cuda.any_sync()`
- `cuda.eq_sync()`
- `cuda.ballot_sync()`

This issue can be observed for `all_sync()` (others are similar) by running:

```python
from numba import cuda, types

def use_shfl_sync_idx(ary):
    i = cuda.grid(1)
    val = cuda.all_sync(0xFFFFFFFF, ary[i])
    ary[i] = val


args = (types.int32[::1],)
ptx, resty = cuda.compile_ptx(use_shfl_sync_idx, args, cc=(10, 0))

print(ptx)
```

with

```
NUMBA_DUMP_LLVM=1 python repro.py
```

The output will contain something similar to:

```
  %".5" = trunc i64 4294967295 to i32
  %".6" = trunc i64 0 to i32
  %".7" = icmp ne i32 %"predicate", 0
  %".8" = call {i32, i1} @"llvm.nvvm.vote.sync"(i32 %".5", i32 %".6", i1 %".7")
```

to set up the parameters for and call `llvm.nvvm.vote.sync()`. The parameter `%.6` is the `mode` parameter. Instead of being a value, it should be an constant `i32 0`. So the generated IR should look something like:

```
  %".5" = trunc i64 4294967295 to i32
  %".6" = icmp ne i32 %"predicate", 0
  %".7" = call {i32, i1} @"llvm.nvvm.vote.sync"(i32 %".5", i32 0, i1 %".6")
```

(note that values are renumbered now that there is no need for a `trunc i64 to 0` value).

The output PTX may be printed by the example; however, prior to the fix there is also a chance of crashing NVVM.

**Expected behavior**

Constant parameters should be generated for the `mode` parameter, as seen in the corrected output in the example above.

**Additional context**

The fix will look very similar for that to #288.

I think that there will be no need for the `intrinsic_wrapper.py` file after this fix, and it can be removed.

NVBug: 5122224.

kaeun97:
Fixes #592. Followed similar pattern as https://github.com/NVIDIA/numba-cuda/pull/231. cc. @gmarkall

### DIFF
--- a/numba_cuda/numba/cuda/tests/cudapy/test_warp_ops.py
+++ b/numba_cuda/numba/cuda/tests/cudapy/test_warp_ops.py
@@ -262,6 +262,10 @@ class TestCudaWarpOperations(CUDATestCase):
                 "Mask type must be an integer",
             ),
             (
+                "void(boolean[:], int32[:], int32[:])",
+                "Mask type must be an integer",
+            ),
+            (
                 "void(float64[:], int32[:], int32[:])",
                 "Mask type must be an integer",
             ),


numba_cuda/numba/cuda/intrinsics.py
```diff
@@ -380,3 +380,139 @@ def codegen(context, builder, sig, args):
...
+
+    sig_outer = signature(types.i4, mask_type, predicate_type)
+    return sig_outer, codegen
+
+
+def vote_sync_intrinsic(typingctx, mask_type, mode_value, predicate_type):
+    # Validate mode value
+    if mode_value not in (0, 1, 2, 3):
+        raise ValueError("mode must be 0 (all), 1 (any), 2 (eq), or 3 (ballot)")
+
```
gmarkall, line 496: The warp vote operations are a little different to the warp sync operations in that we never permitted vote operations to be used for floating point types, whereas sync operations were. We do not need to add support for floating point here (I don't think it would even make sense to do so) but I think we do need to guard against user code passing in floating point types. This is because it will lead to LLVM code generation errors or NVVM verification failures in the current implementation, because the types will be accepted but the `codegen()` function does not handle them.

I had a bit of a fiddle around with this, and I think that the following does manage to reject most of what we want to reject whilst still allowing the types we want:

```suggestion
    if types.unliteral(mask_type) not in types.integer_domain:
        raise NumbaTypeError(f"Mask type must be an integer. Got {mask_type}")

    predicate_types = types.integer_domain | {types.boolean}
    if types.unliteral(predicate_type) not in predicate_types:
        raise NumbaTypeError(
            f"Predicate must be an integer or boolean. Got {predicate_type}"
        )

``` 

My thinking here:
* `membermask` (which has type `mask_type`) is documented as expected to be a 32-bit integer. I added the `unliteral` because `types.integer_domain` holds the different integer types but does not include integer literal types, and if a literal is provided, its type will be `types.IntegerLiteral`, and it will fail the membership test. `unliteral` returns the underlying type of a literal (e.g. `int64`, `int32`, etc), and is idempotent for a non-literal type, so it should handle all cases.
* Same thoughts on literals for `predicate_type` as for `mask_type`, but we handle both integers and booleans for `predicate`, so we need to accept `types.boolean` in addition to the integer domain.

It might be worth a think about if there are other cases that were enforced by the typing removed from [this line](https://github.com/NVIDIA/numba-cuda/pull/606/files#diff-8cad93c828fc0a088291de28a0e188064c061f29d8a86f13b1c9e814e33abd55L111):

```python
signature(types.Tuple((types.i4, types.b1)), types.i4, types.i4, types.b1)
```

in case there's something I've missed here.

It might also be worth adding a test that checks that the incorrect types are correctly rejected.

gmarkall (changes requested):
Many thanks for this PR - this is great work!

I think there is a subtlety around the types that differs a bit from the warp sync intrinsics, and I've outlined my thoughts on it on the diff.

Note that I think CI is good with respect to this PR - the failure is an integration test with a downstream library (Awkward Array) and I think it is a bit sensitive to the versions of its dependencies, so unexpected / unrelated failures sometimes turn up. I'll confirm this and get back to you if there is a real issue though (unlikely, I think).

gmarkall:
> Note that I think CI is good with respect to this PR - the failure is an integration test with a downstream library (Awkward Array) and I think it is a bit sensitive to the versions of its dependencies, so unexpected / unrelated failures sometimes turn up. I'll confirm this and get back to you if there is a real issue though (unlikely, I think).

I've looked into this and it seems like running the Awkward Array tests with CuPy 13.6 for the version of Awkward that we use causes issues. So the fails are unrelated to this PR. I have a potential fix in #607.

numba_cuda/numba/cuda/intrinsics.py
kaeun97: Thank you @gmarkall for this! I think what you have suggested should be able to fulfill the signature:
```
signature(types.Tuple((types.i4, types.b1)), types.i4, types.i4, types.b1)
```

kaeun97:
@gmarkall Thank you for the feedback! I have added type checks for `mask` and `predicate` and some tests to validate the checks. 

Couple things to note:
- Although the document states 32-bit integers, as we do in `shfl_sync_intrinsic`, `vote_sync_intrinsic` also accepts 64 bit signed/unsigned integers as we truncate later
- I've added a check to make sure that `mask` is not boolean (just accepts integers).

Test coverage:
- Rejects float types for both mask and predicate
- Rejects boolean type for mask (while allowing it for predicate)
- Accepts signed/unsigned 32-bit and 64-bit integers for mask and predicate
- Accepts literal values for mask and predicate

Would appreciate if you can take a look!

PR #609, opened 2025-11-25
Migrate numba-cuda driver to use cuda.core.launch API

rparolin:
# PR #609: Migrate kernel launch to cuda.core

**TL;DR**: Migrates numba-cuda's kernel launching from legacy `launch_kernel` API to `cuda.core.experimental.launch`. Performance concerns addressed through optimization work. Tests pass. Removes old abstractions.

---

### Core Files Modified:

1. **`driver.py`** - Major surgery here
   - ❌ **Removed**: `launch_kernel()`, abstract `Module`/`Function` base classes, `CtypesModule`, `CtypesFunction`
   - ✅ **Added**: `_to_core_stream()` helper for stream conversion
   - ✅ **Changed**: `CudaPythonModule` now wraps `ObjectCode` directly
   - Uses `cuLibraryGetGlobal`/`cuLibraryUnload` instead of `cuModuleGetGlobal`/`cuModuleUnload`
   - Function attributes now read from `kernel.attributes` instead of CUDA driver calls

2. **`dispatcher.py`** - Launch infrastructure rewrite
   - Now uses `LaunchConfig` and `cuda.core.experimental.launch()`
   - Stream conversion happens early in `_LaunchConfiguration.__init__`
   - Kernel args preparation simplified (less ctypes overhead)
   - Added `__getstate__`/`__setstate__` for pickle support, but see https://github.com/NVIDIA/numba-cuda/issues/648

3. **Tests** - Updated everywhere
   - Changed `CudaAPIError` → `CUDAError`
   - Module callbacks now receive `ObjectCode` instead of raw handles
   - Stream conversions use `ExperimentalStream.from_handle()` or `_to_core_stream()`

---

## Performance Impact

**Initial concern**: Early benchmarks showed 30-40% regression on single-arg cases.

**Resolution**: Addressed through optimization work in recent commits:
- Removed instance-checking helpers (`prepare_args`) to minimize overhead
- Simplified kernel args preparation (less ctypes wrapping, direct value passing)
- Reduced stream conversion overhead by moving the conversion outside of the kernel invocation.

### DIFF
--- a/numba_cuda/numba/cuda/tests/cudadrv/test_cuda_driver.py
+++ b/numba_cuda/numba/cuda/tests/cudadrv/test_cuda_driver.py
@@ -3,12 +3,7 @@
 
 from ctypes import c_int, sizeof
 
-from numba.cuda.cudadrv.driver import (
-    host_to_device,
-    device_to_host,
-    driver,
-    _to_core_stream,
-)
+from numba.cuda.cudadrv.driver import host_to_device, device_to_host, driver
 from cuda.core.experimental import (
     LaunchConfig,
     Stream as ExperimentalStream,
@@ -120,6 +115,8 @@ class TestCudaDriver(CUDATestCase):
         module.unload()
 
     def test_cuda_driver_stream_operations(self):
+        from numba.cuda.cudadrv.driver import _to_core_stream
+
         module = self.context.create_module_ptx(self.ptx)
         function = module.get_function("_Z10helloworldPi")
 


numba_cuda/numba/cuda/cudadrv/libs.py
```diff
@@ -59,6 +59,13 @@ def get_cudalib(lib, static=False):
     libdir = get_cuda_paths()[dir_type].info
 
     candidates = find_lib(lib, libdir, static=static)
+
```
gmarkall: What is this for?

rparolin: I was seeing a local test runs fail. See below. I'm not sure what I'm doing wrong that I'm seeing this locally if no one else is.  I recently started using pixi so perhaps its installing dependencies in the wrong place?

``` FAILED test_debuginfo.py - LinkerError: libcudadevrt.a not found```

Comment (rparolin) [reply numba_cuda/numba/cuda/cudadrv/libs.py]: Removed. Fixed locally. Verified that it doesn't impact CI.

**Comment by leofang** [comment]
- I suggest to hold off from merging for now. We should avoid importing private attributes and functions if possible. Any gap identified by this PR (which is a good exercise!) needs to be addressed in cuda-core.
- What is the perf number before/after this change? As it stands, it is hard to guess if this PR brings any perf improvement. I would hate to cause perf regression just because we want dogfooding 🙂

**Comment by rparolin** [comment]
> * I suggest to hold off from merging for now. We should avoid importing private attributes and functions if possible. Any gap identified by this PR (which is a good exercise!) needs to be addressed in cuda-core.
> * What is the perf number before/after this change? As it stands, it is hard to guess if this PR brings any perf improvement. I would hate to cause perf regression just because we want dogfooding 🙂

```  ┌─────────────────────────────────────┬──────────────┬──────────────┬──────────────┬────────────┐
  │ Component                           │   Baseline   │   Current    │    Delta     │   Change   │
  ├─────────────────────────────────────┼──────────────┼──────────────┼──────────────┼────────────┤
  │ PYTHON DISPATCH LAYER               │              │              │              │            │
  ├─────────────────────────────────────┼──────────────┼──────────────┼──────────────┼────────────┤
  │   dispatcher.py → launch_kernel()   │     ~2.0 μs  │     ~2.0 μs  │      ~0.0 μs │      ~0%   │
  ├─────────────────────────────────────┼──────────────┼──────────────┼──────────────┼────────────┤
  │ CUDA.CORE OBJECTS (NEW OVERHEAD)    │              │              │              │            │
  ├─────────────────────────────────────┼──────────────┼──────────────┼──────────────┼────────────┤
  │   LaunchConfig creation             │      ---     │     0.411 μs │    +0.411 μs │    +100%   │
  │   ObjectCode stub creation          │      ---     │     0.093 μs │    +0.093 μs │    +100%   │
  │   Kernel._from_obj()                │      ---     │    10.971 μs │   +10.971 μs │    +100%   │
  │   Stream.from_handle()              │      ---     │     2.706 μs │    +2.706 μs │    +100%   │
  ├─────────────────────────────────────┼──────────────┼──────────────┼──────────────┼────────────┤
  │   Subtotal (cuda.core objects)      │      0.0 μs  │    14.181 μs │   +14.181 μs │    +100%   │
  ├─────────────────────────────────────┼──────────────┼──────────────┼──────────────┼────────────┤
  │ C/GPU DISPATCH LAYER                │              │              │              │            │
  ├─────────────────────────────────────┼──────────────┼──────────────┼──────────────┼────────────┤
  │   CUDA driver API call              │    ~25.2 μs  │    ~25.0 μs  │     -0.2 μs  │     -1%    │
  │   GPU kernel dispatch               │  (included)  │  (included)  │       ---    │     ---    │
  ├─────────────────────────────────────┼──────────────┼──────────────┼──────────────┼────────────┤
  │ TOTAL END-TO-END                    │              │              │              │            │
  ├─────────────────────────────────────┼──────────────┼──────────────┼──────────────┼────────────┤
  │   Median launch time                │    27.23 μs  │    39.18 μs  │   +11.95 μs  │   +43.9%   │
  │   Minimum launch time               │    23.70 μs  │    30.15 μs  │    +6.45 μs  │   +27.2%   │
  │   P95 launch time                   │    74.26 μs  │    57.90 μs  │   -16.36 μs  │   -22.0%   │
  └─────────────────────────────────────┴──────────────┴──────────────┴──────────────┴────────────┘
```  
```  
    **🔬 Bottom Line**
  The entire 11.95 μs overhead is now fully explained and measured:
  • 92% from Kernel._from_obj() (11.0 μs)
  • 19% from Stream.from_handle() (2.7 μs)
  • 3% from LaunchConfig (0.4 μs)
  • 1% from ObjectCode stub (0.1 μs)
```

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2434,39 +2467,51 @@ def launch_kernel(
...
+    elif isinstance(hstream, binding.CUstream):
+        hstream = get_cuda_native_handle(hstream)
+    elif not isinstance(hstream, int):
+        hstream = 0
+    stream = ExperimentalStream.from_handle(hstream)
+
+    # Configure kernel launch parameters
+    config = LaunchConfig(
+        grid=(int(gx), int(gy), int(gz)),
+        block=(int(bx), int(by), int(bz)),
```
leofang: This would be a @gmarkall question: Is `driver.launch_kernel` a public API? If not, is it possible that we do not unpack grid/block at the call sites? `cuda.core.LaunchConfig` prefers the 3-tuples.

In the meanwhile... 
```suggestion
        grid=(gx, gy, gz),
        block=(bx, by, bz),
```

leofang: A minor trick is to cache a `LaunchConfig` instance in the dispatcher so that we don't recreate it at every launch.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2434,39 +2467,51 @@ def launch_kernel(
...
-            gy,
-            gz,
-            bx,
-            by,
-            bz,
-            sharedmem,
-            hstream,
-            params_for_launch,
-        )
+    # Convert stream handle to cuda.core Stream object
```
leofang: This seems odd because we just unpacked it at the call site, ex:
https://github.com/NVIDIA/numba-cuda/blob/9212523b2e09f587e151a496bd1cf87f234f56c9/numba_cuda/numba/cuda/dispatcher.py#L478

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2434,39 +2467,51 @@ def launch_kernel(
...
+        hstream = get_cuda_native_handle(hstream)
+    elif not isinstance(hstream, int):
+        hstream = 0
+    stream = ExperimentalStream.from_handle(hstream)
+
+    # Configure kernel launch parameters
+    config = LaunchConfig(
+        grid=(int(gx), int(gy), int(gz)),
+        block=(int(bx), int(by), int(bz)),
+        shmem_size=int(sharedmem) if sharedmem is not None else 0,
```
leofang:
```suggestion
        shmem_size=sharedmem,
```

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2434,39 +2467,51 @@ def launch_kernel(
...
-            hstream,
-            params_for_launch,
-            extra,
-        )
+        cufunction = binding.CUfunction(int(cufunc_handle))
+
+    # Convert arguments: cuda.core.launch expects numpy.uint64 for pointers,
+    # but can handle ctypes scalars directly
+    converted_args = [
+        np.uint64(arg.value if arg.value is not None else 0)
```
leofang:
```suggestion
        (arg.value if arg.value is not None else 0)
```
it'd be nice for the launch dispatcher in cuda.core to support c_void_p.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2434,39 +2467,51 @@ def launch_kernel(
...
+    ]
+
+    # Create Kernel object and launch
+    # We use Kernel._from_obj with an ObjectCode stub since we're wrapping
+    # an existing CUfunction handle rather than creating from object code.
+    from cuda.core.experimental._module import _lazy_init
+
+    _lazy_init()
+    objcode_stub = object.__new__(ObjectCode)
+    kernel = Kernel._from_obj(cufunction, objcode_stub)
```
leofang: I think this happens because again we unbox somewhere in the call site, and then box it back. Same question as above: If `launch_kernel` is not a public API, we should break it to avoid unboxing/boxing.

numba_cuda/numba/cuda/cudadrv/driver.py
gmarkall: It's not a public API. I was going to suggest we should be getting rid of it as part of this PR and directly using `cuda.core.experimental.launch` in dispatcher.py.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2320,6 +2323,36 @@ def get_global_symbol(self, name):
 )
 
 
+def _convert_cuda_core_error(e):
```
gmarkall: I consider `CudaAPIError` a part of the implementation of our ctypes binding to the driver, so we should be looking at just using the `cuda.core` CUDAError rather than converting it to a `CudaAPIError`. This will likely require the update of some test files.

I don't think we'll be able to get rid of `CudaAPIError` in this PR (there are some other uses of it) but we should definitely end up with fewer uses of it.

gmarkall:
> What is the perf number before/after this change?

It would be good to run the launch benchmarks to quantify this change (edit `testing/pytest.ini` to remove `--benchmark-disable`) then run `pytest --benchmark-only`. You should get something like:

```
------------------------------------------------------------------------------------------------- benchmark: 4 tests ------------------------------------------------------------------------------------------------
Name (time in us)                        Min                    Max                   Mean             StdDev                 Median                IQR            Outliers         OPS            Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_one_arg[device_array]          749.1760 (1.0)         998.8480 (1.0)         769.6751 (1.0)      17.1446 (1.0)         765.7590 (1.0)       8.1370 (1.0)        83;102  1,299.2495 (1.0)         943           1
test_one_arg[cupy]                1,896.8900 (2.53)      2,359.8850 (2.36)      1,946.2075 (2.53)     41.6876 (2.43)      1,937.1740 (2.53)     23.5080 (2.89)        36;34    513.8198 (0.40)        444           1
test_many_args[device_array]      6,195.0860 (8.27)      6,472.5740 (6.48)      6,272.8671 (8.15)     44.8138 (2.61)      6,264.1275 (8.18)     50.9895 (6.27)         40;6    159.4167 (0.12)        156           1
test_many_args[cupy]             27,561.5590 (36.79)    27,951.6780 (27.98)    27,702.3494 (35.99)    80.1222 (4.67)     27,687.4050 (36.16)    96.2945 (11.83)         7;1     36.0980 (0.03)         37           1
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

numba_cuda/numba/cuda/cudadrv/driver.py
rparolin: I avoided breaking API in this change because I wanted to ensure that simply routing to cuda.core.launch didn't breaking any fundamental behaviour.  Then we could follow up with a PR that cleans up the boilerplate by breaking API.  Would you prefer to do that all in one PR?  @leofang @gmarkall

numba_cuda/numba/cuda/cudadrv/driver.py
rparolin: Ditto above. I wasn't intending to break API in this change. I want to swap out the driver call for the cuda.core.launch API and ensure everything was working before breaking API.  I think its the safer way to go unless  there is a major performance regression that we can't workaround.

rparolin:
> > What is the perf number before/after this change?
> 
> It would be good to run the launch benchmarks to quantify this change (edit `testing/pytest.ini` to remove `--benchmark-disable`) then run `pytest --benchmark-only`. You should get something like:
> 
> ```
> ------------------------------------------------------------------------------------------------- benchmark: 4 tests ------------------------------------------------------------------------------------------------
> Name (time in us)                        Min                    Max                   Mean             StdDev                 Median                IQR            Outliers         OPS            Rounds  Iterations
> ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
> test_one_arg[device_array]          749.1760 (1.0)         998.8480 (1.0)         769.6751 (1.0)      17.1446 (1.0)         765.7590 (1.0)       8.1370 (1.0)        83;102  1,299.2495 (1.0)         943           1
> test_one_arg[cupy]                1,896.8900 (2.53)      2,359.8850 (2.36)      1,946.2075 (2.53)     41.6876 (2.43)      1,937.1740 (2.53)     23.5080 (2.89)        36;34    513.8198 (0.40)        444           1
> test_many_args[device_array]      6,195.0860 (8.27)      6,472.5740 (6.48)      6,272.8671 (8.15)     44.8138 (2.61)      6,264.1275 (8.18)     50.9895 (6.27)         40;6    159.4167 (0.12)        156           1
> test_many_args[cupy]             27,561.5590 (36.79)    27,951.6780 (27.98)    27,702.3494 (35.99)    80.1222 (4.67)     27,687.4050 (36.16)    96.2945 (11.83)         7;1     36.0980 (0.03)         37           1
> ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
> ```

```--------------------------------------------------------------------------------------- benchmark: 2 tests --------------------------------------------------------------------------------------
Name (time in ms)                    Min                Max               Mean            StdDev             Median               IQR            Outliers       OPS            Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_one_arg[device_array]        2.5915 (1.0)       4.6078 (1.0)       3.1092 (1.0)      0.4016 (1.0)       2.9746 (1.0)      0.4342 (1.0)          15;5  321.6312 (1.0)          92           1
test_many_args[device_array]     16.0708 (6.20)     29.9002 (6.49)     20.1777 (6.49)     2.9787 (7.42)     19.5070 (6.56)     3.1564 (7.27)         14;4   49.5596 (0.15)         54           1
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

I'm having issues getting the cupy benchmarks to run successfully but here is what I can report now.

numba_cuda/numba/cuda/cudadrv/driver.py
rparolin: I migrated away from `CudaAPIError` exception in favor of the cuda.core `CUDAError`

numba_cuda/numba/cuda/cudadrv/driver.py
rparolin: Removed.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2437,54 +2373,67 @@ def read_func_attr_all(self):
...
+    kernel = Kernel._from_obj(cufunction, objcode_stub)
+    _kernel_cache[key] = kernel
+    return kernel
+
+
+def _core_launch(stream, config, cufunction, *args):
+    """
+    Internal helper to launch via cuda.core.experimental.launch ensuring a Kernel object.
+    """
+    kernel = _get_or_create_kernel(_to_cufunction(cufunction))
```
gmarkall: I think cufunction should always be a `binding.CUfunction` at this point - if it is not, we should look into what changes are needed to ensure that is true.

cpcloud: That is correct, `_to_cufunction` isn't needed anywhere.

cpcloud:
@greptile-apps That seems like the wrong tradeoff to make at the current moment. `WeakValueDictionary` and cache invalidation are important but we don't know if the unbounded growth actually matters, so taking on the complexity of either of those solutions doesn't seem worth it right now.

numba_cuda/numba/cuda/tests/cudadrv/test_cuda_driver.py
```diff
@@ -177,18 +165,13 @@ def auto_synchronize(stream):
...
-                0,  # dynamic shared mem
-                stream.handle,  # stream
-                [ptr],
+            config = LaunchConfig(
+                grid=(1, 1, 1),
+                block=(100, 1, 1),
+                shmem_size=0,
+                cooperative_launch=False,
             )
+            drvmod._core_launch(stream, config, function.handle, ptr)
```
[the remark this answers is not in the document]
cpcloud: Does that matter? It's a test case, and if it's passing then it seems fine to leave it as is.

cpcloud:
Benchmarks seem to show little to no effect. Run with `pixi run bench-against upstream/main HEAD`:

image

numba_cuda/numba/cuda/cudadrv/driver.py
cpcloud: Let's avoid breaking for now. We can merge this PR async with adding an API like `Kernel.from_cufunc` (or similar), so that this PR isn't blocked by that.

cpcloud: Hm, that said, importing a private function from a private module doesn't give me warm fuzzies.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2437,54 +2357,57 @@ def read_func_attr_all(self):
...
+
+
+def _get_or_create_kernel(cufunction):
+    key = int(cufunction)
+    kernel = _kernel_cache.get(key)
+    if kernel is not None:
+        return kernel
+    _lazy_init()
+    objcode_stub = object.__new__(ObjectCode)
+    kernel = Kernel._from_obj(cufunction, objcode_stub)
```
cpcloud: @leofang What API should we implement for this use case, in `cuda.core`?

With my `cuda.core` **user** hat on I'd want `Kernel.from_cufunc(cufunction)` (I don't care about naming here).

cpcloud: ... or maybe `from_pointer`/`from_ptr`?

gmarkall: Ideally we would create it starting from a cubin in the `CUDACodeLibrary` object. In `get_cufunc()`

https://github.com/NVIDIA/numba-cuda/blob/bdb4c73e10b07868915ad30f3369a12545e3e1b8/numba_cuda/numba/cuda/codegen.py#L337-L342

instead of using `ctx.create_module_image()` and `module.get_function()`, we should be using [`Objectcode.from_kernel()`](https://nvidia.github.io/cuda-python/cuda-core/latest/generated/cuda.core.experimental.ObjectCode.html) and then using `get_kernel()` to get the kernel we'd launch.

However, I think cuda.core is not a drop-in replacement because modules in Numba-CUDA support module setup and teardown functions (from #145), so the use of `create_module_image` that gets replaced here would also have to have a way to handle module setup and teardown functions.

cpcloud: I _think_ I've got a handle on this, and am close to getting this working as is without any modifications to `cuda.core`. Will report back on progress.

numba_cuda/numba/cuda/cudadrv/driver.py
cpcloud: I was able to remove the hack entirely, as well as the kernel launch helper functions.

The main downside is that everything internally is now dealing with `CUlibrary` instead of `CUmodule` and I don't know whether that detail somehow leaks into the public API.

numba_cuda/numba/cuda/cudadrv/driver.py
cpcloud: The latest commit uses existing `ObjectCode` APIs and continues to support setup and teardown functions, albeit on `CUlibrary` objects instead of `CUmodule` objects.

cpcloud:
Now I am seeing larger performance hits on existing benchmarks (I think I might've been comparison the same commit above):

image

cpcloud:
As of the latest commit, I got everything under 2x by removing some of the helper functions that were doing tons of instance checking (prepare_args for example). The remaining overhead appears to be mostly coming from LaunchConfig construction (which is cythonized on `main`) and from the `launch` function itself.

Here are benchmarks from `upstream/main` against this branch:

image

cpcloud:
@greptile-apps That case is handled by `int(value)`, via the `__int__` protocol.

cpcloud:
@gmarkall @leofang What is an acceptable level of performance regression here? I'd rather this PR not linger on until we achieve parity, but we'll need to adjust timelines if we have to achieve exact performance parity before merging this.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -57,12 +59,7 @@
...
-from cuda.core.experimental import (
-    Linker,
-    LinkerOptions,
-    ObjectCode,
-)
-
+from cuda.core.experimental import Linker, LinkerOptions, ObjectCode
 from cuda.bindings.utils import get_cuda_native_handle
 from cuda.core.experimental import (
     Stream as ExperimentalStream,
```
isVoid, line 68: Not sure if already too late and don't meant to bike shedding too much on this, the name `ExperimentalStream` sounds a bit non-intuitive from the first look. For context because there are separate stream object in cuda core and Numba-CUDA, I too also have to sometime rename `cuda.core.experimental.Stream` to something like `CudaPythonStream` or use the qualified name.

cpcloud:
As of the latest commit, the benchmark results against `upstream/main` are:

image

Personally, I think we should just take the one-argument-case 30-40% hit for now and move on.

cpcloud:
I benchmarked using `main` of `cuda-core` and didn't see a meaningful difference in benchmark numbers.

numba_cuda/numba/cuda/tests/cudadrv/test_module_callbacks.py
```diff
@@ -185,7 +185,7 @@ def setUp(self):
 
         def set_forty_two(handle):
             # Initialize 42 to global variable `num`
-            res, dptr, size = cuModuleGetGlobal(
+            res, dptr, size = cuLibraryGetGlobal(
```
isVoid, line 187: We'll probably have to update the callback signature in the documentation as well:
https://nvidia.github.io/numba-cuda/user/cuda_ffi.html#linking-and-calling-functions

cpcloud:
Note that the single argument case is worse, but the multiple-argument case is better. I'm not sure why. I spent most of the day profiling so far, and basically the only thing I came up with was that there's a bit of additional overhead to all the checking and conversion of stream, config and args in `launch`, but it doesn't seem to capture the full 30-50% difference that shows up on any given benchmarking run.

isVoid (approved):
Thanks for the ping. The cumodule->culibrary change lgtm. The impact downstream is that nvshmem4py will need to update the init/fini method in the callback functions from cumodule_init to culibrary_init:
https://github.com/NVIDIA/nvshmem/blob/9cc869bc28e565e6944c4ddf76976ada4a1ebbf7/nvshmem4py/nvshmem/bindings/nvshmem.pyx#L1609-L1614

cc @benhg

Besides we just need to update the documentation to reflect the change accordingly.

[UPDATE] we should experiment if passing in `ObjectCode` class to callback function is also an option.

numba_cuda/numba/cuda/cudadrv/driver.py
cpcloud: This is an internal use, so I'm not sure it matters that much, though I agree that it's maybe not that intuitive. I'd rather just leave it as is for now.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2434,39 +2467,51 @@ def launch_kernel(
...
+    ]
+
+    # Create Kernel object and launch
+    # We use Kernel._from_obj with an ObjectCode stub since we're wrapping
+    # an existing CUfunction handle rather than creating from object code.
+    from cuda.core.experimental._module import _lazy_init
+
+    _lazy_init()
+    objcode_stub = object.__new__(ObjectCode)
+    kernel = Kernel._from_obj(cufunction, objcode_stub)
```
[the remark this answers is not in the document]
cpcloud: No longer present.

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -540,30 +543,26 @@ def _prepare_args(self, ty, val, stream, retr, kernelargs):
...
+            kernelargs.append(devary.size)
+            kernelargs.append(devary.dtype.itemsize)
+            kernelargs.append(devary.device_ctypes_pointer.value)
+            kernelargs.extend(devary.shape)
+            kernelargs.extend(devary.strides)
 
         elif isinstance(ty, types.CPointer):
             # Pointer arguments should be a pointer-sized integer
-            kernelargs.append(ctypes.c_uint64(val))
+            kernelargs.append(val)
```
cpcloud, line 562: This is okay, because Python `int`s are assumed to be `intptr_t` (the unsigned-less is lost) in `ParamHolder`.

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -1314,37 +1314,43 @@ def unload_module(self, module):
...
+        return Stream(core_stream)
 
     def create_stream(self):
-        # The default stream creation flag, specifying that the created
-        # stream synchronizes with stream 0 (this is different from the
-        # default stream, which we define also as CU_STREAM_DEFAULT when
-        # the NV binding is in use).
+        # Create a cuda.core Stream directly
+        # Note: cuda.core doesn't have direct stream creation from context,
+        # so we use the NVIDIA bindings directly
```
kkraus14: Can we use `self.device` to get a `cuda.core.Device` object and then use `Device.create_stream` instead of manually creating the stream via the driver bindings?

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2046,32 +1973,51 @@ class ManagedOwnedPointer(OwnedPointer, mviewbuf.MemAlloc):
...
+_DEFAULT_STREAM_REPRS = {
+    0: "<Default CUDA stream>",
+    drvapi.CU_STREAM_DEFAULT: "<Default CUDA stream>",
+    drvapi.CU_STREAM_LEGACY: "<Legacy default CUDA stream>",
+    drvapi.CU_STREAM_PER_THREAD: "<Per-thread default CUDA stream>",
+}
+
+
 class Stream:
-    def __init__(self, handle, finalizer=None, external=False):
```
kkraus14, line 2050: This is a breaking change that will definitely be felt and cause downstream issues. I.E. a quick search on GitHub (https://github.com/search?q=%22cudadrv.driver.Stream%28%22&type=code) shows a small number of projects that use this `__init__` signature.

Should we go through a period of deprecation first?

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -109,6 +109,22 @@ def compile(src, name, cc, ltoir=False, lineinfo=False, debug=False):
...
+    #
+    # jit_option = binding.CUjit_option
+    # options = {
+    #     jit_option.CU_JIT_INFO_LOG_BUFFER: jitinfo,
+    #     jit_option.CU_JIT_INFO_LOG_BUFFER_SIZE_BYTES: logsz,
+    #     jit_option.CU_JIT_ERROR_LOG_BUFFER: jiterrors,
+    #     jit_option.CU_JIT_ERROR_LOG_BUFFER_SIZE_BYTES: logsz,
+    #     jit_option.CU_JIT_LOG_VERBOSE: config.CUDA_VERBOSE_JIT_LOG,
+    # }
+    # info_log = jitinfo.decode("utf-8")
```
kkraus14, line 126: Should this be addressed before merging?

numba_cuda/numba/cuda/tests/cudapy/test_dispatcher.py
```diff
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: BSD-2-Clause
 
-from numba.cuda.cudadrv.driver import CudaAPIError
+from cuda.core.experimental._utils.cuda_utils import CUDAError
```
kkraus14, line 4: Is this something that we need to better expose publicly in `cuda.core`?

leofang:
> What is an acceptable level of performance regression here?

This is a question for Graham/Rob/Keith. In this case I would be happy to help on any front but not setting any expectations.

kkraus14:
My 2c and unscientifically:
* Anything less than a 5% performance regression we should just move forward
* With a 5 - 15% performance regression we should analyze and merge if we feel comfortable based on analysis that we can address it in follow up work
* More than a 15% performance regression should probably be addressed before merging

numba_cuda/numba/cuda/cudadrv/nvrtc.py
cpcloud: Dunno. It likely requires some significant changes to `cuda.core` objects and more plumbing. Seeing as how the `info_log` member of `CudaPythonModule` is removed in this PR, and it was never used anywhere, I'd say that we remove it for now to avoid additional things blocking this PR.

cpcloud: cc @leofang

numba_cuda/numba/cuda/cudadrv/driver.py
cpcloud: I think that's out of scope for this PR. It requires dealing with various concepts like default stream, legacy default stream, and per thread default stream in ways that will further stall this PR.

cpcloud: IOW, it's better to just leave this as is and make the move to `Device.create_stream` in a follow-up so that those changes can be properly reviewed without further bogging down this PR.

numba_cuda/numba/cuda/cudadrv/driver.py
cpcloud: Maybe we should just leave this class entirely as is and convert to `cuda.core` Streams on launch. Then this PR will be much easier to review.

cpcloud:
> My 2c and unscientifically:
> 
>     * Anything less than a 5% performance regression we should just move forward
> 
>     * With a 5 - 15% performance regression we should analyze and merge if we feel comfortable based on analysis that we can address it in follow up work
> 
>     * More than a 15% performance regression should probably be addressed before merging

What about in this case where two benchmarks are ~50% faster and two are ~50% slower?

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2440,76 +2288,14 @@ def set_shared_memory_carveout(self, carveout):
...
-            sharedmem,
-            hstream,
-            params_for_launch,
-            extra,
-        )
 
+# Alias for backward compatibility
+CudaPythonFunction = Function
 
-class _LinkerBase(metaclass=ABCMeta):
```
cpcloud, line 2510: I'll remove these in a follow-up to reduce total changes.

numba_cuda/numba/cuda/tests/cudapy/test_dispatcher.py
cpcloud: Yes, but I don't think that should block this PR.

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -598,8 +593,7 @@ def _prepare_args(self, ty, val, stream, retr, kernelargs):
 
         elif isinstance(ty, types.Record):
             devrec = wrap_arg(val).to_device(retr, stream)
-            ptr = devrec.device_ctypes_pointer
-            kernelargs.append(ptr)
+            kernelargs.append(devrec.device_ctypes_pointer.value)
```
cpcloud, line 596: It's actually slightly faster to pull out the `int` in Python and let `ParamHolder` do its usual treatment of `int` -> `intptr_t`.

numba_cuda/numba/cuda/cudadrv/driver.py
cpcloud: I reverted this change.

cpcloud:
Running the following profiling code:

```python
import sys
from numba import cuda
import numpy as np


@cuda.jit
def one_arg(arr1):
    return


def bench(func, arr):
    for _ in range(100_000):
        func(arr)


func = one_arg[128, 128]
arr = cuda.device_array(128, dtype=np.float32)


if __name__ == "__main__":
    import pyinstrument

    pi = pyinstrument.Profiler()
    for _ in range(3):
        func(arr)
    pi.start()
    bench(func, arr)
    pi.stop()
    print(pi.output_html(), file=sys.stdout)
    pi.print(file=sys.stderr)

```

with different numbers of iterations--`1_000`, `10_000`, `100_000`--the ratio of the slowdown is the same, which strongly suggests that there's a fixed overhead to the changes here. These are also the kinds of performance differences that IME are the toughest to squeeze out, because they are often from the accumulation of a bunch of benign, individually-fast operations and function calls, and thus are typically hard to isolate.

numba_cuda/numba/cuda/tests/cudapy/test_dispatcher.py
kkraus14: 100%, just wanted to make sure that my understanding is correct in that there's a gap we need to address

cpcloud:
I modified the benchmarks to run with different numbers of iterations to make it clear that the overhead is fixed and doesn't depend on the number of iterations (e.g., a memory leak might manifest as slowdown over larger and larger numbers of iterations).

image

cpcloud:
The other thing to note is that for "many" arguments, there's a performance **improvement**:

image

kkraus14:
> What about in this case where two benchmarks are ~50% faster and two are ~50% slower?

Unless the ~50% slower use cases are considered niche, I don't think we want to eat that performance regression.

Based on your above benchmarks, I don't think we can take a ~50% regression in the case of launching a kernel against a devicearray object with one arg.

cpcloud:
After some more investigation, it looks like `_to_core_stream` is the main source of the bottleneck.

Is caching the numba Stream -> cuda core Stream conversion a viable option?

I will also look into whether we can move the conversion higher up in the loop.

cpcloud:
It looks like streams are already being cached as part of the `CUDADispatcher.configure` call, so it seems like it's okay to do so (or the existing code is incorrect).

cpcloud:
Ok, moving the core stream conversion into the `_LaunchConfiguration` constructor shaves off most of the difference now.

cpcloud:
Naturally, that breaks pickling `jit` 🥳

cpcloud:
Pushed up a possible implementation of `__getstate__`/`__setstate__` for `_LaunchConfiguration`.

cpcloud:
New single argument benchmarks:

image

cpcloud:
The variance seems to be rather large between runs:

image

Note that in this case there's actually a negligible *improvement*.

cpcloud:
This seems within noise to me, since I can't reproduce a consistent slowdown between 3 and 15%:

image

cpcloud:
One thing that also may have gotten lost in my sea of comments is that there'll be some slight improvements from the next release of `cuda.core` due to Cythonization of some of the launch configuraion stuff.

rparolin:
Just to confirm, we aren't able to reliably reproduce the 30%+ performance regression once we eliminated the overhead by the stream creation helper?

cpcloud:
That is correct. There's still overhead to it, but it's not being invoked on every single launch call. It happens when you slice into the function to get the grid and block dimensions.

Before:

```python
arr = cuda.device_array(...)

func = one_arg[1, 1]

for _ in range(HUGE_NUMBER):
    func(arr) # _to_core_stream invoked here
```

After:

```python
arr = cuda.device_array(...)

func = one_arg[1, 1]  # _to_core_stream invoked here

for _ in range(HUGE_NUMBER):
    func(arr)
```

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -700,6 +694,16 @@ def __call__(self, *args):
...
 
+    def __getstate__(self):
+        state = self.__dict__.copy()
+        state["stream"] = int(state["stream"].handle)
+        return state
+
+    def __setstate__(self, state):
+        handle = state.pop("stream")
+        self.__dict__.update(state)
+        self.stream = driver._to_core_stream(handle)
```
kkraus14, line 705: How is the current `Stream` class handled for pickling? Capturing the `handle` which is a ptr in a pickle feels wrong.

cpcloud: Yes, pickling a pointer is indeed weird, however, that is the status quo, as `Stream` uses the default pickling implementation and its fields are `handle` (stream handle), and `external` (bool).

rparolin: I was chatting with @cpcloud about this and yeah, serializing a stream class handle feels weird to me.  What does that even mean to deserialize a Stream handle? I think we are probably getting lucky that the majority of handles are 0 representing the default stream.

cpcloud: I will create an issue to track this.

numba_cuda/numba/cuda/cudadrv/nvrtc.py
gmarkall: I suspect there were very few users of `info_log`, with me being the primary one - I used it when debugging issues only. It's not high priority to bring back unless someone asks for it, IMHO.

PR #611, opened 2025-11-26
Revert #536 "perf: remove context threading in various pointer abstractions"

gmarkall:
This reverts commit 9a56516cdaf5fe216a3db3db896974e084b4fa23.

This changed the public API of `MemoryPointer` and related classes, and the context that they held was used by Arrow (see https://github.com/apache/arrow/pull/48259#issuecomment-3580532978):

> Numba interop tests fail with:

```
arrow-dev/lib/python3.12/site-packages/pyarrow/tests/test_cuda_numba_interop.py:233:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   TypeError: MemoryPointer.__init__() got multiple values for argument 'pointer'
```

This commit reverts the change, as it was intended to improve performance without changing functionality, but has had a functional change as a side effect. Following the merge of this PR, we should be able to remove some of the `@require_context` decorators with some more targeted changes.

### DIFF
--- a/numba_cuda/numba/cuda/api.py
+++ b/numba_cuda/numba/cuda/api.py
@@ -21,6 +21,7 @@ current_context = devices.get_context
 gpus = devices.gpus
 
 
+@require_context
 def from_cuda_array_interface(desc, owner=None, sync=True):
     """Create a DeviceNDArray from a cuda-array-interface description.
     The ``owner`` is the owner of the underlying memory.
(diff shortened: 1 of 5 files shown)


gmarkall:
Following this PR: create an issue to track the work of re-removing `@require_context` and `get_context()` from `from_cuda_array_interface()`.

Issue #483, opened 2025-09-19
Pass NULL pointer for cffi.FFI().from_buffer(empty array)


ed-o-saurus:
This is a transfer of bug https://github.com/numba/numba/issues/9654. 

In short, empty arrays, when copied to a device, are not recognized as contiguous. Therefore, a pointer to the array cannot be passed.

gmarkall:
@kaeun97 Would you have any interest in this issue?

kaeun97:
Absolutely. On it!

PR #612, opened 2025-11-27
fix: empty array type mismatch between host and device

kaeun97:
Fixes https://github.com/NVIDIA/numba-cuda/issues/483. 

As mentioned [here](https://github.com/numba/numba/issues/9654#issuecomment-2230960793), the bug was due to:
- empty arrays incorrectly identified as broadcast arrays
- incorrect contiguity flags for empty arrays

When running this:
```
from numba import cuda, typeof
import numpy as np

h_values = np.random.randint(low=0, high=2, size=(10, 0))
d_values = cuda.to_device(h_values)

print(typeof(h_values))
print(typeof(d_values))

assert typeof(h_values) == typeof(d_values)
```

Below is the output:
```
array(int64, 2d, C)
array(int64, 2d, C)
```

### DIFF
--- a/numba_cuda/numba/cuda/tests/nocuda/test_dummyarray.py
+++ b/numba_cuda/numba/cuda/tests/nocuda/test_dummyarray.py
@@ -387,5 +387,66 @@ class TestIterate(unittest.TestCase):
             x = val  # noqa: F841
 
 
+@skip_on_cudasim("Tests internals of the CUDA driver device array")
+class TestEmptyArrays(unittest.TestCase):
+    def test_empty_array_flags(self):
+        test_shapes = [
+            (0,),
+            (10, 0),
+            (0, 10),
+            (0, 0),
+            (5, 0, 3),
+            (0, 5, 3),
+            (5, 3, 0),
+            (0, 0, 0),
+        ]
+        for shape in test_shapes:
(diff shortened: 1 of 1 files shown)


numba_cuda/numba/cuda/cudadrv/dummyarray.py
```diff
@@ -279,6 +279,10 @@ def _compute_layout(self):
         if not self.dims:
             return {"C_CONTIGUOUS": True, "F_CONTIGUOUS": True}
 
+        # All 0-size arrays are considered contiguous, even if they are multidimensional
+        if self.size == 0:
+            return {"C_CONTIGUOUS": True, "F_CONTIGUOUS": True}
+
         # If this is a broadcast array then it is not contiguous
         if any([dim.stride == 0 for dim in self.dims]):
```
gmarkall, line 287: I noted that the NumPy implementation that this is following (from [the `_UpdateContiguousFlags` implementation referenced above](https://github.com/numpy/numpy/blob/maintenance/1.19.x/numpy/core/src/multiarray/flagsobject.c#L123-L191)) doesn't have this check for zero strides.

It's not directly related to this PR, but I think it is suspicious that we still differ in our implementation.

Issue #617, opened 2025-11-28
[FEA] Support for L1 cache / shared memory configuration


gmarkall:
**Is your feature request related to a problem? Please describe.**

Numba-CUDA did have some unpolished support for setting the [shared memory configuration](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html?highlight=cudafuncsetcacheconfig#shared-memory-7-x) for a kernel. This would set the cache configuration using [`cuFuncSetCacheConfig`](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__EXEC.html#group__CUDA__EXEC_1g40f8c11e81def95dc0072a375f965681). 

This was accessed like:

```python
from numba import cuda, types

sig = (types.float32[::1], types.float32[::1])

@cuda.jit(sig)
def add_one(r, x):
    i = cuda.grid(1)
    if i < len(r):
        r[i] = x[i] + 1

kernel = add_one.overloads[sig]
cufunc = kernel._codelibrary.get_cufunc()
cufunc.cache_config(prefer_shared=True)
```

However, this blows up now with:

```
Traceback (most recent call last):
  File "/home/gmarkall/numbadev/issues/discourse-3080/repro.py", line 13, in <module>
    cufunc.cache_config(prefer_shared=True)
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/cudadrv/driver.py", line 2418, in cache_config
    flag = attr.CU_FUNC_CACHE_PREFER_SHARED
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: type object 'CUfunction_attribute' has no attribute 'CU_FUNC_CACHE_PREFER_SHARED'
```

**Describe the solution you'd like**

There are a number of improvements we can make here:

- Fix the functionality and add testing so that the example above works, and continues to work
- Update the implementation so that it uses the `cuFuncSetAttribute` mechanism to have greater control over the carveout
- Improve the API so that one doesn't need to have a pre-compiled kernel and then retrieve the underlying function to set the cache configuration. Instead, a new argument to the `@cuda.jit` decorator that ensures the required configuration is applied for all compilations of the kernel would be much more user-friendly.

An example of what usage of the improved API could look like:

```python
from numba import cuda

@cuda.jit(shared_memory_carveout=50)
def add_one(r, x):
    i = cuda.grid(1)
    if i < len(r):
        r[i] = x[i] + 1
[...]

PR #628, opened 2025-12-03
fix: cache config enum

kaeun97:
This will be followed by a sequence of work on improving the l1 cache/shared memory configuration implementation as stated [here](https://github.com/NVIDIA/numba-cuda/issues/617#issuecomment-3599322414).

This PR fixes the following bug:
```
  File "/home/gmarkall/numbadev/issues/discourse-3080/repro.py", line 13, in <module>
    cufunc.cache_config(prefer_shared=True)
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/cudadrv/driver.py", line 2418, in cache_config
    flag = attr.CU_FUNC_CACHE_PREFER_SHARED
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: type object 'CUfunction_attribute' has no attribute 'CU_FUNC_CACHE_PREFER_SHARED'
``` 

with a test that would ensure that this won't regress.


gmarkall (approved):
Many thanks! I've checked with NSight Compute that setting the attributes takes effect.

PR #629, opened 2025-12-03
feat: add set_shared_memory_carveout

kaeun97:
This adds `set_shared_memory_carveout` using `cuFuncSetAttribute` requested [here](https://github.com/NVIDIA/numba-cuda/issues/617). 

Do want to confirm with @gmarkall about this design (and would be nice to chat about it). An alternative is to combine this with`cache_config` in some way. However, I do think separating the two makes more sense mainly because it makes the intent clearer for the user (cache_config sets preference hints while set_shared_memory_carveout sets explicit percentages).


gmarkall:
Thanks for the PR! I'll circle back to this when I get a chance (hopefully soon) and come back to you. Happy to chat about the design as well.

numba_cuda/numba/cuda/tests/cudadrv/test_cuda_driver.py
```diff
@@ -285,6 +285,58 @@ def b2d(bs):
...
+        def add_one(r, x):
+            i = cuda.grid(1)
+            if i < len(r):
+                r[i] = x[i] + 1
+
+        kernel = add_one.overloads[sig]
+        cufunc = kernel._codelibrary.get_cufunc()
+
+        # valid carveout values
+        carveout_values = [0, 50, 100]
```
gmarkall: Is it worth testing with the default as well?


```suggestion
        carveout_values = [-1, 0, 50, 100]
```

gmarkall (approved):
> Do want to confirm with @gmarkall about this design (and would be nice to chat about it).

I think the design for the part of the problem that this PR solves looks great - I can think of no improvement.

> An alternative is to combine this with`cache_config` in some way. However, I do think separating the two makes more sense mainly because it makes the intent clearer for the user (cache_config sets preference hints while set_shared_memory_carveout sets explicit percentages).

I agree, keeping it separate makes sense to me too.

I'm happy to chat if you have unresolved questions, but I'm also happy for this to be merged - let me know how you'd like to proceed.

kaeun97:
@gmarkall Thank you for the review! Good to hear that the design makes sense. Just wanted to make sure that I am interpreting the description from the issue properly :) Would be nice to re-run CI and merge this.

gmarkall:
> ### Additional Comments (1)
> 
>     1. `numba_cuda/numba/cuda/cudadrv/driver.py`, line 2400 ([link](/nvidia/numba-cuda/blob/683e774faebe7254a870818c648c556f24554cf5/numba_cuda/numba/cuda/cudadrv/driver.py#L2400))
>        **logic:** `cuFuncSetAttribute` is not defined in `API_PROTOTYPES` in `drvapi.py`. If `CtypesFunction` is ever used, this will fail with `CudaDriverError: Driver missing function: cuFuncSetAttribute`. While the default code path uses `CudaPythonFunction`, the `CtypesFunction` path should either be removed or have the API prototype added to `drvapi.py`:
> 
> 
> 2 files reviewed, 1 comment
> 
> [Edit Code Review Agent Settings](https://app.greptile.com/review/github) | [Greptile](https://greptile.com?utm_source=greptile_expert&utm_medium=github&utm_campaign=code_reviews&utm_content=numba-cuda_629)

This is dead code, so it won't be hit. We need to remove it at some point.

gmarkall:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/629/commits/683e774faebe7254a870818c648c556f24554cf5

gmarkall:
Many thanks @kaeun97!

PR #632, opened 2025-12-03
ci: pin pixi in ci due to lockfile mismatch in 0.60.0

cpcloud:
Pin pixi to 0.59.0 due to https://github.com/prefix-dev/pixi/issues/5055.


cpcloud:
I'll follow up with a PR to bump the lockfile using 0.60.0

PR #634, opened 2025-12-04
Always use `core` errors if `numba` is present

brandon-b-miller:
This PR fixes an issue where `numba-cuda` may raise a `numba.cuda.core.errors.TypingError` in some situations where typing fails, and a `numba.core.errors.TypingError` in others. In this PR we take the approach of always using `numba.core.errors` in our codebase if numba is present, else use only our own. 

This PR also adds some missing cases of test coverage from the cuDF test suite that exposed this issue when performing the numba upgrade. These tests rely on catching the correct type of `TypingError` which was found to vary.

### DIFF
--- a/greptile.json
+++ b/greptile.json
@@ -1,0 +1,7 @@
+{
+  "sequenceDiagramSection": {
+    "included": false,
+    "collapsible": false,
+    "defaultOpen": false
+  }
+}
--- a/pixi.lock
+++ b/pixi.lock
@@ -38,13 +38,13 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/openssl-3.6.0-h26f9b46_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/packaging-25.0-pyh29332c3_1.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/platformdirs-4.5.0-pyhcf101f3_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/pluggy-1.6.0-pyhd8ed1ab_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/pluggy-1.6.0-pyhf9edf01_1.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/pre-commit-4.5.0-pyha770c72_0.conda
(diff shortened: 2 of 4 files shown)


gmarkall:
This needs a merge from `main`.

gmarkall:
> This needs a merge from `main`.

(After #635 is merged)

brandon-b-miller:
> I'm currently trying to understand the rationale behind the mechanism added here instead of using `redirect_numba_module` as is used for other modules. Can you share any thoughts here?

There's no reason besides having needed to spell it out to make sense of it when implementing things, I've switched it back now.

numba_cuda/numba/cuda/core/cuda_errors.py
```diff
@@ -0,0 +1,917 @@
...
+        def highlight(self, msg):
+            return self._markup(msg, self._highlight)
+
+        def reset(self, msg):
+            return self._markup(msg, self._reset)
+
+    def termcolor():
+        global _termcolor_inst
+        if _termcolor_inst is None:
+            scheme = themes[numba.core.config.COLOR_SCHEME]
```
gmarkall:
```suggestion
            scheme = themes[numba.cuda.core.config.COLOR_SCHEME]
```

numba_cuda/numba/cuda/core/cuda_errors.py
```diff
@@ -0,0 +1,917 @@
...
+        self._warnings.clear()
+
+    def __enter__(self):
+        return
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.flush()
+
+
+class NumbaError(numba.core.errors.NumbaError):
```
gmarkall: `NumbaError` is a subclass of `Exception` upstream. We made our `NumbaError` inherit from `numba.core.errors.NumbaError` with the intent of avoiding the issue that you're fixing here, but it didn't work... Now that this file is imported in the absence of Numba, we should use `Exception` as the base again.
```suggestion
class NumbaError(Exception):
```

gmarkall (changes requested):
Thanks for switching over to the `redirect_numba_module` method. Now that `cuda_errors.py` is imported in the absence, of Numba, it needs to avoid referencing Numba.

PR #635, opened 2025-12-05
dev: update numba-cuda version

cpcloud:
# Update project version

## Summary

- Bumped numba-cuda version from 0.20.1 to 0.22.0 in pixi.toml

## Changes

- Updated pixi.toml to reflect version 0.22.0s

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -248,7 +248,7 @@ NUMBA_ENABLE_CUDASIM = "1"
 
 [package]
 name = "numba-cuda"
-version = "0.20.1"
+version = "0.22.0"
 
 [package.build]
 backend = { name = "pixi-build-python", version = "*" }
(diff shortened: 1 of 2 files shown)


pixi.toml
```diff
@@ -248,7 +248,7 @@ NUMBA_ENABLE_CUDASIM = "1"
 
 [package]
 name = "numba-cuda"
-version = "0.20.1"
+version = "0.22.0"
```
gmarkall, line 251: Is there a way to get this from `numba_cuda/VERSION` so it doesn't need a separate update?

gmarkall (approved):

pixi.toml
cpcloud: Unfortunately, no. We should just create a version bumping script that does it for all the files we care about containing the correct version.

cpcloud: There are a number of tools that do this quite well.

cpcloud: https://github.com/prefix-dev/pixi/issues/2946 is the upstream pixi issue.

cpcloud:
I will do the dependency version updates in a separate PR, and just include the numba-cuda update here.

PR #636, opened 2025-12-05
Disable Greptile sequence diagrams

gmarkall:


cpcloud:
Eagerly awaiting to see whether greptile decides to respect configuration...

cpcloud:
Failing due to the pixi package version, which is fixed in #635.

gmarkall:
> Hard approve.

Thanks! It seems the target branch may need a greptile.json for it to pull from greptile.json, so I'll merge this and hope they stop.

cpcloud:
@gmarkall I don't think we need to wait for the full test suite to run, since you're just changing a configuration file that doesn't enter into any CI processes.

PR #642, opened 2025-12-08
feat: users can pass `shared_memory_carveout` to @cuda.jit

kaeun97:
This PR requires rebase after merging this: https://github.com/NVIDIA/numba-cuda/pull/629.

This PR allows users to pass a new argument `shared_memory_carveout` to the @cuda.jit decorator as described [here](https://github.com/NVIDIA/numba-cuda/issues/617). 

An alternative to consider is to use `enum` for `cudaSharedCarveout` but I think the current implementation is cleaner.


kaeun97:
@gmarkall I didn't know this was still a draft 😅 would appreciate your review! Also, is there anything in particular that I can look into next? Excited to see what's next!

gmarkall:
@kaeun97 Apologies for the delay in getting to this, it was not because it was a draft that it didn't receive a review, but just that things have been a a bit heavy (and I had a little PTO) over the last few days.

gmarkall:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/642/commits/bbdda616e7b864e53471794797fe0b18a0005493

gmarkall (commented):
Many thanks again for your efforts here!

I've had a quick skim - apologies for the delay. Some initial thoughts:

- The API design looks good to me.
- We should probably test with and without signatures to the `@cuda.jit` decorator, because the paths taken through the `jit` decorator are different when `func_or_sig` is a function and when it is a signature.
- As this is a user-facing feature, we should add it to the docstring for `jit`. I think the documentation is then derived from this docstring, so there shouldn't need to be a separate documentation change.

kaeun97:
Thank you for the quick response and feedback after your PTO (+ hope you got some chance to rest and have fun)! Much appreciated. 

- [x] add docstring for `shared_memory_carveout`
- [x] add test with signature

Note that as I was testing with signatures, I realized that the `shared_memory_carveout` value was checked at different times (previous implementation: (without signature) raise error at compilation time, (with signature) raise at decoration time)  , so made it consistent to check it at decoration time. 

Also, would there be some other issue I can work on after this task?

kaeun97:
Hi @gmarkall, would be much appreciated if we can get this merged :)

gmarkall:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/642/commits/373b7c2169cb662988344fed315aedbb91fc8ca6

gmarkall:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/642/commits/e5ba74879dbf011ebdf00c22d8bf4849991b4f97

gmarkall:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/642/commits/1be8209338ddf5b89241e54c026b880d08970ce1

gmarkall:
@kaeun97 Thanks for this, and apologies for the delay here. Things got a bit backlogged over the last month and the holiday period, and I'm aiming to get a bit more back on track again now.

PR #643, opened 2025-12-09
Fix #624: Accept Numba IR nodes in all places Numba-CUDA IR nodes are expected

### LINKED ISSUE #624: [BUG] [REGRESSION] inline="always" does not work on overload
**Describe the bug**
Since 0.21 this code stopped working

```python
import numpy as np
from numba import cuda
from numba.extending import overload

def get_42():
    raise NotImplementedError()

@overload(get_42, target="cuda", inline="always")
def ol_blas_get_accumulator():
    def impl():
        return 42

    return impl

@cuda.jit
def kernel(a):
    a[0] = get_42()

def main():
    a = np.empty(1, dtype=np.float32)

    d_a = cuda.to_device(a)
    kernel[1, 1](d_a)
    d_a.copy_to_host(a)

    assert a[0] == 42.0

if __name__ == "__main__":
    main()

```

Run the code above with numba-cuda 0.21.3

**Expected behavior**
Suppose to work as in 0.20

 - Environment location: Bare-metal
 - Method of numba-cuda install: pypi

**Additional context**
Originally caught with `overload_method`, but I guess the root of the issue should be at the same place

gmarkall:
This PR attempts to fix Issue #624. The root cause there is that Numba's inline pass is used when `numba.extending.overload(inline="always")` is used, and it creates IR nodes from `numba.core.ir` instead of `numba.cuda.core.ir`, which are not recognised in instance checks in Numba-CUDA.

Rather than redirecting the IR module to Numba, this PR aims to modify all instance checks to accept both Numba and Numba-CUDA IR nodes, to allow for flexibility in editing the `numba.cuda.core.ir` module and the changes co-existing with uses of the `numba.core.ir` module.

These changes are mostly performed by a script. The majority of these were done with:

```python
import re
import pathlib
import sys

path = pathlib.Path(sys.argv[1])
text = path.read_text()

pattern = re.compile(r"(isinstance\([^,]+,\s*)ir\.([A-Z][A-Za-z0-9_]+)")


def repl(m):
    head = m.group(1)
    name = m.group(2).lower()
    return f"{head}ir.{name}_types"


new = pattern.sub(repl, text)
path.write_text(new)
```

Which replaces uses like `isinstance(<thing>, ir.<Classname>)` with `isinstance(<thing>, ir.<classname>_types)`. The `<classname>_types` are initialized based on whether Numba is present - if it is then e.g. `(Arg, numba.core.ir.Arg)` would be in `arg_types`, and similar for all the other node types.

The script above did not change everything - some manual edits were made after those changes, which were detected using the following:

```
# Generate a script to grep for all the uses of classes in ir.py
grep "^class [A-Z]" numba_cuda/numba/cuda/core/ir.py | awk '{print $2}' | awk 'BEGIN { FS="(" } { print "grep --color \"ir\\." $1 "\" numba_cuda -R" }' > audit.sh
# Run the script, ignoring instances where we create a new IR node and assign it, e.g "v = ir.Var(...)"
bash audit.sh | grep -v "= ir\.[A-Z]" | sort
```

The bash script above can also be used to check all the remaining uses of the nodes, to ensure that none that need changing were missed.

A test based on the reproducer in #624 is also added. The test did not work initially - this is because the test case adds an implementation of `get_42` to Numba's builtin registry, which was ignored by the CUDA target. The CUDA target would normally install registrations from the Numba builtin registry, but skips those that appear to be internal, as determined by the module name beginning with `"numba."`. Implementations defined in test code should not be considered internal (they provide a model of the use of Numba-CUDA from outside) so the `is_external()` method is updated to consider implementations in the `numba.cuda.tests` module to be external.


replace_checks.py
kkraus14, line 1: Was this intended to be checked in?

kkraus14:
@gmarkall changes LGTM other than checking in the script to automate the replacement

replace_checks.py
gmarkall: It's not intended to be in the code that gets merged. I did have it checked in whilst I was working on the branch but it needs removing.

gmarkall:
Many thanks @kkraus14!

@ZzEeKkAa I believe this addresses the issue as in the reproducer. Would you be able to give it a test with your real code to ensure that it resolves the real issues for you please?

PR #647, opened 2025-12-10
Add support for dependabot

### LINKED ISSUE #646: Add dependabot workflow
Add dependabot to update action versions. This is more necessary as we move to SHAs for third-party approved actions where version numbers are more opaque.

jpascucci-nv:
Closes #646


cpcloud:
/ok to test

Though not actually necessary.

PR #650, opened 2025-12-10
Fix false negative NRT link decision when NRT was previously toggled on

brandon-b-miller:
This PR fixes an NRT caching bug encountered while running groups of tests for a separate task. 

NRT currently must be toggleable, because otherwise the lack of the proper incref/decref pruning pass from upstream numba makes the test suite walltime intractable. When enabled, numba may inject incref/decref calls into generated code. Later on, numba inspects the code to be linked for these calls to determine if the other half of the NRT library needs to be linked. However the current logic answers "no" if NRT is not _currently_ enabled. 

This may cause linker errors down the line if we get a cache hit and the setting is later toggled off. This may be observed via:

```python

def array_reshape1d(arr, newshape, got):
    y = arr.reshape(newshape)
    for i in range(y.shape[0]):
        got[i] = y[i]



def array_reshape(arr, newshape):
    return arr.reshape(newshape)


config.CUDA_ENABLE_NRT=1
@cuda.jit
def kernel(out):
    out[0] = np.min(np.array([1.0, 2.0, 0.0, -0.0, 1.0, -1.5]))

out = cp.zeros(1, dtype=cp.float64)      
kernel[1, 1](out)
config.CUDA_ENABLE_NRT=0


arr = np.arange(24)
kernel2 = cuda.jit(array_reshape1d)
expected = array_reshape(arr, (24,))
got = np.zeros(expected.shape, dtype=arr.dtype)
kernel2[1, 1](arr, (24,), got)
```
```
cuda.bindings.nvjitlink.nvJitLinkError: ERROR_PTX_COMPILE (4)                                                                                                                                                                                              
Linker error log: ptxas fatal   : Unresolved extern function 'NRT_incref'                                                                                                                                                                                  
ERROR NVJITLINK_ERROR_PTX_COMPILE: JIT the PTX (ltoPtx)
```

This PR fixes the issue by remembering if NRT was ever turned on for this python session and avoiding the short circuit that decides not to link NRT if it's not currently enabled.

### DIFF
--- a/numba_cuda/numba/cuda/tests/nrt/test_nrt.py
+++ b/numba_cuda/numba/cuda/tests/nrt/test_nrt.py
@@ -383,8 +383,6 @@ class TestNrtStatistics(CUDATestCase):
             self.assertEqual(stats.mi_free, stats_mi_free)
 
     def test_nrt_toggle_enabled(self):
-        cuda.dispatcher._Kernel._LINKED_NRT = False
-
         def array_reshape1d(arr, newshape, got):
             y = arr.reshape(newshape)
             for i in range(y.shape[0]):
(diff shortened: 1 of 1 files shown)


gmarkall:
Can the reproducer in the description be turned into a test case?

gmarkall (approved):

PR #656, opened 2025-12-11
test: fix bogus `self` argument to `Context`

cpcloud:
Corrects a test that breaks when you start mucking around with `Context` internals. I am experimenting with replacing the guts of `MemoryPointer` with a `cuda.core` `Buffer` implementation, and this came up.


gmarkall:
Set to auto-merge after CI is done.

cpcloud:
Looks like the `FUTURE_TIMEOUT` is perhaps not large enough to capture variability of timeouts. I'll put up a PR to address that.

PR #604, opened 2025-11-20
Set up a new VM-based CI infrastructure 

leofang:
~~TODO: Rebase and write up descriptions~~

See WIP report here https://github.com/NVIDIA/numba-cuda/pull/604#issuecomment-3564463179 and here https://github.com/NVIDIA/numba-cuda/pull/604#issuecomment-3597994387.

***UPDATE***: This PR brings in a new CI infra that is a clone of what cuda-python uses today. The new CI is fully VM-based instead of container-based, except for
- `cibuildwheel` launching a manylinux container
- we launch a vanilla Ubuntu container for testing due to the requirement of nv-gha-runners

This is desirable because containers are the major bottleneck that we should seriously consider moving away from:
- it takes time to pull on a per-PR basis
- it blocks us from performing Day 1 rollout (to support new CUDA or new Python versions) which is now a requirement for CUDA Python
   - related: it takes nontrivial amount of efforts in maintaining our own containers 

Furthermore, my opinion is that we really need to make sure the CUDA Python CI infrastructure is "copy-paste-able" (with some caveats discussed internally, which I am not going to repeat here). It was designed with future application to CuPy and numba-cuda in mind, since at the Python level lots of our projects have similar/same support matrix, and there is no reason for each project to rebuild the wheel from scratch and suffer from maintenance issues 

Currently this PR is made such that it runs in parallel with the old CI. We can have a separate PR to follow up and hook more old CI pieces with the new CI if we decide to move forward.

Below is a detailed breakdown of what this PR entails.
- c1f1cec2ed0939d0e6780a8199e526885c7799a5: copy/paste minimal CI infra from cuda-python, with zero change
- 85566c7002c3dcf8ac367bc3ba8e9f3bf2aadd4f: CI changes needed to tailor for numba-cuda needs
- 1b939c39cd32948b02e4752bff1ac33da0c87d44: Enable the cibuildwheel GHA
- 950078e090e96f8cf200f232717539193c6944d2: Disable Python 3.14 for now 
   - This shows how easy it is to add support when a new Python version is out. For example, once https://github.com/NVIDIA/numba-cuda/pull/599 is merged we can revert this commit to start testing.
- 6fa44fc76ba61f3cc843ceffe38998814434978c: a drive-by fix to update the warning when `cuda-bindings` is not installed
- 593fcf33ab5c33e37be1c83e5fd4de7340f4bdf8: Ensure Linux executables can be found when installed by our custom `fetch_ctk` action 
[...]

PR #662, opened 2025-12-12
Do not manually set DUMP_ASSEMBLY in `nvjitlink` tests

brandon-b-miller:
Fixes an issue surfaced in https://github.com/NVIDIA/numba-cuda/pull/604. These tests are currently written so a dynamic skip causes the DUMP_ASSEMBLY setting to stick. This PR switches to the recommended setup/teardown approach.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,7 +8,15 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      # Check for updates to GitHub Actions every week
-      interval: "weekly"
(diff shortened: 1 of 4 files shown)


cpcloud (approved):
More reason to move things to fixtures like `monkeypatch`, to avoid polluting global state!

gmarkall (approved):
I think the change improves the setup. I don't see any dynamic skips, but I can see a failure in these tests leaving the `DUMP_ASSEMBLY` flag stuck after a failure.

Issue #659, opened 2025-12-11
[FEA] Support capture of device arrays from globals/closures by reference


shwina:
Currently, capturing device arrays (such as CuPy arrays) as globals is unsupported:

```python
"""
Minimal reproducer: Device array captured from global in device function.

Run with: python test_minimal_failing.py
"""

import numpy as np
import cupy as cp
from numba import cuda

# Device array at global scope
GLOBAL_ARRAY = cp.asarray( [1.0, 2.0, 3.0, 4.0], dtype=np.float32)


@cuda.jit(device=True)
def device_func(idx):
    return GLOBAL_ARRAY[idx]


@cuda.jit
def kernel(output):
    idx = cuda.grid(1)
    if idx < output.shape[0]:
        output[idx] = device_func(idx)


if __name__ == "__main__":
    output = cuda.device_array(4, dtype=np.float32)
    kernel[1, 32](output)
    print(output.copy_to_host())
```

```python
Untyped global name 'GLOBAL_ARRAY': Cannot determine Numba type of <class 'cupy.ndarray'>
```

It would be nice to be able to capture these by (non-constant) reference.

shwina:
Some additional context - we need this in [`cuda.compute`](https://nvidia.github.io/cccl/python/compute.html) to support user-provided ops that can have side-effects. This unlocks many interesting and efficient ways to construct algorithms. For an example, see the code snippet in the description for this PR: https://github.com/NVIDIA/cccl/pull/6954.

I've implemented this by extending numba-cuda to include a device-array-like type and defining `lower_constant` for that type. Would a solution like this be acceptable to upstream to `numba-cuda`?

gmarkall:
Further context: https://github.com/numba/numba/issues/9084

gmarkall:
Prior attempt: https://github.com/NVIDIA/numba-cuda/pull/15

PR #666, opened 2025-12-15
Capture global device arrays in kernels and device functions

shwina:
Closes https://github.com/NVIDIA/numba-cuda/issues/659

### DIFF
--- a/numba_cuda/numba/cuda/tests/cudapy/test_device_array_capture.py
+++ b/numba_cuda/numba/cuda/tests/cudapy/test_device_array_capture.py
@@ -220,6 +220,24 @@ class TestDeviceArrayCapture(CUDATestCase):
                 expected = np.array([8.0, 9.0, 10.0], dtype=np.float32)
                 np.testing.assert_array_equal(result, expected)
 
+    def test_zero_dimensional(self):
+        """Test capturing 0-D (scalar) device arrays."""
+        for name, make_array in ARRAY_FACTORIES:
+            with self.subTest(array_type=name):
+                host_0d = np.array(42.0, dtype=np.float32)
+                global_0d = make_array(host_0d)
+
+                @cuda.jit
+                def kernel_0d(output):
+                    output[()] = global_0d[()] * 2.0
+
+                output = cuda.device_array((), dtype=np.float32)
+                kernel_0d[1, 1](output)
+
+                result = output.copy_to_host()
+                expected = 84.0
+                self.assertEqual(result, expected)
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/numba_cuda/numba/cuda/np/numpy_support.py
+++ b/numba_cuda/numba/cuda/np/numpy_support.py
@@ -25,6 +25,9 @@ def strides_from_shape(
     shape: tuple[int, ...], itemsize: int, *, order: str
 ) -> tuple[int, ...]:
     """Compute strides for a contiguous array with given shape and order."""
+    if len(shape) == 0:
+        # 0-D arrays have empty strides
+        return ()
     limits = slice(1, None) if order == "C" else slice(None, -1)
     transform = reversed if order == "C" else lambda x: x
     strides = tuple(


numba_cuda/numba/cuda/np/arrayobj.py
```diff
@@ -3642,10 +3642,75 @@ def record_static_setitem_int(context, builder, sig, args):
...
     """
     Create a constant array (mechanism is target-dependent).
+
+    For objects implementing __cuda_array_interface__,
+    the device pointer is embedded directly as a constant. For other arrays,
+    the target-specific mechanism is used.
     """
+    # Check if this is a device array (implements __cuda_array_interface__)
+    if hasattr(pyval, "__cuda_array_interface__"):
+        return _lower_constant_device_array(context, builder, ty, pyval)
```
cpcloud: It's actually cheaper to structure this as `getattr` since you're only getting the attribute once in that case.

`hasattr` is a C implementation of `try` -> `getattr` -> `except`, and after all that you still have to get the attribute again.

Can you restructure this like the following?

```suggestion
    cai = getattr(pyval, "__cuda_array_interface__", None)
    if cai is not None:
        return _lower_constant_device_array(context, builder, ty, cai)
```

The signature of `_lower_constant_device_array` would then have to change (but only in the fourth argument).

numba_cuda/numba/cuda/np/arrayobj.py
```diff
@@ -3642,10 +3642,75 @@ def record_static_setitem_int(context, builder, sig, args):
...
+    # Calculate strides if not provided (C-contiguous)
+    if strides is None:
+        itemsize = np.dtype(typestr).itemsize
+        ndim = len(shape)
+        strides = []
+        stride = itemsize
+        for i in range(ndim - 1, -1, -1):
+            strides.insert(0, stride)
+            stride *= shape[i]
+        strides = tuple(strides)
```
cpcloud: I think you can tighten this up a bit, something like this:

```suggestion
        strides = map(
            itemsize.__mul__,
            itertools.accumulate(
                reversed(shape[1:]),
                operator.mul,
                initial=1,
            ),
        )
```

numba_cuda/numba/cuda/typing/typeof.py
```diff
@@ -55,6 +55,14 @@ def typeof_impl(val, c):
     if tp is not None:
         return tp
 
+    # Check for __cuda_array_interface__ objects (third-party device arrays)
+
+    # Numba's own DeviceNDArray is handled above via _numba_type_.
+    if hasattr(val, "__cuda_array_interface__"):
```
cpcloud: Same review as above regarding `getattr`. Can you make `_typeof_cuda_array_interface` take the interface `dict` and then just access the attribute once?

cpcloud (changes requested):
Looks good.

I'd like to see a couple changes on the `hasattr` use and computing `strides`.

numba_cuda/numba/cuda/typing/typeof.py
```diff
@@ -299,3 +308,44 @@ def typeof_numpy_polynomial(val, c):
...
+        itemsize = np.dtype(val["typestr"]).itemsize
+        # Quick rejection: C-contiguous has strides[-1] == itemsize,
+        # F-contiguous has strides[0] == itemsize. If neither, it's "A".
+        if strides[-1] == itemsize:
+            c_strides = tuple(
+                itemsize * np.prod(shape[i + 1 :]) for i in range(ndim)
+            )
+            layout = "C" if strides == c_strides else "A"
+        elif strides[0] == itemsize:
+            f_strides = tuple(itemsize * np.prod(shape[:i]) for i in range(ndim))
```
cpcloud: This seems unnecessarily wasteful since you're repeating the accumulative multiplication for each dimension instead of doing a single multiplication for each dimension, and there's the waste of tuple -> ndarray -> ndarray (`*` operator) -> tuple, where you can remove all the ndarray intermediates.

How about this function instead?

```python
def strides_from_shape(shape: tuple[int, ...], itemsize: int, *, order: str) -> tuple[int, ...]:
    limits = slice(1, None) if order == "C" else slice(None, -1)
    transform = reversed if order == "C" else lambda x: x
    strides = tuple(map(itemsize.__mul__, itertools.accumulate(transform(shape[limits]), operator.mul, initial=1)))
    if order == "F":
        return strides
    return strides[::-1]
```


```suggestion
        if strides[-1] == itemsize:
            c_strides = strides_from_shape(shape, itemsize, order="C")
            layout = "C" if strides == c_strides else "A"
        elif strides[0] == itemsize:
            f_strides = strides_from_shape(shape, itemsize, order="F")
```

It's actually much faster than `np.prod` approach:

```
In [112]: import itertools, operator

In [113]: import numpy as np

In [114]: x = np.zeros((2, 3, 4), order="C")

In [115]: itemsize = x.dtype.itemsize

In [116]: shape = x.shape

In [117]: ndim = x.ndim

In [118]: timeit tuple(itemsize * np.prod(shape[i + 1:]) for i in range(ndim))
14 μs ± 39.4 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [119]: def strides_from_shape(shape: tuple[int, ...], itemsize: int, *, order: str) -> tuple[int, ...]:
     ...:     limits = slice(1, None) if order == "C" else slice(None, -1)
     ...:     transform = reversed if order == "C" else lambda x: x
     ...:     strides = tuple(map(itemsize.__mul__, itertools.accumulate(transform(shape[limits]), operator.mul, initial=1)))
     ...:     if order == "F":
     ...:         return strides
     ...:     return strides[::-1]
     ...:

In [120]: timeit strides_from_shape(shape, itemsize, order="C")
1.3 μs ± 9.75 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

shwina: Thanks! I tried `cumprod` to avoid the redundant calculation - and while it looks nicer your approach is still faster. I'll add this helper and use it

numba_cuda/numba/cuda/np/arrayobj.py
shwina: Changed to use `getattr` here. However, I  do `getattr` again in `_lower_constant_device_array`:

We do still need to pass `pyval` because in `_lower_constant_device_array` we have some logic that ensures numba holds a reference to it. It seems awkward to pass both `pyval` _and_ its CAI (what stops them from being unrelated?). So I elected to just pass `pyval`, and keep the second `getattr` in `_lower_constant_device_array`.

I don't feel super strongly about this decision as it's purely a helper. Defer to you whether we want to save an additional `getatttr` call at the cost of a slightly more fragile `_lower_constant_device_array`.

numba_cuda/numba/cuda/np/arrayobj.py
shwina: Fixed

numba_cuda/numba/cuda/typing/typeof.py
shwina: Made this change

shwina:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/666/commits/98799f8e1d835822e70601ad58e6e5a4f87a309a

numba_cuda/numba/cuda/np/arrayobj.py
cpcloud: Let's leave the additional `getattr` in for now. No point in dragging this out for that!

cpcloud (approved):
LGTM, ship it!

numba_cuda/numba/cuda/tests/cudapy/test_device_array_capture.py
```diff
@@ -0,0 +1,318 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: BSD-2-Clause
+
+"""
+Tests for capturing device arrays (objects implementing __cuda_array_interface__)
+from global scope in CUDA kernels and device functions.
+
+This tests the capture of third-party arrays (like CuPy) that implement
```
gmarkall: The docs say this works for Numba device arrays but this implies that it doesn't (because Numba device arrays have `_numba_type_`). I think this docstring rather than the documentation is incorrect - is that right?

numba_cuda/numba/cuda/np/arrayobj.py
```diff
@@ -3642,10 +3643,68 @@ def record_static_setitem_int(context, builder, sig, args):
...
+    Lower objects with __cuda_array_interface__ by embedding the device
+    pointer as a constant.
+
+    This allows device arrays captured from globals to be used in CUDA
+    kernels and device functions.
+    """
+    interface = pyval.__cuda_array_interface__
+
+    # Hold on to the device-array-like object to prevent garbage collection.
+    # The code library maintains a dictionary of referenced objects.
```
gmarkall: Since you showed (elsewhere) that the method I suggested to prevent pickling (patching the active code library to raise an error when it is serialized) doesn't work, I realised that this method for keeping the arrays alive will also not work - the code library that needs to survive is the one associated with the kernel, not a device function that referenced the array. So I think this will also not work for keeping referenced arrays alive.

numba_cuda/numba/cuda/np/arrayobj.py
```diff
@@ -3642,10 +3643,68 @@ def record_static_setitem_int(context, builder, sig, args):
...
+    lib = context.active_code_library
+    referenced_objects = getattr(lib, "referenced_objects", None)
+    if referenced_objects is None:
+        lib.referenced_objects = referenced_objects = {}
```
gmarkall: This approach made sense for an external library like CCCL. For Numba-CUDA itself, we should initialize `lib.referenced_objects` in `CUDACodeLibrary.__init__` instead.

When a code library is linked into another code library, we should also add the referenced objects of the library we're linking in to the current one. This should resolve the issue I mentioned above about the code library of a kernel not necessarily holding references to the device arrays it will access.

shwina: Implemented this solution in https://github.com/NVIDIA/numba-cuda/pull/666/commits/54d5fb315cb2e0087b1de8f1ecc9426f8b656353

numba_cuda/numba/cuda/tests/cudapy/test_device_array_capture.py
shwina: I've fixed this module docstring to say the correct thing.

I also replaced the use of CuPy with `ForeignArray`, and extended the tests to use both `DeviceNDArray` and `ForeignArray`.

shwina:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/666/commits/54d5fb315cb2e0087b1de8f1ecc9426f8b656353

numba_cuda/numba/cuda/tests/cudapy/test_device_array_capture.py
shwina: https://github.com/NVIDIA/numba-cuda/pull/666/commits/0a0b55b07290555336a0313519abc90cea05a7bb

shwina:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/666/commits/8d20e24c7b596893f1a4487be5ab64bc6c3bd6a2

docs/source/user/globals.rst
gmarkall, line 1: The examples in this file look like they should be copy-pastable to execute, but they don't run because of missing declarations. Should they be expected to work? (If they are, it may be good to convert them to doctests, like e.g https://github.com/NVIDIA/numba-cuda/blob/main/numba_cuda/numba/cuda/tests/doc_examples/test_random.py / https://github.com/NVIDIA/numba-cuda/blob/main/docs/source/user/examples.rst?plain=1)

numba_cuda/numba/cuda/tests/cudapy/test_device_array_capture.py
```diff
@@ -0,0 +1,280 @@
...
+    When kernels are compiled with cache=True, the device pointer would be
+    embedded in the cached code. This is problematic because:
+    1. The original device array may have been deallocated
+    2. A new device array may be allocated at a different address
+
+    Numba's serialization correctly rejects pickling of objects implementing
+    __cuda_array_interface__, preventing this issue.
+    """
+
+    def test_caching_rejects_captured_pointer(self):
```
gmarkall: If I run the test code from this test outside of the test suite:

```python
from numba import cuda
import numpy as np

host_data = np.array([1.0, 2.0, 3.0], dtype=np.float32)
captured_arr = cuda.to_device(host_data)

@cuda.jit(cache=True)
def cached_kernel(output):
    i = cuda.grid(1)
    if i < output.size:
        output[i] = captured_arr[i] * 2.0

output = cuda.device_array(3, dtype=np.float32)

cached_kernel[1, 3](output)

print(output.copy_to_host())
```

then it still caches the kernel, and the run fails the second time round:

```
  File "/home/gmarkall/numbadev/issues/numba-cuda-666/test_caching.py", line 17, in <module>
    print(output.copy_to_host())
          ^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/cudadrv/devices.py", line 233, in _require_cuda_context
    return fn(*args, **kws)
           ^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/cudadrv/devicearray.py", line 272, in copy_to_host
    _driver.device_to_host(
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/cudadrv/driver.py", line 2708, in device_to_host
    fn(*args)
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/cudadrv/driver.py", line 358, in safe_cuda_api_call
    return self._check_cuda_python_error(fname, libfn(*args))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/cudadrv/driver.py", line 417, in _check_cuda_python_error
    raise CudaAPIError(retcode, msg)
numba.cuda.cudadrv.driver.CudaAPIError: [700] Call to cuMemcpyDtoH results in CUDA_ERROR_ILLEGAL_ADDRESS
```

I don't think this test is working correctly (either the code is not running or the assertion is not firing) and the code that is intended to prevent caching is not preventing caching.

numba_cuda/numba/cuda/serialize.py
```diff
@@ -197,6 +197,18 @@ def reducer_override(self, obj):
         # Overridden to disable pickling of certain types
         if type(obj) in self.disabled_types:
             _no_pickle(obj)  # noreturn
+
+        # Prevent pickling of objects implementing __cuda_array_interface__
```
gmarkall, line 201: I think this has no effect because it never gets to see an object with the CUDA Array Interface when pickling. The `_reduce_states()` method of the `CUDACodeLibrary` has no reference to `referenced_objects`, so the serialization erroneously succeeds.

To prevent caching, the `CUDACodeLibrary` needs to detect that it holds referenced objects:

```diff
diff --git a/numba_cuda/numba/cuda/codegen.py b/numba_cuda/numba/cuda/codegen.py
index 957dd72e..9ee91e29 100644
--- a/numba_cuda/numba/cuda/codegen.py
+++ b/numba_cuda/numba/cuda/codegen.py
@@ -463,6 +463,10 @@ class CUDACodeLibrary(serialize.ReduceMixin, CodeLibrary):
 
         if not self._finalized:
             raise RuntimeError("Cannot pickle unfinalized CUDACodeLibrary")
+
+        if self.referenced_objects:
+            raise RuntimeError("Cannot pickle...")
+
         return dict(
             codegen=None,
             name=self.name,
```

gmarkall (changes requested):
I started reviewing this but picked up on a couple of issues so I wanted to submit the review so you could look into the comments rather than waiting until I looked everything over.

numba_cuda/numba/cuda/serialize.py
shwina: I see -- the issue is that for _closure_ variables, we are able to raise this `PicklingError` correctly - this happens starting at https://github.com/NVIDIA/numba-cuda/blob/dd396c82f0980e16e59d6e68a7a621b81fb4dd3a/numba_cuda/numba/cuda/core/caching.py#L338. The tests I have currently use closure variables (not globals). 

I followed your suggestion to correctly handle global device arrays in `CUDACodeLibrary`. Now we raise a `PicklingError` for both cases.

https://github.com/NVIDIA/numba-cuda/pull/666/commits/acce40daff134340d378bc58ce93fbb823efff7d

numba_cuda/numba/cuda/tests/cudapy/test_device_array_capture.py
shwina: See my comment below; the issue is that we handle closure variables correctly (and test appropriately), but not globals.

docs/source/user/globals.rst
shwina: They were meant to be representative/illustrative. If you think it's useful, I'll replace them with real doctests; thanks!

numba_cuda/numba/cuda/serialize.py
```diff
@@ -200,13 +200,11 @@ def reducer_override(self, obj):
...
         # These contain device pointers that would become stale after unpickling
-        if hasattr(obj, "__cuda_array_interface__"):
+        if getattr(obj, "__cuda_array_interface__", None) is not None:
             raise pickle.PicklingError(
-                f"Cannot pickle {type(obj).__name__} object: objects "
-                f"implementing __cuda_array_interface__ contain device "
-                f"pointers that cannot be safely serialized. "
-                f"Disable caching (cache=False) for kernels that capture "
-                f"device arrays from global scope."
+                "Cannot cache kernels or device functions referencing "
```
gmarkall: If Numba-CUDA is used with a distributed runtime (like Ray) that serializes kernels to send them to remote workers, then the error message will reference caching ("Cannot cache...") rather than pickling ("Cannot pickle..."). Do you think this could be confusing? Should we stick to referring to pickling here?

gmarkall:
A quick summary (pulling in discussions on this PR and Slack) of items I think are outstanding / noted:

* Modify the caching tests to use the same form as test_caching.py (possibly moving them there for ease of reuse of caching test infrastructure).
* Keep the `PicklingError` in `CUDACodeLibrary.reduce_states()`. The other error case raises `RuntimeError`, but that was a mistake to have used that exception class, so we should not copy it just for consistency.
* Potentially add a test with a 0D array (Greptile seems to think this will be an issue).
* Potentially reword exception messages, due to the serialization use cases that are not caching.
* Resolution of the `test_tanhf_compile_ptx` issue (I do not yet know why this is occurring).

shwina:
> Modify the caching tests to use the same form as test_caching.py (possibly moving them there for ease of reuse of caching test infrastructure) 

See https://github.com/NVIDIA/numba-cuda/pull/666/commits/2c3ead142c15a47eb2267fd2cceae0d4f34a015c

> Potentially add a test with a 0D array (Greptile seems to think this will be an issue).

See  https://github.com/NVIDIA/numba-cuda/pull/666/commits/f848c0ac7669545ffeb9b1a5954afa0f82624090

> Potentially reword exception messages, due to the serialization use cases that are not caching.

See https://github.com/NVIDIA/numba-cuda/pull/666/commits/8313619d04744d4599bc03d0ec22ed9f910e0cd8

> Add doctests (from comment [here](https://github.com/NVIDIA/numba-cuda/pull/666#discussion_r2624609903))

See https://github.com/NVIDIA/numba-cuda/pull/666/commits/2630d25e4ec2df8d0b67b892a4aa3605d2727903

With regards to:

> Resolution of the test_tanhf_compile_ptx issue (I do not yet know why this is occurring).

I wonder if mucking about with the caching infrastructure changes was causing this; especially when tests run in a certain order. Anyway, I'm hoping that with 2c3ead142c15a47eb2267fd2cceae0d4f34a015c this is resolved 🤞

shwina:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/666/commits/f848c0ac7669545ffeb9b1a5954afa0f82624090

Issue #673, opened 2025-12-18
CI: Switch to use `RAPIDSAI/sccache`?


leofang:
In #604 we noticed that the upstream sccache does not work with nvcc for generating fatbin/ltoir
https://github.com/NVIDIA/numba-cuda/blob/2dd7f259687b13606f2819c4351abad2cb32090d/testing/Makefile#L34-L39
and archive 
https://github.com/NVIDIA/numba-cuda/blob/2dd7f259687b13606f2819c4351abad2cb32090d/testing/Makefile#L110-L111
so we disabled using sccache for these outputs.

@trxcllnt mentioned that RAPIDS's fork has LTOIR covered
https://github.com/rapidsai/sccache/blob/58df5348a12b99c0ae726ccc3505a9749c414139/tests/system.rs#L1726-L1836

Currently, we're using `mozilla-actions/sccache-action`
https://github.com/NVIDIA/numba-cuda/blob/2dd7f259687b13606f2819c4351abad2cb32090d/.github/workflows/build-wheel.yml#L47-L51 
which literally just
- find or download `sccache`
- set a bunch of env vars

I suspect if we download the RAPIDS fork before executing this action, it’d use the forked version instead. Alternatively, we can just roll our own action that targets the RAPIDS fork. Should pretty easy either way.

PR #674, opened 2025-12-18
Use `rapidsai/sccache` in CI

trxcllnt:
Use the [`rapidsai/sccache`](https://github.com/rapidsai/sccache) fork in CI because it supports caching fatbins and ltoir.

Closes https://github.com/NVIDIA/numba-cuda/issues/673


cpcloud (approved):
LGTM

PR #676, opened 2025-12-18
Drop `experimental` from cuda.core namespace imports

brandon-b-miller:
This PR drops the `experimental` part of the `cuda.core` namespace.

### DIFF
--- a/.github/workflows/conda-python-tests.yaml
+++ b/.github/workflows/conda-python-tests.yaml
@@ -120,6 +120,7 @@ jobs:
       - name: Setup pixi
         uses: prefix-dev/setup-pixi@fef5c9568ca6c4ff7707bf840ab0692ba3f08293
         with:
+          pixi-version: "v0.61.0"
           environments: ${{ env.PIXI_ENV }}
           cache: false
       - name: Python tests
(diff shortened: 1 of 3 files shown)


brandon-b-miller:
/ok to test [ec8d388](https://github.com/NVIDIA/numba-cuda/pull/676/commits/ec8d388b4c9ac3c08259419c603572e177f42930)

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -54,18 +54,16 @@
...
+from numba.cuda._compat import (
     Linker,
     LinkerOptions,
     ObjectCode,
-)
-
-from cuda.bindings.utils import get_cuda_native_handle
-from cuda.core.experimental import (
     Stream as ExperimentalStream,
     Device as ExperimentalDevice,
```
bdice, line 62: You could drop "Experimental" from these names.

brandon-b-miller: I agree with this, however there's some broader work to be done to disambiguate the `cuda.core` classes with the ones defined later in this file, so I'm going to leave it for a follow up for now.

PR #681, opened 2025-12-19
Pass the -numba-debug flag to libnvvm

### LINKED ISSUE #679: [FEA] Pass -numba-debug flag to libnvvm for debug builds using the 13.1 or later release of the CUDA Toolkit
The CUDA Toolkit 13.1 provides a new libnvvm flag to enable enhanced debugging support for Numba CUDA programs (-numba-debug).

The fix is to update nvvm.py to pass the -numba-debug flag in to libnvvm when compiling with debugging enabled and when CUDA Toolkit 13.1 or later is detected.

mmason-nvidia:
When using the CUDA Toolkit release 13.1 or later for debug builds, we need to pass the -numba-debug flag to libnvvm in order to enable enhanced debug information.

Closes #679

### DIFF
--- a/numba_cuda/numba/cuda/cudadrv/nvvm.py
+++ b/numba_cuda/numba/cuda/cudadrv/nvvm.py
@@ -312,7 +312,7 @@ class CompilationUnit(object):
 
             return f"-{k}={v}".encode("utf-8")
 
-        debug_build = 'g' in options
+        debug_build = "g" in options
         options = [stringify_option(k, v) for k, v in options.items()]
 
         # If the -numba-debug option is supported by libnvvm, pass it to the


numba_cuda/numba/cuda/cudadrv/nvvm.py
```diff
@@ -243,6 +244,18 @@ def stringify_option(k, v):
...
+        # compiler when compiling with a debug build.
+        if 'g' in options:
+            # runtime.getLocalRuntimeVersion() returns (cudaError_t, version_int)
+            # Example: 13010 = CTK 13.1, 13020 = CTK 13.2
+            _, ctk_version_number = runtime.getLocalRuntimeVersion()
+            ctk_major = ctk_version_number // 1000
+            ctk_minor = (ctk_version_number % 1000) // 10
+            ctk_version = (ctk_major, ctk_minor)
+            if ctk_version >= (13, 1):
+                options['numba-debug'] = None
```
leofang: Hi Mark, we have found that this is not reliable on the cuda-python side. I'd suggest that you copy/paste this approach to determine the libNVVM version.
https://github.com/NVIDIA/cuda-python/blob/da7eb1f5a97aa21d8f78098e13e7c4edad013530/cuda_core/tests/test_program.py#L58-L132

mmason-nvidia: I incorporated your suggestion into the NVVM singleton in nvvm.py.

gmarkall:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/681/commits/6764681be58696996c217cde70663d347eb573df

numba_cuda/numba/cuda/cudadrv/nvvm.py
```diff
@@ -195,6 +198,115 @@ def get_ir_version(self):
...
+            print(f"Exception compiling test program: {exception}")
+            raise exception
+        finally:
+            if program.value:
+                # Destroy the NVVM program, not fatal if it fails
+                err = self.nvvmDestroyProgram(byref(program))
+                try:
+                    self.check_error(err, "Failed to destroy test program.")
+                except Exception:
+                    pass
```
gmarkall: I think there's no point checking the error if we're going to swallow the exception the check will raise anyway.
```suggestion
```

numba_cuda/numba/cuda/cudadrv/nvvm.py
```diff
@@ -195,6 +198,115 @@ def get_ir_version(self):
...
+            ptx_size = c_size_t()
+            err = self.nvvmGetCompiledResultSize(program, byref(ptx_size))
+            self.check_error(
+                err, "Failed to get test program compiled result size."
+            )
+            ptx_data = (c_char * ptx_size.value)()
+            err = self.nvvmGetCompiledResult(program, ptx_data)
+            self.check_error(err, "Failed to get test program compiled result.")
+        except Exception as exception:
+            print(f"Exception compiling test program: {exception}")
```
gmarkall: This should probably be a warning rather than a print:
```suggestion
            warnings.warn(
                f"Exception compiling test program: {exception}",
                category=NvvmWarning
            )
```

numba_cuda/numba/cuda/cudadrv/nvvm.py
```diff
@@ -195,6 +198,115 @@ def get_ir_version(self):
...
+            raise exception
```
gmarkall: I don't think we should re-raise the exception, just let it pass - otherwise I'd expect it to propagate all the way back to the user, which we may not want.
```suggestion
```

numba_cuda/numba/cuda/cudadrv/nvvm.py
```diff
@@ -195,6 +198,115 @@ def get_ir_version(self):
...
+        # Extract the PTX version and lookup the corresponding
+        # CUDA Toolkit version. If this fails, the CUDA Toolkit version
+        # cannot be determined and self._libnvvm_cuda_version will remain None
+        # as expected.
+        try:
+            ptx_version = get_ptx_ver(ptx_data[:].decode("utf-8"))
+            self._libnvvm_cuda_version = (
+                get_minimal_required_cuda_ver_from_ptx_ver(ptx_version)
+            )
+        except Exception:
```
gmarkall: It seems that the only likely exception we'd expect from the PTX version functions is a `ValueError` - anything else is a bit more surprising so we should let it manifest to expose the underlying bug instead:
```suggestion
        except ValueError:
```

numba_cuda/numba/cuda/cudadrv/nvvm.py
```diff
@@ -243,6 +355,15 @@ def stringify_option(k, v):
 
             return f"-{k}={v}".encode("utf-8")
 
+        # Starting in r13.1, we must pass in the -numba-debug flag to the
+        # compiler when compiling with a debug build. If the CUDA version
+        # cannot be determined, assume that a newer version is being used and
+        # pass in the -numba-debug flag.
+        if "g" in options:
+            ctk_version = self.driver.get_cuda_version()
+            if ctk_version is None or ctk_version >= (13, 1):
```
gmarkall: If we couldn't determine the version of the CTK, that could be because the version is 12.x because the necessary PTX version functions weren't present in the CUDA bindings for 12.x. So I think it'd be safer to assume we don't pass the `-numba-debug` flag if we can't determine the version:
```suggestion
            if ctk_version is not None and ctk_version >= (13, 1):
```

gmarkall:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/681/commits/f32c2a53f61032396ed5dc8d586b8c960fc5a10e

gmarkall:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/681/commits/56b553b7aabeb0a12fbc7e74d6918550f93fea7e

PR #682, opened 2025-12-19
fix: use freethreading-supported `_PySet_NextItemRef` where possible

cpcloud:
Small change to the `_typeof.cpp` code to use a freethreading-supported API for getting the first item from a set in C.


kkraus14:
Is there any reasonable way to test this?

cpcloud:
> Is there any reasonable way to test this?

AFAICT, there isn't.

PR #686, opened 2025-12-26
Add CUDA FP8 type + conversion bindings (E5M2/E4M3/E8M0), HW-accel detection, and comprehensive tests

isVoid:
## Summary
Adds **FP8 support in `numba-cuda`** by introducing generated bindings for CUDA FP8 types and conversion intrinsics (including saturation + rounding-mode variants), wiring them into CUDA typing/target registries, and adding a comprehensive test suite. Also adds a public runtime check for **FP8 hardware acceleration**.

## Motivation / Context
- Enable first-class use of CUDA FP8 types (`__nv_fp8_*`) and conversion intrinsics from Numba CUDA kernels.
- Provide a capability API (`cuda.is_fp8_accelerated()`) so users can branch on HW-accelerated FP8 vs software-emulated behavior.
- Lock down behavior with thorough tests (constructors, conversions, NaNs, saturation, rounding modes).

### FP8 bindings + API surface
- **Autogenerated bindings**: `numba_cuda/numba/cuda/_internal/cuda_fp8.py`
- **Public registry shim**: `numba_cuda/numba/cuda/fp8.py` (exports `typing_registry` + `target_registry`)
- **Toolkit headers vendored**:
  - `numba_cuda/numba/cuda/include/12/cuda_fp8.h`
  - `numba_cuda/numba/cuda/include/12/cuda_fp8.hpp`
  - `numba_cuda/numba/cuda/include/13/cuda_fp8.h`
  - `numba_cuda/numba/cuda/include/13/cuda_fp8.hpp`
- **Binding generator config**: `configs/cuda_fp8.yml`

### Compiler integration
- Registers FP8 typing/target lowering via `numba_cuda/numba/cuda/target.py` (`fp8.typing_registry`, `fp8.target_registry`).

### Public capability API
- **New**: `numba_cuda/numba/cuda/api.py::is_fp8_accelerated()`
- Backed by:
  - `numba_cuda/numba/cuda/cudadrv/driver.py::Device.accelerates_fp8` (CC >= 8.9)
  - `numba_cuda/numba/cuda/simulator/api.py::is_fp8_accelerated()` → always `False`
  - Simulator `detect()` output reports FP8 HW acceleration status.

### Tests
- **New**: `numba_cuda/numba/cuda/tests/cudapy/test_fp8_bindings.py`
- Covers:
  - Constructors from float/int types
  - Conversions to/from float types
  - NaN handling
  - Saturation (`saturation_t`)
  - Rounding-mode variants using `cuda.bindings.runtime.cudaRoundMode` (notably for E8M0 paths)

## User-facing API changes
- **New**: `numba.cuda.is_fp8_accelerated()` returns `True` when device supports FP8 HW acceleration; otherwise `False`.
- **New internal bindings**: `numba.cuda._internal.cuda_fp8` providing FP8 types and conversion intrinsics (available for advanced use).

## Testing
Added test coverage via `numba_cuda/numba/cuda/tests/cudapy/test_fp8_bindings.py`.

```bash
python -m pytest -q numba_cuda/numba/cuda/tests/cudapy/test_fp8_bindings.py
```

## Notes
- `numba_cuda/numba/cuda/_internal/cuda_fp8.py` is **autogenerated** (see header for generator/version).
- FP8 operations still work on older GPUs via software emulation, but `cuda.is_fp8_accelerated()` is the recommended guard for performance-sensitive paths.

## Follow ups
- Provide "raw-8-bit" constructor from uint8 type. I intend to combine with raw bf16 construction in a single PR.
- Better error handling (unsupported round mode failure message)
- High level API design and documentation


isVoid:
I addressed point 3 and 4. 
1) Will not address since it's vended code.
2) `numpy.testing.assert_array_equal` does not raise when `nan` appears at the same index for two arrays even if elementwise comparison of `nan==nan` is false.

gmarkall:
I had a quick look at the FP8 headers, and I think that things are going wrong in `__nv_cvt_fp8_to_halfraw` but I haven't yet looked into why... I think it should be possible to debug by compiling for the host and using the `-6` input value then inspecting what's going on though.

That said, I wonder if it's acceptable and more straightforward to only support FP8 on Hopper / Ada and beyond?

Implementation of the function lifted from `cuda_fp16.hpp` (from line 472) for reference below:

```c++
__CUDA_HOSTDEVICE_FP8_DECL__ __half_raw
__nv_cvt_fp8_to_halfraw(const __nv_fp8_storage_t x,
                        const __nv_fp8_interpretation_t fp8_interpretation) {
    __half_raw res;
    res.x = 0U;
#if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 890)
    res.x =
        __nv_cvt_fp8x2_to_halfraw2((__nv_fp8x2_storage_t)x, fp8_interpretation)
            .x;
#else
    unsigned short int ur = (unsigned short int)x;
    ur = (unsigned short int)(ur << 8U);

    if (fp8_interpretation == __NV_E5M2) {
        if ((ur & 0x7FFFU) > 0x7C00U) {
            /* If NaN, return canonical NaN */
            ur = 0x7FFFU;
        }
    } else { // __NV_E4M3
        unsigned short int sign = ur & 0x8000U;
        unsigned short int exponent =
            (unsigned short int)(((ur & 0x7800U) >> 1U) + 0x2000U);
        unsigned short int mantissa = (ur & 0x0700U) >> 1U;
        unsigned char absx = 0x7FU & (unsigned char)x;

        if (absx == 0x7FU) // NaN
        {
            ur = 0x7FFFU; // fp16 canonical NaN, discard sign
        } else if (exponent == 0x2000U) {
            // zero or denormal
            if (mantissa != 0U) {
                // normalize
                mantissa = (unsigned short int)(mantissa << 1U);
                while ((mantissa & 0x0400U) == 0U) {
                    mantissa = (unsigned short int)(mantissa << 1U);
                    exponent = (unsigned short int)(exponent - 0x0400U);
                }
                // discard implicit leading bit
                mantissa &= 0x03FFU;
            } else { // Zero
                exponent = 0U;
            }

            ur = (sign | exponent) | mantissa;
        } else {
            ur = (sign | exponent) | mantissa;
        }
    }
    res.x = ur;
#endif
    return res;
}
```

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -625,6 +625,15 @@ def supports_float16(self):
     def supports_bfloat16(self):
         return self.compute_capability >= (8, 0)
 
+    @property
+    def accelerates_fp8(self):
```
isVoid: Rename this back to `supports_fp8` and guard against tests.

isVoid:
> That said, I wonder if it's acceptable and more straightforward to only support FP8 on Hopper / Ada and beyond?

It should be, considering that support on older architecture is likely simulated and no real performance gain is expected.

gmarkall (approved):

PR #695, opened 2026-01-06
Fix DISubprogram line number to point to function definition line

jiel-nv:
Fixes incorrect DW_AT_decl_line where the line number was first_stmt_line - 1 instead of the actual def line, causing off-by-one errors when comments exist between the def line and first statement.

- Add get_func_def_lineno() in firstlinefinder.py to return the actual 'def' line
- Update lowering.py to use get_func_def_lineno() directly without offset calculation
- Add test_DISubprogram_def_line_number covering single/multi-line decorators with/without comments

This change resolves nvbug#5723821.


gmarkall:
Apologies for the delay in reviewing. Whilst testing it, I discovered and investigated #721, which I think needs fixing at some point but could be done separately from this PR. I think this PR is otherwise correct and makes an improvement.

PR #697, opened 2026-01-06
chore: perf lint

cpcloud:
Add `"PERF"` lints to the ruff config. For our current kernel launch benchmarks there is no change. Most of the changes here would not be incredibly impactful on their own.


rparolin (approved):

rparolin:
@gmarkall for visibility

PR #698, opened 2026-01-07
Fix: Pass correct flags to linker when debugging in the presence of LTOIR code

### LINKED ISSUE #696: [BUG]Incorrect linker flags when linking a debuggable Numba CUDA kernel with code compiled for LTO
Attempting to debug a kernel declared with debug=True that calls into externally compiled LTO-IR code causes cuda.bindings.nvjitlink to raise the following exception:

``` File "cuda/bindings/nvjitlink.pyx", line 124, in cuda.bindings.nvjitlink.add_data
  File "cuda/bindings/nvjitlink.pyx", line 143, in cuda.bindings.nvjitlink.add_data
  File "cuda/bindings/nvjitlink.pyx", line 75, in cuda.bindings.nvjitlink.check_status
cuda.bindings.nvjitlink.nvJitLinkError: ERROR_LTO_NOT_ENABLED (18)
Linker error log: ERROR 1: LTO-IR input cannot be used without LTO enabled (-lto)
```

The fix is to only pass in the -lto and -ptx flags to the linker if LTO code is actually present in the object codes list, instead of depending on the value of self.lto specified at _Linker creation time.

mmason-nvidia:
The linker code was passing in -lto to linker invocations that did not involve LTOIR code, and not passing it in some cases where LTOIR code was being linked. When enabling debugging of a Numba CUDA kernel which calls into LTOIR code, an exception was being raised by nvjitlink.

This change corrects that behavior, only passing in -lto for cases where at least one LTOIR code object is in the link list. The lto= parameter to the Linker initialization is still used to control compilation of .cu code with LTO enabled (which will result in the self._has_ltoir flag being set).

A testcase for validating this change and catching regressions is included.

Closes #696

### DIFF
--- a/numba_cuda/numba/cuda/cudadrv/driver.py
+++ b/numba_cuda/numba/cuda/cudadrv/driver.py
@@ -2495,12 +2495,16 @@ class _Linker:
 
     def _get_linker_options(self, ptx):
         # Some linker flags are only valid/required if LTOIR object code is present.
+        # WAR for cuda-core < 0.4.0 where passing False incorrectly appends flags
+        # (fixed in cuda-python PR #989, released in cuda-core v0.4.0)
+        lto_flag = True if self._has_ltoir else None
+        ptx_flag = True if (self._has_ltoir and ptx) else None
         options = LinkerOptions(
             max_register_count=self.max_registers,
             lineinfo=self.lineinfo,
             arch=self.arch,
-            link_time_optimization=self._has_ltoir,
-            ptx=self._has_ltoir and ptx,
+            link_time_optimization=lto_flag,
+            ptx=ptx_flag,
         )
         return options
 


gmarkall:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/698/commits/94e67457874dffcf8ec7ef4be26f95ab09293ad1

numba_cuda/numba/cuda/tests/cudadrv/test_linker.py
```diff
@@ -109,6 +109,18 @@ def simple_lmem(A, B, dty):
         B[i] = C[i]
 
 
+add_from_numba = cuda.declare_device(
+    "add_from_numba",
+    "int32(int32, int32)",
+    link=["testing/test_device_functions.ltoir"],
```
gmarkall: This looks wrong (hardcoding the path will not work in all cases) - instead it should be like in other test cases, e.g.:

https://github.com/NVIDIA/numba-cuda/blob/6a1db644ffa40aefce6651ba171597f38d6555de/numba_cuda/numba/cuda/tests/cudapy/test_compiler.py#L41-L43

numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -2308,18 +2308,9 @@ def __init__(
         self.lineinfo = lineinfo
         self.cc = cc
         self.arch = arch
-        if lto is False:
-            # WAR for apparent nvjitlink issue
-            lto = None
```
gmarkall, line 2317: I wonder if this will be a problem with the 12.9 CUDA Python bindings - there was a version that would only accept `None` and not `False`, but I don't recall what the version was.

gmarkall: As far as I could tell from the CI so far, the removal of the `None` value did not cause an issue - perhaps it is no longer needed.

gmarkall:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/698/commits/ec527109e9bc14a271c601e041060a7b5bf26646

gmarkall:
Since the code changes in the non-test code looked good, I've just pushed the relevant fix for the test code to get this over the line.

numba_cuda/numba/cuda/cudadrv/driver.py
gmarkall: Looking at the tests that are now failing, it seems that this workaround is probably still needed.

gmarkall:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/698/commits/78e063485105769face0e5c4a37398216dbc3b03

jiel-nv:
@mmason-nvidia, I have a fix commited to my local repo, but not able to push to your fork. Could you please add me as a collaborator on your fork?

gmarkall:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/698/commits/fa5ad371860d25df2fbf604610e0bec77358949c

gmarkall:
I think I see the fix so I just pushed it so ti can get tested whilst CI is (hopefully) not too busy.

jiel-nv:
> I think I see the fix so I just pushed it so ti can get tested whilst CI is (hopefully) not too busy.

Thanks a lot, @gmarkall, for helping to push the fix in.

Just wondering how you did it, because I tried to push and were not successful.

gmarkall:
I did

```
git push mmason-nvidia mmason-nvidia/bugfix/lto-linker-flags
```

where that remote is

```
$ git remote -v
mmason-nvidia	git@github.com:mmason-nvidia/numba-cuda.git (push)
```

PR #699, opened 2026-01-07
Backport today's pathfinding logic to `0.19.*`

brandon-b-miller:
Backport of a few bits of https://github.com/NVIDIA/numba-cuda/pull/464 needed to handle `libnvvm` and `libdevice` for cuda 13+

### DIFF
--- a/ci/test_wheel_deps_wheels.sh
+++ b/ci/test_wheel_deps_wheels.sh
@@ -52,8 +52,8 @@ if [[ -z "$nvrtc_pkgs" ]]; then
     exit 1
 fi
 
-apt remove --purge -y "$nvvm_pkgs"
-apt remove --purge -y "$nvrtc_pkgs"
+apt remove --purge -y $nvvm_pkgs
+apt remove --purge -y $nvrtc_pkgs
 
 rapids-logger "Run Tests"
 NUMBA_CUDA_TEST_BIN_DIR=$NUMBA_CUDA_TEST_BIN_DIR python -m pytest --pyargs numba.cuda.tests -v


numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -2,29 +2,35 @@
...
-from pathlib import Path
-from numba.core.config import IS_WIN32
-from numba.misc.findlib import find_lib
+import importlib.metadata
+from numba.cuda.misc.findlib import find_lib
 from numba import config
-import ctypes
+
+IS_WIN32 = config.IS_WIN32
+
```
leofang, line 13: nit: I was tracking diff and it seems these two lines are the extra change compared to what is shown in https://github.com/NVIDIA/numba-cuda/compare/v0.19-patches..97ce4b1525de474efbc93b45eea0668167fcce3b#diff-682ac5ffbee833445b5101e759557a35163bda936522251370e29f9625147554.

brandon-b-miller:
maybe @kkraus14 might be a good person to take a quick look here? This PR attempts to backport today's pathfinding logic into 0.19.

ci/coverage_report.sh
```diff
@@ -13,8 +13,9 @@ python -m pip install \
     "${package}[test]" \
     "cuda-python==${CUDA_VER_MAJOR_MINOR%.*}.*" \
     "cuda-core==0.3.*" \
+    "numba<0.62.0" \
```
brandon-b-miller, line 16: In relevant CI jobs, this patch release restricts the tested environment to numba <0.62.0, which was released after numba-cuda 0.19.1 was released. numba-cuda 0.19.1 is already incompatible with numba 0.62+, this PR makes no attempt to fix this situation but adjusts CI jobs at the patch commit to pass tests.

ci/test_thirdparty.sh
```diff
@@ -45,7 +45,7 @@ rapids-logger "Run GroupBy UDF tests"
 python -m pytest python/cudf/cudf/tests/groupby/test_apply.py -k test_groupby_apply_jit -W ignore::UserWarning
 
 rapids-logger "Run NRT Stats Counting tests"
-python -m pytest python/cudf/cudf/tests/test_nrt_stats.py -W ignore::UserWarning
+python -m pytest python/cudf/cudf/tests/private_objects/test_nrt_stats.py -W ignore::UserWarning
```
brandon-b-miller, line 48: These tests moved in between the last time CI ran for 0.19.1 and the final rapids release targeted by this job.

ci/test_wheel.sh
```diff
@@ -5,20 +5,24 @@
...
+CUDA_VER_MAJOR=${CUDA_VER%.*.*}
+
 
 rapids-logger "Install wheel with test dependencies"
 package=$(realpath wheel/numba_cuda*.whl)
 echo "Package path: ${package}"
 
 DEPENDENCIES=(
-    "${package}[test]"
+    "${package}[cu${CUDA_VER_MAJOR},test-cu${CUDA_VER_MAJOR}]"
```
brandon-b-miller, line 16: These tests never installed the cuda toolkit dependencies and passed via system libraries. This PR doesn't remove the system libraries from these CI jobs, but does install the actual wheel deps now which get picked up instead.

Issue #152, opened 2025-03-10
[FEA] Make `cuda.core.Buffer` recognized by numba-cuda


leofang:

brandon-b-miller:
@leofang just getting my head around the requirements here. Is the goal to make it so that one can pass `cuda.core.Buffer` objects as arguments of numba kernels? Or are you envisioning a world where one can write `x = cuda.core.Buffer(...)` in a kernel and expect back a heap allocated buffer for the cuda thread to use for something downstream in the kernel? Or something else?

If it's the former I think all we might need to do is support `__cuda_array_interface__` for `cuda.core.Buffer`.

Issue #128, opened 2025-02-20
[RFC] Make `cuda.core` a required dependency


leofang:
Recent works in `cuda.core`/`cuda.bindings` have reached to a point where we can replace most, if not all, `numba-cuda` internal wrappers to driver, nvvm, linker, … by cuda.core. It has been the plan to make `cuda.core` a required (not just optional) dependency of `numba-cuda` (and `cuda.bindings` would become a transitive dependency). There are multiple benefits, such as
- separation of concerns: `numba-cuda` focuses on the compiler stack and leave CUDA’s Python exposure to CUDA Python
- support CUDA wheels out of box (so [issue 65](https://github.com/NVIDIA/numba-cuda/issues/65) and [issue 66](https://github.com/NVIDIA/numba-cuda/issues/66) do not need any effort on the numba-cuda side)
- better performance (since cuda-bindings is based on Cython instead of ctypes)
- support Windows (which [pynvjitlink does not, as of today](https://github.com/rapidsai/pynvjitlink/pull/79))
- coverage of latest CUDA features
- …

This RFC is created to track necessary efforts.

realarnavgoel:
@leofang Does `cuda.core._launcher` support option in `LaunchConfig` to select `cuLaunchCooperativeKernel` or `cuLaunchKernel` depending on the user needs for `grid.sync()` or other deadlock-free guarantees when performing non-local operations in each GPU (e.g. barrier between GPUs) of a multi-GPU application ?

`numba-cuda` supports this automatically and selects the appropriate mechanism in the dispatcher: https://github.com/NVIDIA/numba-cuda/blob/main/numba_cuda/numba/cuda/dispatcher.py#L409

leofang:
> Does `cuda.core._launcher` support option in `LaunchConfig` to select `cuLaunchCooperativeKernel` or `cuLaunchKernel` depending on the user needs for `grid.sync()` or other deadlock-free guarantees when performing non-local operations in each GPU (e.g. barrier between GPUs) of a multi-GPU application ?

The beauty of the new `cuLaunchKernelEx` API is it is not our (cuda.core, and maybe even Numba) job to choose which API to use. We just need to add a new option to `LaunchConfig` to hook with `CU_LAUNCH_ATTRIBUTE_COOPERATIVE` as you suggested, so that the driver knows to launch it cooperatively. Please open an issue in NVIDIA/cuda-python to track it.

realarnavgoel:
[...]

PR #703, opened 2026-01-07
feat: swap out internal device array usage with `StridedMemoryView`

cpcloud:
# Summary

Refactor kernel argument handling to use `StridedMemoryView` internally,
enabling direct support for `__dlpack__` objects and improving interoperability
with libraries like CuPy.

**Closes:** #152
**Tracking issue:** #128

## Key Changes

**New capability:** Kernel arguments now accept objects with `__dlpack__`
protocol directly (e.g., CuPy arrays).

**Internals:** Replaced array interface handling with
`cuda.core.utils.StridedMemoryView` for:
- `__dlpack__` objects (new)
- `__cuda_array_interface__` objects
- Device arrays (existing `DeviceNDArray`)
- NumPy arrays (copied to device as before)

**Performance:**
- CuPy arrays: **~3x improvement** on kernel launch (initial measurements)
- `device_array()` arrays: **~2.5x regression** (initial measurements)
- torch shows regression as well, probably because its DLPack implementation is
  slow. Previously it was going through CAI but its CAI version isn't supported
  by ~StridedMemoryView~

## Performance Trade-off Discussion

The 2.5x slowdown for `device_array()` is worth discussing (and perhaps the
torch regression is as well):

**Arguments for accepting this regression:**
1. CuPy and other `__dlpack__` libraries represent the primary ecosystem (or at
   least the *end goal*) for GPU computing in Python
2. The 3x speedup benefits the common interoperability use case for objects
   that we are prioritizing
3. `device_array()` is primarily used in legacy code and tests and is
   deprecated

**Why this might be worth merging despite the regression:**
- Blocking ecosystem integration to optimize a legacy API doesn't align with
  the project's direction
- The gains where it matters (external arrays) are substantial
- Performance parity for device_arrays could be addressed in follow-up work if
  it proves important

## Implementation Details

- Added `_to_strided_memory_view()` and `_make_strided_memory_view()` helper
  functions (numba_cuda/numba/cuda/cudadrv/devicearray.py:247-359)
- Updated kernel argument marshaling in dispatcher.py:541-614
- Added LRU caching to `typeof` for CAI objects to reduce type inference
  overhead (typing/typeof.py:315-365)

## Testing

Existing test suite passes.

---

**TL;DR:** Adds `__dlpack__` support (~3x faster for CuPy), with ~2.5x
regression on legacy `device_array()`. Trade-off favors ecosystem integration,
but open to discussion.

### DIFF
--- a/numba_cuda/numba/cuda/typing/typeof.py
+++ b/numba_cuda/numba/cuda/typing/typeof.py
@@ -323,7 +323,7 @@ def _typeof_cuda_array_interface_cached(
     elif strides is None:
         layout = "C"
     else:
-        itemsize = dtype.itemsize
+        itemsize = dtype.bitwidth
         # Quick rejection: C-contiguous has strides[-1] == itemsize,
         # F-contiguous has strides[0] == itemsize. If neither, it's "A".
         if strides[-1] == itemsize:


cpcloud:
This PR can't be merged until the next release of cuda-core, because I depend on some unreleased features there.

However, it's still worth reviewing.

cpcloud:
Current benchmarks versus `main` (`NOW` is this PR, the other is `upstream/main` as of when I ran the command):

image

<details><summary><b>Full text of benchmark results</b></summary>

```
------------------------------------------------------------------------------ benchmark 'test_many_args[dispatch-cupy]': 2 tests ------------------------------------------------------------------------------
Name (time in ms)                                    Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_many_args[dispatch-cupy] (NOW)              28.5491 (1.0)      32.0701 (1.0)      29.3013 (1.0)      0.8509 (3.39)     29.0451 (1.0)      0.7380 (2.20)          3;1  34.1282 (1.0)          20           1
test_many_args[dispatch-cupy] (0001_459b8c0)     81.2223 (2.85)     82.0122 (2.56)     81.6227 (2.79)     0.2509 (1.0)      81.5803 (2.81)     0.3350 (1.0)           3;0  12.2515 (0.36)          8           1
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------- benchmark 'test_many_args[dispatch-device_array]': 2 tests ------------------------------------------------------------------------------
Name (time in ms)                                            Min                Max               Mean            StdDev             Median               IQR            Outliers       OPS            Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_many_args[dispatch-device_array] (0001_459b8c0)      6.5039 (1.0)       6.6605 (1.0)       6.5602 (1.0)      0.0347 (1.0)       6.5567 (1.0)      0.0391 (1.0)           5;1  152.4334 (1.0)          21           1
test_many_args[dispatch-device_array] (NOW)              16.1680 (2.49)     18.7992 (2.82)     16.4472 (2.51)     0.5087 (14.68)    16.3114 (2.49)     0.1429 (3.65)          2;3   60.8005 (0.40)         28           1
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------ benchmark 'test_many_args[dispatch-torch]': 2 tests ------------------------------------------------------------------------------
Name (time in ms)                                     Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_many_args[dispatch-torch] (NOW)              61.4096 (1.0)      73.2956 (1.03)     65.9567 (1.0)      3.3014 (6.78)     65.8106 (1.0)      3.7330 (7.42)          4;1  15.1615 (1.0)          13           1
test_many_args[dispatch-torch] (0001_459b8c0)     69.6396 (1.13)     71.3505 (1.0)      70.7103 (1.07)     0.4867 (1.0)      70.8821 (1.08)     0.5033 (1.0)           3;1  14.1422 (0.93)         10           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------ benchmark 'test_many_args[signature-cupy]': 2 tests ------------------------------------------------------------------------------
Name (time in ms)                                     Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_many_args[signature-cupy] (NOW)              20.3893 (1.0)      22.5916 (1.0)      20.7331 (1.0)      0.2960 (1.79)     20.6997 (1.0)      0.1337 (1.0)           3;3  48.2320 (1.0)          48           1
test_many_args[signature-cupy] (0001_459b8c0)     61.7753 (3.03)     62.3395 (2.76)     61.9959 (2.99)     0.1649 (1.0)      61.9470 (2.99)     0.1919 (1.44)          5;0  16.1301 (0.33)         17           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------ benchmark 'test_many_args[signature-device_array]': 2 tests -------------------------------------------------------------------------------
Name (time in ms)                                             Min                Max               Mean            StdDev             Median               IQR            Outliers       OPS            Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_many_args[signature-device_array] (0001_459b8c0)      6.7572 (1.0)       6.9647 (1.0)       6.8144 (1.0)      0.0366 (1.0)       6.8028 (1.0)      0.0550 (1.0)          38;2  146.7473 (1.0)         141           1
test_many_args[signature-device_array] (NOW)              16.7750 (2.48)     18.1621 (2.61)     17.2505 (2.53)     0.2715 (7.42)     17.1982 (2.53)     0.2356 (4.28)         12;5   57.9693 (0.40)         58           1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------ benchmark 'test_many_args[signature-torch]': 2 tests ------------------------------------------------------------------------------
Name (time in ms)                                      Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_many_args[signature-torch] (0001_459b8c0)     49.9804 (1.0)      50.5731 (1.0)      50.1364 (1.0)      0.1298 (1.0)      50.1156 (1.0)      0.1224 (1.0)           4;1  19.9456 (1.0)          20           1
test_many_args[signature-torch] (NOW)              51.7247 (1.03)     53.2339 (1.05)     52.3643 (1.04)     0.4227 (3.26)     52.5188 (1.05)     0.7008 (5.73)          6;0  19.0970 (0.96)         20           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

----------------------------------------------------------------------------- benchmark 'test_one_arg[dispatch-cupy]': 2 tests ----------------------------------------------------------------------------
Name (time in ms)                                 Min               Max              Mean            StdDev            Median               IQR            Outliers       OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_one_arg[dispatch-cupy] (NOW)              3.1364 (1.0)      3.2981 (1.0)      3.2120 (1.0)      0.0336 (1.17)     3.2084 (1.0)      0.0362 (1.09)         18;3  311.3370 (1.0)          63           1
test_one_arg[dispatch-cupy] (0001_459b8c0)     5.6052 (1.79)     5.7344 (1.74)     5.6561 (1.76)     0.0286 (1.0)      5.6565 (1.76)     0.0333 (1.0)           8;2  176.7993 (0.57)         35           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

----------------------------------------------------------------------------- benchmark 'test_one_arg[dispatch-device_array]': 2 tests ----------------------------------------------------------------------------
Name (time in ms)                                         Min               Max              Mean            StdDev            Median               IQR            Outliers       OPS            Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_one_arg[dispatch-device_array] (0001_459b8c0)     1.2796 (1.0)      1.3412 (1.0)      1.3013 (1.0)      0.0261 (2.85)     1.2922 (1.0)      0.0400 (3.88)          1;0  768.4784 (1.0)           5           1
test_one_arg[dispatch-device_array] (NOW)              1.9005 (1.49)     1.9257 (1.44)     1.9143 (1.47)     0.0091 (1.0)      1.9148 (1.48)     0.0103 (1.0)           2;0  522.3939 (0.68)          5           1
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

---------------------------------------------------------------------------- benchmark 'test_one_arg[dispatch-torch]': 2 tests -----------------------------------------------------------------------------
Name (time in ms)                                  Min               Max              Mean            StdDev            Median               IQR            Outliers       OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_one_arg[dispatch-torch] (NOW)              4.9126 (1.0)      5.9787 (1.16)     5.0667 (1.0)      0.2255 (15.06)    4.9729 (1.0)      0.0839 (3.53)        10;11  197.3664 (1.0)          58           1
test_one_arg[dispatch-torch] (0001_459b8c0)     5.1037 (1.04)     5.1586 (1.0)      5.1313 (1.01)     0.0150 (1.0)      5.1302 (1.03)     0.0238 (1.0)          13;0  194.8823 (0.99)         32           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

---------------------------------------------------------------------------- benchmark 'test_one_arg[signature-cupy]': 2 tests -----------------------------------------------------------------------------
Name (time in ms)                                  Min               Max              Mean            StdDev            Median               IQR            Outliers       OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_one_arg[signature-cupy] (NOW)              2.1288 (1.0)      2.7003 (1.0)      2.2073 (1.0)      0.0588 (1.0)      2.1964 (1.0)      0.0452 (1.0)         41;20  453.0358 (1.0)         401           1
test_one_arg[signature-cupy] (0001_459b8c0)     3.8849 (1.82)     4.4341 (1.64)     4.0874 (1.85)     0.1246 (2.12)     4.0341 (1.84)     0.2278 (5.04)        109;0  244.6541 (0.54)        231           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

---------------------------------------------------------------------------- benchmark 'test_one_arg[signature-device_array]': 2 tests -----------------------------------------------------------------------------
Name (time in ms)                                          Min               Max              Mean            StdDev            Median               IQR            Outliers       OPS            Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_one_arg[signature-device_array] (0001_459b8c0)     1.2409 (1.0)      1.9163 (1.0)      1.2736 (1.0)      0.0424 (1.0)      1.2686 (1.0)      0.0137 (1.0)         23;32  785.1979 (1.0)         646           1
test_one_arg[signature-device_array] (NOW)              1.8690 (1.51)     2.8435 (1.48)     1.9830 (1.56)     0.1331 (3.14)     1.9313 (1.52)     0.1464 (10.69)       32;12  504.2930 (0.64)        453           1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

----------------------------------------------------------------------------- benchmark 'test_one_arg[signature-torch]': 2 tests ----------------------------------------------------------------------------
Name (time in ms)                                   Min               Max              Mean            StdDev            Median               IQR            Outliers       OPS            Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_one_arg[signature-torch] (0001_459b8c0)     3.4580 (1.0)      3.7149 (1.0)      3.5206 (1.0)      0.0280 (1.0)      3.5193 (1.0)      0.0324 (1.0)          75;6  284.0446 (1.0)         259           1
test_one_arg[signature-torch] (NOW)              4.1100 (1.19)     4.6233 (1.24)     4.2768 (1.21)     0.1268 (4.53)     4.3273 (1.23)     0.2216 (6.83)         87;0  233.8189 (0.82)        224           1
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
======================================================================================== 12 passed in 11.54s =========================================================================================
```

</details>

cpcloud:
I managed to recover a good amount of perf of devicearray by avoiding the SMV conversion entirely and spoofing the interface.

cpcloud:
However there is still a slowdown of ~60%, but only in the many-args case (it's about 15% in the single argument case). This is much better than the previous commit which was upwards of 2.5x.

numba_cuda/numba/cuda/cudadrv/devicearray.py
```diff
@@ -392,13 +394,28 @@ def view(self, dtype):
             gpu_data=self.gpu_data,
         )
 
-    @property
+    @functools.cached_property
```
rparolin, line 398: Question: I'm surprised the cost of the cache fetch is less than performing the multiply.

cpcloud: It's 3 attribute accesses + a multiply versus a single attribute access after the first computation.

numba_cuda/numba/cuda/cudadrv/devicearray.py
```diff
@@ -934,6 +951,123 @@ def auto_device(obj, stream=0, copy=True, user_explicit=False):
...
+def _to_strided_memory_view(
+    obj, stream=0, copy: bool = True, user_explicit: bool = False
+) -> tuple[StridedMemoryView, bool]:
+    if _driver.is_device_memory(obj):
+        return obj._strided_memory_view_shim, False
+    elif not isinstance(
+        obj, (np.ndarray, _UNSUPPORTED_DLPACK_TYPES)
+    ) and hasattr(obj, "__dlpack__"):
+        # numpy arrays need to be copied to the device
+        # so we can't view them as SMVs until then
```
kkraus14, line 1023: SMVs can be used with host memory as opposed to device memory as well and have support for copying between host and device which would likely be more optimized then doing the checking and handling via Python code here?

May need to add some attributes to SMV / Buffer in order to properly introspect if it's backed by device memory or not, but I imagine it would be most efficient if we could essentially just always try to create an SMV from `obj` and then check if it's not on device and copy it to a newly allocated on device SMV if necessary?

cpcloud: Yes, that would probably be better, though I don't think it blocks this PR. My implementation here reflects the status quo, which was doing all the numpy checking and copying dance as well.

kkraus14:
So the results of where we are is that using CuPy Arrays has ~3x less latency, but using device arrays or torch tensors has ~60% more latency?

On the torch front https://github.com/NVIDIA/cuda-python/issues/1439 may help in bypassing the slow `__dlpack__` implementation.

cpcloud:
> So the results of where we are is that using CuPy Arrays has ~3x less latency, but using device arrays or torch tensors has ~60% more latency?

Almost. Will post new numbers in a bit.

> On the torch front [NVIDIA/cuda-python#1439](https://github.com/NVIDIA/cuda-python/issues/1439) may help in bypassing the slow `__dlpack__` implementation.

Passing `stream_ptr=-1` removes most of the additional torch overhead versus cupy, which I believe indicates that it's the stream synchronization done by torch (or `StridedMemoryView`, or both) that causes the slowness. I think that also means removing Python overhead won't help with that source of performance loss.

numba_cuda/numba/cuda/np/numpy_support.py
```diff
@@ -40,7 +49,7 @@ def strides_from_shape(
             ),
         )
     )
-    if order == "F":
+    if f_contiguous:
         return strides
     return strides[::-1]
```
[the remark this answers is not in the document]
cpcloud: This is fine.

numba_cuda/numba/cuda/cudadrv/devicearray.py
```diff
@@ -934,6 +951,123 @@ def auto_device(obj, stream=0, copy=True, user_explicit=False):
...
+        )
+        if config.CUDA_ARRAY_INTERFACE_SYNC and external_stream_ptr != -1:
+            ctx = devices.get_context()
+            ext_stream = ctx.create_external_stream(external_stream_ptr)
+            ext_stream.synchronize()
+        return smv, False
+    else:
+        array_obj = np.asanyarray(obj)
+        nbytes = array_obj.nbytes
+        stream_ptr = getattr(stream, "handle", stream)
```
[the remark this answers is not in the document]
cpcloud: These are always converted to a cuda core Stream before ever hitting this code path.

numba_cuda/numba/cuda/args.py
```diff
@@ -23,40 +26,61 @@ def to_device(self, retr, stream=0):
...
+                devptr = devary.ptr
+                if stream_ptr:
+                    driver.cuMemcpyDtoHAsync(
+                        hostptr, devptr, nbytes, stream_ptr
+                    )
+                else:
+                    driver.cuMemcpyDtoH(hostptr, devptr, nbytes)
+                    ctx = devices.get_context()
+                    stream = ctx.get_default_stream()
+                    stream.synchronize()
```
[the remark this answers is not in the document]
cpcloud: I believe this was the previous code's behavior, but I'll verify.

numba_cuda/numba/cuda/typing/typeof.py
```diff
@@ -311,40 +312,62 @@ def typeof_numpy_polynomial(val, c):
...
+@functools.lru_cache
+def _typeof_cuda_array_interface_cached(
+    *, dtype, shape, strides, readonly: bool
+):
     # Determine layout
+    ndim = len(shape)
     if not ndim:
         layout = "C"
     elif strides is None:
         layout = "C"
```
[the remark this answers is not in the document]
cpcloud: Are you sure about shape and strides being "very large"? NumPy has an upper bound of 64 on the number of dimensions, and therefore at most each of shape and strides can be 64 elements long (128 elements in total).

However, the cache key **space** is large because shape and strides elements are integers, but it's not because the tuples themselves are large.

128 is probably an okay starting point. I suspect that most kernels do not have many more than that number of arrays with unique shape and strides.

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -552,10 +551,18 @@ def _prepare_args(self, ty, val, stream, retr, kernelargs):
...
+            kernelargs.append(devary.ptr)
+            kernelargs.extend(shape := devary.shape)
+            kernelargs.extend(
+                (layout := devary._layout).strides_in_bytes
+                or numpy_support.strides_from_shape(
+                    shape=shape,
+                    itemsize=itemsize,
+                    c_contiguous=layout.is_contiguous_c,
+                    f_contiguous=layout.is_contiguous_f,
+                )
```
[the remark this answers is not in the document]
cpcloud: *grumbles* ... fine.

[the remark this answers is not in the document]
cpcloud: Actually, the fallback computation would also return an empty tuple, right?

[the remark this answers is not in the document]
cpcloud: Yes, if `strides_in_bytes` is the empty tuple, then `shape` will be the empty tuple, and the fallback computation would return the empty tuple.

Other than wasting a function call, why is that a "critical" error?

cpcloud:
@kkraus14 

- `-` is an improvement
- `+` is a regression

`cupy`:
- one arg: -2x
- many args: -3x

`torch`:
- one arg: +17%
- many args: +7% for explicit signature, -12% for dispatch

`device_array`:
- one arg: roughly +9%
- many args: roughly +60%

image

cpcloud:
I also added a benchmark demonstrating that the additional overhead with `device_array` is entirely due to having to construct a shim to fit device array information into an object that is API-compatible with the attributes we pull out of `StridedMemoryView` when preparing arguments for kernel launch.

Maybe there's some way that we can reduce that further, but I haven't looked into it.

numba_cuda/numba/cuda/args.py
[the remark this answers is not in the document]
cpcloud: Previously this called `_driver.device_to_host` which is this:

```python
def device_to_host(dst, src, size, stream=0):
    """
    NOTE: The underlying data pointer from the host data buffer is used and
    it should not be changed until the operation which can be asynchronous
    completes.
    """
    fn = driver.cuMemcpyDtoH
    args = (host_pointer(dst), device_pointer(src), size)

    if stream:
        fn = driver.cuMemcpyDtoHAsync
        args += (_stream_handle(stream),)

    fn(*args)
```

So, it's effectively the same code, just inlined and made to work with `cuda.core.StridedMemoryView`.

kkraus14:
The devicearray regressions are somewhat concerning, but given we are actively working towards deprecating it, I think it would still be worth it.

Do we have a sense on follow up work from here that helps to ameliorate the performance overheads related to torch?

cpcloud:
> The devicearray regressions are somewhat concerning, but given we are actively working towards deprecating it, I think it would still be worth it.
> 
> Do we have a sense on follow up work from here that helps to ameliorate the performance overheads related to torch?

At least some of the remaining overhead is related to stream synchronization, but that may be justified/useful in some cases I'm guessing.

After that, I'm not sure. It will require more investigation.

Just to make sure we're on the same page, our expectation is that if an array is on device then we should expect the kernel launch overhead to amount to a collection of relatively cheap attribute accesses. Is that correct?

kkraus14:
> At least some of the remaining overhead is related to stream synchronization, but that may be justified/useful in some cases I'm guessing.

My 2c: numba-cuda shouldn't be in the business of handling stream synchronization and that if someone is passing an array on a different stream through dlpack / CAI, it becomes their responsibility to launch the kernel on a stream that is synchronized with respect to the passed stream. This is likely a breaking change that would need to be clearly and loudly deprecated and subsequently removed.

> Just to make sure we're on the same page, our expectation is that if an array is on device then we should expect the kernel launch overhead to amount to a collection of relatively cheap attribute accesses. Is that correct?

Yes. Kernel launch latency is quite important where we should aim for less than 1us overhead.

cpcloud:
> My 2c: numba-cuda shouldn't be in the business of handling stream synchronization and that if someone is passing an array on a different stream through dlpack / CAI, it becomes their responsibility to launch the kernel on a stream that is synchronized with respect to the passed stream. This is likely a breaking change that would need to be clearly and loudly deprecated and subsequently removed.

Got it, yeah I don't really know enough about how this functionality is used or assumed to be used to have an informed opinion (yet!), but simply removing sync (in the torch case by passing -1 as the stream pointer to `__dlpack__`) saves a noticeable amount of time.

> Yes. Kernel launch latency is quite important where we should aim for less than 1us overhead.

Roger that, I _think_ we can get there if not very close.

numba_cuda/numba/cuda/cudadrv/devicearray.py
cpcloud: Going to resolve this! Let me know if there's still a question here.

numba_cuda/numba/cuda/cudadrv/devicearray.py
```diff
@@ -934,6 +951,123 @@ def auto_device(obj, stream=0, copy=True, user_explicit=False):
...
+def _make_strided_memory_view(obj, *, stream_ptr) -> StridedMemoryView:
+    if isinstance(obj, _UNSUPPORTED_DLPACK_TYPES) or (
+        isinstance(obj, np.ndarray)
+        and issubclass(obj.dtype.type, _UNSUPPORTED_DLPACK_TYPES)
+    ):
+        return StridedMemoryView.from_array_interface(obj)
+    return StridedMemoryView.from_any_interface(obj, stream_ptr=stream_ptr)
+
+
+class _StridedLayoutShim:
```
cpcloud, line 973: Seeing as how we're blocked on a cuda-core release, I'm going to try one last thing to speed this up.

kkraus14:
> Got it, yeah I don't really know enough about how this functionality is used or assumed to be used to have an informed opinion (yet!), but simply removing sync (in the torch case by passing -1 as the stream pointer to `__dlpack__`) saves a noticeable amount of time.

An option available to use is instead of actually synchronizing the stream (assuming we're calling `cuStreamSynchronize`), we could record a CUDA event on that stream and wait on that event in our launching stream. This waiting would then be done device side without blocking the CPU thread as oppposed to `cuStreamSynchronize` which does block the CPU thread. This would retain the current behavior, but again I would argue this is out of the scope of numba-cuda and should be done by a higher level library / framework using numba-cuda.

cpcloud:
I tried rolling some C-API classes for the two `Shim` objects I introduced here but there was virtually no difference in the benchmark versus defining them in Python.

numba_cuda/numba/cuda/cudadrv/devicearray.py
cpcloud: Defining classes using the C API doesn't meaningfully speed things up here.

cpcloud:
I included the pure C classes here for reference, but I am going to revert those changes because they don't really improve anything.

numba_cuda/numba/cuda/cext/_helpermod.c
```diff
@@ -54,17 +54,185 @@ build_c_helpers_dict(void)
...
+    StridedMemoryViewShim *self = (StridedMemoryViewShim *)op;
+    Py_XDECREF(self->ptr);
+    Py_XDECREF(self->shape);
+    Py_XDECREF(self->dtype);
+    Py_XDECREF(self->size);
+    Py_XDECREF(self->layout);
+    Py_TYPE(op)->tp_free(op);
+}
+
+static int StridedMemoryViewShim_init(PyObject *op, PyObject *args, PyObject *kwds) {
```
kkraus14: Not sure if any of the seen overhead is coming from here, but with `StridedMemoryView` being a cython cdef class, all of this information is available via the underlying C struct as native types and we could avoid the whole boxing, unboxing, boxing, unboxing of Python objects that I suspect is happening here.

numba_cuda/numba/cuda/cudadrv/devicearray.py
```diff
@@ -934,6 +952,123 @@ def auto_device(obj, stream=0, copy=True, user_explicit=False):
...
+        return smv, False
+    else:
+        array_obj = np.asanyarray(obj)
+        nbytes = array_obj.nbytes
+        stream_ptr = getattr(stream, "handle", stream)
+
+        ctx = devices.get_context()
+        # TODO: this might need to be slotted into the extensible memory
+        # manager API (EMM) somehow.
+        buf = ctx.device._dev.memory_resource.allocate(nbytes)
```
kkraus14: Does this get hit in the path someone calls `cuda.to_device(...)` or any of the other common host --> device codepaths? If so, I don't think we can merge this before addressing the TODO comment here.

numba_cuda/numba/cuda/cext/_helpermod.c
cpcloud: There's no boxing or unboxing here, these all remain as untouched pyobjects. The pure python version of this code isn't measurably slower, so I'm going to revert back to that.

numba_cuda/numba/cuda/cudadrv/devicearray.py
cpcloud: This is called:

- when launching kernels and copying from host to device in that context
- when copying from device to host for output arrays **from** a kernel.

Everything else is powered by `auto_device`, which is unchanged.

numba_cuda/numba/cuda/cext/_helpermod.c
cpcloud: I implemented this in Cython as well and construction is faster by 1.5x-2x, but attribute access is slower by 1.5x-2x.

Both operations are performed, but attribute access is more common. The constructor is only called once per instance because it's a `functools.cached_property` whereas these various attributes are accessed once per kernel launch call.

Slots is apparently the best we can do in this particular situation.

kkraus14: Sorry, I meant in the overlap pipeline leading to kernel launch, eventually we need to lower these Python objects down to native types in order to pass into the kernel via the numba-cuda ABI and that if we were able to keep things as native types from the beginning and never raise them back up to Python objects before the conversion down to the numba-cuda ABI for kernel launch that there would be a potential win. That's probably a mountain of work though.

numba_cuda/numba/cuda/cudadrv/devicearray.py
kkraus14: Summarizing offline conversation:
* Users using EMM are generally advanced users that wouldn't pass numpy arrays to numba-cuda kernels due to performance pitfalls so we're okay with this not respecting the EMM plugin
* Using the cuda.core memory_resource allocation is a change in behavior from the default numba-cuda allocator where the memory_resource allocation uses stream ordered allocations vs default numba-cuda uses `cuMemAlloc` which is a synchronous allocator. We need to make sure that we pass `stream` to the allocation here, otherwise we have a race in this allocation happening asynchronously on the default stream vs the `cuMemcpyHtoDAsync` below happening on the passed `stream`.

cpcloud: Fixed, stream is now being passed into `allocate`.

numba_cuda/numba/cuda/args.py
```diff
@@ -23,40 +26,61 @@ def to_device(self, retr, stream=0):
...
+        )
         if conv:
-            retr.append(lambda: devary.copy_to_host(self.value, stream=stream))
+            stream_ptr = int(getattr(stream, "handle", stream))
+
+            def copy_to_host(devary=devary, value=value, stream_ptr=stream_ptr):
+                hostary = _make_strided_memory_view(
+                    value, stream_ptr=stream_ptr
+                )
+                nbytes = value.nbytes
```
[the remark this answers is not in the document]
cpcloud: Fixed.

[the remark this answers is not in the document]
cpcloud: No longer valid.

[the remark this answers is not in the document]
cpcloud: No longer valid.

kkraus14 (approved):

kkraus14:
The implementation generally looks good to me. Would be good to get a final round of benchmarks to evaluate where the improvements / regressions are.

numba_cuda/numba/cuda/cext/_helpermod.c
cpcloud: If all the argument preparation can be done in cuda-core, down in Cython land, then yeah, that would be ideal.

It might be worth it to start planning that effort now if we think it will bear fruit.

kkraus14: Yes, 100%. We don't currently have `StridedMemoryView` plumbed into how we lower Python objects for kernel launch: https://github.com/NVIDIA/cuda-python/blob/main/cuda_core/cuda/core/_kernel_arg_handler.pyx but it absolutely could be added. I'm sure there's ABI convention things that need to be handled is the big thing.

cpcloud:
Ran this through an LLM to get a nice summary of the final benchmark results.

## High-level takeaways (NOW vs `0001_c91948d`)

- **CuPy:** big wins across the board (**~2–3× faster**).
- **device_array:** consistent regressions (**~10% slower** for one-arg, **~55–56% slower** for many-args).
- **Torch:** mixed — **dispatch got slightly faster**, but **signature got slower**.

---

## Mean time changes (lower is better)

### `test_many_args`

| Variant | Baseline mean (ms) | NOW mean (ms) | Change |
|---|---:|---:|---:|
| dispatch-cupy | 82.40 | 27.21 | **-67% (3.0× faster)** |
| dispatch-device_array | 6.77 | 10.48 | **+55% (1.55× slower)** |
| dispatch-torch | 70.60 | 64.31 | **-9% (1.10× faster)** |
| signature-cupy | 62.79 | 19.80 | **-68% (3.17× faster)** |
| signature-device_array | 6.91 | 10.78 | **+56% (1.56× slower)** |
| signature-torch | 49.94 | 54.77 | **+10% (1.10× slower)** |

### `test_one_arg`

| Variant | Baseline mean (ms) | NOW mean (ms) | Change |
|---|---:|---:|---:|
| dispatch-cupy | 5.91 | 2.90 | **-51% (2.0× faster)** |
| dispatch-device_array | 1.27 | 1.40 | **+11% slower** |
| dispatch-torch | 5.32 | 4.96 | **-7% faster** |
| signature-cupy | 4.38 | 2.27 | **-48% (1.9× faster)** |
| signature-device_array | 1.27 | 1.41 | **+11% slower** |
| signature-torch | 3.62 | 4.19 | **+16% slower** |

---

## Current absolute performance (NOW, mean time)

- **Many args:** `device_array` (~10–11 ms) is still fastest, then **CuPy (~20–27 ms)**, then **Torch (~55–64 ms)**.
- **One arg:** `device_array` (~1.4 ms) fastest, then **CuPy (~2.3–2.9 ms)**, then **Torch (~4.2–5.0 ms)**.

<details>
<summary><h2>Raw data</h2></summary>

```
------------------------------------------------------------------------------ benchmark 'test_many_args[dispatch-cupy]': 2 tests ------------------------------------------------------------------------------
Name (time in ms)                                    Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_many_args[dispatch-cupy] (NOW)              26.5499 (1.0)      28.4821 (1.0)      27.2123 (1.0)      0.5466 (1.0)      27.0107 (1.0)      0.7143 (1.0)           6;0  36.7481 (1.0)          22           1
test_many_args[dispatch-cupy] (0001_c91948d)     80.1594 (3.02)     84.9900 (2.98)     82.3980 (3.03)     1.5929 (2.91)     82.2436 (3.04)     2.3078 (3.23)          4;0  12.1362 (0.33)         10           1
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------- benchmark 'test_many_args[dispatch-device_array]': 2 tests ------------------------------------------------------------------------------
Name (time in ms)                                            Min                Max               Mean            StdDev             Median               IQR            Outliers       OPS            Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_many_args[dispatch-device_array] (0001_c91948d)      6.4306 (1.0)       7.0619 (1.0)       6.7681 (1.0)      0.1830 (1.06)      6.7565 (1.0)      0.3377 (1.82)         15;0  147.7515 (1.0)          30           1
test_many_args[dispatch-device_array] (NOW)              10.1953 (1.59)     10.9627 (1.55)     10.4819 (1.55)     0.1723 (1.0)      10.4611 (1.55)     0.1854 (1.0)           8;2   95.4026 (0.65)         33           1
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------ benchmark 'test_many_args[dispatch-torch]': 2 tests ------------------------------------------------------------------------------
Name (time in ms)                                     Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_many_args[dispatch-torch] (NOW)              63.9991 (1.0)      65.4999 (1.0)      64.3109 (1.0)      0.3777 (1.0)      64.2809 (1.0)      0.2062 (1.0)           1;1  15.5495 (1.0)          13           1
test_many_args[dispatch-torch] (0001_c91948d)     68.3296 (1.07)     73.8380 (1.13)     70.5973 (1.10)     1.8827 (4.98)     70.1094 (1.09)     2.6950 (13.07)         5;0  14.1648 (0.91)         11           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------ benchmark 'test_many_args[signature-cupy]': 2 tests ------------------------------------------------------------------------------
Name (time in ms)                                     Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_many_args[signature-cupy] (NOW)              19.3784 (1.0)      21.0086 (1.0)      19.7990 (1.0)      0.4120 (1.0)      19.6576 (1.0)      0.4164 (1.0)           9;4  50.5075 (1.0)          51           1
test_many_args[signature-cupy] (0001_c91948d)     60.1597 (3.10)     67.1960 (3.20)     62.7940 (3.17)     2.0793 (5.05)     63.2831 (3.22)     3.3902 (8.14)          6;0  15.9251 (0.32)         17           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------ benchmark 'test_many_args[signature-device_array]': 2 tests -------------------------------------------------------------------------------
Name (time in ms)                                             Min                Max               Mean            StdDev             Median               IQR            Outliers       OPS            Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_many_args[signature-device_array] (0001_c91948d)      6.6276 (1.0)      10.7366 (1.0)       6.9080 (1.0)      0.4227 (2.08)      6.7669 (1.0)      0.2620 (1.91)        16;15  144.7594 (1.0)         143           1
test_many_args[signature-device_array] (NOW)              10.5675 (1.59)     11.8660 (1.11)     10.7840 (1.56)     0.2029 (1.0)      10.7284 (1.59)     0.1375 (1.0)          10;6   92.7299 (0.64)         87           1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------ benchmark 'test_many_args[signature-torch]': 2 tests ------------------------------------------------------------------------------
Name (time in ms)                                      Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_many_args[signature-torch] (0001_c91948d)     48.2905 (1.0)      52.6185 (1.0)      49.9353 (1.0)      1.3081 (1.0)      49.6528 (1.0)      2.2560 (1.0)           8;0  20.0259 (1.0)          20           1
test_many_args[signature-torch] (NOW)              51.6698 (1.07)     57.1802 (1.09)     54.7654 (1.10)     1.8941 (1.45)     55.9016 (1.13)     3.7764 (1.67)          7;0  18.2597 (0.91)         19           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

----------------------------------------------------------------------------- benchmark 'test_one_arg[dispatch-cupy]': 2 tests ----------------------------------------------------------------------------
Name (time in ms)                                 Min               Max              Mean            StdDev            Median               IQR            Outliers       OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_one_arg[dispatch-cupy] (NOW)              2.8639 (1.0)      3.1819 (1.0)      2.9041 (1.0)      0.0440 (1.0)      2.8976 (1.0)      0.0236 (1.0)           2;4  344.3451 (1.0)          66           1
test_one_arg[dispatch-cupy] (0001_c91948d)     5.4198 (1.89)     6.7414 (2.12)     5.9068 (2.03)     0.3580 (8.14)     5.6831 (1.96)     0.5460 (23.13)         7;0  169.2971 (0.49)         43           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

----------------------------------------------------------------------------- benchmark 'test_one_arg[dispatch-device_array]': 2 tests ----------------------------------------------------------------------------
Name (time in ms)                                         Min               Max              Mean            StdDev            Median               IQR            Outliers       OPS            Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_one_arg[dispatch-device_array] (0001_c91948d)     1.2494 (1.0)      1.3038 (1.0)      1.2674 (1.0)      0.0213 (1.04)     1.2601 (1.0)      0.0212 (1.0)           1;0  789.0171 (1.0)           5           1
test_one_arg[dispatch-device_array] (NOW)              1.3719 (1.10)     1.4275 (1.09)     1.4005 (1.10)     0.0206 (1.0)      1.3975 (1.11)     0.0260 (1.23)          2;0  714.0520 (0.90)          5           1
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

---------------------------------------------------------------------------- benchmark 'test_one_arg[dispatch-torch]': 2 tests -----------------------------------------------------------------------------
Name (time in ms)                                  Min               Max              Mean            StdDev            Median               IQR            Outliers       OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_one_arg[dispatch-torch] (NOW)              4.9318 (1.0)      4.9933 (1.0)      4.9609 (1.0)      0.0142 (1.0)      4.9585 (1.0)      0.0149 (1.0)          15;2  201.5761 (1.0)          55           1
test_one_arg[dispatch-torch] (0001_c91948d)     4.9826 (1.01)     5.4021 (1.08)     5.3169 (1.07)     0.0916 (6.43)     5.3375 (1.08)     0.0466 (3.12)          3;3  188.0811 (0.93)         46           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

---------------------------------------------------------------------------- benchmark 'test_one_arg[signature-cupy]': 2 tests -----------------------------------------------------------------------------
Name (time in ms)                                  Min               Max              Mean            StdDev            Median               IQR            Outliers       OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_one_arg[signature-cupy] (NOW)              2.0933 (1.0)      2.5494 (1.0)      2.2743 (1.0)      0.0755 (1.0)      2.3032 (1.0)      0.0573 (1.0)        104;85  439.6938 (1.0)         412           1
test_one_arg[signature-cupy] (0001_c91948d)     3.9292 (1.88)     5.7554 (2.26)     4.3756 (1.92)     0.4172 (5.53)     4.2204 (1.83)     0.5594 (9.77)         61;4  228.5381 (0.52)        235           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

---------------------------------------------------------------------------- benchmark 'test_one_arg[signature-device_array]': 2 tests -----------------------------------------------------------------------------
Name (time in ms)                                          Min               Max              Mean            StdDev            Median               IQR            Outliers       OPS            Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_one_arg[signature-device_array] (0001_c91948d)     1.2242 (1.0)      1.9068 (1.0)      1.2712 (1.0)      0.0389 (1.08)     1.2642 (1.0)      0.0191 (1.0)         36;35  786.6426 (1.0)         621           1
test_one_arg[signature-device_array] (NOW)              1.3621 (1.11)     2.0377 (1.07)     1.4072 (1.11)     0.0359 (1.0)      1.4019 (1.11)     0.0281 (1.47)        52;13  710.6116 (0.90)        597           1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

----------------------------------------------------------------------------- benchmark 'test_one_arg[signature-torch]': 2 tests ----------------------------------------------------------------------------
Name (time in ms)                                   Min               Max              Mean            StdDev            Median               IQR            Outliers       OPS            Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_one_arg[signature-torch] (0001_c91948d)     3.3064 (1.0)      5.4537 (1.14)     3.6175 (1.0)      0.2671 (2.00)     3.5855 (1.0)      0.1975 (1.0)          14;8  276.4351 (1.0)         196           1
test_one_arg[signature-torch] (NOW)              3.9787 (1.20)     4.7722 (1.0)      4.1866 (1.16)     0.1335 (1.0)      4.2446 (1.18)     0.2394 (1.21)         95;1  238.8596 (0.86)        227           1
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

</details>

numba_cuda/numba/cuda/dispatcher.py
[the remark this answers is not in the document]
cpcloud: Meh

numba_cuda/numba/cuda/simulator/cudadrv/devicearray.py
```diff
@@ -99,6 +99,16 @@ def convert_fakes(obj):
         return call(*args, **kwargs)
 
 
+def _to_strided_memory_view(*args, **kwargs):
+    import pytest
+
+    raise pytest.skip("strided memory views not supported in the simulator")
+
+
+def _make_strided_memory_view(*args, **kwargs):
```
cpcloud: This is some pretty heavy cheating here, since many of the simulator tests no longer run because of the skip here.

That said, I'm a bit loathe to duplicate the work from `devicearray.py` here, as it's unclear that that adds any value.

@gmarkall What's the procedure for cases like this?

numba_cuda/numba/cuda/np/numpy_support.py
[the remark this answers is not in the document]
cpcloud: Fine already!

kkraus14:
From offline conversation: the torch benchmark may be creating torch tensors on a different stream than the CUDA default stream and causing stream synchronization.

cpcloud:
> the torch benchmark may be creating torch tensors on a different stream than the CUDA default stream and causing stream synchronization.

Traced through the pytorch `__dlpack__` implementation and confirmed this is **NOT** what is happening.

The current implementation uses the default stream from torch and does **NO** synchronization.

Whatever additional overhead there is appears to be from the various setup code and checks that happen before actually getting to the C API call that builds the DLPack capsule.

cpcloud:
The cudf tests are failing because their `__cuda_array_interface__` implementation `version` was zero for a bunch of objects. That's been fixed in the latest tag (`v26.02.00a` as of this comment). I'm going to try cloning that in our CI instead of the latest release's tag to see if we can that part of the CI here passing.

cpcloud:
I ended up trying to patch in the places where the version was 0 or 1 in cudf's `__cuda_array_interface__` instead, because there was a large amount of internal code that's changed in between 25.12 and 26.02, so installing 25.12 and running the (reduced) test suite of 26.02 seemed unlikely to work.

numba_cuda/numba/cuda/np/numpy_support.py
```diff
@@ -40,7 +51,7 @@ def strides_from_shape(
...
```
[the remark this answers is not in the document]
cpcloud: AFAIK they can only both be true in the 1D and 0D cases.

pyproject.toml
```diff
@@ -17,21 +17,19 @@ requires-python = ">=3.9"
 dependencies = [
     "numba>=0.60.0",
     "cuda-bindings>=12.9.1,<14.0.0",
-    "cuda-core>=0.3.2,<1.0.0",
+    "cuda-core>=0.5.1,<1.0.0",
```
cpcloud, line 20: The lower bound increased because the various APIs that this PR uses are only available in `cuda-core >= 0.5.1`.

The `cu12` and `cu13` constraints are removed, since the new `cuda-core >=0.5.1` constraint invalidates the previously declared constraints.

Issue #515, opened 2025-10-10
Incomplete Test Coverage in Numba-CUDA


atmnp:
Numba itself has a rather comprehensive test suite for testing functionality, but it's still unclear that all of the functionality provided by Numba works on CUDA devices via Numba-CUDA. This issue is meant to be a catch-all issue for tracking tests that can and should be ported to CUDA to flesh out the current limitations of Numba-CUDA vs. Numba, including features that require Numba-CUDA's experimental NRT. This list is not comprehensive and will be updated as more tests are determined to be desirable to be ported to the CUDA target.

Porting the following tests to the CUDA target would be beneficial for ensure a more comprehensive test coverage of Numba-CUDA:
- `test_array_manipulation`
- `test_asnumbatype`
- `test_auto_constants`
- `test_builtins`
- `test_casting`
- `test_cgutils`
- `test_chained_assign`
- `test_compile_cache`
- `test_compiler_flags`
- `test_compiler_lock`
- `test_complex`
- `test_conditions_as_predicates`
- `test_cli`
- `test_closure`
- `test_conditions_as_predicates`
- `test_config`
- `test_conversion`
- `test_copy_propagate`
- `test_datamodel`

CodersAcademy006:
Hey @atmnp is this issue still working? I already have some test cases which are ready to be converted as PR.

atmnp:
If you have PRs with some of the test files from this list or any others from the numba test-suite that should be ported, please feel free to submit them for review!

maifeeulasad:
Willing to work on this one. Can I get assigned to this?

CodersAcademy006:
@maifeeulasad Hey!! thank you for your interest. I have made some path for the test cases above and would love to have a review from you. Also i have not covered all the test cases so you are free to choose one and continue. But first i think it's better to resolve the previous ones and then look for further changes.

Thank You!!

CodersAcademy006:
Hi @atmnp,

I’ve been working through the test modules listed in this issue and have submitted PRs covering a large subset where the functionality has clear, device-side CUDA semantics (array manipulation, control flow, casting, builtins, complex numbers, constants, closures, etc.).

[...]

PR #706, opened 2026-01-08
cuda/tests: add array manipulation tests for cudasim and grid-stride correctness

CodersAcademy006:
# cuda/tests: add array manipulation tests for cudadrv and cudapy

## Summary

Adds two new test suites for CUDA array manipulation as part of #515.

### `cudadrv/test_device_array_manipulation.py`
Tests for the `DeviceArray` API (no kernels):
- Shape, dtype, size, ndim attributes
- `to_device` / `copy_to_host` roundtrip
- `copy_to_device` into a pre-allocated array
- Device-to-device copy
- `device_array_like`
- Multidimensional shape consistency
- Memory lifetime across multiple kernel launches
- Device pointer stability across kernel launches (`@skip_on_cudasim`)
- `reshape`, `ravel`, `view`
- Step slicing (`arr[::2]`) and multidimensional slicing
- `xfail`: boolean indexing, fancy indexing

### `cudapy/test_array_manipulation.py`
Tests for array operations inside CUDA kernels:
- Basic fill via kernel
- Elementwise addition
- Integer `getitem`/`setitem` inside a kernel
- 2D multidimensional indexing
- Dtype casting (`int32 → float32`, `float32 → int32`, mixed)
- Shape attribute access inside a kernel
- Grid-stride loop correctness (intentionally over-provisioned grid)

## Notes

- All tests pass on the CUDA simulator except `test_memory_pointer_stability`
  (simulator does not model real device pointers).
- No production code changes — test coverage only.
- Contributes to #515.


gmarkall (commented):
Thanks for the PR! I see you have several PRs in-flight - I'll work through one at a time, and it may be worth checking if the feedback in the current PR applies to any of the other PRs too. 

Some initial comments and questions on this PR:

- The test is in the wrong location - it is in `numba/cuda/tests/test_array_manipluation.py`. Tests for Numba-CUDA are all in `numba_cuda/numba/cuda/tests`. We don't put tests directly in this directory though, but instead a subdirectory depending on what the test is testing. Most of the tests in this PR would be appropriate for the `cudapy` subdirectory, but the tests that test parts of the `DeviceArray` API would make more sense under `cudadrv` as its implementation is part of the `cudadrv` module.
- The tests marked as skipping on the CUDA simulator don't actually fail when run under the CUDA simulator. Generally if a test can pass on the CUDA simulator, it should be allowed to run under it, in order to ensure parity between the simulator and real hardware as much as possible.
- The PR description seems to be at odds with what the PR does. To my reading, the PR is adding some new tests; the description claims it is fixing tests, and then the descriptions that follow from that introduction then seem not to have any connection with addition of code. Can you clarify this discrepancy please?
- The comments in the test code suggest they are adapted from "the CPU array manipulation suite". I see in the CPU target for Numba, there is a class called `TestArrayManipulation`, but I don't see much correlation between it and the tests added here. What part of the CPU test suite is this derived from?

CodersAcademy006:
Hey @gmarkall, thanks a ton for the review and for going through this one step by step – really helps a lot, especially since I've got a few PRs open right now.

I pushed the updates you asked for:

- **Split the tests** the way you suggested:  
  `cudadrv/test_device_array_manipulation.py` for the pure DeviceArray stuff (shape, dtype, size, copies, lifetime)  
  `cudapy/test_array_manipulation.py` for the kernel + indexing + elementwise ops side  

- **Dropped almost all** the `@skip_on_cudasim` lines. Kept just two where it actually makes sense (memory lifetime and the grid-stride one), everything else should run happily in the simulator now.

- **Fixed up** the PR title and description so it actually says what it does (adding new tests, not patching old ones). Also threw in a couple inline comments about where the CPU test inspiration came from.

About the CPU connection – yeah, it's mostly taking the shape/structure/intent from `numba/tests/test_array_manipulation.py` (the TestArrayManipulation class), but rewritten to fit CUDA realities (kernels instead of Python loops, no views or reshapes, etc.). Not a 1:1 copy, but the spirit is the same for the parts that actually work on device arrays.

Locally I'm on Windows without a GPU, so the editable install is fighting me a bit with the C extensions and I couldn't run the full sim suite yet. But the code sticks pretty close to how the existing tests are written (like test_device_array.py and test_datetime.py), so I'm pretty confident the non-skipped stuff will be fine in CI.

If this looks alright to you now, would love to get your 👍 so we can land it and move on to the ufunc/reduction tests for #515.

Thanks again man, Please correct me otherwise!

gmarkall:
> Locally I'm on Windows without a GPU, so the editable install is fighting me a bit with the C extensions and I couldn't run the full sim suite yet.

If you're having issues getting the install to work, feel free to open an issue with the details of what you're encountering... Perhaps there is some issue with editable installs on Windows that we never noticed because most development is done on Linux.

numba_cuda/numba/cuda/tests/cudadrv/test_device_array_manipulation.py
```diff
@@ -0,0 +1,129 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
gmarkall:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

numba_cuda/numba/cuda/tests/cudapy/test_array_manipulation.py
```diff
@@ -0,0 +1,212 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
gmarkall:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

numba_cuda/numba/cuda/tests/cudadrv/test_device_array_manipulation.py
```diff
@@ -0,0 +1,129 @@
...
+    i = cuda.grid(1)
+    if i < a.size:
+        a[i] = v
+
+
+class TestDeviceArrayManipulation(CUDATestCase):
+    """Tests for DeviceArray memory behavior and shape semantics."""
+
+    @skip_on_cudasim("Device memory lifetime tracking requires real device memory")
+    def test_memory_lifetime(self):
```
gmarkall, line 29: I think if I were aiming to verify the lifetime of the memory, I'd probably check the pointer is unchanged after each kernel as well. I can imagine this would still pass if the second kernel operated on some different memory to the first.

You can get the pointer value with `arr.gpu_data.device_ctypes_pointer.value`.

numba_cuda/numba/cuda/tests/cudadrv/test_device_array_manipulation.py
```diff
@@ -0,0 +1,129 @@
...
+def set_val_kernel(a, v):
+    i = cuda.grid(1)
+    if i < a.size:
+        a[i] = v
+
+
+class TestDeviceArrayManipulation(CUDATestCase):
+    """Tests for DeviceArray memory behavior and shape semantics."""
+
+    @skip_on_cudasim("Device memory lifetime tracking requires real device memory")
```
gmarkall: For info: the test as-written does not need to be skipped in the simulator. However, if the suggestion to check the pointer values is followed, then this skip would need to remain because I don't think the simulator models pointers to the data.

numba_cuda/numba/cuda/tests/cudadrv/test_device_array_manipulation.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def test_copy_to_device_existing_array(self):
+        """Verify copy_to_device into pre-allocated device array."""
+        host = np.arange(50, dtype=np.int32)
+        dev = cuda.device_array(50, dtype=np.int32)
+        dev.copy_to_device(host)
+        result = dev.copy_to_host()
+        self.assertTrue(np.array_equal(host, result))
+
+    @skip_on_cudasim("Device-to-device copy requires real device memory")
```
gmarkall: This skip does not seem necessary - the test should pass on the simulator.

numba_cuda/numba/cuda/tests/cudadrv/test_device_array_manipulation.py
```diff
@@ -0,0 +1,129 @@
...
+    def test_multidimensional_shape_consistency(self):
+        """Verify shape consistency for multidimensional arrays."""
+        shapes = [(10,), (5, 8), (3, 4, 5), (2, 3, 4, 5)]
+        for shape in shapes:
+            arr = cuda.device_array(shape, dtype=np.float32)
+            self.assertEqual(arr.shape, shape)
+            self.assertEqual(arr.ndim, len(shape))
+            self.assertEqual(arr.size, np.prod(shape))
+
+    @pytest.mark.xfail(reason="reshape not supported on CUDA device arrays")
```
gmarkall: All of these tests marked `xfail` pass ("unexpectedly"). Tests need to either fail and be marked xfail, or not be marked xfail.

numba_cuda/numba/cuda/tests/cudapy/test_array_manipulation.py
```diff
@@ -0,0 +1,212 @@
...
+            if cuda.threadIdx.x == 0 and cuda.blockIdx.x == 0:
+                out[0] = a.shape[0]
+                out[1] = a.shape[1]
+
+        out = cuda.device_array(2, dtype=np.int32)
+        check_shape[1, 1](arr, out)
+        host = out.copy_to_host()
+        self.assertEqual(tuple(host), shape)
+
+    @skip_on_cudasim("True grid-stride loop semantics require real device execution")
```
gmarkall: This test passes on the  simulator, so the skip is not necessary.

numba_cuda/numba/cuda/tests/cudapy/test_array_manipulation.py
```diff
@@ -0,0 +1,212 @@
...
+
+        threads = 64
+        blocks = (n + threads - 1) // threads
+        fill_kernel[blocks, threads](arr, 4.25)
+
+        result = arr.copy_to_host()
+        self.assertTrue(np.all(result == 4.25))
+
+    def test_elementwise_add(self):
+        """Elementwise addition - basic arithmetic works in simulator."""
```
gmarkall: This test is not constrained to the simulator:
```suggestion
        """Elementwise addition - basic arithmetic works"""
```

numba_cuda/numba/cuda/tests/cudapy/test_array_manipulation.py
```diff
@@ -0,0 +1,212 @@
...
+    i = cuda.grid(1)
+    if i < out.size:
+        out[i] = a[i] + b[i]
+
+
+class TestArrayManipulation(CUDATestCase):
+    """Tests for array manipulation operations in CUDA kernels."""
+
+    def test_fill_basic(self):
+        """Basic elementwise fill operation - works in simulator."""
```
gmarkall:
```suggestion
        """Basic elementwise fill operation"""
```

numba_cuda/numba/cuda/tests/cudapy/test_array_manipulation.py
```diff
@@ -0,0 +1,212 @@
...
+
+        threads = 64
+        blocks = (n + threads - 1) // threads
+        add_kernel[blocks, threads](da, db, dout)
+
+        result = dout.copy_to_host()
+        self.assertTrue(np.allclose(result, a + b))
+
+    def test_integer_getitem_setitem(self):
+        """Direct integer indexing in kernel - works in simulator."""
```
gmarkall: I noticed the "works in simulator" comment is in every docstring - please remove it from all of them.

numba_cuda/numba/cuda/tests/cudapy/test_array_manipulation.py
```diff
@@ -0,0 +1,212 @@
...
+        threads = (4, 4)
+        blocks = ((shape[0] + threads[0] - 1) // threads[0],
+                  (shape[1] + threads[1] - 1) // threads[1])
+
+        write_2d[blocks, threads](dev)
+        result = dev.copy_to_host()
+        self.assertEqual(result[2, 3], 2 * shape[1] + 3)
+
+    @pytest.mark.xfail(reason="Advanced slicing not supported on CUDA device arrays")
+    def test_advanced_slicing(self):
```
gmarkall: This test unexpectedly passes.

numba_cuda/numba/cuda/tests/cudapy/test_array_manipulation.py
```diff
@@ -0,0 +1,212 @@
...
+        """Advanced slicing (e.g., ::2) is not supported on device arrays."""
+        arr = cuda.to_device(np.arange(10))
+        _ = arr[::2]
+
+    @pytest.mark.xfail(reason="Multidimensional slicing not supported on CUDA device arrays")
+    def test_multidimensional_slicing(self):
```
gmarkall: This test unexpectedly passes.

numba_cuda/numba/cuda/tests/cudapy/test_array_manipulation.py
```diff
@@ -0,0 +1,212 @@
...
+        """Multidimensional slicing not supported on device arrays."""
+        arr = cuda.to_device(np.arange(16).reshape(4, 4))
+        _ = arr[:, ::-1]
```
gmarkall: This is not testing functionality in a kernel as the docstring for the test module suggests.

numba_cuda/numba/cuda/tests/cudapy/test_array_manipulation.py
```diff
@@ -0,0 +1,212 @@
...
+
+        write_2d[blocks, threads](dev)
+        result = dev.copy_to_host()
+        self.assertEqual(result[2, 3], 2 * shape[1] + 3)
+
+    @pytest.mark.xfail(reason="Advanced slicing not supported on CUDA device arrays")
+    def test_advanced_slicing(self):
+        """Advanced slicing (e.g., ::2) is not supported on device arrays."""
+        arr = cuda.to_device(np.arange(10))
+        _ = arr[::2]
```
gmarkall: This is not testing functionality in a kernel as the docstring for the test module suggests.

numba_cuda/numba/cuda/tests/cudapy/test_array_manipulation.py
```diff
@@ -0,0 +1,212 @@
...
+
+    @pytest.mark.xfail(reason="Multidimensional slicing not supported on CUDA device arrays")
+    def test_multidimensional_slicing(self):
+        """Multidimensional slicing not supported on device arrays."""
+        arr = cuda.to_device(np.arange(16).reshape(4, 4))
+        _ = arr[:, ::-1]
+
+    @pytest.mark.xfail(reason="Boolean indexing not supported on CUDA device arrays")
+    def test_boolean_indexing(self):
+        """Boolean mask indexing is not supported on device arrays."""
```
gmarkall:
```suggestion
        """Boolean mask indexing is not supported in kernels."""
```

numba_cuda/numba/cuda/tests/cudapy/test_array_manipulation.py
```diff
@@ -0,0 +1,212 @@
...
+        arr = cuda.to_device(np.arange(10))
+        mask = np.array([True, False] * 5)
+        _ = arr[mask]
+
+    @pytest.mark.xfail(reason="Fancy indexing with arrays not supported on CUDA device arrays")
+    def test_fancy_indexing(self):
+        """Fancy indexing with integer arrays not supported on device arrays."""
```
gmarkall:
```suggestion
        """Fancy indexing with integer arrays not supported in kernels."""
```

numba_cuda/numba/cuda/tests/cudapy/test_array_manipulation.py
```diff
@@ -0,0 +1,212 @@
...
+            i = cuda.grid(1)
+            if i < src.size:
+                dst[i] = int(src[i])
+
+        threads = 32
+        blocks = (n + threads - 1) // threads
+        cast_int_to_float[blocks, threads](da, dout1)
+        cast_float_to_int[blocks, threads](db, dout2)
+
+        self.assertTrue(np.allclose(dout1.copy_to_host(), a.astype(np.float32)))
```
gmarkall: It's better to use the `np.testing` functions, as they provide better information the event of a failure, e.g.:


```suggestion
        np.testing.assert_allclose(dout1.copy_to_host(), a.astype(np.float32))
```

(All checks like this should be modified to use `np.testing`)

numba_cuda/numba/cuda/tests/cudapy/test_array_manipulation.py
```diff
@@ -0,0 +1,212 @@
...
+                out[1] = a.shape[1]
+
+        out = cuda.device_array(2, dtype=np.int32)
+        check_shape[1, 1](arr, out)
+        host = out.copy_to_host()
+        self.assertEqual(tuple(host), shape)
+
+    @skip_on_cudasim("True grid-stride loop semantics require real device execution")
+    def test_grid_stride_correctness(self):
+        """Grid-stride loop pattern - requires actual CUDA grid execution."""
```
gmarkall:
```suggestion
        """Grid-stride loop pattern"""
```

gmarkall (changes requested):
Thanks for the updates. I think there are still some items to address here, which I have marked on the diff. Some issues are functional, and others are about clarifying the intent of the tests - it's important that the commentary surrounding test code is clear and accurate, so as to prevent confusion when reading test cases (or debugging failures).

numba_cuda/numba/cuda/tests/cudadrv/test_device_array_manipulation.py
CodersAcademy006: I see test_reshape/view/ravel pass on the simulator — should I remove xfail, or keep it until verified on hardware?

CodersAcademy006:
I see `test_reshape`, `test_view`, and `test_ravel` may pass on the simulator — should I remove `xfail`, or keep it until verified on hardware?

numba_cuda/numba/cuda/tests/cudadrv/test_device_array_manipulation.py
CodersAcademy006: Good point — I've added the pointer check as a separate test, `test_memory_pointer_stability`, which captures `arr.gpu_data.device_ctypes_pointer.value` before the first kernel launch and asserts it's unchanged after each subsequent launch. I kept `test_memory_lifetime` focused on value correctness and put the allocation-identity check in its own test so the failure modes are distinct — a value failure vs. pointer drift point to different bugs. `test_memory_pointer_stability` is decorated @skip_on_cudasim since the simulator doesn't model real device pointers, as you noted earlier.

CodersAcademy006:
@gmarkall — pushed a squashed update addressing all remaining review items:

`test_boolean_indexing` / `test_fancy_indexing`: restored `@pytest.mark.xfail` and removed incorrect positive assertions — these operations are unsupported so the tests document the failure, not assert success
Replaced all `assertTrue(np.array_equal/allclose)` with `np.testing.assert_array_equal` / `np.testing.assert_allclose` throughout both files
`test_reshape` and `test_ravel` now seed with real data and assert content, not just shape
`test_device_to_device_copy`: removed unnecessary `@skip_on_cudasim`
`test_grid_stride_correctness`: removed @skip_on_cudasim, fixed docstring
Slicing and indexing tests moved from `cudapy` to `cudadrv` where they belong
Replied inline on `test_memory_lifetime` explaining the split into two tests

One open question: `test_reshape`, `test_ravel`, `test_view` pass on the simulator — should these remain as positive tests, or revert to `xfail` until verified on real hardware?

CodersAcademy006:
Pushed an update addressing the review. Moved the tests into `cudapy`/`cudadrv` per the split you described, dropped the `skip_on_cudasim` decorators on tests that pass under the simulator (kept only the pointer one), and fixed the `xfail` tests so they actually fail — boolean/fancy indexing now hit the unsupported path. Also added a pointer-stability check to the lifetime test, switched the asserts to `np.testing`, and corrected the description (it adds new tests rather than fixing existing ones).

Re-review whenever you get a chance, thanks.

CodersAcademy006:
Closing this in light of Numba-CUDA entering maintenance mode (#902) — new test-coverage additions are out of scope here going forward. The array-manipulation tests from this PR have already been ported to Numba-CUDA-MLIR as the first module (see NVIDIA/numba-cuda-mlir#215). Thanks to @gmarkall for the thorough review — the feedback carried over directly to the ported version.

PR #708, opened 2026-01-08
fix: enable flake8-bugbear lints and fix found problems

cpcloud:
Enable [`flake8-bugbear`](https://docs.astral.sh/ruff/rules/#flake8-bugbear-b) ruff lints and fix problems.

### DIFF
--- a/numba_cuda/numba/cuda/deviceufunc.py
+++ b/numba_cuda/numba/cuda/deviceufunc.py
@@ -131,7 +131,7 @@ def _multi_broadcast(*shapelist):
         for i, each in enumerate(others, start=1):  # noqa: B007
             result = _pairwise_broadcast(result, each)
     except ValueError:
-        raise ValueError("failed to broadcast argument #{i:d}")
+        raise ValueError(f"failed to broadcast argument #{i:d}")
     else:
         return result
 


cpcloud:
Actually a decent amount of unnecessary API calls happening in these modules, mostly `enumerate`s and pulling out keys or values of `dict`s without using them.

numba_cuda/numba/cuda/np/npyfuncs.py
kkraus14, line 41: Should these throw instead of asserting as well?

numba_cuda/numba/cuda/tests/cudadrv/test_streams.py
```diff
@@ -94,34 +94,5 @@ async def test_cancelled_future(self):
...
 
-@skip_on_cudasim("CUDA Driver API unsupported in the simulator")
-class TestFailingStream(CUDATestCase):
-    # This test can only be run in isolation because it corrupts the CUDA
-    # context, which cannot be recovered from within the same process. It is
-    # left here so that it can be run manually for debugging / testing purposes
-    # - or may be re-enabled if in future there is infrastructure added for
-    # running tests in a separate process (a subprocess cannot be used because
-    # CUDA will have been initialized before the fork, so it cannot be used in
-    # the child process).
```
kkraus14, line 105: Why did we remove this test?

numba_cuda/numba/cuda/deviceufunc.py
```diff
@@ -128,10 +128,10 @@ def _multi_broadcast(*shapelist):
     result = shapelist[0]
     others = shapelist[1:]
     try:
-        for i, each in enumerate(others, start=1):
+        for i, each in enumerate(others, start=1):  # noqa: B007
             result = _pairwise_broadcast(result, each)
     except ValueError:
-        raise ValueError("failed to broadcast argument #{0}".format(i))
+        raise ValueError("failed to broadcast argument #{i:d}")
```
kkraus14: Should this be an f-string?

numba_cuda/numba/cuda/tests/cudadrv/test_streams.py
cpcloud: It never runs because it's skipped unconditionally with `unittest.skip`.

numba_cuda/numba/cuda/np/npyfuncs.py
cpcloud: Depends on whether these are validating input for users or making assertions about invariants. If they are asserting invariants then I would leave them. If users can provide "typical" inputs that can violate these assertions then we should make them proper `raise`s.

numba_cuda/numba/cuda/deviceufunc.py
cpcloud: Yes, thanks.

cpcloud: Done.

PR #713, opened 2026-01-12
chore(dev): build pixi using rattler

### LINKED ISSUE #685: CI: Review/align conda test strategy
I've been curious how the conda tests are done in this repo, and IIUC there are serious issues that we should rectify.

## Issues
1. The conda test workflow (`conda-python-tests.yaml`) does not guarantee to reuse the build artifacts generated from the build workflow (`conda-python-build.yaml`). In fact, the former and the latter run in parallel, instead of depending on the latter. It is a waste of CI resources.
<details>
<summary>Click me for Copilot-generated analysis</summary>

Looking at this workflow file, I can see it's designed to run tests, but I need to check how it actually depends on `conda-python-build.yaml` to get the build artifacts.  Let me search for related workflow files and understand the dependency chain. Now let me look at how these workflows are typically called together: Based on my analysis of the workflow files and scripts, I can now explain how `conda-python-tests. yaml` depends on `conda-python-build.yaml` to get build artifacts: 

## How the Dependency Works

The `conda-python-tests.yaml` workflow **does NOT have an explicit dependency** on `conda-python-build.yaml` in the traditional GitHub Actions sense (using `needs:`). Instead, they work together through a **shared artifact storage pattern**:

### 1. **Orchestrated by Parent Workflows**

Looking at the calling workflows (`ci.yaml` and `pr.yaml`), the two workflows are called separately:

```yaml
build-conda: 
  needs: 
    - compute-matrix
  uses: ./.github/workflows/conda-python-build.yaml
  with:
    build_type: pull-request
    script: "ci/build_conda.sh"
    matrix: ${{ needs.compute-matrix.outputs.BUILD_MATRIX }}

test-conda:
  needs: 
    - compute-matrix  # Note: does NOT depend on build-conda
  uses: ./. github/workflows/conda-python-tests.yaml
  with:
    build_type: pull-request
    script: "ci/test_conda.sh"
    run_codecov: false
    matrix: ${{ needs.compute-matrix.outputs.TEST_MATRIX }}
```

**Key observation**: `test-conda` doesn't list `build-conda` in its `needs: ` section, meaning they can run in parallel! 

### 2. **How Artifacts are Actually Retrieved**

The dependency is handled through **conda channel repositories** and **pixi package management**:

1. **Build stage** (`conda-python-build.yaml`):
   - Runs `ci/build_python.sh` (or `ci/build_conda.sh`)
   - Uploads artifacts to `/tmp/conda-bld-output` (line 110 in the build workflow)
   - Publishes as GitHub Actions artifact named `conda-repo-py${{ matrix.PY_VER }}-${{ matrix.ARCH }}`

2. **Test stage** (`conda-python-tests.yaml`):
   - Uses **pixi** to manage the conda environment
   - The key dependency is in `pixi.toml` line 13: `numba-cuda = { path = "." }`
   - This tells pixi to build/install `numba-cuda` from the local source
   - Pixi builds the package fresh in the test environment rather than downloading pre-built artifacts

### 3. **The Real Mechanism**

Looking at the `pixi.toml` configuration, the test workflow: 
- Sets up a pixi environment (line 112-114 in test workflow)
- Installs `numba-cuda` as an editable/local package via `[dependencies] numba-cuda = { path = "." }` (pixi.toml line 13)
- **Builds the package from source in the test container** rather than using pre-built binaries

### Summary

The workflows **don't directly share artifacts** in the traditional sense. Instead:

1. Both workflows check out the repository source code
2. The build workflow creates conda packages and uploads them as artifacts (for potential anaconda publishing)
3. The test workflow **rebuilds the package from source** using pixi in its own environment
4. This is why they can run in parallel - tests don't actually need the build artifacts from the build job

The architecture follows a "build-from-source" testing pattern rather than "build-once-test-artifacts" pattern, which ensures tests validate the actual build process rather than just pre-built binaries. 

</details>

2. More importantly, [the recipe that we test](https://github.com/NVIDIA/numba-cuda/blob/main/conda/recipes/numba-cuda/meta.yaml) deviates too much from [the source-of-truth (conda-forge) recipe](https://github.com/conda-forge/numba-cuda-feedstock/blob/main/recipe/recipe.yaml). Even if we fix Issue No.1 above, the build workflow still does not reflect the end user experience. 

## Root cause analysis
I think what we're seeing is a technical debt due to reusing RAPIDS CI. The RAPIDS packages
- are not published on `conda-forge`, only on `rapidsai`
- are built in RAPIDS CI, instead of conda-forge public CI

whereas within CUDA Python program we 
- only publish on `conda-forge`, and copy the artifacts to `nvidia` (https://github.com/NVIDIA/cuda-python/issues/1203)
- always use conda-forge public CI
   - currently the NVIDIA/cuda-python repo lacks CI coverage for conda packages (https://github.com/NVIDIA/cuda-python/issues/280)

Since the old assumptions no longer hold, the gap to reality is widened.

## Discussions
I think the RAPIDS-based workflows need to be nuked, but we should discuss how to build a new CI for CUDA Python projects. I have some thoughts but let's reconvene after the holidays 🙂

cpcloud:
Move pixi build to use upstream recipe. The only change here is to remove the stdlib(c) bit, because pixi fails to parse it. We may also be able to remove it upstream.

Closes #685.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -8,6 +8,12 @@ preview = ["pixi-build"]
 [workspace.build-variants]
 python = ["3.10.*", "3.11.*", "3.12.*", "3.13.*", "3.14.*"]
 
+[workspace.target.linux.build-variants]
+c_stdlib = ["sysroot"]
+
+[workspace.target.win-64.build-variants]
+c_stdlib = ["vs"]
+
 [dependencies]
 # source
 numba-cuda = { path = "." }
--- a/recipe.yml
+++ b/recipe.yml
@@ -20,6 +20,7 @@ build:
 
 requirements:
   build:
+    - ${{ stdlib('c') }}
     - ${{ compiler('c') }}
     - ${{ compiler('cxx') }}
   host:
--- a/pixi.lock
+++ b/pixi.lock
@@ -213,7 +213,6 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-nvcc_linux-64-12.0.76-hba56722_12.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-nvrtc-12.0.76-hd3aeb46_2.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-pathfinder-1.3.3-pyhcf101f3_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-python-12.9.5-pyh698daf1_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-version-12.0-hffde075_3.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/distlib-0.4.0-pyhd8ed1ab_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/exceptiongroup-1.3.1-pyhd8ed1ab_0.conda
@@ -286,7 +285,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/yaml-0.2.5-h280c20c_3.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/zstd-1.5.7-hb78ec9c_6.conda
       - conda: .
-        build: py310h04c9772_0
+        build: py310h08333af_0
       - pypi: https://files.pythonhosted.org/packages/5c/40/69ca9ea803303e14301fff9d4931b6d080b9603e134df0419c55e9764df4/filecheck-1.0.3-py3-none-any.whl
       - pypi: https://files.pythonhosted.org/packages/10/b1/8938e8830b0ee2e167fc75a094dea766a1152bde46752cd9bfc57ee78a82/ml_dtypes-0.5.4-cp310-cp310-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl
       linux-aarch64:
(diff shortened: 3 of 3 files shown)


recipe.yml
```diff
@@ -0,0 +1,64 @@
...
+
+extra:
+  recipe-maintainers:
+    - rparolin
+    - conda-forge/cuda-python
+    - leofang
+    - bdice
+    - gmarkall
+    - vyasr
+    - kkraus14
```
kkraus14, line 66: Is this needed for here or is this just copy pasted from conda-forge?

cpcloud: Copy pasted

cpcloud: Trying to get this to a point where we can copy paste this back into conda-forge -> compute the hash -> submit a PR to conda-forge.

pixi.toml
```diff
@@ -8,6 +8,12 @@ preview = ["pixi-build"]
 [workspace.build-variants]
 python = ["3.10.*", "3.11.*", "3.12.*", "3.13.*", "3.14.*"]
 
+[workspace.target.linux.build-variants]
+c_stdlib = ["sysroot"]
+
+[workspace.target.win-64.build-variants]
+c_stdlib = ["vs"]
```
cpcloud, line 15: @kkraus14 Are these the build variants we want here?

kkraus14: Based on https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/conda_build_config.yaml#L11-L14, yes, this is what we want

kkraus14 (approved):

cpcloud:
Hm, this shouldn't have gone in, tests were still failing. Reverting.

PR #717, opened 2026-01-13
Move `CallConv` from `CUDAContext` to `FunctionDescriptor`

isVoid:
Today, calling conventions are defined globally per compilation context. This makes it hard to switch flexibly between the Numba ABI and the C ABI when declaring external functions. It also explains the need for the kernel “fixup” logic: CUDA kernels are fundamentally C-ABI, but have historically been forced through the Numba ABI path.

This PR moves calling-convention selection to a more granular level, removing these limitations and eliminating the kernel fixup workaround. It also lays the groundwork for users to plug in additional calling-convention implementations in the future.

### DIFF
--- a/ci/test_thirdparty_awkward.sh
+++ b/ci/test_thirdparty_awkward.sh
@@ -74,6 +74,21 @@ index 39080a34..0eb3940f 100644
      array = rng.integers(50, size=1000)
 EOF
 
+patch -p1 <<'EOF'
+diff --git a/pyproject.toml b/pyproject.toml
+index 78ecfba9..80a25474 100644
+--- a/pyproject.toml
++++ b/pyproject.toml
+@@ -136,6 +136,7 @@ filterwarnings = [
+     "ignore:.*np\\.MachAr.*:DeprecationWarning",
+     "ignore:module 'sre_.*' is deprecated:DeprecationWarning",
+     "ignore:Jitify is performing a one-time only warm-up",
++    "ignore:Context.call_conv is deprecated.",
+ ]
+ log_cli_level = "INFO"
+ testpaths = ["tests", "tests-cuda", "tests-cuda-kernels", "tests-cuda-kernels-explicit"]
+EOF
+
 rapids-logger "Generate awkward tests"
 nox -s prepare -- --tests
 


numba_cuda/numba/cuda/tests/cudapy/test_multithreads.py
```diff
@@ -37,6 +37,7 @@ def use_foo(x):
 
 @skip_under_cuda_memcheck("Hangs cuda-memcheck")
 @skip_on_cudasim("disabled for cudasim")
+@unittest.skip
```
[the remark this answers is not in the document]
isVoid: Some of the multithread / multigpu test were suspected to be not tested properly in the current CI system as they are all single-GPU runners. They show up as failures in my local environment but was not related this PR.

numba_cuda/numba/cuda/compiler.py
```diff
@@ -1116,21 +1124,16 @@ def _compile_pyfunc_with_fixup(
         nvvm_options=nvvm_options,
         cc=cc,
         forceinline=forceinline,
+        abi=abi,
+        abi_info=abi_info,
     )
     resty = cres.signature.return_type
 
     if resty and not device and resty != types.void:
         raise TypeError("CUDA kernel must have void return type.")
```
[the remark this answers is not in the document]
isVoid: This is indeed intentional, because function / kernels can bare its own calling convention, rather than using a target-wide calling convention. `cabi_wrap_function` was meant to patch a function with Numba ABI to a caller that expects a CABI function. The fix in this PR is imply switching the kernel to bare a `CUDACABICallConv` and should work as expected.

numba_cuda/numba/cuda/compiler.py
```diff
@@ -1116,21 +1073,16 @@ def _compile_pyfunc_with_fixup(
...
-
     if device:
         lib = cres.library
-        if abi == "c":
-            wrapper_name = abi_info.get("abi_name", pyfunc.__name__)
-            lib = cabi_wrap_function(
-                tgt, lib, cres.fndesc, wrapper_name, nvvm_options
-            )
     else:
         lib = cres.library
```
gmarkall: Since this statement is on both branches (and the only one left on this one), we could tidy up the logic a little bit here. It won't let me add a suggestion on deleted lines, but I'm thinking along the lines of

```python
lib = cres.library
if not device:
    kernel = lib.get_function(...
```

numba_cuda/numba/cuda/core/base.py
```diff
@@ -469,9 +469,11 @@ def get_external_function_type(self, fndesc):
         return fnty
 
     def declare_function(self, module, fndesc):
```
gmarkall: I think this need no longer be a method of the context as it doesn't use `self` anymore? (I think you alluded to this "code smell" during our pairing session last week).

gmarkall: In fact it uses so much of `fndesc` I wonder if it makes sense to be a method of `FunctionDescriptor`.

numba_cuda/numba/cuda/core/imputils.py
```diff
@@ -215,13 +215,18 @@ def user_function(fndesc, libs):
...
-            context.call_conv.return_status_propagate(builder, status)
+
+        if status is not None:
+            with cgutils.if_unlikely(builder, status.is_error):
+                context.call_conv.return_status_propagate(builder, status)
+
         assert sig.return_type == fndesc.restype
+
         # Reconstruct optional return type
+        # XXX: What do we do now when we use CABI?
```
gmarkall: We don't do anything, because there will never be an optional return type with a C ABI function. It's safe to delete this comment.

I think we had this thought when we first modified this function, but realised later it wasn't an issue.

numba_cuda/numba/cuda/core/typed_passes.py
```diff
@@ -325,6 +326,17 @@ def run_pass(self, state):
...
+        if call_conv is None:
+            call_conv = CUDACallConv(state.targetctx)
+
+        if isinstance(call_conv, CUDACABICallConv):
+            mangler = targetctx.c_abi_mangler
+        else:
+            mangler = targetctx.mangler
+
+        # mangler = targetctx.mangler
+
```
gmarkall: A little tidy-up:

```suggestion
```

numba_cuda/numba/cuda/core/typed_passes.py
```diff
@@ -325,6 +326,17 @@ def run_pass(self, state):
...
         pre_stats = passmanagers.dump_refprune_stats()
 
+        call_conv = flags.call_conv
+        if call_conv is None:
+            call_conv = CUDACallConv(state.targetctx)
+
+        if isinstance(call_conv, CUDACABICallConv):
+            mangler = targetctx.c_abi_mangler
+        else:
+            mangler = targetctx.mangler
```
gmarkall: Now I think about it, maybe the mangler should be part of the call conv. Does it look like it would be straightforward to make that change?

numba_cuda/numba/cuda/decorators.py
```diff
@@ -276,6 +276,9 @@ def declare_device(name, sig, link=None, use_cooperative=False):
     :param link: External code to link when calling the function.
     :param use_cooperative: External code requires cooperative launch.
```
gmarkall, line 297: Need to document the `abi` kwarg as well.

numba_cuda/numba/cuda/target.py
```diff
@@ -266,15 +265,22 @@ def nonconst_module_attrs(self):
         )
         return nonconsts_with_mod
 
-    @cached_property
+    @property
     def call_conv(self):
-        return CUDACallConv(self)
+        return self.fndesc.call_conv
```
gmarkall: Hopefully we can delete this property now, and all uses of `call_conv` should come from `fndesc` directly?

numba_cuda/numba/cuda/target.py
```diff
@@ -266,15 +265,22 @@ def nonconst_module_attrs(self):
...
 
     def mangler(self, name, argtypes, *, abi_tags=(), uid=None):
         return itanium_mangler.mangle(
             name, argtypes, abi_tags=abi_tags, uid=uid
         )
 
+    def c_abi_mangler(self, name, argtypes, *, abi_tags=None, uid=None):
```
gmarkall: As per the above comment, https://github.com/NVIDIA/numba-cuda/pull/717/files#r2710201612, maybe these could move to the calling convention?

numba_cuda/numba/cuda/tests/cudapy/test_compiler.py
```diff
@@ -179,7 +179,6 @@ def check_debug_info(self, ptx):
     # in that module anyway. So this test can only be expected to fail until we
     # have a proper way of generating device functions with the C ABI without
     # requiring the hack of generating a wrapper.
-    @unittest.expectedFailure
```
gmarkall, line 182: :tada:

numba_cuda/numba/cuda/tests/cudapy/test_device_func.py
```diff
@@ -496,5 +496,112 @@ def kernel(r, x):
         np.testing.assert_equal(r, x * 5)
 
 
+# XXX: Range from 0 - 2 input arguments
+# XXX: void return type?
```
gmarkall: Looks like these are now implemented below?
```suggestion
```

numba_cuda/numba/cuda/tests/cudapy/test_device_func.py
```diff
@@ -496,5 +496,112 @@ def kernel(r, x):
...
+{
+  return a + b;
+}
+""")
+
+consume_cabi_cu = cuda.CUSource("""
+extern "C" __device__
+void consume(int a)
+{
+  // Prevent trivial removal while still having no externally-visible effect.
```
gmarkall, line 528: Would it be better to have a test function with an `int*` here so we could write to a memory location and check the effect was as-intended?

numba_cuda/numba/cuda/core/compiler.py
```diff
@@ -66,7 +68,10 @@ def _make_subtarget(targetctx, flags):
         subtargetoptions["enable_nrt"] = True
     if flags.fastmath:
         subtargetoptions["fastmath"] = flags.fastmath
-    error_model = callconv.create_error_model(flags.error_model, targetctx)
+
+    # FIXME: should update everywhere uses error_model to use callconv from fundesc
```
gmarkall: Presumably if we delete the `callconv` property from the context, that should show up any places where we're not using the `fndesc` for the callconv, and then it should be OK to remove this comment.

gmarkall (changes requested):
In terms of the design, this is pretty much 90% of the way there. I made a couple of comments on how to resolve the XXX / FIXME items, and had one other suggestion for where the mangler should live - generally I'd think of mangling as part of the calling convention anyway.

I think with the resolution of the items noted on the diff, the design of this should be good to merge (give or take any minor tweaks following the changes).

numba_cuda/numba/cuda/core/base.py
isVoid: > In fact it uses so much of `fndesc` I wonder if it makes sense to be a method of `FunctionDescriptor`.

Makes sense to me.

isVoid: [f972e09](https://github.com/NVIDIA/numba-cuda/pull/717/commits/f972e09be58e0f2c46c663a5c04ed564b909c1b0)

numba_cuda/numba/cuda/core/typed_passes.py
isVoid: Moved in [45f90f6](https://github.com/NVIDIA/numba-cuda/pull/717/commits/45f90f691d7188662983a1f0fc949786ced1dd05)

numba_cuda/numba/cuda/decorators.py
isVoid: [3475e35](https://github.com/NVIDIA/numba-cuda/pull/717/commits/3475e35b631f8f5e38655b9703cefb1319b22d27)

numba_cuda/numba/cuda/target.py
isVoid: Yes, this is moved in [684ad4f](https://github.com/NVIDIA/numba-cuda/pull/717/commits/684ad4f82a46078626f2a8cd94c155f4b4ad64f5). Though in many cases it's still a pass through via `context.fndesc.call_conv.`

numba_cuda/numba/cuda/target.py
isVoid: Moved in [45f90f6](https://github.com/NVIDIA/numba-cuda/pull/717/commits/45f90f691d7188662983a1f0fc949786ced1dd05)

numba_cuda/numba/cuda/tests/cudapy/test_device_func.py
isVoid: Added in [b93f046](https://github.com/NVIDIA/numba-cuda/pull/717/commits/b93f046d6cec0d0fd9ba206f5303e9774c6fe491)

numba_cuda/numba/cuda/core/compiler.py
isVoid: I'm a bit missing the connection here. I can see the first half - when we remove directly using the context property we can see all places that we don't directly use `fndesc`. But for error_model, how does it connect to where fndesc's call site?

isVoid:
@gmarkall I believe I addressed most of the review comments above. A few things that still stands out to me: 
- The handling of error model is still arbitrated to CUDACallConv, not CUDACABICallConv.
- A few other functions in BaseContext like `call_internal` and `call_internal_no_propagate` have similar pattern to `declare_function`. Should they also get moved to FunctionDescriptor?

numba_cuda/numba/cuda/core/compiler.py
```diff
@@ -66,7 +68,9 @@ def _make_subtarget(targetctx, flags):
         subtargetoptions["enable_nrt"] = True
     if flags.fastmath:
         subtargetoptions["fastmath"] = flags.fastmath
-    error_model = callconv.create_error_model(flags.error_model, targetctx)
+
+    call_conv = CUDACallConv(targetctx)
+    error_model = callconv.create_error_model(flags.error_model, call_conv)
```
gmarkall, line 75:
```suggestion
    # Only the CUDA Calling convention can raise exceptions, so we assume here
    # that it is suitable for creating the error model.
    cuda_call_conv = CUDACallConv(targetctx)
    error_model = callconv.create_error_model(flags.error_model, cuda_call_conv)
```

numba_cuda/numba/cuda/core/generators.py
```diff
@@ -76,7 +76,7 @@ def __init__(self, lower):
         self.geninfo = lower.generator_info
         self.gentype = self.get_generator_type()
         self.gendesc = GeneratorDescriptor.from_generator_fndesc(
-            lower.func_ir, self.fndesc, self.gentype, self.context.mangler
+            lower.func_ir, self.fndesc, self.gentype, self.call_conv.mangler
```
gmarkall: Where is `self.call_conv` set?

I suspect the answer is that it's not, but that this code is dead in the context of Numba-CUDA.

For "correctness", should it be using `fndesc.call_conv.mangler` instead?

gmarkall:
> The handling of error model is still arbitrated to CUDACallConv, not CUDACABICallConv.
Although this isn't as neat as I'd like, I think it's functionally OK because only the `CUDACallConv` can raise an exception, and the only difference between error models is where or not floating point division by zero raises an exception.

I think for Numba-CUDA I'd just get rid of the error model stuff entirely, but I think it's a bit too much upheaval right now (even for a separate PR).

> A few other functions in BaseContext like `call_internal` and `call_internal_no_propagate` have similar pattern to declare_function. Should they also get moved to FunctionDescriptor?

CCCL is using `call_internal`: https://github.com/NVIDIA/cccl/blob/90899d560b599e1faf01412e9bffd31a57d38c31/python/cuda_cccl/cuda/compute/_odr_helpers.py#L118

I think it'd be good to eventually move them, but I think I might leave it as a TODO item here.

gmarkall:
Note to self: I need to run this locally with the changes from #770 merged to see how they interact.

PR #724, opened 2026-01-16
Adding `pixi run test` and `pixi run test-par` support

rparolin:
# Summary

- Add pixi run test-par command to run the test suite in parallel using pytest-xdist

### Parallel Testing
- Added new `test-par` task that uses `pytest -n auto` to automatically detect and utilize all available CPU cores
- Maintains same environment variables and build dependencies as the serial test task
- Provides significant speedup for running the full test suite (2m45s -> 36s on my machine)

### Run tests in parallel:
`pixi run test-par`

### Run tests serially (existing):
`pixi run test`

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -105,7 +105,7 @@ numpydoc = ">=1.9.0"
 nvidia-sphinx-theme = "*"
 
 [environments]
-default = { features = ["test"], solve-group = "default" }
+default = { features = ["cu-13-1", "test", "cu", "cu-13", "cu-rt", "nvvm", "py314"], solve-group = "default" }
 bench-against = { features = ["test"], no-default-feature = true }
 # CUDA 12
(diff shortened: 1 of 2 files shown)


pixi.toml
```diff
@@ -105,6 +105,7 @@ numpydoc = ">=1.9.0"
 nvidia-sphinx-theme = "*"
 
 [environments]
+default = { features = ["test"], solve-group = "default" }
```
cpcloud: What version of cuda and python does this include?

rparolin: I figured it would use the default environment dependencies like cuda-python does.

`default = { features = ["cu-13-1", "test", "cu", "cu-13", "cu-rt", "nvvm", "py314"], solve-group = "default" }`

Should I add something explicit like this?

cpcloud (approved):
LGTM.

PR #727, opened 2026-01-20
[WIP] Second pass at LaunchConfig

tpn:


scripts/bench-launch-overhead.py
cpcloud, line 1: Not entirely sure what the purpose of this file is beyond what's happening in the existing `test_kernel_launch.py` benchmarks.

scripts/plot-launch-overhead.py
cpcloud, line 1: I think all of this can be done with the existing pytest-benchmark plugin.

scripts/bench-launch-overhead.py
cpcloud, line 1: I think all or most of this functionality can be done with the existing pytest-benchmark plugin.

Really would like to avoid duplicating functionality, especially if it's AI generated duplication.

LAUNCH-CONFIG-CODEX-PROMPT.md
```diff
@@ -0,0 +1,87 @@
...
+Your goal is to benchmark CUDA kernel launch overhead across all three.
+
+## Key Artifacts
+- `scripts/bench-launch-overhead.py` (micro-benchmark for launch overhead)
+- `scripts/bench-against.py` (two-ref compare helper)
+- Pixi tasks in `pixi.toml`:
+  - `bench-launch-overhead`
+  - `bench`
+  - `benchcmp`
+  - `bench-against`
```
cpcloud: While this script doesn't do a three way comparison, it also doesn't require writing any new code to run it.

Can we try to reuse `bench-against` instead of reinventing a lot of what that already does?

scripts/bench-launch-overhead.py
```diff
@@ -0,0 +1,308 @@
...
+    @cuda.jit("void(float32[:],float32[:])")
+    def some_kernel_3(arr1, arr2):
+        return
+
+    @cuda.jit("void(float32[:],float32[:],float32[:])")
+    def some_kernel_4(arr1, arr2, arr3):
+        return
+
+    @cuda.jit("void(float32[:],float32[:],float32[:],float32[:])")
+    def some_kernel_5(arr1, arr2, arr3, arr4):
```
cpcloud, line 105: These are nearly identical to the existing benchmarks. Let's avoid repeating existing benchmarks and tools that run them.

PR #742, opened 2026-01-22
Fix missing line info in Jupyter notebooks

### LINKED ISSUE #721: [BUG] Lineinfo missing in Jupyter notebook cells
Lineinfo for jupyter notebook cells is missing because the source code of the function cannot be found. For example:

Image

The `DebugInfoWarning` is produced for both lineinfo and debuginfo generation. This does not occur with upstream Numba, and is likely some fallout from the vendoring.

jiel-nv:
Fixes #721.

The linecache scenario (which Jupyter uses) was missed during vendoring via PR #457. Port in the upstream Numba approach of adding inspect.getsourcelines() to handle this case (original PR numba/numba#9861) with modifications so that it serves as the main/only approach instead of fallback.

Jupyter notebook cells have virtual filenames (e.g., <ipython-input-X-Y>) that don't exist on disk. IPython registers cell source in Python's linecache module, which inspect.getsourcelines() uses internally. The previous implementation only used open() to read source files, causing lookups to fail for notebook cells.

Using inspect.getsourcelines() as the main approach rather than a fallback simplifies the code since it handles both disk files and linecache entries uniformly.

Added test_linecache_source() in test_debuginfo.py to simulate the Jupyter scenario by registering source in linecache and verifying that no NumbaDebugInfoWarning is raised when compiling a kernel with debug=True.

### DIFF
--- a/numba_cuda/numba/cuda/misc/firstlinefinder.py
+++ b/numba_cuda/numba/cuda/misc/firstlinefinder.py
@@ -7,6 +7,8 @@ body.
 """
 
 import ast
+import inspect
+import textwrap
 
 
 class FindDefFirstLine(ast.NodeVisitor):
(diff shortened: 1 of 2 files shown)


gmarkall:
~~I'm getting a really weird traceback when I try to run the reproducer from the bug report in a Jupyter notebook.~~

~~It would be surprising if this PR was the cause of the issue, but perhaps it exposes another strange issue.~~

**Edit:** my local environment needed an update for recent changes. There is no issue.

(traceback deleted)

gmarkall:
It's OK, it turns out my local `main` was out of date. ~~We have some other regression which I'll look into separately.~~

PR #749, opened 2026-01-27
Eliminate duplicate DWARF entries for boolean kernel parameters

jiel-nv:
Fixes NVBug#5805171.

Ref-only boolean parameters were generating duplicate DWARF entries (both `DW_TAG_formal_parameter` and `DW_TAG_variable`). This occurs in the code path that extends `dbg.value` coverage per basic block to accommodate NVVM. The debug info emission used LLVM-level argument names (`fnargs`) which differ from source-level names after calling convention transformation, causing the `argidx` lookup to fail and default to 0. This created a new `DILocalVariable` with `arg: 0` instead of reusing the existing formal parameter metadata.

The fix uses source-level argument names (`fndesc.args`) to get the correct `argidx`, allowing LLVM to deduplicate the metadata.

Added test `test_bool_param_ref_only`.

### DIFF
--- a/numba_cuda/numba/cuda/lowering.py
+++ b/numba_cuda/numba/cuda/lowering.py
@@ -1786,10 +1786,9 @@ class CUDALower(Lower):
                             # Not yet covered by the dbg.value range
                             and src_name not in self.dbg_val_names
                         ):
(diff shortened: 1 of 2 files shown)


cpcloud:
Thanks!

PR #756, opened 2026-01-28
Generate line info for PHI exporters in terminator block

jiel-nv:
Fixes nvbug5811432.

When stepping through code in `cuda-gdb`, the debugger would jump to unrelated lines (where variables were last assigned) before reaching the `continue` statement when the condition is TURE. For example, with this code snippet:

```
for i in range(5):
    if i == 3:        # line 280
        continue      # line 281
    ...
    x = x + i         # line 295
```

The debugger would step: 280 → 295 → 281, instead of the expected 280 → 281.

The root cause is during stripping off PHI nodes and creating exporters, a uniformed `rhs.loc` (the location where the RHS variable was last assigned) is used as the source location. Despite of the existing WITH and WITHOUT assignments cases already handing the place to insert the exporters, the location info are not well handled.

The fix is to conditionally set the `loc` for PHI exporter assignments:
If the variable is assigned in the current block: Use `rhs.loc` to preserve the assignment site's location.
If the variable is NOT assigned in the current block (pass-through in `continue`/`break`): Use `term_loc` (the block terminator's location, i.e., the `continue`/`break` statement's line number).

Also adds a test `test_terminator_line_number` into `test_debuginfo`.

### DIFF
--- a/numba_cuda/numba/cuda/tests/cudapy/test_matmul.py
+++ b/numba_cuda/numba/cuda/tests/cudapy/test_matmul.py
@@ -5,7 +5,6 @@ import numpy as np
 
 from numba import cuda
 from numba.cuda import float32, void
-from numba.cuda.testing import unittest, CUDATestCase
 from numba.cuda.core import config
 
 # Ensure the test takes a reasonable amount of time in the simulator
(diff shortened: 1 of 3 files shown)


jiel-nv:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/756/commits/69f94c26f993b462d15002173c82d4eb58eee19e

jiel-nv:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/756/commits/2b137989f1597d95e3037a5faa4e3af1f3b9ae29

jiel-nv:
Gentle ping @gmarkall @cpcloud

cpcloud:
Thanks @jiel-nv!

PR #761, opened 2026-01-28
CI cleanup, extend no-ci skip logic, and add GitHub Release uploads

kkraus14:
### [no-ci] skip logic
- Extend the `[no-ci]` PR title check to skip `test-conda`, `test-simulator`, and `build-docs` jobs
- Add `github.repository_owner == 'nvidia'` check to these jobs for consistency with other CI jobs

### CI cleanup
- Delete unused CI scripts: `check_style.sh`, `upload_conda.sh`, `build_docs.sh`
- Delete unused GitHub Action: `get_pr_number`
- Delete unused CI tool scripts: `download-wheels`, `lookup-run-id`
- Remove unused `container-options` input from `conda-python-tests.yaml`
- Remove unused `RAPIDS_AUX_SECRET_1` secret from `wheels-test.yaml`
- Remove `SKIP_NUMBA_CUDA_TEST` conditionals (variable was never set)

### GitHub Release wheel uploads
- Add `upload-release` job to publish workflow that uploads wheels to GitHub Releases
- Creates the GitHub release with auto-generated notes if it doesn't already exist
- Downloads wheels built by the publish workflow and uploads them to the release
- Source archives are not uploaded since GitHub automatically generates them for every release

## Test plan

- CI should pass with these changes
- No functional changes to the actual build/test logic
- Release uploads will be tested on the next tag push

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -81,6 +81,49 @@ jobs:
         with:
           verbose: true
 
+  upload-release:
+    name: Upload wheels to GitHub Release
+    needs: [build-wheels-linux-64, build-wheels-linux-aarch64, build-wheels-windows]
+    if: ${{ !github.event.repository.fork }}
+    permissions:
+      contents: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download wheels
+        uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131  # v7.0.0
+        with:
+          pattern: numba-cuda-python*
+          path: wheels/
+          merge-multiple: true
+
+      - name: Create GitHub Release if not exists
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          # Check if release already exists, create if not
+          if ! gh release view "${{ github.ref_name }}" --repo "${{ github.repository }}" > /dev/null 2>&1; then
+            echo "Creating release ${{ github.ref_name }}"
+            gh release create "${{ github.ref_name }}" \
+              --repo "${{ github.repository }}" \
+              --title "${{ github.ref_name }}" \
+              --generate-notes \
+              --verify-tag
+          else
+            echo "Release ${{ github.ref_name }} already exists"
+          fi
+
+      - name: Upload wheels to GitHub Release
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          if [[ -d "wheels" && $(ls -A wheels 2>/dev/null | wc -l) -gt 0 ]]; then
+            echo "Uploading wheels to release ${{ github.ref_name }}"
+            gh release upload --clobber "${{ github.ref_name }}" --repo "${{ github.repository }}" wheels/*
+          else
+            echo "No wheels found to upload"
+            exit 1
+          fi
+
   build-docs:
     uses: ./.github/workflows/docs-build.yaml
 


.github/workflows/publish.yaml
```diff
@@ -81,6 +81,59 @@ jobs:
...
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          # Upload source archive and checksum
+          gh release upload --clobber "${{ github.ref_name }}" --repo "${{ github.repository }}" release/${{ env.ARCHIVE_NAME }}.tar.gz release/${{ env.ARCHIVE_NAME }}.tar.gz.sha256sum
+
+          # Upload wheels
+          if [[ -d "release/wheels" && $(ls -A release/wheels 2>/dev/null | wc -l) -gt 0 ]]; then
+            echo "Uploading wheels to release ${{ github.ref_name }}"
+            gh release upload --clobber "${{ github.ref_name }}" --repo "${{ github.repository }}" release/wheels/*
+          fi
```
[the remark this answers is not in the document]
kkraus14: Resolved

.github/workflows/publish.yaml
```diff
@@ -81,6 +81,75 @@ jobs:
...
+      - name: Checkout Source
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8  # v6.0.1
+        with:
+          fetch-depth: 0
+          ref: ${{ github.ref_name }}
+
+      - name: Create Release Directory
+        run: mkdir -p release
+
+      - name: Archive Source
```
cpcloud: GitHub automatically creates zip and gzipped tarballs for releases (and every commit). Why are we doing that manually here?

.github/workflows/publish.yaml
```diff
@@ -81,6 +81,75 @@ jobs:
...
+          pattern: numba-cuda-python*
+          path: release/wheels/
+          merge-multiple: true
+
+      - name: Create GitHub Release if not exists
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          # Check if release already exists, create if not
+          if ! gh release view "${{ github.ref_name }}" --repo "${{ github.repository }}" > /dev/null 2>&1; then
```
cpcloud, line 104: Why not just 

```
gh release create --fail-on-no-commits ... || true
```
?

.github/workflows/publish.yaml
```diff
@@ -81,6 +81,75 @@ jobs:
...
+          else
+            echo "Release ${{ github.ref_name }} already exists"
+          fi
+
+      - name: Upload to GitHub Release
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          # Upload source archive and checksum
+          gh release upload --clobber "${{ github.ref_name }}" --repo "${{ github.repository }}" release/${{ env.ARCHIVE_NAME }}.tar.gz release/${{ env.ARCHIVE_NAME }}.tar.gz.sha256sum
```
cpcloud: Ditto, I don't think we actually need to upload source archives (wheel uploading remains, of course).

cpcloud (changes requested):
Ideally we can avoid the manual archiving since GitHub automatically does that for us.

.github/workflows/publish.yaml
kkraus14: I think there's situations where you want to cut a tag on the same commit as another tag and that would cause it to fail? I.E. some kind of packaging hiccup upstream where it's cleaner to cut a new patch release, though I suspect we'd always have a version bump commit in the least?

cpcloud: Fair enough, I care less about this review comment than avoiding the duplicate archiving (less code to maintain if we let the default GitHub behavior do its thing).

cpcloud (approved):

PR #769, opened 2026-02-02
Add `cuda-core` to `oldest` tests

brandon-b-miller:
This PR pins `cuda-core` to the current minimum supported version as part of the jobs that already test the `oldest` upstream `numba` version.

### DIFF
--- a/ci/test_conda.sh
+++ b/ci/test_conda.sh
@@ -11,7 +11,6 @@ if [ "${NUMBA_VERSION:-*}" != "*" ]; then
 fi
 
 if [ "${CUDA_CORE_VERSION:-*}" != "*" ]; then
-    # add to the default environment's dependencies
     pixi add --feature "${PY_VER_PART}" "cuda-core=${CUDA_CORE_VERSION}"
 fi
 


brandon-b-miller:
This scheme (adding `cuda-core` oldest testing the same way we do numba upstream) works for our conda packages, but we still have a divergence between conda and wheel/local CI that I believe we inherit from the `cuda-python` CI, xref https://github.com/NVIDIA/cuda-python/issues/280. This may serve as a stopgap for catching some issues, but we should figure out how to reconcile this and ultimately the wheel/conda jobs should maybe leverage the same matrix so we can test `oldest` across wheel jobs as well.

ci/test_conda.sh
```diff
@@ -10,6 +10,11 @@ if [ "${NUMBA_VERSION:-*}" != "*" ]; then
     pixi add --feature "${PY_VER_PART}" "numba=${NUMBA_VERSION}"
 fi
 
+if [ "${CUDA_CORE_VERSION:-*}" != "*" ]; then
+    # add to the default environment's dependencies
```
cpcloud: nit: this isn't installing to the default environment, and I don't see the comment adding any clarity to what `pixi add` is doing, so perhaps we can remove it? (again, not blocking)


```suggestion
```

Issue #157, opened 2025-03-13
[BUG] Compilation of function comparing two int32 inputs fails to compile with NvvmError exception


oleksandr-pavlyk:
**Describe the bug**

Compilation of simple comparison function for input integral types unexpectedly fails with `NvvmError`.

**Steps/Code to reproduce bug**

See https://github.com/NVIDIA/cccl/pull/3949/files#r1993739985

```python
import numba, numba.cuda
import numpy as np

def cmp1(a, b):
    return numba.types.uint8(a == b)

def cmp2(a, b):
    return np.uint8(a == b)

def cmp3(a, b):
    return a == b

def compile_and_print(func):
    code, retty = numba.cuda.compile(func, (numba.types.int32, numba.types.int32), output="ptx")
    print(code)
    print("Return type: ", retty)
    return code, retty
```

`compile_and_print(cmp3)` fails with `NvvmError`, with code `NVVM_ERROR_INVALID_IR`, while both `compile_and_print(cmp1)` and `compile_and_print(cmp2)` compile fine.

**Expected behavior**

The `compile_and_print(cmp3)` should compile as well.

**Environment details (please complete the following information):**

I am running it in DevContainer from CCCL project with CTK 12.8 and GCC 13, and running on Python 3.10 and 

```
In [7]: numba.__version__
Out[7]: '0.61.0'
```

**Additional context**

kkraus14:
@oleksandr-pavlyk as a workaround in `c.parallel` for the time being if you specify the output type it properly compiles, i.e. use `numba.types.int32(numba.types.int32, numba.types.int32)` as the signature

gmarkall:
Somehow I missed this before. The issue can be seen in the NVVM IR:

```llvm
define i8 @"cmp3"(i32 %".1", i32 %".2")
{
// ...
  ret i1 %".25"
}
```

The data model for a boolean value uses a 1-bit value, but the smallest integer we can return is 8 bits. I suspect the correct fix is for the implementation of returning a value in `CUDACABICallCallConv` to cast the returned value to match the inferred return type.

gmarkall:
@kaeun97 Does this issue look like it's of interest?

kaeun97:
Yes, of course! Still working on https://github.com/NVIDIA/numba-cuda/issues/623#issuecomment-3784700231 at the moment, but should be able to jump into this.

PR #770, opened 2026-02-03
fix: fix boolean return type mismatch in C ABI wrapper

kaeun97:
Fixes https://github.com/NVIDIA/numba-cuda/issues/157.


numba_cuda/numba/cuda/target.py
```diff
@@ -453,6 +453,21 @@ def _make_call_helper(self, builder):
...
 
     def return_value(self, builder, retval):
+        expected_type = builder.function.ftype.return_type
+        actual_type = retval.type
+
+        # If types don't match, we need to cast
+        if actual_type != expected_type:
+            if isinstance(actual_type, ir.IntType) and isinstance(
+                expected_type, ir.IntType
+            ):
```
kaeun97: Should we consider other types as well?

kaeun97: Or throw an error if there is an unexpected type mismatch?

gmarkall:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/770/commits/3ec7629fae2aeced755325a93931fc4c8c3c4296

gmarkall:
Will also check this locally with #717 to see how they interact.

gmarkall:
> Will also check this locally with #717 to see how they interact.

With this merged into #717 (and fixed up conflicts) all tests still pass.

numba_cuda/numba/cuda/target.py
gmarkall: I'm hoping the `i1` boolean / `i8` return type is the only edge case we need to handle.

For some background thoughts: it's a bit fiddly because usually if you're trying to accept as an argument or return something that's not quite the right type, you would be able use `CUDATargetContext.cast()` (i.e. `self.context.cast()` just here) to use the target context's mechanism to handle any case appropriately. But it needs to be told the Numba type of the types to cast from / to, and we don't have them here.

I'm inclined to go with this fix as-is, but if we start butting up against other edge cases, perhaps we should think about how we could use `context.cast()` instead of further customisation here.

gmarkall (approved):
The change looks good, it just needs a merge of `main` and the conflict resolving.

Because of #717:
* `CUDACABICallConv` moved to `numba_cuda/numba/cuda/core/callconv.py`.
* `cabi_wrap_function` was deleted (so that part of the fix shouldn't be needed anymore). `cabi_wrap_function` was a hack to handle the fact we couldn't compile with different calling conventions within one compilation, but #717 removed that limitation.

I did the merge the other way when testing, but if it helps to see how I resolved conflicts, I pushed that up here: https://github.com/gmarkall/numba-cuda/tree/pr-717-770 ([merge commit](https://github.com/gmarkall/numba-cuda/commit/3bbda4dec72e137fd0ac6184bd45ae7c3b6224d6))

gmarkall:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/770/commits/c099a7c3ff49a5a67b6bd65ac0fd1c1ab104ea77

gmarkall:
@kaeun97 I hope you don't mind, I did the merge of `main` because we're hoping to get this merged then make a release incorporating it.

kaeun97:
Thanks for stepping in @gmarkall ! I was planning to do it tonight - was not aware that this was urgent. Appreciate it!

PR #771, opened 2026-02-03
Find CUDA headers via pathfinder

brandon-b-miller:
Xref https://github.com/NVIDIA/numba-cuda/issues/302, https://github.com/NVIDIA/cuda-python/issues/1036

### DIFF
--- a/numba_cuda/numba/cuda/cudadrv/devices.py
+++ b/numba_cuda/numba/cuda/cudadrv/devices.py
@@ -158,12 +158,10 @@ class _Runtime:
                     # Get primary context for the active device
                     ctx = self.gpus[ac.devnum].get_primary_context()
                     # Is active context the primary context?
-                    ctx_handle = ctx.handle.value
-                    ac_ctx_handle = ac.context_handle.value
-                    if ctx_handle != ac_ctx_handle:
+                    if ctx.handle != ac.context_handle:
                         raise RuntimeError(
                             "Numba cannot operate on non-primary"
-                            f" CUDA context {ac_ctx_handle:x}"
+                            f" CUDA context {int(ac.context_handle):x}"
                         )
                     # Ensure the context is ready
                     ctx.prepare_for_use()
--- a/numba_cuda/numba/cuda/cudadrv/driver.py
+++ b/numba_cuda/numba/cuda/cudadrv/driver.py
@@ -29,7 +29,15 @@ import tempfile
 import re
 from itertools import product
 from abc import ABCMeta, abstractmethod
-from ctypes import c_int, byref, c_void_p, c_uint8
+from ctypes import (
+    c_int,
+    byref,
+    c_void_p,
+    c_uint8,
+    CFUNCTYPE,
+    py_object,
+    c_size_t,
+)
 import contextlib
 import importlib
 import numpy as np
@@ -40,12 +48,10 @@ from numba.cuda.cext import mviewbuf
 from numba.cuda.core import config
 from numba.cuda import utils, serialize
 from .error import CudaSupportError, CudaDriverError
-from .drvapi import API_PROTOTYPES
-from .drvapi import cu_stream_callback_pyobj
 from .mappings import FILE_EXTENSION_MAP
 from .linkable_code import LinkableCode, LTOIR, Fatbin, Object
 from numba.cuda.utils import cached_file_read
-from numba.cuda.cudadrv import drvapi, nvrtc
+from numba.cuda.cudadrv import nvrtc
 
 from cuda.bindings import driver as binding
 from cuda.core import (
@@ -67,7 +73,7 @@ CU_STREAM_DEFAULT = 0
 
 MIN_REQUIRED_CC = (3, 5)
 SUPPORTS_IPC = sys.platform.startswith("linux")
-
+cu_stream_callback_pyobj = CFUNCTYPE(None, c_void_p, c_int, py_object)
 
 _py_decref = ctypes.pythonapi.Py_DecRef
 _py_incref = ctypes.pythonapi.Py_IncRef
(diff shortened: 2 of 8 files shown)


numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -108,20 +108,10 @@ def compile(src, name, cc, ltoir=False, lineinfo=False, debug=False):
     numba_cuda_path = os.path.dirname(cudadrv_path)
 
     nvrtc_ver_major = version[0]
-    if nvrtc_ver_major == 12:
-        numba_include = f"{os.path.join(numba_cuda_path, 'include', '12')}"
```
cpcloud, line 112: This is still used on line 133, so that's a guaranteed `NameError` I believe

cpcloud: If we're not using our own anymore, then shouldn't we just remove the `numba/cuda/include` directory entirely?

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -108,20 +108,10 @@ def compile(src, name, cc, ltoir=False, lineinfo=False, debug=False):
...
-        cuda_includes.append(
-            f"{os.path.join(cuda_include_dir, '..', '..', 'cuda_cccl', 'include')}"
-        )
-    elif nvrtc_ver_major == 13:
-        numba_include = f"{os.path.join(numba_cuda_path, 'include', '13')}"
-        # For CUDA 13 wheels, `cuda_include_dir` is `site-packages/nvidia/cu13/include`
-        # We need to find CCCL at `site-packages/nvidia/cu13/include/cccl`
-        # For CUDA 13 conda / system install, CCCL is in the `include/cccl` directory
-        cuda_includes.append(f"{os.path.join(cuda_include_dir, 'cccl')}")
+    cccl_include_dir = pathfinder.locate_nvidia_header_directory(
```
cpcloud: `locate_nvidia_header_directory` can return `None`, and if it does, that needs to be handled with something more informative than an `AttributeError` on `NoneType`.

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -493,44 +437,23 @@ def get_current_cuda_target_name():
...
+            )
+        )
+        return _env_path_tuple(
+            located_header_dir.found_via, located_header_dir.abs_path
+        )
+    else:
+        if config.CUDA_INCLUDE_PATH:
+            return _env_path_tuple(
+                "CUDA_INCLUDE_PATH Config entry", config.CUDA_INCLUDE_PATH
+            )
```
cpcloud, line 443: This function can fall off the end if `located_header_dir` is `False`y and `CUDA_INCLUDE_PATH` is `False`y.

numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -493,44 +437,23 @@ def get_current_cuda_target_name():
...
-        ("NVIDIA NVCC Wheel", get_wheel_include()),
-        ("CUDA_HOME", get_cuda_home_include()),
-        ("System", get_system_ctk_include()),
-        ("CUDA_INCLUDE_PATH Config Entry", config.CUDA_INCLUDE_PATH),
-    ]
-    by, include_dir = _find_valid_path(options)
-    return _env_path_tuple(by, include_dir)
+    located_header_dir = pathfinder.locate_nvidia_header_directory("cudart")
+    if located_header_dir:
+        assert os.path.exists(
```
cpcloud: This large of an expression makes me think this should be a proper error type, like a real `FileNotFoundError` instead of an `assert`.

numba_cuda/numba/cuda/cudadrv/nvrtc.py
```diff
@@ -108,20 +108,10 @@ def compile(src, name, cc, ltoir=False, lineinfo=False, debug=False):
     numba_cuda_path = os.path.dirname(cudadrv_path)
 
     nvrtc_ver_major = version[0]
```
cpcloud, line 110: This isn't used anymore.

numba_cuda/numba/cuda/cuda_paths.py
cpcloud: Seems like this might be dead code now.

numba_cuda/numba/cuda/cuda_paths.py
brandon-b-miller: Yeah, I wasn't sure what to do here. The old functions anchor around `cuda_device_runtime_api.h`, which isn't something pathfinder supports yet. If it couldn't find _that_ header, it returned `None`, else it returned whatever directory that header resided in.

The other helpers still returning `None` when they can't find the object or library they're looking for pushes me towards returning `None` here rather than erroring, but I'd be curious your thoughts on if the current header check is as "strong" as the old one.

brandon-b-miller:
Trying to figure out why the conda jobs aren't pulling the latest pathfinder here, updates hopefully soon.

numba_cuda/numba/cuda/cudadrv/nvrtc.py
brandon-b-miller: Circling back here, this ended up being needed still. This information is still used to decide the `numba_include_path`, which contains `fp8`, `pf16` and `bf16` headers.

numba_cuda/numba/cuda/cudadrv/nvrtc.py
brandon-b-miller: Same story as https://github.com/NVIDIA/numba-cuda/pull/771/changes#r2805165036

brandon-b-miller:
@cpcloud some updates here, I think things are ready for another round of review.

rparolin (approved):

numba_cuda/numba/cuda/cuda_paths.py
cpcloud: I'd lean towards preserving existing behavior instead of introducing an `assert` in a seemingly arbitrary place (i.e., why introduce it here versus anywhere else?)

PR #775, opened 2026-02-05
Remove `enums` and unused ctypes code that required it

brandon-b-miller:
`enums.py` served to centralize the lookup of CUDA constant values that could come from ctypes or the cuda python bindings, now that we don't support ctypes bindings we can just use the cuda python values directly. This does that plus removes the only code that relied on it, some likely unreachable ctypes code.

### DIFF
--- a/numba_cuda/numba/cuda/cudadrv/driver.py
+++ b/numba_cuda/numba/cuda/cudadrv/driver.py
@@ -458,18 +458,6 @@ class _ActiveContext:
 driver = Driver()
 
 
-def _build_reverse_device_attrs():
-    prefix = "CU_DEVICE_ATTRIBUTE_"
-    map = utils.UniqueDict()
-    for name in dir(enums):
-        if name.startswith(prefix):
-            map[name[len(prefix) :]] = getattr(enums, name)
-    return map
-
-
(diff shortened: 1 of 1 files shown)


numba_cuda/numba/cuda/cudadrv/driver.py
```diff
@@ -217,12 +217,10 @@ def _raise_driver_error(e):
...
 def _build_reverse_error_map():
-    prefix = "CUDA_ERROR"
+    """Build error code to name mapping from cuda.bindings.driver.CUresult"""
     map = utils.UniqueDict()
-    for name in dir(enums):
-        if name.startswith(prefix):
-            code = getattr(enums, name)
-            map[code] = name
+    for name, member in binding.CUresult.__members__.items():
+        map[int(member)] = name
```
kkraus14: We shouldn't need to reach for `__members__` here, though this may be slower:
```suggestion
    for enum_val in binding.CUresult:
        map[int(enum_val)] = enum_val.name
```

cc @mdboom since he's reworking the `cuda.bindings` enum implementation if he has guidance here

brandon-b-miller:
Couple fixups needed here, addressing

PR #777, opened 2026-02-05
Fix CI checks job to also fail on dependency failures

kkraus14:
## Summary

- The `checks` job in `ci-new.yaml` only checked for `cancelled` dependencies but not `failure`, meaning a failing CI job would not cause the overall status check to fail.
- Add `failure` result checks alongside existing `cancelled` checks for all dependencies.

Mirrors the fix from https://github.com/NVIDIA/cuda-python/pull/1547.

## Test plan

- Verify that the `checks` job now correctly reports failure when any dependency job fails (not just when cancelled).


.github/workflows/ci-new.yaml
```diff
@@ -326,16 +326,27 @@ jobs:
...
                  needs.test-conda.result == 'cancelled' ||
+                 needs.test-conda.result == 'failure' ||
                  needs.test-simulator.result == 'cancelled' ||
+                 needs.test-simulator.result == 'failure' ||
                  needs.test-thirdparty-cudf.result == 'cancelled' ||
+                 needs.test-thirdparty-cudf.result == 'failure' ||
                  needs.test-thirdparty-nvmath.result == 'cancelled' ||
+                 needs.test-thirdparty-nvmath.result == 'failure' ||
                  needs.test-thirdparty-awkward.result == 'cancelled' ||
+                 needs.test-thirdparty-awkward.result == 'failure' ||
```
kkraus14, line 345: Is there any of these that we don't want to be required?

gmarkall: I think we want them all, if these break then it's indicating we broke a public API.

kkraus14:
I cherry-picked the simulator fixes from https://github.com/NVIDIA/numba-cuda/pull/776 to ensure that it passes as expected as well.

Issue #781, opened 2026-02-06
[BUG] Wrong ABI is chosen by `_compile_subroutine_no_cache` when compiling top-level function for C ABI


gmarkall:
@shwina reported that the following testcase (reduced from CCCL):

```python
from numba import cuda
from numba import types
from numba.core.extending import intrinsic


def foo(x):
    return 2 * x


# This works - compiling a plain function
print("Compiling plain function...")
cuda.compile(foo, (types.int32,), output="ltoir")
print("✓ Plain function compilation succeeded")


# Now create a wrapper that takes void* arguments
# This is what breaks in newer numba versions
def create_void_ptr_wrapper():
    """Create a wrapper that takes void* input and output pointers."""

    # Make foo a device function
    foo_device = cuda.jit(device=True)(foo)

    # The inner signature: int32 -> int32
    inner_sig = types.int32(types.int32)

    # The wrapper signature: void(void*, void*) - input ptr, output ptr
    wrapper_sig = types.void(types.voidptr, types.voidptr)

    @intrinsic
    def wrapper_impl(typingctx, arg0, arg1):
        def codegen(context, builder, sig, args):
            input_ptr, output_ptr = args

            # Cast input void* to int32*, load value
            int32_llvm_type = context.get_value_type(types.int32)
            typed_input_ptr = builder.bitcast(input_ptr, int32_llvm_type.as_pointer())
            input_val = builder.load(typed_input_ptr)

            # Call the inner function
            cres = context.compile_subroutine(builder, foo_device, inner_sig, caching=False)
            result = context.call_internal(builder, cres.fndesc, inner_sig, [input_val])

            # Cast output void* to int32*, store result
            typed_output_ptr = builder.bitcast(output_ptr, int32_llvm_type.as_pointer())
            builder.store(result, typed_output_ptr)

            return context.get_dummy_value()

        return wrapper_sig, codegen

    def wrapper_func(input_ptr, output_ptr):
        return wrapper_impl(input_ptr, output_ptr)

    return wrapper_func, wrapper_sig


print("\nCreating void* wrapper...")
wrapper, wrapper_sig = create_void_ptr_wrapper()

# This is what breaks - trying to compile the wrapper
print("\nCompiling void* wrapper to ltoir...")
cuda.compile(wrapper, wrapper_sig.args, output="ltoir")
```

results in an exception with the following end of the traceback:

```
[...]

PR #782, opened 2026-02-06
Fix mixed C ABI / Numba ABI internal-call lowering (#781, #789)

isVoid:
This PR hardens Numba-CUDA internal-call lowering when C ABI and Numba ABI are mixed, and documents the resulting behavior.

## Why

Compiling C ABI wrappers that call Numba-ABI device functions exposed failures around:

- status/exception propagation when no status channel exists (C ABI)
- optional-return handling across ABI boundaries
- call-convention leakage into internal subroutine compilation

## What changed

- Explicitly thread `call_conv` and `abi_info` through compilation entry points (`compile_extra`, `compile_ir`, `compile_internal`) and compiler state, instead of storing them on `CUDAFlags`.
- `compile_cuda()` now selects call convention explicitly (`CUDACallConv` vs `CUDACABICallConv`) and passes `abi_info` through.
- `compile_subroutine()` now always compiles subroutines with `CUDACallConv` (Numba ABI), preventing outer C ABI wrappers from changing subroutine ABI unexpectedly.
- Refactored internal-call lowering:
  - added `BaseCallConv.call_internal_no_propagate()`
  - implemented ABI-specific `call_internal()` in `MinimalCallConv` and `CUDACABICallConv`
  - C ABI paths now no-op on status/exception propagation and safely flatten `Optional[T]` return values at C ABI boundaries.
- Updated lowering/context call sites to dispatch through the active function descriptor's call convention (`self.fndesc.call_conv`), so each call edge is lowered by callee ABI while propagation behavior follows caller ABI.
- Added user docs: `docs/source/user/callconv.rst` and linked it from the user guide index.

## Tests

Added regression coverage in `numba_cuda/numba/cuda/tests/cudapy/test_compiler.py`:

- `test_compile_jitted_subroutine` (reproducer for #781; C ABI `void*` wrapper calling Numba ABI subroutine)
- `test_compile_CABI_calling_device_function_returning_optional` (C ABI caller + Numba ABI callee returning `Optional`)
- `test_compile_complex_div_c_abi` (coverage for #789)

## Impact

- Mixed C ABI / Numba ABI nested device-call graphs now compile more robustly.
- Existing Numba ABI behavior is preserved.
- C ABI behavior remains explicit about lack of Python exception/status propagation.


numba_cuda/numba/cuda/core/compiler.py
```diff
@@ -116,6 +125,15 @@ def __init__(
         # hold this for e.g. with_lifting, null out on exit
         self.state.pipeline = self
 
+        # Default to CUDAFlag's default
```
gmarkall: Whilst this explains why we do what we do in the context of this PR, it will make no sense when read in the context of the code in the future. I think it's probably less confusing to just omit it.

```suggestion
```

numba_cuda/numba/cuda/tests/cudapy/test_ir_utils.py
```diff
@@ -65,6 +66,8 @@ def mk_pipeline(
                     return_type,
                     flags,
                     locals,
+                    None,
+                    None,
```
gmarkall: I think these are both kwargs / optional arguments? Can they be omitted?

```suggestion
```

numba_cuda/numba/cuda/typing/templates.py
```diff
@@ -689,6 +689,8 @@ def generic(self, args, kws):
             # spoof a compiler pipline like the one that will be in use
             tyctx = fcomp.targetdescr.typing_context
             tgctx = fcomp.targetdescr.target_context
+
+            # XXX: Should we use None for call_conv and abi_info?
```
gmarkall: I think it makes sense that we don't specify them here - they should be the default as we're typing the function with no other information, IIUC. I don't think we should explicitly pass `None` for them as positional arguments though, because it invites the question "What is `None` for?" to someone reading the code later.

numba_cuda/numba/cuda/tests/cudapy/test_ir_utils.py
isVoid: Yes - originally I didn't set the ctor of `CompilerBase` with default args. Only to find that it breaks pretty much the entire test suite. Later when I add them back in it makes these arguments omittable.

numba_cuda/numba/cuda/tests/cudapy/test_compiler.py
```diff
@@ -701,6 +705,69 @@ def f(z, x, y):
...
+                    # Call the inner function
+                    cres = context.compile_subroutine(
+                        builder, foo_device, inner_sig, caching=False
+                    )
+
+                    # Wrapper function is compiled with cabi, but inner function
+                    # is compiled with numba-abi. So cres should have CUDACallConv.
+                    assert isinstance(cres.fndesc.call_conv, CUDACallConv)
+
+                    _, result = context.call_internal_no_propagate(
```
isVoid: Because wrapper function is CABI function, lower level calls are unable to propagate to top level. Hence using `call_internal_no_propagate`.

numba_cuda/numba/cuda/core/callconv.py
```diff
@@ -167,6 +168,18 @@ def mangler(self, name, argtypes, *, abi_tags=(), uid=None):
             name, argtypes, abi_tags=abi_tags, uid=uid
         )
 
+    def call_internal_no_propagate(self, builder, fndesc, sig, args):
+        """Similar to `.call_internal()` but does not handle or propagate
+        the return status automatically.
```
gmarkall, line 173: I think the name and docstring were at least plausibly sensible in their original location, but I think we're running the risk of having functionality with names and docstrings that make it hard to understand what the flow and meaning is. Now we've moved things a bit, I think something like:


```suggestion
    def call_function(self, builder, fndesc, sig, args):
        """Given a function descriptor, emit a call to that function with the
        given arguments.
```

in conjunction with other suggested changes below fits the intention of what's being done more closely.

gmarkall: (I meant to start a review so I could write all comments, but hit "comment" by mistake, so I'm still working on suggestions)

numba_cuda/numba/cuda/core/callconv.py
```diff
@@ -449,18 +507,40 @@ def mangler(self, name, argtypes, *, abi_tags=None, uid=None):
         return name.split(".")[-1]
 
 
+def maybe_return_user_exc(
```
gmarkall: These "maybe" functions solve the problem, but I think they've pushed the behaviour of the calling convention out of the classes into two functions that then have class-specific behaviour - instead of defining these functions, I think we should just make `return_user_exc` and `return_status_propagate` do nothing for the C ABI calling convention (with a comment in the code of each explaining that we do nothing because it's not possible to do anything).

numba_cuda/numba/cuda/core/callconv.py
```diff
@@ -298,6 +311,23 @@ def call_function(self, builder, callee, resty, argtys, args):
         out = self.context.get_returned_value(builder, resty, retval)
         return status, out
 
+    def call_internal(self, builder, fndesc, sig, args):
+        """
+        Given the function descriptor of an internally compiled function,
+        emit a call to that function with the given arguments.
+        """
+        status, res = self.call_internal_no_propagate(
```
gmarkall, line 319: In conjunction with the suggestion above on `call_internal_no_propagate`:

```suggestion
    def call_function_and_propagate(self, builder, fndesc, sig, args):
        """
        Given the function descriptor, call the function and propagate
        exception status
        """
        status, res = self.call_function(
```

docs/source/user/callconv.rst
```diff
@@ -0,0 +1,107 @@
...
+
+Behavioral caveats
+------------------
+
+- The C ABI has no status channel for Python exception propagation.
+- When a C ABI caller invokes a Numba ABI callee returning ``Optional[T]``,
+  the optional is flattened to ``T`` at the C ABI boundary. A ``None`` result
+  is represented as the default-initialized value of ``T``.
+- Kernels must still use the Numba ABI entry model; compiling kernels with
+  ``abi="c"`` is unsupported.
```
isVoid, line 104: I suppose we can now relax this limitation, and always default to C ABI when compiling a kernel after this refactor?

numba_cuda/numba/cuda/tests/cudapy/test_compiler.py
isVoid: This is now reverted since there are separate implementations of `call_internal` for both Numba ABI and C ABI. User should find this working out of the box.

docs/source/user/callconv.rst
gmarkall: The Numba ABI for kernels is actually different and undocumented - there's a side channel for exception information when debug is enabled. I think making them C ABI needs a bit of work to decide what to do about passing arrays (which are an array struct usually in Numba rather than just the pointer) and what to do about exception statuses.

Issue #573, opened 2025-11-03
[BUG] Use of `weakref.proxy` for `rtsys._memsys_module` appears to be incorrect


cpcloud:
**Describe the bug**

I am setting up `pytest-randomly` in order to catch test-order-dependent bugs and this was the first one I found.

When a context is reset, and there's only a single reference to the `rtsys._memsys_module` object, it gets collected, even though its APIs are still available for use.

The following code is the smallest reproducer I could come up with.

**Steps/Code to reproduce bug**

```python
import pytest

from numba.cuda.cudadrv import devices
from numba.cuda.memory_management.nrt import rtsys


@pytest.fixture
def alloc_init():
    rtsys.ensure_allocated()
    rtsys.ensure_initialized()


@pytest.fixture
def ctx(alloc_init):
    ctx = devices.get_context()
    yield ctx
    ctx.reset()

    # this fails with a ReferenceError
    str(rtsys._memsys_module)


def test_nothing(ctx):
    """
    The sequence that causes the failure is:

    1. the alloc_init fixture creates a weakref proxy to a CudaPythonModule
       (rtsys._memsys_module) from devices.get_context()
    2. ctx.reset() calls dict.clear() on the context's module dict
    3. since there's only a single strong ref to `rtsys._memsys_module` it gets collected
       during the clear
    4. All ops on the module (a weakref proxy) fail, because it is now
       referencing a dead object
    """
```

**Expected behavior**

Since `rtsys` is a module global, with APIs that liberally use `_memsys_module`, I would not expect that clearing my current context would immediately make them unusable.

gmarkall:
Is this the root cause of #517?

cpcloud:
I think so, yes. That is the same failure mode.

Zooming out, I'm not sure that a weakref here is correct. That said, "just" returning a (non-weakref) object causes other errors that I haven't yet debugged.

From my debugging it looks like the lifetimes are:

0. `Program` (Python process)
1. `Runtime` (`~Program`; it's a singleton)
2. `Context` (`<= Program`; generally `< Program`)
3. `Module` (`<= Context`; since they are owned by `Context` via the `modules` member)

This means that `Runtime` holding onto modules would be invalid under these constraints, since `Runtime` methods potentially use these objects that are liable to be cleaned up by `Context` while `Runtime` is still in use or available for use.

Perhaps `Module`s should be owned by `Runtime` instead of `Context`?

cpcloud:
[...]

PR #783, opened 2026-02-09
Clear `rtsys` during context reset

brandon-b-miller:
This PR fixes an issue that intermittently surfaces in CI where the `rtsys` singleton remains valid after the context containing it's module has been reset. 

This fixes the immediate problem when running tests while leaving some wider questions from https://github.com/NVIDIA/numba-cuda/issues/573 open.

### DIFF
--- a/numba_cuda/numba/cuda/cudadrv/driver.py
+++ b/numba_cuda/numba/cuda/cudadrv/driver.py
@@ -1066,6 +1066,11 @@ class Context:
         self.modules.clear()
         # Clear trash
         self.deallocations.clear()
+        # Reset NRT runtime so it does not hold a dangling weakref to a
+        # module that was just removed from self.modules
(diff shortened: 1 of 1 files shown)


gmarkall (approved):

gmarkall:
> while leaving some wider questions from https://github.com/NVIDIA/numba-cuda/issues/573 open.

It looks like this is a proper solution to the problem - What else am I missing / what else is there to address in #573?

brandon-b-miller:
> > while leaving some wider questions from #573 open.
> 
> It looks like this is a proper solution to the problem - What else am I missing / what else is there to address in #573?

It seemed like [this conversation](https://github.com/NVIDIA/numba-cuda/issues/573#issuecomment-3486863647) had some lingering questions around the how cuda resource objects should own each other.

PR #784, opened 2026-02-10
Remove most of `drvapi.py` in favor of direct `cuda-python` usage

brandon-b-miller:
This PR removes `drvapi.py`, which contains some dead code that was used to support both bindings historically and now contains mostly inlineable function calls. This also allows us to remove some superfluous and potentially error prone handle->int->pointer casting when working with resource handles like `CUstream, CUevent`, etc.

### DIFF
--- a/numba_cuda/numba/cuda/cudadrv/driver.py
+++ b/numba_cuda/numba/cuda/cudadrv/driver.py
@@ -540,7 +540,7 @@ class Device:
             ctx_handle = self._dev.context._handle
         self.primary_context = ctx = Context(
             weakref.proxy(self),
-            ctypes.c_void_p(int(ctx_handle)),
+            ctx_handle,
         )
         return ctx
 


brandon-b-miller:
Couple of local test failures still, following up

numba_cuda/numba/cuda/cudadrv/drvapi.py
kkraus14, line 1: There is some usage of this in the wild, primarily to reach for `cu_stream`. Seems like outdated dask forks / vendoring.

Should we consider a deprecation cycle for this?

brandon-b-miller: I think that's reasonable, I'll modify things so usage of `drvapi` results in a warning but move off of it internally.

brandon-b-miller: @kkraus14 were the usages you found primarily projects using upstream `numba` or true `numba-cuda`? That'd give some extra signal of what might be impacted.

kkraus14: It was mostly old / unmaintained code with the exception of: https://github.com/xorbitsai/xoscar/blob/main/python/xoscar/backends/communication/ucx.py#L38-L55

We could probably get away with just opening an issue / PR to that repo?

brandon-b-miller: Opened an issue there!

numba_cuda/numba/cuda/cudadrv/drvapi.py
```diff
@@ -16,6 +18,12 @@
     Structure,
 )
 
+warnings.warn(
+    "Importing from numba.cuda.cudadrv.drvapi is deprecated.",
+    DeprecationWarning,
+    stacklevel=2,
+)
```
kkraus14, line 25: Just to confirm, this warning will only pop up if someone tries to explicitly use this module because it won't be used elsewhere?

kkraus14 (approved):

numba_cuda/numba/cuda/cudadrv/drvapi.py
brandon-b-miller: Yeah, the expected behavior is for a warning to emerge when this module is imported, which shouldn't occur anywhere internally anymore or in our tests. So it has to be user code importing from the module.

Issue #200, opened 2025-04-11
[FEA] Support fp8 data types


kkraus14:
- [ ] e4m3
- [ ] e5m2
- [ ] e8m0

PR #788, opened 2026-02-11
Create Highlevel Bindings for FP8 Datatype

isVoid:
This PR adds high level exposure of fp8 data type to Numba-CUDA.

Supported features include:
- FP8 constructors from existing data type (elementwise and packed)
- Conversion intrinsics that provide finer control of saturation type

Supported FP8 variants (element wise):
- fp8_[e5m2, e4m3, e8m0]

Supported packed FP8 variants:
- fp8[x2, x4]_[e5m2, e4m3, e8m0]

This PR also adds tests for packed type bindings introduced in #686.

closes https://github.com/NVIDIA/numba-cuda/issues/200


numba_cuda/numba/cuda/fp8.py
```diff
@@ -3,10 +3,101 @@
...
+    cvt_e8m0_to_bf16raw as _cvt_e8m0_to_bf16raw,
 )
+from numba.cuda.bf16 import (
+    _bfloat16_as_bfloat16_raw,
+    _bfloat16_raw_as_bfloat16,
+)
+from numba.cuda.extending import register_jitable
+
+
+@register_jitable
```
isVoid: The use of `@register_jittable` is recommended by code agents. Is this a good choice in today's Numba?

gmarkall: It only makes sense if you want to call the function as a pure Python function.

gmarkall: (So `register_jitable` just makes no sense in the context of Numba-CUDA)

isVoid: This PR shows that `register_jittable` may enable jitting function that's not directly callable in pure python. As shown by

```python
@register_jitable
def bfloat16_to_e8m0(x, saturate, rounding):
    return _cvt_bfloat16raw_to_e8m0(
        _bfloat16_as_bfloat16_raw(x), saturate, rounding
    )
```

Where `_bfloat16_as_bfloat16_raw` is written as numba intrinsics and may not be called with these arguments as-is.

Proposing updating the docstring of `register_jittable` and keeping the function.

[EDIT]: We can replace with `cuda.jit` to avoid confusion that `bfloat16_to_e8m0` is unusable on host.

isVoid: 203c3a8

PR #790, opened 2026-02-13
Support Interop with External Objects Via DLPack

isVoid:
Depends on https://github.com/NVIDIA/cuda-python/pull/1623

### DIFF
--- a/numba_cuda/numba/cuda/tests/cudapy/test_cupy_bfloat16_interop.py
+++ b/numba_cuda/numba/cuda/tests/cudapy/test_cupy_bfloat16_interop.py
@@ -8,8 +8,11 @@ from numba import cuda
 from numba.cuda.testing import unittest, CUDATestCase, skip_on_cudasim
 
 
-pytest.importorskip("cuda.core", minversion="0.6.0",
-                    reason="requires cuda.core 0.6.0 or later for DLPack + bf16 support")
+pytest.importorskip(
+    "cuda.core",
+    minversion="0.6.0",
+    reason="requires cuda.core 0.6.0 or later for DLPack + bf16 support",
+)
 cp = pytest.importorskip(
     "cupy",
     minversion="14.0.0",


leofang:
@isVoid let's resolve the conflicts and get this merged! I think this is the only missing piece; cuda-core and CuPy are both ready.

ci/tools/run-tests
leofang, line 1: Let's add `ml_dtypes` to the `test` group (since it's a CPU-only project) and then we can undo all changes to this file:
https://github.com/NVIDIA/numba-cuda/blob/5df7dcbc0efb6406a092331628bdf4c8b6ca8b5f/pyproject.toml#L62-L72

leofang: Oh. It's already there. Then I think these changes are not needed? The `test-cuXX` groups also include `test`, so it should already be installed everywhere.

numba_cuda/numba/cuda/np/numpy_support.py
```diff
@@ -74,6 +74,15 @@ def strides_from_shape(
...
 }
 
+# Optional dtype support (not provided by NumPy itself).
+try:
+    import ml_dtypes  # type: ignore
+except ImportError:
+    ml_dtypes = None
+
+if ml_dtypes is not None:
+    FROM_DTYPE[np.dtype(ml_dtypes.bfloat16)] = types.bfloat16
```
leofang, line 84: For my own understanding: `types.bfloat16` came from a previous work where bf16 was already registered to numba's type system?

numba_cuda/numba/cuda/tests/cudapy/test_cupy_bfloat16_interop.py
leofang, line 1: Note to self: I dunno very well about the test infrastructure so I assume making a new file is OK.

numba_cuda/numba/cuda/typing/typeof.py
```diff
@@ -351,6 +358,29 @@ def _typeof_cuda_array_interface_cached(
     return types.Array(dtype, ndim, layout, readonly=readonly)
 
 
+def _typeof_dlpack(val, c):
+    obj = getattr(val, "__self__", None)
+    if obj is not None:
+        smv = StridedMemoryView.from_dlpack(
+            obj, stream_ptr=-1
+        )  # stream is probably unimportant here?
```
leofang: yes

numba_cuda/numba/cuda/typing/typeof.py
```diff
@@ -351,6 +358,29 @@ def _typeof_cuda_array_interface_cached(
...
+
+        smv_layout = smv._layout
```
leofang, line 366: Q: Is it a must for `types.Array` to get `layout="C"` or `"F"`? Can we pass it shapes/strides instead? I am a bit nervous about using cuda.core's private attributes given recent incidents.

leofang (changes requested):
Thanks, @isVoid! LGTM overall. Left a few comments and suggestions.

numba_cuda/numba/cuda/typing/typeof.py
```diff
@@ -351,6 +358,29 @@ def _typeof_cuda_array_interface_cached(
     return types.Array(dtype, ndim, layout, readonly=readonly)
 
 
+def _typeof_dlpack(val, c):
+    obj = getattr(val, "__self__", None)
+    if obj is not None:
+        smv = StridedMemoryView.from_dlpack(
```
leofang: For this to work: I think we need cuda.core >=0.6.0?

isVoid: Yeah - specifically, for bfloat16 support.

numba_cuda/numba/cuda/np/numpy_support.py
isVoid: Yes that's correct. This is mapping ml_dtypes.bfloat16 back into numba.bfloat16

ci/tools/run-tests
isVoid: Ah good find!

numba_cuda/numba/cuda/typing/typeof.py
isVoid: It looks like array's field on layout is a must. It has to be one of the following value:

https://github.com/NVIDIA/numba-cuda/blob/5df7dcbc0efb6406a092331628bdf4c8b6ca8b5f/numba_cuda/numba/cuda/types/cuda_common.py#L49

isVoid: There's existing logic where contiguity is inferred. I'm reusing that for this part to avoid the private field access.

pyproject.toml
```diff
@@ -17,7 +17,7 @@ requires-python = ">=3.9"
 dependencies = [
     "numba>=0.60.0",
     "cuda-bindings>=12.9.1,<14.0.0",
-    "cuda-core>=0.5.1,<1.0.0",
+    "cuda-core>=0.6.0,<1.0.0",
```
leofang: I'd suggest that we keep 0.5.1 as the lower bound and add a version check in the library and test code. The only purpose is to enable bf16 which is optional. Bumping the lower bound seems like an overkill.

pixi.toml
```diff
@@ -304,7 +304,7 @@ numpy = ">=2.1"
 python = "*"
 packaging = "*"
 numba = ">=0.60"
-cuda-core = ">=0.5.1,<1"
+cuda-core = ">=0.6.0,<1"
```
leofang, line 307: Since pixi is developer-oriented, I am fine keeping this tight lower bound.

numba_cuda/numba/cuda/typing/typeof.py
leofang: My apology, I think commit [6701e1d](https://github.com/NVIDIA/numba-cuda/pull/790/commits/6701e1d2422d3a749ba771a117c2ad61f7d2e629) might have introduced bugs (3 CI failed) plus I recognize the layout inference adds perf overhead. Let me revert it for now 🙇

leofang (changes requested):

numba_cuda/numba/cuda/typing/typeof.py
leofang: Done in commit c52e5029

pyproject.toml
leofang: done in commit cd1530b26

leofang:
/ok to test [cd1530b](https://github.com/NVIDIA/numba-cuda/pull/790/commits/cd1530b26c30f1161222dcf1fe7593edc259b91e)

leofang:
/ok to test [e72db43](https://github.com/NVIDIA/numba-cuda/pull/790/commits/e72db43105af6431a01b17f683a169bb5a5e912b)

numba_cuda/numba/cuda/tests/cudapy/test_datetime.py
```diff
@@ -62,7 +62,11 @@ def timediff(start, end):
 
     @skip_on_cudasim("API unsupported in the simulator")
     def test_datetime_cupy_inputs(self):
-        cp = pytest.importorskip("cupy")
+        cp = pytest.importorskip(
+            "cupy",
+            minversion="14.0.0",
+            reason="requires CuPy >= 14 for datetime DLPack interop",
```
leofang, line 68: Not a blocker but I think datetime goes through CAI not DLPack

numba_cuda/numba/cuda/typing/typeof.py
leofang: Note to self: `typeof_strided_memory_view` is already using `_layout`, so if things break in the future it's not because of this PR... lol

numba_cuda/numba/cuda/tests/cudapy/test_datetime.py
isVoid: Ah this is hallucination.

PR #794, opened 2026-02-18
Use only public `.handle` to access `cuda-core` objects

brandon-b-miller:
Fix an issue where we're erroneously accessing the `_handle` of `cuda-core` objects rather than the `.handle`.


kkraus14:
@brandon-b-miller it looks like we only introduced it as a public attribute in the upcoming 0.6 release where we'd need to handle it conditionally it seems.

PR #795, opened 2026-02-18
Fix mixed-IR liveness for inline overload DCE

cpcloud:
## Summary
- fix use/def and DCE liveness tracking for mixed `numba.core` and vendored CUDA IR nodes created by `inline="always"` overload inlining
- preserve inlined overload dependencies so DCE no longer drops the live `static_getitem` feeding the return value in issue #718
- add a user-style regression with a module-level overload and kernel launch using `cuda.local.array`
- refresh `pixi.lock` local package entries so `numba-cuda` variants resolve to `0.27.0`

Related: the separate `np.dtype` signature compatibility fix was split into #797.

Fixes #718

PR #797, opened 2026-02-18
Fix np.dtype overload signature drift

cpcloud:
## Summary
- align CUDA `np.dtype` overload parameters with NumPy (`dtype`, `align`, `copy`) while avoiding unsupported `**kwargs` in typing templates
- align CUDA `np.dot` and `np.vdot` overload parameter names with NumPy (`a`, `b`, `out`) so signature-compatibility checks pass across NumPy variants
- preserve existing lowering behavior for supported dtype and BLAS-backed dot/vdot paths

Extracted from #795 to keep the issue-718 DCE fix focused.

### DIFF
--- a/ci/test_thirdparty_cudf.sh
+++ b/ci/test_thirdparty_cudf.sh
@@ -5,6 +5,7 @@
 set -euo pipefail
 
 CUDA_VER_MAJOR_MINOR=${CUDA_VER%.*}
+CUDA_VER_MAJOR=${CUDA_VER%.*.*}
 
 rapids-logger "Install cuDF Wheel"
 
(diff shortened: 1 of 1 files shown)


brandon-b-miller (approved):
cuDF failures are just a broken nightly I think, lets rerun CI in the morning

brandon-b-miller:
> cuDF failures are just a broken nightly I think, lets rerun CI in the morning

Sorry @cpcloud this was wrong, I forgot we're on the stable release of rapids at this point and it looks like cupy 14 is breaking that release. Can we try installing cupy<0.14.0 in the test script before the tests are run?

https://github.com/NVIDIA/numba-cuda/blob/main/ci/test_thirdparty_cudf.sh

We can undo this later after the cupy patch version comes out.

PR #800, opened 2026-02-20
Fix cache invalidation logic.

tpn:
This PR incorporates the numba-cuda version into the cache key to ensure caches are invalidated when *either* numba or numba-cuda versions change.  (This is required by some upcoming `cuda.coop` changes, which will require cache invalidation upon numba-cuda version changing.)

### DIFF
--- a/numba_cuda/numba/cuda/tests/nocuda/test_cache_version.py
+++ b/numba_cuda/numba/cuda/tests/nocuda/test_cache_version.py
@@ -5,43 +5,67 @@ import os
 import pickle
 import tempfile
 import unittest
+from unittest import mock
 
 import numba
 import numba_cuda
 
+from numba.cuda.core.caching import Cache, IndexDataCacheFile
+
+
+def dummy_func(x):
+    return x
+
+
+class DummyCodegen:
+    def magic_tuple(self):
+        return ("dummy",)
+
 
 class TestCacheVersion(unittest.TestCase):
-    def test_cache_version_includes_numba_cuda(self):
-        with tempfile.TemporaryDirectory() as tmp:
-            from numba.cuda.core.caching import IndexDataCacheFile
+    @staticmethod
+    def make_key(version=None):
+        cache = object.__new__(Cache)
+        cache._py_func = dummy_func
+        codegen = DummyCodegen()
+        if version is None:
+            return Cache._index_key(cache, ("sig",), codegen)
+        with mock.patch.object(numba_cuda, "__version__", version):
+            return Cache._index_key(cache, ("sig",), codegen)
 
+    def test_cache_index_version_matches_numba(self):
+        with tempfile.TemporaryDirectory() as tmp:
             cache = IndexDataCacheFile(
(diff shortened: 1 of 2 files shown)


testing/test_cache_version.py
gmarkall, line 1: Why is this in `testing` and not a subdir of `numba_cuda/numba/cuda/tests`?

numba_cuda/numba/cuda/tests/nocuda/test_cache_version.py
```diff
@@ -0,0 +1,81 @@
...
+import os
+import pickle
+import sys
+import tempfile
+import unittest
+
+import numba
+import numba_cuda
+
+os.environ.setdefault("NUMBA_ENABLE_CUDASIM", "1")
```
gmarkall: What is this for?

testing/test_cache_version.py
tpn: Turns out the behavior being tested here is already handled by the tests we introduce in the right location.  So, I've removed this test.

numba_cuda/numba/cuda/tests/nocuda/test_cache_version.py
tpn: Removed.

numba_cuda/numba/cuda/tests/nocuda/test_cache_version.py
```diff
@@ -0,0 +1,74 @@
...
+import sys
+import tempfile
+import unittest
+
+import numba
+import numba_cuda
+
+
+@contextlib.contextmanager
+def _cuda_bindings_stub():
```
gmarkall: Why does this exist? It appears to have no bearing on whether the test passes or fails (I replaced its definition with `yield; return` and it still fails on `main` and passes on this branch.

tpn: Good catch. `_cuda_bindings_stub()` was leftover scaffolding from an earlier version of the test and it isn’t needed for the current coverage.

I’ve removed the helper and the `with _cuda_bindings_stub():` wrappers so the test just exercises `IndexDataCacheFile` directly. I also re-ran the cache-version tests in an isolated pixi env, and they passed.

leofang:
I don't understand what the cache test failure is about...?

gmarkall:
I think the change has broken the cache keying, but I'm just having a look locally now.

leofang:
oh nvm, I did not mean to review this particular PR, was looking for something else, sorry

gmarkall:
The caching test fails appearing in CI are happening for me locally too.

@tpn Have you tried the tests locally? Have you observed this behaviour?

gmarkall:
I suspect the fails occur because we have a divergence in the cache keying between Numba-CUDA and upstream Numba, so sharing the cached code between the targets no longer works.

tpn:
@gmarkall I haven't seen anything like this locally.  I'll investigate.

Issue #648, opened 2025-12-10
figure out exactly what pickling a stream means


cpcloud:
Yes, pickling a pointer is indeed weird, however, that is the status quo, as `Stream` uses the default pickling implementation and its fields are `handle` (stream handle), and `external` (bool).

_Originally posted by @cpcloud in https://github.com/NVIDIA/numba-cuda/pull/609#discussion_r2607944476_

kkraus14:
Given that we can't safely pickle a Stream, what is dependent on pickling a Stream currently and could we pull the Stream out of that in some way?

cpcloud:
`CUDADispatcher` is picklable, which means `_LaunchConfiguration` must be picklable, which means all of its fields must be picklable and one of those fields is a `cuda.core.experimental.Stream`, which has a non-trivial `__cinit__` and is thus not picklable (without the current underwhelming workaround).

cpcloud:
One option is to not pickle the stream at all, and construct the default stream when unpickling.

kkraus14:
Does `CUDADispatcher` need to be picklable? If it really does, then a Stream probably shouldn't be a part of it. Constructing the default stream when unpickling is really ugly behavior.

cpcloud:
> Does `CUDADispatcher` need to be picklable?

Not sure, but [there's a test for this behavior](https://github.com/NVIDIA/numba-cuda/blob/main/numba_cuda/numba/cuda/tests/cudapy/test_serialize.py?plain=1#L52).

> If it really does, then a Stream probably shouldn't be a part of it. Constructing the default stream when unpickling is really ugly behavior.

To clarify, it's just that `_LaunchConfiguration` is a part of `CUDADispatcher`, and `_LaunchConfiguration` has a `stream` field.

kkraus14:
> Not sure, but [there's a test for this behavior](https://github.com/NVIDIA/numba-cuda/blob/main/numba_cuda/numba/cuda/tests/cudapy/test_serialize.py?plain=1&rgh-link-date=2025-12-11T14%3A22%3A40.000Z#L52).

I think it's this specific test that's problematic? https://github.com/NVIDIA/numba-cuda/blob/main/numba_cuda/numba/cuda/tests/cudapy/test_serialize.py?plain=1#L36-L39

[...]

PR #804, opened 2026-02-20
Implement launch config infrastructure.

tpn:
## Summary
This PR introduces two related pieces of launch-config infrastructure needed by
`cuda.coop` single-phase work:

1. A low-overhead launch-config API that can be consumed from arbitrary
   numba-cuda compilation stages (including rewrites), plus pre-launch callback
   registration on configured launches.
2. Launch-config-sensitive (LCS) plumbing so kernels that depend on launch
   configuration are specialized and cached correctly across launch configs.

## Background and motivation
`cuda.coop` single-phase rewriting needs compile-time access to launch
configuration details (grid/block/shared memory/launch args) and a way to
register pre-launch hooks from rewrite time (for launch-time kernel argument
handling without requiring `@cuda.jit(extensions=...)`).

An earlier implementation (PR #288) provided this via Python `contextvars`, but
review feedback showed launch overhead was too high. This branch reimplements
the mechanism through C-extension TLS plumbing in `_dispatcher.cpp`, with
negligible overhead in the launch micro-benchmark.

From `bench-launch-overhead.out` (us/launch, baseline vs contextvar vs v2):
- 0 args: `5.56` vs `7.29` (+31.1%) vs `5.56` (+0.0%)
- 1 arg: `7.53` vs `9.18` (+21.8%) vs `7.55` (+0.2%)
- 2 args: `8.90` vs `10.64` (+19.5%) vs `8.97` (+0.8%)
- 3 args: `10.31` vs `12.50` (+21.3%) vs `10.37` (+0.5%)
- 4 args: `11.82` vs `13.56` (+14.7%) vs `11.92` (+0.8%)

### 1) Launch-config API with low launch overhead
- C-extension (`numba_cuda/numba/cuda/cext/_dispatcher.cpp`) now carries the
  active launch config in thread-local storage only during compilation paths.
- Python API in `numba_cuda/numba/cuda/launchconfig.py`:
  - `current_launch_config()`
  - `ensure_current_launch_config()`
  - `capture_compile_config()`
- Configured launches expose:
  - launch metadata (`griddim`, `blockdim`, `sharedmem`, `args`, `dispatcher`)
  - `pre_launch_callbacks` for just-in-time launch-time hook registration.

### 2) Launch-config-sensitive compilation/caching
- Explicit LCS marker API on `_LaunchConfiguration`
  (`numba_cuda/numba/cuda/dispatcher.py`):
  - `mark_kernel_as_launch_config_sensitive()`
  - `get_kernel_launch_config_sensitive()`
  - `is_kernel_launch_config_sensitive()`
- CUDA backend (`numba_cuda/numba/cuda/compiler.py`) promotes that mark into
  compile metadata (`state.metadata["launch_config_sensitive"] = True`).
- Dispatcher/cache behavior for LCS kernels:
  - per-launch-config dispatcher specialization routing
  - per-launch-config disk-cache keys
  - `.lcs` marker file indicating launch-config-sensitive cache entries.

## Why the LCS piece is required
Without LCS, cache keys are signature-based only, so a kernel compiled once for
launch config A can be reused for launch config B without rerunning rewrite.
That breaks launch-config-dependent rewrite behavior.

Concrete observed behavior:
- Runtime cache (single process):
  - Launch `[1, 32]`: rewrite runs, callback registered.
  - Launch `[1, 64]` without LCS: existing kernel reused, rewrite does not run,
    callback for the 64-config path is never registered.
  - With LCS marking: second launch recompiles under a distinct launch-config
    specialization, so rewrite/callback registration runs for 64.
- Disk cache (cross process):
  - Process 1 compiles and caches launch `[1, 32]`.
  - Process 2 launches `[1, 64]` without LCS: 32-config artifact can be reused
    from disk (no rewrite for 64 path).
  - With LCS marking: process 2 misses on 64-specific cache key and compiles a
    64-specific variant.
- LCS intentionally preserves exact cache hits for matching launch configs. It
  does not force recompilation when the launch-config key already matches.

So the LCS plumbing is what makes launch-config-dependent rewrite decisions
correct under both in-memory and disk cache reuse.

Scope note for `cuda.coop` today:
- `cuda.coop` frequently injects LTO-IR/linking files during compilation.
- numba-cuda currently does not disk-cache kernels with linking files, so for
  those paths the immediate LCS correctness benefit is runtime/in-memory cache
  behavior across launch configs.
- Disk-cache LCS behavior applies to launch-config-sensitive kernels that are
  otherwise disk-cacheable (and remains relevant for future linked-code cache
  support).

## Safety behavior
- If an LCS kernel is loaded from disk but the `.lcs` marker is missing, we
  treat that cache state as unsafe, force recompile, and re-mark.
- If marking fails (e.g. filesystem error), disk caching is disabled for safety
  (fallback to `NullCache`) to avoid unsafe reuse.

## Out of scope
- Cache invalidation keyed on `numba_cuda.__version__` (handled by PR #800).
  Note that PR #800 should be merged and presumably a release cut before this
  PR is merged--that allows downstream projects like cuda.coop to pin accordingly.

### DIFF
--- a/docs/source/reference/kernel.rst
+++ b/docs/source/reference/kernel.rst
@@ -131,6 +131,10 @@ Pre-launch callbacks
 Configured launches expose ``pre_launch_callbacks``. Each callback is called
 immediately before launch with ``(kernel, launch_config)``.
 
+.. warning:: Pre-launch callbacks must not invoke CUDA APIs or launch CUDA
+   work. This use is not supported, is not tested, and has undefined
+   behavior; it may deadlock or fail in other ways.
+
 .. code-block:: python
 
    cfg = f[1, 1]
--- a/numba_cuda/numba/cuda/dispatcher.py
+++ b/numba_cuda/numba/cuda/dispatcher.py
@@ -697,6 +697,17 @@ class ForAll:
 
 
 class _LaunchConfiguration:
+    """Configured CUDA kernel launch metadata.
+
+    The ``pre_launch_callbacks`` list may be used to register host-side
+    callbacks that run immediately before kernel launch with
+    ``(kernel, launch_config)``.
+
+    Callbacks must not invoke CUDA APIs or launch CUDA work. That use is not
+    supported, is not tested, and has undefined behavior; it may deadlock or
+    fail in other ways.
+    """
+
     def __init__(self, dispatcher, griddim, blockdim, stream, sharedmem):
         self.dispatcher = dispatcher
         self.griddim = griddim


tpn:
I've left my WIP PR up for this v2 work, which includes some plots re kernel launch overhead, e.g.: https://github.com/NVIDIA/numba-cuda/pull/727/changes#diff-0a7ee1c443b1d66342149e5f9b416d2817f22c97d8c6664da3507c15a0e5454b

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -693,6 +702,9 @@ def __init__(self, dispatcher, griddim, blockdim, stream, sharedmem):
         self.blockdim = blockdim
         self.stream = driver._to_core_stream(stream)
         self.sharedmem = sharedmem
+        self.pre_launch_callbacks = []
+        self.args = None
+        self._kernel_launch_config_sensitive = None
```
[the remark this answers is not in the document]
tpn: I'm going to defer to numba-cuda developers here regarding whether or not this is actually problematic; I vaguely recall numba-cuda not being thread-safe re this particular code path in the first place.

docs/source/reference/kernel.rst
```diff
@@ -57,6 +57,87 @@ This is similar to launch configuration in CUDA C/C++:
 .. note:: The order of ``stream`` and ``sharedmem`` are reversed in Numba
    compared to in CUDA C/C++.
 
+Launch configuration access (advanced)
```
gmarkall, line 70: The placement of this section cuts the documentation of the Dispatcher objects in two. I think this issue could be resolved by placing these additions just before the "Kernel Arguments" section.

docs/source/reference/kernel.rst
```diff
@@ -57,6 +57,87 @@ This is similar to launch configuration in CUDA C/C++:
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If code generation depends on launch configuration (for example, a rewrite
+pass that inspects ``cfg.blockdim`` and emits different IR), mark the active
+launch as launch-config sensitive:
+
+.. code-block:: python
+
+   cfg = launchconfig.ensure_current_launch_config()
+   cfg.mark_kernel_as_launch_config_sensitive()
```
gmarkall: Why do we use a configuration to mark a kernel as launch config sensitve? Shouldn't the dispatcher object be marked as launch-config-sensitive?

docs/source/reference/kernel.rst
tpn: Moved in fe9d49ab6451ac7d3e6ebd29c9a53c3b567dbac6. The launch-config section now sits just before "Kernel Arguments", so the dispatcher-object documentation stays contiguous.

docs/source/reference/kernel.rst
tpn: Changed on the current branch (rebased head `e3dafbc`): the primary API is now dispatcher-owned via `CUDADispatcher.mark_launch_config_sensitive()`, and the docs/tests now use `cfg.dispatcher.mark_launch_config_sensitive()`. I kept `_LaunchConfiguration.mark_kernel_as_launch_config_sensitive()` as a thin compatibility wrapper so existing callers such as `cuda.coop` do not break immediately.

tpn: Follow-up on the current branch (`d5828580`): I dropped that compatibility wrapper. `_LaunchConfiguration` no longer exposes `mark_kernel_as_launch_config_sensitive()`; `CUDADispatcher.mark_launch_config_sensitive()` is now the only marker API on this branch, and the compiler-side flag check is private.

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -1585,6 +1678,80 @@ def __getitem__(self, args):
...
+            key = self._cache_launch_config_key(launch_config)
+            self._cache.set_launch_config_key(key)
+            return
+        if self._cache.is_launch_config_sensitive():
+            if launch_config is None:
+                key = None
+            else:
+                key = self._launch_config_key(launch_config)
+            self._cache.set_launch_config_key(key)
+            return
```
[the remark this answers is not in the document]
tpn: Fixed on the current branch in `82ddc75e`. LCS cache lookups with no active launch config now use a dedicated sentinel key instead of falling back to `None` / the default launch-config key, so `dispatcher.compile(sig)` no longer reuses a launch-config-specialized disk artifact. I also added a regression test in `test_caching.py` that exercises `compile(sig)` after populating LCS cache entries and asserts that it raises rather than silently loading from disk.

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -1632,16 +1799,36 @@ def __call__(self, *args, **kwargs):
...
+        blockdim = launch_config.blockdim
+        stream = launch_config.stream
+        sharedmem = launch_config.sharedmem
 
-        kernel.launch(args, griddim, blockdim, stream, sharedmem)
+        launch_config.args = args
+        try:
+            dispatcher = self._select_launch_config_dispatcher(launch_config)
+            if dispatcher is not self:
+                return dispatcher.call(args, launch_config)
```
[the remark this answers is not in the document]
tpn: Addressed by the same fix in `7d1e97c2`: the per-call args are now kept in thread-local state for the duration of `CUDADispatcher.call()`, rather than mutating the shared cached launch-config object. The new regression test in `test_dispatcher.py` exercises the exact two-thread interleaving that would previously have made `cfg.args` observe the wrong launch arguments.

numba_cuda/numba/cuda/dispatcher.py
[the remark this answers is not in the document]
tpn: Fixed on the current branch in `7d1e97c2`. `launch_config.args` is no longer stored on the cached `_LaunchConfiguration` instance; it now reads from thread-local per-call state, so concurrent launches of the same configured object do not stomp each others arguments. I also added a deterministic two-thread regression in `test_dispatcher.py` that forces both threads into `_compile_for_args` before reading `cfg.args`.

numba_cuda/numba/cuda/launchconfig.py
```diff
@@ -0,0 +1,59 @@
...
+    @wraps(original)
+    def wrapped(*args, **kws):
+        record["config"] = current_launch_config()
+        return original(*args, **kws)
+
+    dispatcher._compile_for_args = wrapped
+    try:
+        yield record
+    finally:
+        dispatcher._compile_for_args = original
```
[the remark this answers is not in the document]
tpn: I have not changed this helper on the branch. If we keep scope tight for this PR, my suggestion is to document `capture_compile_config()` as a convenience/testing helper that is **not** safe for concurrent use on the same dispatcher, rather than expanding this PR into a dispatcher-hook redesign. Concretely, I would put that warning in both the `launchconfig.py` docstring and the `kernel.rst` example/docs for the helper.

tpn:
@gmarkall oof, that sea of red with CI!  I'll start investigating.

tpn:
FWIW I can't repro the test deallocation failure locally, either on this branch or a pristine main env+tree.

numba_cuda/numba/cuda/dispatcher.py
```diff
@@ -716,19 +727,35 @@ def __init__(self, dispatcher, griddim, blockdim, stream, sharedmem):
...
+    def _push_args(self, args):
+        return _dispatcher.swap_current_launch_args(args)
+
+    def _pop_args(self, prev):
+        _dispatcher.swap_current_launch_args(prev)
 
     def __getstate__(self):
         state = self.__dict__.copy()
         state["stream"] = int(state["stream"].handle)
+        # Avoid serializing callables that may not be picklable.
```
gmarkall: The rationale behind this `__getstate__` and `__setstate__` method seems confusing at best: https://github.com/NVIDIA/numba-cuda/issues/648#issuecomment-3670591317

I think the side-effect of not serializing pre-launch callbacks is that they won't work for dispatchers that are used in distributed applications (and will silently not be called on remote workers). Would it be better to raise if there are pre-launch callbacks?

tpn: Changed in 9ff1eade: `_LaunchConfiguration.__getstate__()` now raises `pickle.PicklingError` when `pre_launch_callbacks` is non-empty, instead of silently dropping callbacks during pickling. That keeps the existing callback-free configured-launch pickling behavior, without broadening this PR into a claim about distributed callback support.

I also added a regression in `test_serialize.py` covering the callback-bearing configured-launch case.

Verified with:
`PYTHONPATH=/home/trentn/src/numba-cuda/numba_cuda python -m unittest numba.cuda.tests.cudapy.test_serialize`
which passed with 4 tests.

docs/source/reference/kernel.rst
```diff
@@ -63,9 +63,91 @@ creating a specialized instance:
...
+exposes launch metadata and callback hooks that can be consumed by advanced
+tooling (for example, rewrite passes and extension integrations).
+
+Compile-time launch-config access
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The active launch configuration is available only while compilation is in
+progress for a kernel launch.
+
+.. note:: This is compile-time state. If a launch reuses an existing compiled
```
tpn, line 85: Review and remove this bit with the `with launchconfig.capture_compile_config(f)` stuff.

docs/source/reference/kernel.rst
```diff
@@ -63,9 +63,91 @@ creating a specialized instance:
...
+immediately before launch with ``(kernel, launch_config)``.
+
+.. code-block:: python
+
+   cfg = f[1, 1]
+
+   def log_launch(kernel, cfg):
+       print(cfg.griddim, cfg.blockdim)
+
+   cfg.pre_launch_callbacks.append(log_launch)
```
tpn, line 145: Put comments somewhere re not using CUDA APIs in these callbacks.

PR #806, opened 2026-02-24
Fix incorrect DWARF type encodings for i8 and discriminator types

jiel-nv:
This change fixes incorrect DWARF type encodings in debug info generation:

1. `i8` types without a datamodel (e.g., `i8*` pointers for `meminfo` and `parent` fields in array structs) were incorrectly encoded as `DW_ATE_boolean`, now encoded as `DW_ATE_unsigned`.

2. The discriminator field in polymorphic variant types was named `"int"` but had a size of 8 bits (1 byte), renamed to `"uint8"` to accurately reflect the type without confusion.

3. Updated `test_ditypes` wiyh corresponding type encoding and type name changes.

Fixes nvbug5809146.

### DIFF
--- a/numba_cuda/numba/cuda/debuginfo.py
+++ b/numba_cuda/numba/cuda/debuginfo.py
@@ -654,11 +654,7 @@ class CUDADIBuilder(DIBuilder):
         m = self.module
 
         if isinstance(lltype, ir.IntType):
-            if datamodel is None:
-                if size == 1:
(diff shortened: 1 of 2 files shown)


jiel-nv:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/806/commits/d191325b8e11f668cd133c143c3c3035cc55363b

PR #807, opened 2026-02-24
Fix int64 elements DWARF encoding in UniTuple types

jiel-nv:
This change fixes incorrect DWARF type encoding for `int64` elements in UniTuple types. The element type was encoded as `DW_ATE_unsigned` with name `"i64"` instead of `DW_ATE_signed` with name `"int64"`.

The fix is to retrieve the element's data model and pass to the type creation methond.

Also updated `test_ditypes` with necessary changes.

Fixes another issue in nvbug5809146.


jiel-nv:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/807/commits/b3384e1a84544192f6baf25bc2a8f5604e80913f

PR #808, opened 2026-02-26
Fix missing DWARF debug entries for scalar kernel parameters

jiel-nv:
The nvbug5905507 exposes an issue, when a kernel contains control-flow branches, scalar kernel parameters are missing `DW_TAG_formal_parameter` DIEs in the DWARF debug info. This causes `cuda-gdb`'s `info args` to show an incomplete argument list.

The existing code relies on `llvm.dbg.value` alone for scalars, which suffices for tracking local variable values but does not survive control flow — the SSA value from the entry block does not dominate successor blocks, so NVVM may drop the associated debug metadata during optimization.

The fix is to guard the early return in `mark_variable()` with `argidx is None` so that scalar function parameters still get `llvm.dbg.declare`, while scalar local variables continue to use only `llvm.dbg.value`.

Also adds a new test `test_llvm_dbg_declare` and updates existing test `test_llvm_dbg_value` accordingly.

Fixes nvbug5905507.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,12 +26,36 @@ cu12 = [
     "cuda-bindings>=12.9.1,<13.0.0",
     "cuda-pathfinder>=1.3.4,<2.0.0",
     # install nvcc for libNVVM
-    "cuda-toolkit[cudart,nvcc,nvrtc,nvjitlink,cccl]==12.*",
+    "cuda-toolkit[cudart,nvcc,nvrtc,cccl]==12.*",
+
+    # Older nvjitlink is not supported by cuda-bindings. It is okay to be out
+    #  of sync with cuda-toolkit here, as long as the version compatibilty rules
+    #  are followed. See the compatibility rules for nvjitlink below.
+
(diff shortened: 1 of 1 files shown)


jiel-nv:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/808/commits/fff1548fa01b851d78b96ba1515104f3105e13aa

jiel-nv:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/808/commits/9b155c6a1acebd16b7c2861a8a8f471967ea408c

jiel-nv:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/808/commits/6afbbd87d6c814eaee4d35bad52df1bd506d2848

jiel-nv:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/808/commits/de117eb42652008ea55ee856d8d597dea8368d87

jiel-nv:
Decided to proceed with another approach. Closing this one.

PR #798, opened 2026-02-19
Find libdevice via pathfinder

brandon-b-miller:
This PR uses the recently merged [pathfinder improvements](https://github.com/NVIDIA/cuda-python/pull/1218) to locate the libdevice bitcode file rather than using custom pathfinding logic.

PR #810, opened 2026-02-26
Delegate all path discovery in cuda_paths.py to pathfinder

cpcloud:
## Summary

- Delegate all path discovery in `cuda_paths.py` to `cuda.pathfinder`, removing ~360 lines of bespoke conda/wheel/CUDA_HOME/system search logic
- Use `pathfinder.locate_bitcode_lib("device")` for libdevice and `pathfinder.locate_static_lib("cudadevrt")` for cudadevrt, replacing the `cudalib_dir` / `static_cudalib_dir` indirection layer
- Simplify `libs.py` to a thin wrapper over `get_cuda_paths()` with no directory-scanning fallback

Requires `cuda-pathfinder >= 1.4.0` for `locate_bitcode_lib` and `locate_static_lib` (NVIDIA/cuda-python#1690).

Closes NVIDIA/cuda-python#1036

xref NVIDIA/cuda-python#716

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -212,7 +212,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-nvcc-tools-12.0.76-h59595ed_1.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-nvcc_linux-64-12.0.76-hba56722_12.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-nvrtc-12.0.76-hd3aeb46_2.conda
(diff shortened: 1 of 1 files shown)


cpcloud:
Perhaps this should be in draft mode until the next release of `cuda-pathfiner`, since this PR requires being able to find `libcudadevrt.a`.

brandon-b-miller:
Some overlap here with https://github.com/NVIDIA/numba-cuda/pull/798. 

Do we need a CUDA_HOME fallback here to search the `/usr/local/cuda` dirs as is implemented there? Note the comment there `# Fallback: system CTK (pathfinder does not search /usr/local/cuda)`

cpcloud:
We should be able to use the canary fallback now.

PR #811, opened 2026-02-26
Fix empty DW_AT_location for reused loop variables

jiel-nv:
Nvbug5886953 exposes an issue that reused loop index variables have a DWARF location list with corrupted entries.

The root cause is in the base class `Lower.loadvar`, which suppresses `!dbg` metadata on loads of function argument variables. This causes NVVM to emit `.loc line 0` between the loops, and `ptxas` builds a multi-entry location list where the line-0 ranges inserted between entries corrupt it, producing an empty `DW_AT_location` in the final DWARF.

This PR overrides `loadvar` in `CUDALower` to skip the suppression. The suppression is unnecessary because `loadvar` is never called during the prologue — arg unpacking in `lower_assign` and `storevar` already handle suppression there.

Also add `test_arg_load_has_dbg_location` regression test.

Fixes nvbug5886953.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,12 +26,36 @@ cu12 = [
     "cuda-bindings>=12.9.1,<13.0.0",
     "cuda-pathfinder>=1.3.4,<2.0.0",
     # install nvcc for libNVVM
-    "cuda-toolkit[cudart,nvcc,nvrtc,nvjitlink,cccl]==12.*",
+    "cuda-toolkit[cudart,nvcc,nvrtc,cccl]==12.*",
+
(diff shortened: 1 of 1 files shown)


jiel-nv:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/811/commits/3c986d5039a787e2b90aafffa3d0d3ba563e208c

jiel-nv:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/811/commits/518d94b9f69bd1af2e55862d34250b03e49137fd

PR #813, opened 2026-02-27
Extend dbg.value coverage to loadvar for scalar kernel parameters

jiel-nv:
Nvbug5905507 exposes an issue: when a kernel contains control-flow branches, scalar kernel parameters may be missing `DW_TAG_formal_parameter` DIEs in DWARF debug info, causing `cuda-gdb`'s `info args` to show an incomplete argument list.

PR #461 added per-block `dbg.value` coverage via `storevar`, but that only triggers when the variable appears on the RHS of a direct assignment to an internal variable. When a scalar parameter is used directly as an arithmetic operand (e.g. `arr[idx] + scalar`), no `storevar` fires for it and the entry-block `dbg.value` alone is insufficient to guarantee the emission of `DW_TAG_formal_parameter`.

This PR adds a `loadvar` override in `CUDALower` that emits `llvm.dbg.value` for scalar user variables when they are loaded in a basic block but have not yet been covered by a `dbg.value` in that block. This ensures every scalar parameter has `dbg.value` coverage across all blocks that reference it.

Add `test_llvm_dbg_value_loadvar_coverage` to verify the extended coverage path. Also update `test_llvm_dbg_value_range` accordingly.

Fixes nvbug5905507.


jiel-nv:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/813/commits/a820f3cef1b6b48d95bc0a9625cecd9dd17bdbcd

jiel-nv:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/813/commits/68beb10645ccc7de3fe5d94cc0bfbcccaaf6f7f9

jiel-nv:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/813/commits/088597db144d92b20bb1729726a502c35c2b4b27

Issue #755, opened 2026-01-28
[BUG] Slow compile times for string arguments in device functions (0.22 vs 0.24)


Jlisowskyy:
**Bug description**
When using strings as device function arguments, compilation time explodes

**Steps to reproduce bug**
1. Prepare a clean python environment
2. Execute `pip install numba-cuda`
3. The code below demonstrates the behaviour:

```
import numpy as np
from numba import cuda
import sys

@cuda.jit(device=True, forceinline=True)
def load_slow(gmem, value):
    if value == "1":
        gmem[cuda.threadIdx.x] = 1.0
    else:
        gmem[cuda.threadIdx.x] = 0.0

@cuda.jit(device=True, forceinline=True)
def load(gmem, value):
    if value == 1.0:
        gmem[cuda.threadIdx.x] = 1.0
    else:
        gmem[cuda.threadIdx.x] = 0.0

def get_kernel(use_slow):
    if use_slow:
        @cuda.jit
        def slow(buff):
            load_slow(buff, "1")
            
        return slow
        
    @cuda.jit
    def quick(buff):
        load(buff, 1.0)
        
    return quick

def main():
    arg = sys.argv[1] if len(sys.argv) > 1 else "slow"
    kernel = get_kernel(arg == "slow")
    
    threads = 32

    buff = np.zeros((threads,), dtype=np.float32)
    buff_d = cuda.to_device(buff)

    kernel[1, threads](buff_d)
    cuda.synchronize()
    
    buff_result = buff_d.copy_to_host()
    
    assert np.linalg.norm(buff_result - np.ones((threads,), dtype=np.float32)) == 0.0
    print("finished")

if __name__ == "__main__":
    main()
```

**Expected behavior**
Using string arguments slows down compilation significantly (~2s vs ~2min):

```
time python .test.dir.c/slow.py quick

real    0m2.536s
user    0m2.442s
sys     0m0.166s

time python .test.dir.c/slow.py slow

real    1m57.233s
user    1m31.162s
sys     0m26.162s
```

**Environment details:**
 - Environment location: wsl local machine
 - Method of numba-cuda install: pip install
 - Python env:
```
numba-cuda==0.24.0
numba==0.63.1
Python 3.11.14
```

**Additional context**
This issue does not appear in `numba-cuda==0.22`. It appears that a redirect (in the error module), added in  `numba-cuda==0.22.1` triggers this behaviour.

Command log:
```
$ python3.11 -m venv .slow
[...]

PR #816, opened 2026-03-02
Revert exception hierarchy unification (#634) and its fix (#796)

cpcloud:
## Summary

- Revert #796 ("fix(errors): restore CUDA exception hierarchy to avoid slow string compilation"), which dynamically mirrored the CUDA error hierarchy as subclasses of core error classes.
- Revert #634 ("Always use `core` errors if `numba` is present"), which redirected `numba.cuda.core.errors` to `numba.core.errors` and introduced `cuda_errors.py` as a fallback.
- Add a regression test for #755 that compiles a kernel calling a device function with a string argument under a 30-second timeout, guarding against the ~2-minute compilation regression caused by the error module redirect.

## Motivation

PR #634 unified the CUDA and core error hierarchies by aliasing `numba.cuda.core.errors` to `numba.core.errors`. This broadened exception checks in CUDA typing/overload resolution, causing some compile paths (notably string-heavy cases) to stop short-circuiting, resulting in ~60x slower compilation (#755). PR #796 attempted to fix this with dynamic subclass mirroring, but the simplest and most maintainable fix is to revert both and restore the original independent CUDA error class definitions.

## Test plan

- [x] Regression test added: `TestStringArgCompileTime.test_string_device_function_arg_compiles_quickly` runs the #755 reproducer in a subprocess with a 30s timeout
- [ ] Existing `test_errors.py` tests pass
- [ ] CI passes

### DIFF
--- a/testing/pytest.ini
+++ b/testing/pytest.ini
@@ -9,7 +9,7 @@ addopts = --benchmark-disable --pyargs numba.cuda.tests
 
 filterwarnings =
     error
-    ignore:.*Grid size 1 will likely result in GPU under-utilization.+:numba.cuda.core.errors.NumbaPerformanceWarning
+    ignore:.*Grid size .* will likely result in GPU under-utilization.+:numba.cuda.core.errors.NumbaPerformanceWarning
     ignore:.*Grid size .* will likely result in GPU under-utilization.+:numba.core.errors.NumbaPerformanceWarning
     ignore:.*dynamic initialization is not supported for a function-scope static __shared__.*:UserWarning
     ignore:.*The keyword argument .nopython=False. was supplied.+:numba.core.errors.NumbaDeprecationWarning


ci/test_thirdparty_cudf.sh
```diff
@@ -46,8 +46,7 @@ rapids-logger "Run Scalar UDF tests"
 python -m pytest python/cudf/cudf/tests/dataframe/methods/test_apply.py -W ignore::UserWarning
 
 rapids-logger "Run GroupBy UDF tests"
-# Run JIT engine tests and tests that check jittability before falling back
-python -m pytest python/cudf/cudf/tests/groupby/test_apply.py -k test_groupby_apply -W ignore::UserWarning
+python -m pytest python/cudf/cudf/tests/groupby/test_apply.py -k test_groupby_apply_jit -W ignore::UserWarning
```
gmarkall, line 49: I get the impression from the description in #634 that the correct thing to do was to run all of the `test_groupby_apply` tests and not just the `test_groupby_apply_jit` tests - do you know if some of the `test_groupby_apply` tests are failing in conjunction with this PR?

cpcloud: I believe this is part of the revert.

gmarkall: Note from Slack discussion - these tests will be added back in a follow-up PR.

gmarkall (approved):

PR #818, opened 2026-03-02
Use `cuda-python` for `nvvm` bindings

brandon-b-miller:
NVVM is now available through `cuda-bindings`, we can drop our homegrown nvvm binding now.


cpcloud:
Merging, the backcompat (if desired) can be done in a follow-up.

PR #828, opened 2026-03-03
Use dbg.declare for scalar kernel parameters

cpcloud:
## Summary
- Describe scalar kernel parameters with `dbg.declare` on their stack slots to avoid unstable parameter-space locations in downstream DWARF (boolean parameters remain on `dbg.value` to avoid an NVVM crash).
- Avoid emitting additional `dbg.value` entries for function arguments once declared.
- Add a regression test for scalar formal parameters.


jiel-nv:
@cpcloud I appreciate the PR, this is a nice improvement!

I actually attempted the same `dbg.declare` direction in PR#808 but abandoned it after hitting the NVVM crash on boolean parameters. Ended up pivoting to the `loadvar` coverage approach in PR#813 instead.

Thanks again for the change!

cpcloud:
pixi.lock changes are fine to include

Issue #833, opened 2026-03-11
[BUG] CI is blocked after cuda-bindings 13.2.0


kaeun97:
It seems like CI is blocked after the new release.

The most recent PR that ran CI successfully was [this](https://github.com/NVIDIA/numba-cuda/pull/818). However, CI fails in [this PR](https://github.com/NVIDIA/numba-cuda/actions/runs/22950868694/job/66615482144?pr=774) and the difference seems to be cuda-bindings version (the failed tests use `13.2.0`).

Here is the test failure:
```
=================================== FAILURES ===================================
_________ FP8Storage_CVT_Intrinsics_Tests.test_cvt_bfloat16raw_to_e8m0 _________
[gw3] linux -- Python 3.12.13 /opt/hostedtoolcache/Python/3.12.13/x64/bin/python

self = <numba.cuda.tests.cudapy.test_fp8_bindings.FP8Storage_CVT_Intrinsics_Tests testMethod=test_cvt_bfloat16raw_to_e8m0>

    def test_cvt_bfloat16raw_to_e8m0(self):
        @cuda.jit
        def kernel(result, exponent):
            # NOTE: We currently don't have a good way to construct/modify a
            # bfloat16_raw value directly in kernels, so we skip SATFINITE vs
            # NOSAT behavior testing for bf16raw here.
            raw = cvt_e8m0_to_bf16raw(exponent[0])
            result[0] = cvt_bfloat16raw_to_e8m0(
                raw, saturation_t.NOSAT, cudaRoundMode.cudaRoundZero
            )
            result[1] = cvt_bfloat16raw_to_e8m0(
                raw, saturation_t.NOSAT, cudaRoundMode.cudaRoundPosInf
            )
    
        exponent = np.array([126], dtype=np.uint8)  # 0.5
        result = np.zeros(2, dtype=np.uint8)
>       kernel[1, 1](result, exponent)

/opt/hostedtoolcache/Python/3.12.13/x64/lib/python3.12/site-packages/numba_cuda/numba/cuda/tests/cudapy/test_fp8_bindings.py:557: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/hostedtoolcache/Python/3.12.13/x64/lib/python3.12/site-packages/numba_cuda/numba/cuda/dispatcher.py:719: in __call__
[...]

PR #834, opened 2026-03-11
fix(ci): cudaRoundMode typing failure in FP8 test

kaeun97:
Attempt to fix this [issue](https://github.com/NVIDIA/numba-cuda/issues/833).

cuda-bindings `13.2.0` changed `cudaRoundMode` from a standard Python `IntEnum` to a `FastEnumMetaclass` type. Numba's type inference cannot resolve FastEnumMetaclass types, causing three FP8 tests to fail ([ref](https://github.com/NVIDIA/cuda-python/pull/1581)). 

This PR replaces adds local IntEnum in `cuda_fp8.py`, matching the pattern already used for `saturation_t` and `fp8_interpretation_t.

If this works, let file a issue on the Numbast side (if needed).

### DIFF
--- a/numba_cuda/numba/cuda/_internal/cuda_fp8.py
+++ b/numba_cuda/numba/cuda/_internal/cuda_fp8.py
@@ -20,7 +20,6 @@ from numba.cuda.extending import make_attribute_wrapper
 from numba.cuda.types import bool_
 from numba.cuda.types import uint64
 from llvmlite import ir
-from cuda.bindings.runtime import cudaRoundMode
 from numba.cuda.typing.templates import Registry as TypingRegistry
 from numba.cuda.types import uint32
 from numba.cuda.types import float32
@@ -228,6 +227,13 @@ class saturation_t(IntEnum):
 class fp8_interpretation_t(IntEnum):
     E4M3 = 0
     E5M2 = 1
+
+
+class cudaRoundMode(IntEnum):
+    cudaRoundNearest = 0
+    cudaRoundZero = 1
+    cudaRoundPosInf = 2
+    cudaRoundMinInf = 3
 
 
 # Structs:
@@ -3931,6 +3937,6 @@ _FUNCTION_SYMBOLS = [
 ]
 
 
-_ENUM_SYMBOLS = ["saturation_t", "fp8_interpretation_t"]
+_ENUM_SYMBOLS = ["saturation_t", "fp8_interpretation_t", "cudaRoundMode"]
 
 __all__ = _NBTYPE_SYMBOLS + _RECORD_SYMBOLS + _FUNCTION_SYMBOLS + _ENUM_SYMBOLS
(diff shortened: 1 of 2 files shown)


kaeun97:
Thanks in advance - unable to reproduce the error locally so would have to use CI to test this. @gmarkall

numba_cuda/numba/cuda/_internal/cuda_fp8.py
```diff
@@ -230,6 +229,13 @@ class fp8_interpretation_t(IntEnum):
     E5M2 = 1
 
 
+class cudaRoundMode(IntEnum):
```
leofang, line 232: @gmarkall @isVoid what does it take for numba-cuda to understand `cudaRoundMode` from cuda-bindings as a legal enum, without numba-cuda having to repeat the definitions?

On the cuda-bindings side, we switched to a custom fast enum from the builtin `IntEnum` to reduce Python overhead (cc @mdboom), and we can evaluate if a patch on cuda-bindings makes better sense, or if we could just register the type in numba-cuda (or both).

leofang: if it only needs a simple patch to cuda-bindings (and @mdboom deems it's acceptable), it'd be better to fix it on the cuda-bindings side and publish 13.2.1/12.9.7, instead of breaking numba-cuda.

cc @rparolin for vis

gmarkall: We'd need to add typing to Numba for the `FastEnumMetaclass` type for Numba to recognise it.

leofang: Would it help if we add `__int__` to the fast enum class?

gmarkall: I don't think Numba-CUDA will see that. Apart from a small set of cases:

https://github.com/NVIDIA/numba-cuda/blob/893dde7ed75ade4ef44b42ed0163e53982fbcf49/numba_cuda/numba/cuda/typing/typeof.py#L50-L97

Numba-CUDA looks at the Python type to map to the Numba type.

Maybe we could / should add recognition of `__int__` as well, but it would need a Numba-CUDA change in addition to adding `__int__` to the fast enum class.

leofang: Ah, it's done by the same function I just touched recently... Got it. So this is a corner case where Python types do matter and duck-typing does not work.

I would suggest that we merge this PR as a workaround to unblock the CI, and discuss a long-term fix. (My 2c is if the surface area grows to other enums we should register the fast enum type, but one-off patches like this are not bad.) WDYT?

leofang:
/ok to test [cfc3787](https://github.com/NVIDIA/numba-cuda/pull/834/commits/cfc3787ec09297f389e673f800f3d83a6395e954)

numba_cuda/numba/cuda/_internal/cuda_fp8.py
isVoid: Sorry I got around to this a bit late. I saw this error yesterday in my Numbast CI but didn't trace it to the bottom. The `cuda_fp8.py` file was auto generated by Numbast. If we decided on a fast enum type for cuda types we should probably inform Numbast to generate corresponding logics.

PR #836, opened 2026-03-12
Support cuda.core.GraphBuilder as a kernel-launch stream

### LINKED ISSUE #832: [FEA] Support using `cuda.core.GraphBuilder` as a kernel-launch stream
**Is your feature request related to a problem? Please describe.**
We need to support building a CUDA graph that includes launching numba-cuda kernels. 

**Describe the solution you'd like**
The proper way (as established in cuda.core) is to pass `GraphBuilder` to wherever a `Stream` is accepted. Under the hood, we let the stream capture mechanism to kick in.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

Andy-Jost:
Closes #832.

## Summary

- Update `_to_core_stream()` in `driver.py` to accept `cuda.core.GraphBuilder`, extracting its internal capture stream so that numba-cuda kernel launches can participate in CUDA graph construction.
- This allows `kernel[grid, block, graph_builder](args)` to work without the user needing to manually unwrap the builder's stream.

## Changes

- Import `GraphBuilder` from `cuda.core` in `driver.py`
- Add `isinstance(stream, GraphBuilder)` branch to `_to_core_stream()`, returning `stream.stream`
- Add test `test_cuda_core_graph_builder_launch_user_facing` that captures a numba kernel launch into a graph via `GraphBuilder` and verifies execution

## Open questions

1. Should `_stream_handle()` (used by event and memcpy code paths) also be updated to support `GraphBuilder`?
2. `CUDADispatcher.configure` is decorated with `@functools.lru_cache` -- is caching `GraphBuilder` as a key a concern?
3. Can we use `cuda.core.GraphBuilder` directly (as in this change), or does it need a numba wrapper like `ExperimentalStream`?

## Test plan

- [x] `pytest -xvs -k graph_builder` passes
- [x] Existing stream tests (`test_cuda_driver_stream_operations`, `test_cuda_core_stream_operations`, `test_cuda_core_stream_launch_user_facing`) still pass
- [x] `pre-commit run --all-files` passes

### DIFF
--- a/numba_cuda/numba/cuda/tests/cudadrv/test_cuda_driver.py
+++ b/numba_cuda/numba/cuda/tests/cudadrv/test_cuda_driver.py
@@ -113,8 +113,7 @@ class TestCudaDriver(CUDATestCase):
         launch(exp_stream, config, function.kernel, ptr)
 
         device_to_host(array, memory, sizeof(array))
-        for i, v in enumerate(array):
-            self.assertEqual(i, v)
+        np.testing.assert_equal(array, np.arange(len(array)))
 
         module.unload()
 
@@ -144,9 +143,7 @@ class TestCudaDriver(CUDATestCase):
             launch(_to_core_stream(stream), config, function.kernel, ptr)
 
         device_to_host(array, memory, sizeof(array), stream=stream)
-
-        for i, v in enumerate(array):
-            self.assertEqual(i, v)
+        np.testing.assert_equal(array, np.arange(len(array)))
 
     def test_cuda_core_stream_operations(self):
         module = self.context.create_module_ptx(self.ptx)
@@ -178,8 +175,7 @@ class TestCudaDriver(CUDATestCase):
             launch(stream, config, function.kernel, ptr)
 
             device_to_host(array, memory, sizeof(array), stream=stream)
-        for i, v in enumerate(array):
-            self.assertEqual(i, v)
+        np.testing.assert_equal(array, np.arange(len(array)))
 
     def test_cuda_core_stream_launch_user_facing(self):
         @cuda.jit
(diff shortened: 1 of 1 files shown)


numba_cuda/numba/cuda/tests/cudadrv/test_cuda_driver.py
```diff
@@ -202,6 +202,33 @@ def kernel(a):
...
+        graph_builder = dev.create_graph_builder()
+        graph_builder.begin_building()
+        kernel[1, 100, graph_builder](ary)
+        graph = graph_builder.end_building().complete()
+
+        # Execute the captured graph
+        graph.launch(stream)
+        stream.sync()
+
+        result = ary.copy_to_host(stream=stream)
```
gmarkall: If this copy is on a stream, is it possibly asynchronous with respect to host execution, and the assertion below could be checked before the data is copied back to the host? Do we need to sync on the stream before checking the host data?

numba_cuda/numba/cuda/tests/cudadrv/test_cuda_driver.py
```diff
@@ -202,6 +202,33 @@ def kernel(a):
...
+        for i, v in enumerate(result):
+            self.assertEqual(i, v)
```
gmarkall: Generally for numerical tests I prefer to use `np.testing.assert_*`, because it prints out the array values that differ. I know in this case the assert would also print out the first differing values, but I think the following is a little more idiomatic:
```suggestion
        np.testing.assert_equal(result, np.arange(len(result))
```

numba_cuda/numba/cuda/tests/cudadrv/test_cuda_driver.py
leofang: Yes

numba_cuda/numba/cuda/tests/cudadrv/test_cuda_driver.py
Andy-Jost: Good call. Latest commit fixes 5 instances.

numba_cuda/numba/cuda/tests/cudadrv/test_cuda_driver.py
Andy-Jost: Good catch. I also fixed the test I copied this from. Same problem.

leofang:
Thanks!

PR #844, opened 2026-04-02
Fix CABI calling convention: skip env global and escape mangled names

isVoid:
## Summary

- **Skip emitting the `NumbaEnv` global for CABI functions.** The `CUDACABICallConv` has no error-status channel and never references the environment, yet the lowering pass unconditionally emitted it. A new `needs_env` class attribute on `BaseCallConv` (default `True`, overridden to `False` in `CUDACABICallConv`) now guards `emit_environment_object()`.
- **Fix `CUDACABICallConv.mangler` to escape special characters.** The mangler used a naive `name.split(".")[-1]` which passed raw `<lambda>` (with angle brackets) into LLVM global names, causing `ERROR_INVALID_IR` from NVVM. It now uses `itanium_mangler.escape_string()` to hex-escape invalid characters (e.g. `<` → `_3c`).
- **Promote `_escape_string` to public API** as `escape_string` with a backward-compatible alias.

## Test plan

- [x] New `TestEscapeString` tests in `test_itanium_mangler.py` (6 cases: plain ASCII, digits/underscores, angle brackets, dot, empty string, backward-compat alias)
- [x] New `test_c_abi_lambda_with_abi_name` in `test_compiler.py` -- compiles a lambda with `abi_info={"abi_name": "blah"}` via `compile_ptx` and `compile_all`, verifies PTX contains the correct function name
- [x] New `test_c_abi_no_env_global` -- verifies CABI PTX does not contain `NumbaEnv`
- [x] New `test_numba_abi_has_env_global` -- verifies Numba-ABI PTX still contains `NumbaEnv`
- [x] All existing `c_abi` tests pass without regression

### DIFF
--- a/numba_cuda/numba/cuda/itanium_mangler.py
+++ b/numba_cuda/numba/cuda/itanium_mangler.py
@@ -62,25 +62,32 @@ N2CODE = {
 }
 
 
-def _escape_string(text):
+def escape_string(text):
     """Escape the given string so that it only contains ASCII characters
-    of [a-zA-Z0-9_$].
+    valid in mangled names: ``[a-zA-Z0-9_]``.
 
-    The dollar symbol ($) and other invalid characters are escaped into
-    the string sequence of "$xx" where "xx" is the hex codepoint of the char.
+    Invalid characters are hex-escaped as ``_xx`` where *xx* is the two-digit
+    hex code point.  Multibyte (non-ASCII) characters are first encoded to
+    UTF-8 and each byte is escaped individually.
 
-    Multibyte characters are encoded into utf8 and converted into the above
-    hex format.
+    Examples::
+
+        >>> escape_string("hello")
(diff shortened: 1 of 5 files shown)


numba_cuda/numba/cuda/itanium_mangler.py
```diff
@@ -62,27 +62,34 @@
...
         return "".join(("_%02x" % ch) for ch in m.group(0).encode("utf8"))
 
     ret = re.sub(_re_invalid_char, repl, text)
-    # Return str if we got a unicode (for py2)
     if not isinstance(ret, str):
         return ret.encode("ascii")
     return ret
 
 
+_escape_string = escape_string
```
rparolin, line 90: Nit: Maybe leave comment saying this is for backwards compatibility.

rparolin (approved):

PR #847, opened 2026-04-03
Map types.NoneType return type to LLVM void instead of i8*

### LINKED ISSUE #845: [BUG] `types.void` is translated to `i8*` instead of LLVM `void` type, causing ABI and LTO mismatches
### Description

When using Numba's `types.void` (e.g., as a return type for a device function or in signatures), Numba's datamodel translates it to an LLVM `i8*` (opaque pointer `ptr` in LLVM 15+) instead of the actual LLVM `void` type (`ir.VoidType()`). 

**Note:** This issue originates in the core `numba` package itself (specifically in `numba.core.datamodel` where `types.void` is registered to `OpaqueModel`), but it has a severe impact on `numba-cuda` when linking device code.

This causes the generated LLVM IR to treat `void` as a pointer type. In the Numba calling convention, this results in device functions taking an `i8**` return pointer argument rather than simply returning `void`. While this might accidentally work at runtime on some architectures, it breaks strict ABI requirements. More importantly, it causes fatal errors during Link Time Optimization (LTO) when linking Numba-generated IR with C/C++ device code that correctly declares the function as returning `void`.

### Reproducer

```python
import numba
from numba import types, cuda
from numba.core import datamodel
from numba.cuda.descriptor import cuda_target

# 1. Numba's default void type is translated to i8* via OpaqueModel
void_type = types.void
default_model = datamodel.default_manager.lookup(void_type)
print(f"LLVM type for void: {default_model.get_value_type()}")
# Output: LLVM type for void: i8*

# 2. CUDA Context return types also reflect this mismatch
ctx = cuda_target.target_context
print(f"Context return type: {ctx.get_return_type(types.void)}")
# Output: Context return type: i8*

# 3. Demonstrate this in a compiled CUDA device function
@cuda.jit(device=True)
def foo():
    pass

@cuda.jit
def kernel():
    foo()

kernel[1, 1]()

# The device function signature takes an `i8**` return pointer instead of returning `void`
for k, v in foo.inspect_llvm().items():
    for line in v.splitlines():
        if line.startswith("define") and "foo" in line:
            print(line)
# Output: define linkonce_odr i32 @"_ZN8__main__3foo..."(i8** %".ret")
```

### Expected Behavior

`types.void` should be mapped to a model that returns `ir.VoidType()` (i.e., `void` in LLVM IR), so that functions returning `void` have the correct LLVM IR signature without an unnecessary return pointer argument:

```llvm
define void @"_ZN8__main__3foo..."()
```

And context return types should reflect `void` instead of `i8*`.

### Actual Behavior

`types.void` is mapped to `OpaqueModel`, which returns `ir.IntType(8).as_pointer()` (`ptr` / `i8*`).

Because of this, the calling convention treats the return type as a pointer, generating an `i8**` return argument:
```llvm
define linkonce_odr i32 @"_ZN8__main__3foo..."(i8** %".ret")
```

### Impact

When exporting Numba JIT code to LLVM IR and linking it via LTO with external C/C++ device libraries, the LLVM linker detects a signature mismatch between the C++ declaration (`void foo()`) and the Numba definition (`i32 foo(i8**)`). This results in LTO linking failures and ABI correctness issues for CUDA device code.

isVoid:
## Summary

Fixes #845: `types.void` / `types.NoneType` was mapped to `OpaqueModel`, which used `i8*` for all LLVM type queries including the return type. This caused C-ABI device functions returning `void` to emit `i8* foo(...)` instead of `void foo(...)`, breaking ABI compatibility and LTO linking with external C/C++ device code.

**NoneTypeModel** (`numba_cuda/numba/cuda/models.py`):

- Registered in `cuda_data_manager`, which shadows the upstream `OpaqueModel` for `types.NoneType` in `default_manager` via `ChainMap` priority.
- `get_return_type()` returns `ir.VoidType()` — used when building `ir.FunctionType` for function signatures.
- `get_value_type()` returns `i8*` (opaque pointer) — LLVM void cannot represent values, so `None` still needs a concrete type for variable assignment, alloca, store/load, and boxing/unboxing.

**Calling convention updates** (`numba_cuda/numba/cuda/core/callconv.py`):

- Numba ABI (`CUDACallConv.get_return_type`): uses a dummy `i8**` return slot when the data model return type is void, preserving the existing `i32 func(retslot*, args...)` signature.
- C ABI (`CUDACABICallConv.return_value`): emits `ret_void` when the function return type is void.
- C ABI (`CUDACABICallConv.call_function`): synthesizes a null `i8*` value for void call results instead of using the void instruction as a value.

**Cleanup** (`numba_cuda/numba/cuda/datamodel/cuda_models.py`):

- Removed the `types.NoneType` registration from `OpaqueModel` so there is no stale fallback in `default_manager` for the standalone path.

## Test plan

- [x] `test_nonetype_model_return_type_is_void` — data model `get_return_type()` returns `VoidType`
- [x] `test_nonetype_model_value_type_is_opaque_ptr` — `get_value_type()` returns `PointerType`
- [x] `test_cabi_void_device_function_signature` — C-ABI void function links and runs correctly
- [x] `test_void_device_function_numba_abi` — Numba-ABI void device function compiles and runs
- [x] `test_void_device_function_with_side_effect` — void device function with array mutation
- [x] 160 existing tests across device_func, compiler, inspect, extending, overload, dispatcher pass with no regressions

### DIFF
--- a/numba_cuda/numba/cuda/core/callconv.py
+++ b/numba_cuda/numba/cuda/core/callconv.py
@@ -104,6 +104,8 @@ class BaseCallConv:
         Get the actual type of the return argument for Numba type *ty*.
         """
         restype = self.context.data_model_manager[ty].get_return_type()
+        if isinstance(restype, ir.VoidType):
+            return ir.IntType(8).as_pointer().as_pointer()
         return restype.as_pointer()
 
     def init_call_helper(self, builder):
@@ -414,6 +416,10 @@ class CUDACABICallConv(BaseCallConv):
 
     def return_value(self, builder, retval):
         expected_type = builder.function.ftype.return_type
+
+        if isinstance(expected_type, ir.VoidType):
+            return builder.ret_void()
+
         actual_type = retval.type
 
         # If types don't match, we need to cast
@@ -473,7 +479,11 @@ class CUDACABICallConv(BaseCallConv):
         # No status required as we don't support exceptions or a distinct None
         # value in a C ABI.
         status = None
-        out = self.context.get_returned_value(builder, resty, code)
+        ret_model = self.context.data_model_manager[resty]
+        if isinstance(ret_model.get_return_type(), ir.VoidType):
+            out = cgutils.get_null_value(ret_model.get_value_type())
+        else:
+            out = self.context.get_returned_value(builder, resty, code)
         return status, out
 
     def call_internal(self, builder, fndesc, sig, args):
--- a/numba_cuda/numba/cuda/tests/cudapy/test_device_func.py
+++ b/numba_cuda/numba/cuda/tests/cudapy/test_device_func.py
@@ -630,5 +630,83 @@ class TestDeclareDeviceCABI(CUDATestCase):
         np.testing.assert_equal(x, 42)
 
 
+@skip_on_cudasim("Data model inspection unsupported in the simulator")
+class TestNoneTypeModel(CUDATestCase):
+    """Tests for the NoneTypeModel data model (issue #845).
+
+    Verifies that types.void / types.NoneType uses ir.VoidType() as the
+    LLVM return type, fixing ABI mismatches when linking with external
+    C/C++ device code via LTO.
+    """
+
+    def test_nonetype_model_return_type_is_void(self):
+        from llvmlite import ir
+        from numba.cuda.descriptor import cuda_target
+
+        dm = cuda_target.target_context.data_model_manager
+        model = dm.lookup(types.void)
+
+        self.assertEqual(type(model).__name__, "NoneTypeModel")
+        self.assertIsInstance(model.get_return_type(), ir.VoidType)
+
+    def test_nonetype_model_value_type_is_opaque_ptr(self):
+        from llvmlite import ir
+        from numba.cuda.descriptor import cuda_target
+
+        dm = cuda_target.target_context.data_model_manager
+        model = dm.lookup(types.void)
+        vt = model.get_value_type()
+
+        self.assertIsInstance(vt, ir.PointerType)
+
+    def test_cabi_void_device_function_signature(self):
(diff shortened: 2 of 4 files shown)


numba_cuda/numba/cuda/core/callconv.py
```diff
@@ -104,6 +104,8 @@ def get_return_type(self, ty):
         Get the actual type of the return argument for Numba type *ty*.
         """
         restype = self.context.data_model_manager[ty].get_return_type()
+        if isinstance(restype, ir.VoidType):
+            return ir.IntType(8).as_pointer().as_pointer()
```
isVoid, line 108: We need to return a placeholder value type because void type cannot be used in declaration for argument type. This also aligns with how the rest of Numba internally passes arguments around.

numba_cuda/numba/cuda/tests/cudapy/test_device_func.py
```diff
@@ -630,5 +630,83 @@ def kernel(x):
...
+
+        dm = cuda_target.target_context.data_model_manager
+        model = dm.lookup(types.void)
+        vt = model.get_value_type()
+
+        self.assertIsInstance(vt, ir.PointerType)
+
+    def test_cabi_void_device_function_signature(self):
+        consume = cuda.declare_device(
+            "consume", "void(int32)", link=consume_cabi_cu, abi="c"
```
ZzEeKkAa, line 664: This by itself does not guarantee that it was inlined. In fact that code used to work properly before. Could we add tests that verify ir signature both for callee and caller

gmarkall (changes requested):
It's not clear to me that this is a fix that should be made - see https://github.com/NVIDIA/numba-cuda/issues/845#issuecomment-4198326507

PR #853, opened 2026-04-09
Support interop with active green contexts

caugonnet:
Teach the CUDA runtime to borrow externally activated green-context-derived execution contexts while continuing to reject ordinary non-primary contexts, and scope loaded kernels to the active execution context so resets do not reuse stale handles. Protect destructive reset/close paths from touching borrowed contexts, add CUDA 13 interop coverage, and document the supported phase-1 behavior.


numba_cuda/numba/cuda/cuda_paths.py
```diff
@@ -82,6 +85,20 @@ def _find_valid_path(options):
...
+    if locator is None:
+        locator = getattr(pathfinder, "find_nvidia_header_directory", None)
+    if locator is None:
+        return None
+    located = locator(name)
+    if located is None:
+        return None
+    if hasattr(located, "abs_path") and hasattr(located, "found_via"):
+        return located
+    return _header_dir_info("cuda.pathfinder", located)
```
leofang, line 99: Q: Did you do this because you did not want to raise the lower bound of cuda-pathfinder (to the version that offers these APIs?)

leofang:
Hi @caugonnet, update: We're adding the green context support to cuda-core (https://github.com/NVIDIA/cuda-python/pull/1976). Once merged, I expect that you can just create a "green" stream there, and pass it to a numba-cuda kernel when launching it. There should be no need to make any change in numba-cuda (we have been busy moving CUDA runtime-specific work to cuda-core, and keep numba-cuda slim to only handle JIT compilation). Please give it a shot and let us know how it goes! 🙂

PR #855, opened 2026-04-13
Fix CUDA context leak in `with cuda.gpus[N]:` context manager

isVoid:
## Summary

- **Fix**: `Device.get_primary_context()` calls `self._dev.set_current()` on first invocation, which leaves the primary context active on the thread. The caller `_activate_context_for` then calls `push()`, saving that already-active context onto the stack. On exit, `pop()` restores it — so the context remains current after the `with` block (a leak). The fix pops the context left by `set_current()` immediately after obtaining the handle, so `get_primary_context()` upholds its documented contract: *"Note: it is not pushed to the CPU thread."*
- **Tests**: Adds two regression tests in `test_context_stack.py` — one verifying no context remains after `with cuda.gpus[0]: pass` on a clean stack, and another verifying the previous context is properly restored.

## Test plan

- [x] `test.py` repro script now prints `None` (previously printed a live context handle)
- [x] `test_context_stack.py`: all 11 tests pass (9 existing + 2 new)
- [x] `test_cuda_driver.py`: all 18 tests pass
- [x] `test_cuda_memory.py`: all 15 tests pass
- [x] `test_reset_device.py`: 1 test passes
- [x] `test_module_callbacks.py`: all 8 tests pass
- [x] Full cudadrv + cudapy suite (134 modules, 1819 tests): all pass, 0 failures

### DIFF
--- a/numba_cuda/numba/cuda/tests/cudadrv/test_context_stack.py
+++ b/numba_cuda/numba/cuda/tests/cudadrv/test_context_stack.py
@@ -78,6 +78,51 @@ class TestContextAPI(CUDATestCase):
 
 
 @skip_on_cudasim("CUDA HW required")
+class TestContextLeak(CUDATestCase):
+    """Regression tests for context leaks from the gpu context manager."""
+
+    def test_gpus_context_manager_does_not_leak(self):
+        # Regression test: ``with cuda.gpus[N]`` must not leave a CUDA
+        # context on the thread after the block exits.
+        the_driver = driver.driver
+
+        # Drain any pre-existing contexts from the stack.
+        while the_driver.pop_active_context() is not None:
+            pass
+
+        with cuda.gpus[0]:
+            pass
+
+        # After exiting the context manager the current context must be null.
+        with the_driver.get_active_context() as ac:
+            self.assertIsNone(
+                ac.context_handle,
+                "CUDA context leaked after exiting cuda.gpus context manager",
+            )
+
+    def test_gpus_context_manager_restores_previous_context(self):
+        # If a context is already active before entering the context manager,
+        # it must be restored on exit.
+        the_driver = driver.driver
+
+        # Ensure device-0 context exists and is pushed.
+        outer_ctx = cuda.current_context()
+        outer_handle = int(outer_ctx.handle)
+
+        with cuda.gpus[0]:
+            pass
+
+        with the_driver.get_active_context() as ac:
+            self.assertIsNotNone(ac.context_handle)
+            self.assertEqual(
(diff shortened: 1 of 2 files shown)


numba_cuda/numba/cuda/tests/cudadrv/test_context_stack.py
```diff
@@ -77,6 +77,51 @@ def switch_gpu():
         self.assertEqual(int(devid), 1)
 
 
+@skip_on_cudasim("CUDA HW required")
+class TestContextLeak(CUDATestCase):
```
gmarkall, line 81: These tests seem to pass for me on `main`. Do they pass on `main` when you run locally?

gmarkall (changes requested):
I'm finding that on `main` in a multi-GPU system, `cudapy/test_multigpu.py::TestMultiGPUContext::test_multigpu_context` fails on `main` because of an `AssertionError` when popping the context:

```
  File "/home/gmarkall/miniforge3/envs/numbadev/lib/python3.11/unittest/case.py", line 57, in testPartExecutor
    yield
  File "/home/gmarkall/miniforge3/envs/numbadev/lib/python3.11/unittest/case.py", line 623, in run
    self._callTestMethod(testMethod)
  File "/home/gmarkall/miniforge3/envs/numbadev/lib/python3.11/unittest/case.py", line 579, in _callTestMethod
    if method() is not None:
       ^^^^^^^^
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/tests/cudapy/test_multigpu.py", line 35, in test_multigpu_context
    with cuda.gpus[0]:
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/cudadrv/devices.py", line 85, in __exit__
    self.get_primary_context().pop()
  File "/home/gmarkall/numbadev/numba-cuda/numba_cuda/numba/cuda/cudadrv/driver.py", line 1160, in pop
    assert int(popped) == int(self.handle)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

With this PR, the test returns to passing again. So I think the fix helps here, but the `TestContextLeak` test class appears to be adding nothing - it passes on `main` for me on both single- and multi-GPU systems.

Do you see the added tests here serving a purpose that I've missed?

gmarkall (approved):
Actually I have had a re-think about the tests - although they seem to work on `main`, they do encode the kind of behaviour we'd like to preserve in a way that may not be explicitly represented anywhere else, so it makes sense to keep them. It looks like CI didn't run so I'll just make sure all is well with it before merging.

PR #857, opened 2026-04-14
Fix flaky test_max_pending_count by preventing premature GC

### LINKED ISSUE #856: Test: `TestDeallocation.test_max_pending_count` is flaky
https://github.com/NVIDIA/numba-cuda/actions/runs/24417845685/job/71332295933#step:15:4517
```
=================================== FAILURES ===================================
___________________ TestDeallocation.test_max_pending_count ____________________
[gw11] linux -- Python 3.13.13 /opt/hostedtoolcache/Python/3.13.13/x64/bin/python

self = <numba.cuda.tests.cudadrv.test_deallocations.TestDeallocation testMethod=test_max_pending_count>

    @skip_if_external_memmgr("Deallocation specific to Numba memory management")
    def test_max_pending_count(self):
        # get deallocation manager and flush it
        deallocs = cuda.current_context().memory_manager.deallocations
        deallocs.clear()
        self.assertEqual(len(deallocs), 0)
        # deallocate to maximum count
        for i in range(config.CUDA_DEALLOCS_COUNT):
            cuda.to_device(np.arange(1))
>           self.assertEqual(len(deallocs), i + 1)
E           AssertionError: 10 != 9

/opt/hostedtoolcache/Python/3.13.13/x64/lib/python3.13/site-packages/numba_cuda/numba/cuda/tests/cudadrv/test_deallocations.py:30: AssertionError
```

xyaz1313:
## Problem

TestDeallocation.test_max_pending_count is flaky — intermittently fails with .

## Root Cause

The test creates DeviceNDArray objects without storing references:

```python
for i in range(config.CUDA_DEALLOCS_COUNT):
    cuda.to_device(np.arange(1))  # no reference stored
    self.assertEqual(len(deallocs), i + 1)
```

Each `cuda.to_device()` returns a DeviceNDArray → OwnedPointer → weakref.finalizer. When Python's GC non-deterministically collects these temporaries between loop iterations, their finalizers fire early and add deallocations to the pending queue. This makes `len(deallocs)` exceed the expected `i + 1`.

## Fix

1. Store all device arrays in a list during creation
2. Delete them one by one in a controlled loop
3. Call `gc.collect()` after each deletion to ensure finalizers fire

This makes deallocation timing fully deterministic.

## Reproduction

The flakiness depends on GC timing — it happens more frequently with high memory pressure or when running the full test suite (where previous tests accumulate GC-tracked objects).

Fixes #856

### DIFF
--- a/numba_cuda/numba/cuda/tests/cudadrv/test_deallocations.py
+++ b/numba_cuda/numba/cuda/tests/cudadrv/test_deallocations.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: BSD-2-Clause
 
 from contextlib import contextmanager
+import gc
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


gmarkall:
I don't understand the root cause analysis here. My understanding is that by storing no reference to the `DeviceNDArray` returned by `cuda.to_device()`, its refcount would immediately drop to 0 and cause the finalizer to add its pointer to the list of deallocations. I would not have expected the GC to be involved here, because my understanding is that it's for cleaning up lost objects that participate in reference cycles.

My concern is that the real underlying issue could be that with all the changes we've made to leverage cuda-core / cuda-python more, we've accidentally introduced reference cycles that keep device memory alive even when nothing "downstream" of the call to `cuda.to_device()` is holding a reference to it anymore. If that were the case, then these test changes will make the test no longer flaky, but completely mask the underlying issue.

Perhaps I have some misunderstanding about the Python GC or the deallocation behaviour is supposed to work - if so, could you help me understand how things fit together please?

PR #863, opened 2026-04-20
Flip to stable rapids release in thirdparty tests

brandon-b-miller:
This PR flips us to the latest stable release of rapids in our thirdparty tests. Previously we were on a nightly after an issue surfaced that required a cudf side fix that we needed to pull here, this fix is now in a stable version.

### DIFF
--- a/ci/test_thirdparty_cudf.sh
+++ b/ci/test_thirdparty_cudf.sh
@@ -9,8 +9,7 @@ CUDA_VER_MAJOR=${CUDA_VER%.*.*}
 
 rapids-logger "Install cuDF Wheel"
 
-pip install \
-    "cudf-cu12==26.4.*"
+pip install cudf-cu12
 
 
 rapids-logger "Remove Extraneous numba-cuda"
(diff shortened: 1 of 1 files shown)


gmarkall (approved):

ci/test_thirdparty_cudf.sh
```diff
@@ -10,8 +10,7 @@ CUDA_VER_MAJOR=${CUDA_VER%.*.*}
 rapids-logger "Install cuDF Wheel"
 
 pip install \
-    --extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple \
-    "cudf-cu12>=26.4.0a0,<=26.4" "dask-cuda>=26.4.0a0,<=26.4"
+    "cudf-cu12==26.4.*"
```
leofang: Q: Should we pin, or should we let it float forward?

brandon-b-miller: I've pinned so far but I see no reason we couldnt generalize to "latest stable" unless a nightly is needed for some specific reason 👍

PR #871, opened 2026-04-28
Bump rapidsai/sccache tag to v0.14.0-rapids.9

leofang:
## Summary
- Bump pinned `SCCACHE_TAG` from `v0.14.0-rapids.3` to `v0.14.0-rapids.9` in `build-wheel.yml`

-- Leo's bot

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #869, opened 2026-04-23
Remove Conflicting Lowering for FP8->Int8 Conversion

isVoid:
FP8 bindings lower casts are generated twice by Numbast. One time for `operator char()` and another time for `operator singed char()`. These C APIs are defined unambiguously for x86 and arm machines. But when registering both in the lowering registry, `operator char()` shadows the lowering for `signed char`. Therefore, in arm machines, when casting fp8 to int8, `operator char()` is always resolved to. Therefore on arm machine negative fp8 values are always truncated.

Closes #866

PR #867, opened 2026-04-23
Fix Windows path parsing in generate_raw_ltoir.py

rluo8:
This is to fix path parsing issue in generate_raw_ltoir.py on Windows platform. 

Build failed:
make -j%NUMBER_OF_PROCESSORS% "NVCC_FLAGS=-O3 -rdc true -std=c++17 -Xcompiler=/Zc:preprocessor"
...
Using CUDA include flags: ['-IC:\\Program', 'Files\\NVIDIA', 'GPU', 'Computing', 'Toolkit\\CUDA\\v13.2\\bin/../include', '-IC:\\Program', 'Files\\NVIDIA', 'GPU', 'Computing', 'Toolkit\\CUDA\\v13.2\\bin/../include/cccl', '-IC:\\Program\\cccl', 'Files\\NVIDIA\\cccl', 'GPU\\cccl', 'Computing\\cccl', 'Toolkit\\CUDA\\v13.2\\bin/../include\\cccl', '-IC:\\Program\\cccl', 'Files\\NVIDIA\\cccl', 'GPU\\cccl', 'Computing\\cccl', 'Toolkit\\CUDA\\v13.2\\bin/../include/cccl\\cccl']
NVRTC error, code 5: NVRTC_ERROR_INVALID_OPTION
make: *** [Makefile:84: test_device_functions.ltoir] Error 1
make: *** Waiting for unfinished jobs....
make: *** [Makefile:117: nrt_extern.ltoir] Error 1
...

On Windows, CUDA typically lives under C:\Program Files\..., so the INCLUDES= line  include flags with spaces inside double quotes, e.g. -I"C:\Program Files\NVIDIA GPU Computing Toolkit\...\include". The str.split() splits on whitespace and doesn't know about quotes, so it splits the path in half. shlex.split could treat quoted substrings as a single token, so paths with spaces could be split correctly on Windows.

After the fix, build on Windows could pass.

PR #864, opened 2026-04-21
Alternate half of linux arm64 CI runners from a100 to l4

leofang:
Change every other arm64 entry in the pull-request test matrix to use l4 GPU instead of a100, giving a ~50/50 split in an alternating pattern (8 a100, 7 l4 active; plus 2 commented-out 3.14t entries updated to match). This reduces pressure on a100 runners while maintaining arm64 coverage across all Python and CUDA version combinations.

Companion to NVIDIA/cuda-python#1956.

### Changes

- 7 of 15 active linux arm64 entries changed from `a100` to `l4` (every even-numbered row)
- 2 of 3 commented-out 3.14t arm64 entries also updated (continuing the alternation)
- Updated copyright year (2025 → 2026) and header comment to reflect mixed GPU usage

### Resulting arm64 matrix

| # | PY_VER | CUDA_VER | GPU | Active? |
|---|--------|----------|-----|---------|
| 1 | 3.10 | 12.9.1 | a100 | yes |
| 2 | 3.10 | 13.0.2 | **l4** | yes |
| 3 | 3.10 | 13.2.1 | a100 | yes |
| 4 | 3.11 | 12.9.1 | **l4** | yes |
| 5 | 3.11 | 13.0.2 | a100 | yes |
| 6 | 3.11 | 13.2.1 | **l4** | yes |
| 7 | 3.12 | 12.9.1 | a100 | yes |
| 8 | 3.12 | 13.0.2 | **l4** | yes |
| 9 | 3.12 | 13.2.1 | a100 | yes |
| 10 | 3.13 | 12.9.1 | **l4** | yes |
| 11 | 3.13 | 13.0.2 | a100 | yes |
| 12 | 3.13 | 13.2.1 | **l4** | yes |
| 13 | 3.14 | 12.9.1 | a100 | yes |
| 14 | 3.14 | 13.0.2 | **l4** | yes |
| 15 | 3.14 | 13.2.1 | a100 | yes |
| 16 | 3.14t | 12.9.1 | **l4** | commented |
| 17 | 3.14t | 13.0.2 | a100 | commented |
| 18 | 3.14t | 13.2.1 | **l4** | commented |

-- Leo's bot

PR #873, opened 2026-04-28
Bump Version to 0.30.1

isVoid:
## Changes since v0.30.0

- Fix indentation bug in test_linker.py (#875)
- Bump rapidsai/sccache tag to v0.14.0-rapids.9 (#871)
- Fix Windows path parsing in generate_raw_ltoir.py (#867)
- build(deps): bump the actions-monthly group across 1 directory with 5 updates (#860)
- Alternate half of linux arm64 CI runners from a100 to l4 (#864)
- Flip to stable rapids release in thirdparty tests (#863)
- Remove Conflicting Lowering for FP8->Int8 Conversion (#869)
- Sync CI with cuda-python (#858)
- sync with cuda-python and add 2-gpu runners (#814)
- fix: support np.count_nonzero in device functions (#854)
- Fix CUDA context leak in `with cuda.gpus[N]:` context manager (#855)
- Fix CABI return type mismatch when calling Numba-ABI subroutines (#852)


leofang:
@isVoid can we include #875? Otherwise we need to carry the patch in cuda-python, which is annoying.

leofang:
#871 is ready too

isVoid:
@leofang Sounds good. It seems like there's another flaky test in the CI which I want to also address.

leofang (approved):

PR #875, opened 2026-04-29
Fix indentation bug in test_linker.py

leofang:
`add_from_numba` and `debuggable_kernel` reference `test_device_functions_ltoir` which is only defined inside `if TEST_BIN_DIR:`. When `NUMBA_CUDA_TEST_BIN_DIR` is not set, this causes a `NameError` at module load time.

This PR:
1. Moves `add_from_numba` and `debuggable_kernel` inside the `if TEST_BIN_DIR:` block
2. Adds `@unittest.skipUnless(TEST_BIN_DIR, ...)` to `test_debug_kernel_with_lto` so it is skipped when the env var is unset

-- Leo's bot

### DIFF
--- a/numba_cuda/numba/cuda/tests/cudadrv/test_linker.py
+++ b/numba_cuda/numba/cuda/tests/cudadrv/test_linker.py
@@ -116,17 +116,15 @@ if TEST_BIN_DIR:
         TEST_BIN_DIR, "test_device_functions.ltoir"
     )
 
+    add_from_numba = cuda.declare_device(
+        "add_from_numba",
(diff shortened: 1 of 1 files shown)


leofang:
This patch is migrated from https://github.com/NVIDIA/cuda-python/pull/1987.

leofang:
/ok to test [5a1eeff](https://github.com/NVIDIA/numba-cuda/pull/875/commits/5a1eeff96a5bf0005619852cc2e7903f56f14f24)

isVoid (approved):

PR #882, opened 2026-05-09
Migrate test suite to completely utilize pytest (#619)

acosmicflamingo:
Addresses #619

PR #881, opened 2026-05-08
Fix minor issues in CONTRIBUTING.md

acosmicflamingo:
Fixes #880 

* Modify the CLA hyperlink to point to the right URL

* Update 'pip install' commands to support dependency groups usage in pyproject.toml
   + test-cu12 and test-cu13 were moved from 'optional-dependencies' section to 'dependency-groups' section

PR #883, opened 2026-05-10
Refactor TestCase subclasses to directly inherit from unittest.TestCase

acosmicflamingo:
A cleaner approach to migrating numba-cuda's test suite to pytest is to decouple `TestCase` methods from the class. Doing so will make removing the TestCase inheritance a trivial matter, and thus pave the way to using pytest. Many instances (no pun intended) where `self` is being used is just to access functions (like `assertEqual`) that can simply be replaced with an assert statement. Of course, there are some other functions that do require a bit of care (and I'm regularly running the test suite to make sure I'm not introducing any regressions).

Once this is done, the rest of the migration is pretty much a matter of utilizing pytest API in appropriate places (fingers crossed).


numba_cuda/numba/cuda/tests/support.py
```diff
@@ -178,6 +193,20 @@ def ignore_internal_warnings():
     )
 
 
+def reset_module_warnings(module):
```
gmarkall, line 197: I suspect this is archaic now, but I'm OK with leaving it in on the basis that it's just part of a refactor anyway.

numba_cuda/numba/cuda/tests/support.py
```diff
@@ -237,552 +425,366 @@ def captured_stderr():
...
+        ]
 
-        @unbox(DummyType)
-        def unbox_dummy(typ, obj, c):
-            return NativeValue(c.context.get_dummy_value())
 
-        return Dummy, DummyType
+class TestCase(unittest.TestCase):
+    pass
+    # TODO: safe to delete since it's not used anywhere in numba_cuda?
```
gmarkall: Indeed, these are safe to delete.

numba_cuda/numba/cuda/tests/support.py
```diff
@@ -193,6 +222,50 @@ def override_config(name, value):
         setattr(config, name, old_value)
 
 
+# TODO: how to generate self.id() without access to instance?
```
gmarkall: I suspect that you can use a monotonically-increasing counter for the test IDs instead of the instance ID.

gmarkall:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/883/commits/3eeb4aef7607ee1e6c94b4c9a0a1b5c3f53520e7

gmarkall:
Thanks for the PR! I've had a quick skim, and started the tests to see how it goes.

acosmicflamingo:
@gmarkall no problem! Thanks for the feedback; I'll implement suggestions in a bit.

Whoops, forgot to mark this PR as a draft; would not want you to have to review the PR for the nth time because you thought I had finally pushed that last commit (or maybe you'd prefer it be in its current state and you periodically check to see how it's going?).

By the way, what is the status of upstream numba's cuda-specific test suite, and what's the relationship between this codebase and theirs? I was finding that some functions from support.py marked for deletion reside in upstream numba's codebase. I was curious as to whether some of the work in this PR should eventually make its way upstream, or whether it's not really supported now that numba-cuda is in the picture.

numba_cuda/numba/cuda/tests/support.py
acosmicflamingo: Ah, if that's archaic I could remove it too. This was also under the category of functions I considered removing given their lack of usage; just forgot to add the TODO there.

acosmicflamingo: Oh wait a minute...no this is used in test_ufuncs. My apologies; it will stay :)

numba_cuda/numba/cuda/tests/support.py
acosmicflamingo: Great! Although I accidentally deleted `longMessage = True`, will re-add.

numba_cuda/numba/cuda/tests/support.py
acosmicflamingo: Good thinking! I pushed a solution in the newest commit fyi (although I don't know if you want to incrementally look through it or wait until the end; I can mark this PR as "Draft" to indicate as such).

acosmicflamingo:
/ok to test [65d755b](https://github.com/NVIDIA/numba-cuda/pull/883/commits/65d755bb03f11055fbb5e1d847fba733d7928cf4)

gmarkall:
> would not want you to have to review the PR for the nth time because you thought I had finally pushed that last commit (or maybe you'd prefer it be in its current state and you periodically check to see how it's going?).

I'm quite happy if we just iterate a few times, as long as you're OK with that. I find it hard to do "one perfect review" that outlines perfectly the exact set of suggestions I want to make anyway.

> By the way, what is the status of upstream numba's cuda-specific test suite, and what's the relationship between this codebase and theirs? I was finding that some functions from support.py marked for deletion reside in upstream numba's codebase. I was curious as to whether some of the work in this PR should eventually make its way upstream, or whether it's not really supported now that numba-cuda is in the picture.

`numba.cuda` in upstream is in maintenance mode for backward compatibility - that is, code written to use it should still work with the latest version of `numba` installed, but Numba-CUDA is where all new development has happened since we separated the CUDA target from upstream Numba. So things will stay quite static upstream, whilst evolution of the code in this repo can proceed independently.

There's a note about it in the Numba documentation (https://numba.readthedocs.io/en/stable/cuda/overview.html#built-in-cuda-target-deprecation-and-maintenance-status), but I realise now we don't explicitly spell it out in the Numba-CUDA docs.

numba_cuda/numba/cuda/tests/support.py
```diff
@@ -713,78 +724,7 @@ def _fix_strides(arr):
 
 
 class TestCase(unittest.TestCase):
```
gmarkall: Do we even need the `TestCase` class here anymore? Is there a usage of it I didn't see / understand?

numba_cuda/numba/cuda/tests/support.py
```diff
@@ -222,15 +223,25 @@ def override_config(name, value):
...
 
 
-# TODO: how to generate self.id() without access to instance?
-def make_dummy_type(test_obj: unittest.TestCase):
+# This can certainly be divided into two pytest fixtures. However, doing so
+# means developers will have to import both 'test_id_generator' and the fixture
+# returning the itertools.count() generator. Importing could be avoided
+# altogether by adding the fixtures in conftest.py, but that potentially
+# pollutes the namespace
+@pytest.fixture(name="test_id")
```
gmarkall, line 228: Ah, this is an interesting idea, and maybe better style than what I was thinking. I was imagining just something like:

```python

_dummy_type_counter = 0

def make_dummy_type():
    global _dummy_type_counter
    test_id = dummy_type_counter
    dummy_type_counter += 1

    # ...
```

(but I'm not suggesting you change it to this!)

gmarkall:
/ok to test https://github.com/NVIDIA/numba-cuda/pull/883/commits/65d755bb03f11055fbb5e1d847fba733d7928cf4

gmarkall:
> /ok to test [65d755b](https://github.com/NVIDIA/numba-cuda/pull/883/commits/65d755bb03f11055fbb5e1d847fba733d7928cf4)

Thanks for trying to start CI - unfortunately, only NVIDIA staff can run CI. Please ping me if you're waiting to run it to get feedback from the CI test suite though.

acosmicflamingo:
> > would not want you to have to review the PR for the nth time because you thought I had finally pushed that last commit (or maybe you'd prefer it be in its current state and you periodically check to see how it's going?).
> 
> I'm quite happy if we just iterate a few times, as long as you're OK with that. I find it hard to do "one perfect review" that outlines perfectly the exact set of suggestions I want to make anyway.
> 
> > By the way, what is the status of upstream numba's cuda-specific test suite, and what's the relationship between this codebase and theirs? I was finding that some functions from support.py marked for deletion reside in upstream numba's codebase. I was curious as to whether some of the work in this PR should eventually make its way upstream, or whether it's not really supported now that numba-cuda is in the picture.
> 
> `numba.cuda` in upstream is in maintenance mode for backward compatibility - that is, code written to use it should still work with the latest version of `numba` installed, but Numba-CUDA is where all new development has happened since we separated the CUDA target from upstream Numba. So things will stay quite static upstream, whilst evolution of the code in this repo can proceed independently.
> 
> There's a note about it in the Numba documentation (https://numba.readthedocs.io/en/stable/cuda/overview.html#built-in-cuda-target-deprecation-and-maintenance-status), but I realise now we don't explicitly spell it out in the Numba-CUDA docs.

Got it! I'm okay with incremental reviews; was actually thinking it's the better approach after seeing that just this PR alone has 22 files that have been modified and I have a loooong way to go.

Thank you for the background information and pointing me to the documentation; that really helps! I wonder whether I should add a mention of it in this separate low-priority [PR](https://github.com/NVIDIA/numba-cuda/pull/881) I had made tweaking CONTRIBUTING.md.

numba_cuda/numba/cuda/tests/support.py
acosmicflamingo: I suppose what I technically could do is just search all usage of `TestCase` and just replace it with `unittest.TestCase`; that would allow me to remove the `TestCase` class from `support.py`. I figured it'd be easier to just leave it there so tests can keep utilizing unittest API. Then when I refactor each test module, I just remove `(TestCase)` and change all the `unittest` API to simple assert statements.

gmarkall:
> I have a loooong way to go.

Maybe...? It looks most of the way there to me, I was expecting we'd just be doing a few tweaks (assuming that fixing CI is not too onerous). Let me know if you need any advice on reproducing the failures locally.

> I wonder whether I should add a mention of it in this separate low-priority https://github.com/NVIDIA/numba-cuda/pull/881 I had made tweaking CONTRIBUTING.md.

Thank you for the kind offer - I think that there are a few directional changes that need to be made to our guidance on Numba vs Numba-CUDA soon that will make it hard to align with an external contribution - could we please keep the other PR as-is for now?

numba_cuda/numba/cuda/tests/support.py
acosmicflamingo: Hahah yes, I had that idea too since it's not like the value would ever overflow (although the chance is still greater than zero). But this kind of thing seemed like something pytest fixtures were meant for :)

acosmicflamingo:
> > I have a loooong way to go.
> 
> Maybe...? It looks most of the way there to me, I was expecting we'd just be doing a few tweaks (assuming that fixing CI is not too onerous). Let me know if you need any advice on reproducing the failures locally.
> 
I see. This PR is about decoupling `unittest.TestCase` from the test suite, and removing `TestCase` (a `unittest.TestCase` subclass) will break all tests that use unittest API (e.g. `self.assertEqual`, etc.) through inheritance. This top-down approach of removing functionality like that was what I was doing [here](https://github.com/NVIDIA/numba-cuda/pull/882/changes), but felt like it was a lot messier than just taking the bottom-up approach of moving functionality out of the class, where removing the base class inheritance would just leave me with errors about missing `unittest` API. The approach I'm taking also allows me to run pytest on the suite to make sure that I haven't broken anything along the way :D

> > I wonder whether I should add a mention of it in this separate low-priority #881 I had made tweaking CONTRIBUTING.md.
> 
> Thank you for the kind offer - I think that there are a few directional changes that need to be made to our guidance on Numba vs Numba-CUDA soon that will make it hard to align with an external contribution - could we please keep the other PR as-is for now?

Yes that's fine with me!

acosmicflamingo:
> > /ok to test [65d755b](https://github.com/NVIDIA/numba-cuda/pull/883/commits/65d755bb03f11055fbb5e1d847fba733d7928cf4)
> 
> Thanks for trying to start CI - unfortunately, only NVIDIA staff can run CI. Please ping me if you're waiting to run it to get feedback from the CI test suite though.

Ha! That explains the disobedient bot ;) I'm going to try and reproduce the issue by creating a conda environment, since the issue seems to stem from there. I don't want to ping you every 4 minutes to try and fix that particular issue; I will save your patience for another time.

numba_cuda/numba/cuda/tests/support.py
acosmicflamingo: I have an idea. Given how extremely large a PR would be to remove unittest.TestCase inheritance from the test suite, this PR can be renamed "Refactor TestCase subclasses to directly inherit from unittest.TestCase". I just pushed a commit that removes TestCase usage, so I suppose once I fix that issue occurring in the conda environment, this PR would be ready to be ~~merged~~ reviewed!

acosmicflamingo:
@gmarkall never used `pixi` before; pretty neat package manager! After tweaking the subprocess command to address the `cuda.tests` not being found, I'm hitting another issue, a false positive that only appears when tests are executed in parallel (`pixi run -e cu-13-0-py313 pytest $PIXI_PROJECT_ROOT/testing -n auto --dist loadscope --loadscope-reorder -v`) and not sequentially (`pytest $PIXI_PROJECT_ROOT/testing -n auto --dist loadscope --loadscope-reorder -v`):
```python
assert status.stderr == "", status.stderr
                   ^^^^^^^^^^^^^^^^^^^
E           AssertionError: .../pytest_benchmark/logger.py:44: PytestBenchmarkWarning: Benchmarks are automatically disabled because xdist plugin is active.Benchmarks cannot be performed reliably in a parallelized environment.
E             warner(PytestBenchmarkWarning(text))

numba_cuda/numba/cuda/tests/support.py:416: AssertionError
```
The original code would assert that "OK" was present in stderr, which I figured was just a way to ensure the test really did pass:
```python
if no_tests_ran in status.stderr:
    self.skipTest(no_tests_ran)
else:
    self.assertIn("OK", status.stderr)
return status
```
After switching to `pytest` instead of `numba.runtests`, the assumption I made was that `status.stderr` would always be empty in a successful run, but I forgot about warnings that could still end up appearing in the error stream. 

Do you have a favorite out of the potential solutions below?
```diff
### Option A  ###
if no_tests_ran in status.stderr:
    self.skipTest(no_tests_ran)
-else:
-    self.assertIn("OK", status.stderr)
return status



### Option B  ###
+_IGNORED_STDERR_PATTERNS = [
+    "PytestBenchmarkWarning",
+    "Benchmarks are automatically disabled",
+]
+
+def _unexpected_stderr_found(stderr):
+    lines = stderr.splitlines()
+    return any(
+        line and not any(pat in line for pat in _IGNORED_STDERR_PATTERNS)
+        for line in lines
+    )
+
no_tests_ran = "NO TESTS RAN"
if no_tests_ran in status.stderr:
    pytest.skip(no_tests_ran)
+elif _unexpected_stderr_found(status.stderr):
+    pytest.fail(f"Unexpected stderr output:\n{status.stderr}")
return status

### Option C ###
???
```

acosmicflamingo:
Closing since numba-cuda is now marked as a maintenance-only product

PR #877, opened 2026-05-05
Migration to cuda-core 1.0RC

isVoid:
## Summary

Updates numba-cuda for cuda-core 1.0 RC API changes while retaining the cuda-core 0.7 API path:

- Add `CUDA_CORE_GE_1_0` in `numba.cuda._compat`.
- Use cuda-core 1.0 property-style kernel attributes when `CUDA_CORE_GE_1_0` is true.
- Keep cuda-core 0.7 callable kernel attribute access otherwise.
- Use `LaunchConfig(is_cooperative=...)` for cuda-core >= 1.0 and `LaunchConfig(cooperative_launch=...)` for older cuda-core versions.

## Validation

The installed candidate wheel still reports `cuda-core 0.7.1.dev...` while exposing the 1.0-style API, so the local validation forced `cuda.core.__version__ = "1.0.0"` through a temporary `sitecustomize.py` to exercise the gated >= 1.0 path in all xdist workers.

Ran the full CUDA test suite outside the sandbox against real GPUs:

```text
1984 passed, 63 skipped, 13 xfailed in 33.50s
```

JUnit report from the local run:

```text
/tmp/numba-cuda-cuda-core-rc-pytest-outside-ge1-versiongate.xml
```

PR #884, opened 2026-05-12
Bump Version to 0.30.2

isVoid:
## Changes since v0.30.1

- Migration to cuda-core 1.0RC (#877)


numba_cuda/VERSION
```diff
@@ -1 +1 @@
-0.30.1
+0.31.0
```
leofang: Based on the offline discussion, I think we want 0.30.2 instead of 0.31.0 (since there is only one combat-related change).

isVoid: Got it. I may missed the discussion at some point.

Issue #879, opened 2026-05-07
[BUG] Unreachable condition in interprtation of `SET_FUNCTION_ATTRIBUTE` on Python 3.14


gmarkall:
Reported by @shwina 

Minimal reproducer:

```python
import numpy as np
import numba.cuda

def outer(idx: np.int64) -> np.uint8:
    def inner(x: np.int64) -> np.uint8:   # ← nested fn with annotations
        return np.uint8(x % 5)
    return inner(idx)

sig = numba.types.uint8(numba.types.int64)
numba.cuda.compile(outer, sig=sig, output="ltoir")
```

This seems to have something to do with a nested function with annotations.

acosmicflamingo:
Looks like this issue came up on upstream side a few months ago (https://github.com/numba/numba/issues/10319) and was addressed accordingly: https://github.com/numba/numba/pull/10321.

Is there a formal process for syncing numba-cuda with upstream version? I could make a PR that cherry-picks the changes if you'd like.

gmarkall:
> I could make a PR that cherry-picks the changes if you'd like.

That would be great - many thanks!

acosmicflamingo:
@gmarkall Here you go! https://github.com/NVIDIA/numba-cuda/pull/886

PR #886, opened 2026-05-13
Cherry-pick fix for unreachable condition issue from upstream numba 

acosmicflamingo:
Addresses https://github.com/NVIDIA/numba-cuda/issues/879

A couple of months ago, this same issue appeared in upstream numba: https://github.com/numba/numba/issues/10319. They fixed it in this [PR](https://github.com/numba/numba/pull/10321), so all that's needed is to just cherry-pick their fixes. I verified that the reproducer @shwina created for us does work with these changes. 

Only question I have is whether `numba/tests/test_nested_calls.py` has any place in numba-cuda. I deliberately didn't cherry-pick https://github.com/numba/numba/pull/10321/changes/b6a2a10214064db5e79e54499a1d731ffb2d6c62 because I figure this repository is meant for CUDA-specific functionality and it's upstream's responsibility for ensuring this kind of functionality is tested. If my assumption is correct, then this PR is ready to be merged.

If it is needed, I wonder where it should be placed because numba/tests directory doesn't exist. I'm 99% sure it's not needed but just want an opinion from someone with authority ;)


acosmicflamingo:
I'd close this, but this seems critical for Python 3.14 users

PR #887, opened 2026-05-13
Add missing vendor-in changes to numba-cuda

acosmicflamingo:
Changes from the following commit (from upstream numba) didn't make it through to numba-cuda: https://github.com/numba/numba/commit/3bac3bf9d7479107a3819b480abee8e57f59ea96

Re-adding the work Todd A. Anderson had made, where he had commented "Extend lazy_pformat to work on arbitary string producing functions for use in other places where we need lazy but it doesn't use pformat."

Want to mention that I deliberately didn't add a test to this PR because I felt like this fell under the category of tests that upstream numba is responsible for. I'd be happy to add a test; wanted to avoid giving the impression that I don't think about adding tests until a maintainer asks me to ;)


atmnp (approved):
I had a similar change like this locally at some point, ran into the same bug it seemed to be transient before I got around to submitting the fix. Thanks for taking the time!

acosmicflamingo:
@atmnp woohoo! Thanks for taking the time to review and merge! I imagine you are already aware but I also have a PR that fixes it in numba-cuda-mlir, just so you know :)

acosmicflamingo:
@atmnp oh no, the dreaded `E1` appeared :(

acosmicflamingo:
Could anyone with admin privileges please run `/ok to test 3bf0ffa07d15437b3e6dad269becaf43a862c799` for me 🙇🏽?

PR #683, opened 2025-12-19
Dropping bits in the old CI & Propagating recent changes from cuda-python

leofang:
Part of #671.

This PR removes the old wheel build/test workflows in favor of the new ones. In particular, this allows us to drop the GH-hosted Windows GPU runners (`cuda-python-windows-gpu-github`) entirely and to also reduce the amount of CI resource that we use in this repo (we currently use more than we should, because the old/new CIs are running in parallel).

Then, all other test jobs, such as conda & 3rd libs, in the old CI (`ci.yaml` runs on `main`, while `pr.yaml` runs per PR, but the content of both are almost identical) are hooked to the new CI as-is (`ci-new.yaml` runs on both `main` and per-PR). Further consolidation would require some nontrivial changes (ex: commit 0d5e440 proved that we still need to use RAPIDS containers for these jobs, so I reverted it).

Some recent cuda-python CI changes are propagated to here, including transition from JSON to YAML files. However, the two YAML files `test-matrix.yaml` and `matrix.yaml` still cannot be merged yet, due to some workflows still using the latter (ex: https://github.com/NVIDIA/numba-cuda/issues/680).

Below is an LLM-generated summary:

<details>
<summary> ➡️ CLICK ME ⬅️ </summary>

This pull request significantly refactors and consolidates the project's CI/CD workflows. The main focus is on migrating jobs from legacy workflow files (`ci.yaml`, `pr.yaml`) into the unified `ci-new.yaml` workflow, modernizing configuration files, and improving maintainability by removing redundancy and unused scripts.

**CI/CD Workflow Consolidation and Refactoring:**

[...]

PR #889, opened 2026-05-27
docs: add Windows test commands

### LINKED ISSUE #862: [DOC] Add steps for running tests on Windows
The [documentation for running tests](https://github.com/NVIDIA/numba-cuda?tab=readme-ov-file#running-tests) in the readme only describes the process on Linux.

We should also add the equivalent steps for running tests on Windows.

omribz156:
## Summary

- Add Windows commands for building test binaries with the existing `testing/build.bat` helper
- Show the PowerShell form of `NUMBA_CUDA_TEST_BIN_DIR`
- Add language hints to the existing README command blocks

Fixes #862.

## Verification

- `git diff --check`
- `rg -n "build.bat|NUMBA_CUDA_TEST_BIN_DIR|pytest -n auto|pixi run" README.md testing pixi.toml`

Full test execution was not run because this is a docs-only change and this local machine is not a CUDA-capable Windows test environment.

This was prepared with Codex assistance and manually reviewed, with the patch kept to the README test-running instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ configuration and code to generate binaries used during the tests. The test
 binaries need to be built on the system on which the tests are run, so that
 they are compiled for the appropriate compute capability.
 
-```
+```bash
 cd testing
 # Optionally, build test binaries and point to their location for the test suite
 make -j $(nproc)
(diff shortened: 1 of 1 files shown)


acosmicflamingo:
Copying what I had written in original ticket so maintainers are aware while reviewing the PR:
> so that build.bat script was called by ci/test_wheel.ps1 a few months ago in this https://github.com/NVIDIA/numba-cuda/pull/683, and it was the only thing referencing build.bat before it was removed from the codebase. Additionally, it doesn't look like it was getting as many updates as Makefile currently does. With that being said, before giving build.bat a well-needed makeover, I think we first need to know whether we are better off maintaining two distinct files, or whether we should leverage unix-tools in Windows so we can just use Makefile

TL;DR: the build.bat script hasn't been maintained for a while, so the PR in its current form isn't enough to address the issue

PR #895, opened 2026-06-03
Fix RMM EMM plugin module path

bdice:
## Summary
Updates the RMM EMM plugin example to use the module that exposes `_numba_memory_manager`.

### DIFF
--- a/docs/source/user/external-memory.rst
+++ b/docs/source/user/external-memory.rst
@@ -282,7 +282,7 @@ Numba test suite with an EMM Plugin, e.g.:
 
 .. code::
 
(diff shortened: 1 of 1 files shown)


kkraus14 (approved):

leofang:
@bdice could you rebase?

docs/source/user/external-memory.rst
leofang, line 1: @bdice can we just keep this diff and remove all other diff?

leofang (approved):

leofang:
Since this is a doc-only change and we have CI issues today, let me admin-merge.

PR #905, opened 2026-07-02
Add support for Numba 0.66

bdice:
Numba 0.66 moved `NPDatetime` and `NPTimedelta` from
`numba.core.types.scalars` to `numba.np.types`. This updates numba-cuda to
re-export the types from their new public location while preserving the
existing behavior with older Numba releases. It also replaces the remaining
uses of the old private `_NPDatetimeBase` path with the public concrete types.

xref: https://github.com/NVIDIA/numba-cuda-mlir/issues/172
xref: https://github.com/rapidsai/build-planning/issues/301

Tests:

- `pre-commit run --files ...`
- Numba 0.66.0 / Python 3.14: no-CUDA import tests (2 passed)
- Numba 0.66.0 / Python 3.14 on an NVIDIA GeForce RTX 2070 SUPER: datetime tests (7 passed)
- Numba 0.60.0 / Python 3.12: no-CUDA import tests (2 passed)
- Numba 0.60.0 / Python 3.12 on an NVIDIA GeForce RTX 2070 SUPER: datetime tests (7 passed)

### DIFF
--- a/numba_cuda/numba/cuda/cgutils.py
+++ b/numba_cuda/numba/cuda/cgutils.py
@@ -22,6 +22,19 @@ int32_t = ir.IntType(32)
 intp_t = ir.IntType(utils.MACHINE_BITS)
 voidptr_t = int8_t.as_pointer()
 
+
+def add_no_capture_attribute(arg):
+    """Mark an LLVM function argument as not capturing pointers."""
+    try:
+        arg.add_attribute("nocapture")
+    except ValueError:
+        # llvmlite >= 0.48 only supports the newer ``captures(none)``
+        # spelling, but NVVM does not accept that syntax. The attribute is an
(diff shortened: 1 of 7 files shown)


brandon-b-miller (approved):

gmarkall (approved):

kkraus14:
I think we should also pin to numba <0.67 in addition to this change. Things will inevitably continue to break as there's future numba releases.

gmarkall (approved):

bdice:
> I think we should also pin to numba <0.67 in addition to this change. Things will inevitably continue to break as there's future numba releases.

That would force fallback to _older_ numba-cuda versions with no upper bound, right?

leofang (approved):

Issue #900, opened 2026-06-25
CI: Check job status aggregator passes when build jobs fail (port of NVIDIA/cuda-python#2208)


leofang:
Same CI gating-check regression as NVIDIA/cuda-python#2208 — the `Check job status` aggregator only catches `cancelled || failure` results on its `test-*` / `build-docs` / `coverage-report` deps, so an upstream `build-*` failure that cascades to `'skipped'` on the downstream jobs slips through silently and the gate reports success. The single required status check on `main` (per the repo ruleset) is exactly this aggregator.

Affected code on `main` (HEAD `5013eaf5d5d8237028fd97dcfee3386ee2efd7a0`):
https://github.com/NVIDIA/numba-cuda/blob/5013eaf5d5d8237028fd97dcfee3386ee2efd7a0/.github/workflows/ci.yaml#L292-L352

Fix design and reference implementations (CCCL's `check_result` pattern with explicit `expected="success"` per dep):
- Root cause + design discussion: NVIDIA/cuda-python#2208
- Source-of-truth fix: NVIDIA/cuda-python#2210
- Port to sibling `numba-cuda-mlir` repo: NVIDIA/numba-cuda-mlir#142
- Underlying CCCL gate pattern: https://github.com/NVIDIA/cccl/blob/8c0e6cb1b6412d7bd0070f9ac3f55fa80231961a/.github/workflows/ci-workflow-pull-request.yml#L463-L526

Filing for visibility — this repo is in maintenance mode, no active fix planned.

PR #908, opened 2026-07-03
Revert "build(deps): bump the actions-monthly group across 1 directory with 4 updates"

leofang:
Reverts NVIDIA/numba-cuda#904. The automerge kicked in because of a CI bug https://github.com/NVIDIA/numba-cuda/issues/900 that could not mark the CI failed  when the build failed 😩


leofang:
Admin-merging...

PR #915, opened 2026-07-10
Migrates resolution cache code from numba fixing a JIT speed regression.

c200chromebook:
This is a straight-shot move of resolution cache from the original numba to correct a compile time regression. We have it monkeypatched in now and it works ok. Anyone using the sol'n would probably benefit from it. We are working to get MLIR up and running for use case as well (note recent bug report).

### DIFF
--- a/numba_cuda/numba/cuda/core/config.py
+++ b/numba_cuda/numba/cuda/core/config.py
@@ -424,6 +424,12 @@ class _EnvReloader:
         # Disable jit for debugging
         DISABLE_JIT = _readenv("NUMBA_DISABLE_JIT", int, 0)
 
+        # Disable caching of failed type inferences.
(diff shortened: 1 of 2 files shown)


c200chromebook:
@gmarkall  would you mind having a look at this one? it is kind of annoying for us

gmarkall:
@c200chromebook Thanks for the PR and the ping. I'm no longer working on Numba-CUDA, so I will defer to @kkraus14 on where to go with this PR.

kkraus14:
@c200chromebook please see https://github.com/NVIDIA/numba-cuda#maintenance-notice

We are only accepting security issues and critical bug fixes at this time for numba-cuda. I would recommend trying to move to numba-cuda-mlir and if the same JIT speed regression exists there we will happily try to resolve it.

PR #916, opened 2026-07-14
Guard np.row_stack overload for numpy < 2.5 (#907)

### LINKED ISSUE #907: [FEA] Support numpy 2.5
See https://github.com/numba/numba/pull/10645

This patch should do it:

```patch
diff --git a/numba_cuda/numba/cuda/np/arrayobj.py b/numba_cuda/numba/cuda/np/arrayobj.py
index 299810d..3587683 100644
--- a/numba_cuda/numba/cuda/np/arrayobj.py
+++ b/numba_cuda/numba/cuda/np/arrayobj.py
@@ -6856,7 +6856,7 @@ def impl_np_vstack(tup):
         return impl


-if numpy_version >= (2, 0):
+if numpy_version >= (2, 0) and numpy_version < (2, 5):
     overload(np.row_stack)(impl_np_vstack)
```

CodersAcademy006:
## Summary
`numpy` 2.5 removes `np.row_stack`. The overload was registered unconditionally for `numpy_version >= (2, 0)`, so it breaks under numpy 2.5. This bounds it to `[2.0, 2.5)`, mirroring the upstream numba fix in numba/numba#10645 (as suggested by the reporter).

## Change
`numba_cuda/numba/cuda/np/arrayobj.py`:
```python
-if numpy_version >= (2, 0):
+if numpy_version >= (2, 0) and numpy_version < (2, 5):
     overload(np.row_stack)(impl_np_vstack)
```

Closes #907

cc @gmarkall

### DIFF
--- a/numba_cuda/numba/cuda/np/arrayobj.py
+++ b/numba_cuda/numba/cuda/np/arrayobj.py
@@ -6856,7 +6856,7 @@ def impl_np_vstack(tup):
         return impl
 
 
-if numpy_version >= (2, 0):
+if numpy_version >= (2, 0) and numpy_version < (2, 5):
     overload(np.row_stack)(impl_np_vstack)
 
 


leofang:
Thank you for your contribution! Unfortunately we don't plan to keep up with newer NumPy versions, see https://github.com/NVIDIA/numba-cuda/issues/907#issuecomment-5620678407 for further detail.

PR #922, opened 2026-08-26
Fix numpy 2.5 deprecation warnings

brandon-b-miller:
Inherits a fix already on upstream that emits a warning for an deprecated method of datetime construction we're still using. This is strictly for numpy >= 2.5 environments.


seberg (approved):
Thanks, sorry about this deprecation noise.   Clearly good, since in this context the actual time unit doesn't matter (it really is just `INT64_MIN` in the end)

bdice (approved):

brandon-b-miller:
Some additional fixups needed here, following up shortly.

leofang:
xref https://github.com/NVIDIA/numba-cuda/issues/907#issuecomment-5620678407

